PR #37, opened 2024-12-14
A compat wrapper that converts `TransactionInstruction` to `IInstruction`

shubhiscoding:
solves https://github.com/anza-xyz/solana-web3.js/issues/10


shubhiscoding:
need your review @steveluscher

PR #38, opened 2024-12-14
Enable ESLint rule and suppress errors for data flexibility

kox:
Reopening this PR as the repository has been moved. This PR intends to close another mini task from https://github.com/anza-xyz/solana-web3.js/issues/13

I suppressed the errors due keeping data type flexibility


steveluscher (approved):
Ooh, gross. I guess the problem is that the `stringify` method doesn't actually return `string` all the time. Rather than to mess with the implementation, I agree with your assessment to suppress lint here.

Issue #13, opened 2024-10-26
Re-enable `@typescript-eslint` rules one at a time


steveluscher:
Create one PR for each of these rules re-enabled:

https://github.com/solana-labs/solana-web3.js/blob/612a4d8688fc2a481d24ec7f415701c6aa92c707/eslint.config.mjs#L9-L20

https://github.com/solana-labs/solana-web3.js/blob/612a4d8688fc2a481d24ec7f415701c6aa92c707/examples/react-app/eslint.config.mjs#L27-L30

PR #42, opened 2024-12-15
Enable eslint rule and reject with errors instead of strings

kox:
Following #13 , this PR intends to enable the eslint rule: `prefer-promise-reject-errors` as error.

I believe that treating promise rejections as errors ensures better handling and propagation of the rejection reason to the final components. The affected package tests have been updated accordingly.

### DIFF
--- a/examples/rpc-transport-throttled/src/example.ts
+++ b/examples/rpc-transport-throttled/src/example.ts
@@ -94,7 +94,8 @@ function getThrottledTransport<TClusterUrl extends ClusterUrl>(
             } as QueuedRequest<TClusterUrl>);
             if (config.signal) {
                 config.signal.addEventListener('abort', function () {
-                    reject(new Error(this.reason));
+                    // eslint-disable-next-line @typescript-eslint/prefer-promise-reject-errors
+                    reject(this.reason);
                 });
             }
             processQueue();
--- a/examples/rpc-transport-throttled/src/example.ts
+++ b/examples/rpc-transport-throttled/src/example.ts
@@ -94,7 +94,7 @@ function getThrottledTransport<TClusterUrl extends ClusterUrl>(
             } as QueuedRequest<TClusterUrl>);
             if (config.signal) {
                 config.signal.addEventListener('abort', function () {
-                    reject(this.reason);
+                    reject(new Error(this.reason));
                 });
             }
             processQueue();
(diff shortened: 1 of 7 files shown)


steveluscher (changes requested):
The `reason` property of an `AbortSignal` can actually be an `Error`.

Consider the difference between:

```ts
const ac = new AbortController();
ac.signal.addEventListener('abort', function () { console.log(this.reason.constructor, this.reason); })
ac.abort();
```

…and…

```ts
const ac = new AbortController();
ac.signal.addEventListener('abort', function () { console.log(this.reason.constructor, this.reason); })
ac.abort('o no');
```

`reason` is a `DOMException` in the first, and a `String` in the second.

I think there are only two ways forward here:

* Leave this as is, suppress the lint.
* If the `reason` is `typeof string` then wrap it in an `Error`. Otherwise, reject with `reason` directly.

I think the second raises more questions than it answers, so I'm inclined to do the first.

kox:
Following your concerns and for simplicity, I suppressed the lints and introduced Errors inside of a test and an example.

examples/rpc-transport-throttled/src/example.ts
```diff
@@ -94,7 +94,7 @@ function getThrottledTransport<TClusterUrl extends ClusterUrl>(
             } as QueuedRequest<TClusterUrl>);
             if (config.signal) {
                 config.signal.addEventListener('abort', function () {
-                    reject(this.reason);
+                    reject(new Error(this.reason));
```
steveluscher: Again, if the call to `AbortController#abort` had no `reason` argument, the `reason` would materialize as a `DOMException`. The code that you wrote here would be the equivalent of:

```ts
reject(
    new Error(
        (new DOMException(...)).toString(),
    ),
);
```

…when what you want to do is to throw the `DOMException` as is.

packages/rpc/src/__tests__/rpc-request-coalescer-test.ts
```diff
@@ -110,7 +110,7 @@ describe('RPC request coalescer', () => {
                     return await new Promise((resolve, reject) => {
                         transportResponsePromise = resolve;
                         signal?.addEventListener('abort', (e: AbortSignalEventMap['abort']) => {
-                            reject((e.target as AbortSignal).reason);
+                            reject(new Error((e.target as AbortSignal).reason));
```
steveluscher: I think the aim of the tests is to demonstrate that the `reason` gets thrown as-is. All this conversion does is to get in the way of that.

steveluscher (changes requested):
I still think the `Error` conversions need to go.

kox:
Sorry about that. I took a deeper look into the AbortController and AbortSignal and you are right about it. It doesn't make sense to wrap it in an extra Error.

steveluscher (approved):
Thanks for improving our code quality!

PR #43, opened 2024-12-16
Add hash to frozen slot update

diman-io:
This is a corresponding PR to https://github.com/anza-xyz/agave/pull/3881

This reopens https://github.com/solana-labs/solana-web3.js/pull/3662 as development has transitioned from Solana Labs to Anza.


diman-io:
@steveluscher
The errors don’t look like something I could have caused, do they? Sorry, I’m not very familiar with npm and all that.
Could you please take a look?

diman-io:
I see that `main` has the same issues, but `changeset-release/main` doesn’t, so maybe I just rebased at an unfortunate time. I’ll redo it later, no problem. Anyway, the waiting time for a PR in the main repository is likely to be longer.

PR #48, opened 2025-01-03
Add documentation to the subscription channel websocket package

kox:
This PR intend to include documentation for the subscription channel websocket package. I don't know if there is already an issue created but i thought it would be interesting to remove another TODO from the list.

### DIFF
--- a/packages/rpc-subscriptions-channel-websocket/README.md
+++ b/packages/rpc-subscriptions-channel-websocket/README.md
@@ -11,4 +11,136 @@
 
 # @solana/rpc-subscriptions-channel-websocket
 
-TODO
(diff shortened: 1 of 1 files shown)


steveluscher:
Did you use an LLM to produce this PR?

kox:
> Did you use an LLM to produce this PR?

Hi, I  used another package Readme as a template and I created a first draft. Then I used a LLM to improve the writting and fillup more information. Therefore yes, I used it.

steveluscher (changes requested):
Well, the LLM led you astray. Methods like `subscribe` do not exist on this interface, and your example does not run.

If you still want to add documentation here, please use the documentation from https://github.com/anza-xyz/solana-web3.js/tree/main/packages/rpc-transport-http as a starting point, and don't use machine assistance.

kox:
My bad!! It really went wrong this time, sorry about it. I'm closing this PR as it doesn't reflect correctly and I will take a look into your recommendation.  Thanks and sorry for the confusion

Issue #50, opened 2025-01-04
Move `README` content into code as TypeDoc


steveluscher:
* Familiarize yourself with TypeDoc **Tags** https://typedoc.org/documents/Tags.html
* Preview the documentation by running `pnpm turbo compile:ghpages --output-logs=errors-only` and `python3 -m http.server -d docs` in a separate terminal
* Upload the docs somewhere and attach the link to your PR (eg. upload a ZIP through https://app.netlify.com/)
* Submit one or more PR per package

PR #51, opened 2025-01-04
Document `@solana/accounts` with TypeDoc

steveluscher:
This PR moves content from the `README` into the code.

Preview here: https://illustrious-gumption-b4d2fe.netlify.app/modules/_solana_accounts.html

Addresses #50

### DIFF
--- a/packages/accounts/src/account.ts
+++ b/packages/accounts/src/account.ts
@@ -2,10 +2,32 @@ import type { Address } from '@solana/addresses';
 import { ReadonlyUint8Array } from '@solana/codecs-core';
 import type { Lamports } from '@solana/rpc-types';
 
-/** The amount of bytes required to store the base account information without its data. */
+/**
+ * The number of bytes required to store the {@link BaseAccount} information without its data.
+ *
+ * @example
+ * ```ts
+ * const myTotalAccountSize = myAccountDataSize + BASE_ACCOUNT_SIZE;
+ * ```
+ */
 export const BASE_ACCOUNT_SIZE = 128;
 
-/** Describe the generic account details applicable to every account. */
+/**
+ * Defines the attributes common to all Solana accounts. Namely, it contains everything stored
+ * on-chain except the account data itself.
+ *
+ * @interface
+ *
+ * @example
+ * ```ts
+ * const BaseAccount: BaseAccount = {
+ *     executable: false,
+ *     lamports: lamports(1_000_000_000n),
+ *     programAddress: address('1111..1111'),
+ *     space: 42n,
+ * };
+ * ```
+ */
 export type BaseAccount = {
     readonly executable: boolean;
     readonly lamports: Lamports;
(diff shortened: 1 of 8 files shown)


steveluscher:
* **#51** <a href="https://app.graphite.dev/github/pr/anza-xyz/solana-web3.js/51?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/solana-web3.js/51?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/accounts/typedoc.json
```diff
@@ -1,5 +1,6 @@
 {
     "$schema": "https://typedoc.org/schema.json",
     "extends": ["../typedoc.base.json"],
-    "entryPoints": ["src/index.ts"]
+    "entryPoints": ["src/index.ts"],
+    "readme": "none"
```
steveluscher, line 5: This stops TypeDoc from vaccuming up the `README` for display in the package docs' index, in favour of the comment with `@packageDocumentation`.

packages/accounts/src/account.ts
```diff
@@ -2,22 +2,93 @@ import type { Address } from '@solana/addresses';
...
+ * @interface
+ *
+ * @example
+ * ```ts
+ * const BaseAccount: BaseAccount = {
+ *     executable: false,
+ *     lamports: lamports(1_000_000_000n),
+ *     programAddress: address('1111..1111'),
+ * };
+ * ```
```
buffalojoec: This is missing `space`.

packages/accounts/src/account.ts
```diff
@@ -2,22 +2,93 @@ import type { Address } from '@solana/addresses';
...
+ *
+ * @typeParam TAddress - Supply a string literal to define an account having a particular address.
+ * @typeParam TData - The nature of this account's data. It can be represented as either a
+ * `Uint8Array` &ndash; meaning the account is encoded &ndash; or a custom data type &ndash; meaning
+ * the account is decoded.
+ *
+ * @example
+ * ```ts
+ * // Encoded
+ * const myEncodedAccount: Account<Uint8Array, '1234..5678'> = {
```
buffalojoec, line 52: I think it would be slightly more clear if the example string literal looked like a base58 address, instead of integers.

packages/accounts/src/decode-account.ts
```diff
@@ -38,7 +59,39 @@ function accountExists<TData extends object>(account: Account<TData> | MaybeAcco
     return !('exists' in account) || ('exists' in account && account.exists);
 }
 
-/** Asserts that an account has been decoded. */
+/**
+ * Asserts that an account stores decoded data, ie. not a `Uint8Array`.
+ *
+ * Note that it does not check the shape of the data matches the decoded type, only that it is not a
+ * `Uint8Array`.
```
buffalojoec, line 66: Interesting. I find the function signature a bit misleading then, since it requires the type parameter `TData`, which would imply that it's checking against the provided type. If it's just asserting it's not a `Uint8Array`, the type parameter can probably be elided. Perhaps we just rename the generic parameter.

packages/accounts/src/decode-account.ts
```diff
@@ -55,7 +108,25 @@ export function assertAccountDecoded<TData extends object, TAddress extends stri
     }
 }
 
-/** Asserts that all accounts have been decoded. */
+/**
+ * Asserts that all input accounts store decoded data, ie. not a `Uint8Array`.
+ *
+ * As with {@link assertAccountDecoded} it does not check the shape of the data matches the decoded
+ * type, only that it is not a `Uint8Array`.
```
buffalojoec, line 115: Does this function work for an array of accounts that are all different `TData`? Say you have a response from `getProgramAccounts` with token accounts and mint accounts, this should work fine, right?

packages/accounts/src/maybe-account.ts
```diff
@@ -7,15 +7,88 @@ import {
...
 
-/** Asserts that an account that may or may not exists, actually exists. */
+/**
+ * Given a {@link MaybeAccount}, asserts that the account exists and allows it to be used as an
+ * {@link Account} type going forward.
+ *
+ * @typeParam TAddress - Supply a string literal to define an account having a particular address.
+ * @typeParam TData - The nature of this account's data. It can be represented as either a
+ * `Uint8Array` &ndash; meaning the account is encoded &ndash; or a custom data type &ndash; meaning
+ * the account is decoded.
```
buffalojoec, line 79: Calling out that, similar to `assertAccountDecoded`, this function does not validate that the account has the provided `TData` layout.

packages/accounts/typedoc.json
buffalojoec: Is this what we want, though? Because it looks like you just pasted the README anyway. Otherwise we will have to maintain both, right?

buffalojoec (commented):
Nice!! Looking super slick.

packages/accounts/src/decode-account.ts
lorisleiva: I doesn't require a `TData` really because it is inferred from the first argument. So you give me a `Account<Token | Uint8Array>` and you want to narrow that down to `Account<Token>`, you use this function.

packages/accounts/src/decode-account.ts
lorisleiva: Yes, that can work if `TData` equals `Token | Mint` basically.

packages/accounts/typedoc.json
lorisleiva: I have a similar concern here. I think this is all amazing work. I'm just not sure about how we should proceed going forward with this information redundancy.

steveluscher: Pasted the README? The idea here is that _all_ documentation should be:

1. In the code, if nowhere else
2. Autogenable from the code

In a world where the README prose lives in the `index.ts` file as `@packageDocumentation` and the types/functions documentation lives on the types/functions themselves, then the README is fully codegennable from the code.

If you look at the preview site, the _autogenerated_ [package documentation](https://illustrious-gumption-b4d2fe.netlify.app/modules/_solana_accounts.html) has everything the README has now (and a little bit more).

packages/accounts/src/account.ts
steveluscher: We'll have to change that in a separate PR because this style is used in multiple packages.

packages/accounts/src/decode-account.ts
steveluscher: It doesn't, actually. https://github.com/anza-xyz/solana-web3.js/issues/49

steveluscher (commented):
Alright, this is going to be an absolute ton of work, I'm going to need help, and this isn't the terminal state of the docs by any means, but rather a better starting point (ie. with the docs finally beside the functions / values / types to which it pertains). I'm going to land this and we'll keep moving forward.

lorisleiva (approved):
👌

steveluscher:
### Merge activity

* **Jan 14, 3:21 PM PST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/solana-web3.js/51).

PR #62, opened 2025-01-16
Document `@solana/addresses` with TypeDoc

steveluscher:
This PR moves content from the README into the code.

Addresses #50

# Test Plan

```shell
cd packages/addresses
pnpm typedoc --watch --plugin typedoc-plugin-missing-exports
python3 -m http.server -d docs
```

Preview here: https://storied-raindrop-7eacf2.netlify.app/

### DIFF
--- a/packages/addresses/src/address.ts
+++ b/packages/addresses/src/address.ts
@@ -16,6 +16,13 @@ import {
     SolanaError,
 } from '@solana/errors';
 
+/**
+ * Represents a string that validates as a Solana address. Functions that require well-formed
+ * addresses should specify their inputs in terms of this type.
+ *
+ * Whenever you need to validate an arbitrary string as a base58-encoded address, use the
+ * {@link address}, {@link assertIsAddress}, or {@link isAddress} functions in this package.
+ */
 export type Address<TAddress extends string = string> = TAddress & {
     readonly __brand: unique symbol;
 };
@@ -33,6 +40,24 @@ function getMemoizedBase58Decoder(): Decoder<string> {
     return memoizedBase58Decoder;
 }
 
+/**
+ * A type guard that returns `true` if the input string conforms to the {@link Address} type, and
+ * refines its type for use in your program.
+ *
+ * @example
+ * ```ts
+ * import { isAddress } from '@solana/addresses';
+ *
+ * if (isAddress(ownerAddress)) {
+ *     // At this point, `ownerAddress` has been refined to a
+ *     // `Address` that can be used with the RPC.
+ *     const { value: lamports } = await rpc.getBalance(ownerAddress).send();
+ *     setBalanceLamports(lamports);
+ * } else {
+ *     setError(`${ownerAddress} is not an address`);
+ * }
+ * ```
+ */
 export function isAddress(putativeAddress: string): putativeAddress is Address<typeof putativeAddress> {
     // Fast-path; see if the input string is of an acceptable length.
     if (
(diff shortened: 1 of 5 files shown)


steveluscher:
* **#62** <a href="https://app.graphite.dev/github/pr/anza-xyz/solana-web3.js/62?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/solana-web3.js/62?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
FYI the preview isn't working correctly, I'm getting a 404 on all the type/function pages, eg https://peaceful-sorbet-e44996.netlify.app/types/Address.html

packages/addresses/src/address.ts
```diff
@@ -33,6 +40,23 @@ function getMemoizedBase58Decoder(): Decoder<string> {
     return memoizedBase58Decoder;
 }
 
+/**
+ * A type guard that returns `true` if the input string conforms to the {@link Address} type, and
+ * refines its type for use in your program.
+ *
+ * ```ts
```
mcintyre94, line 48: Missing `@example`?

packages/addresses/src/address.ts
```diff
@@ -75,21 +124,87 @@ export function assertIsAddress(putativeAddress: string): asserts putativeAddres
...
+ *
+ * await transfer(address(fromAddress), address(toAddress), lamports(100000n));
+ * ```
+ *
+ * > [!TIP]
+ * > When starting from a known-good address as a string, it's more efficient to typecast it rather
+ * than to use the {@link address} helper, because the helper unconditionally performs validation on
+ * its input.
+ * >
+ * > ```ts
```
mcintyre94, line 144: Missing `@example` I think
(though these don't seem to affect rendering)

mcintyre94, line 144: Is the `[!TIP]` supposed to render as-is?

Screenshot 2025-01-16 at 10 45 32

packages/addresses/src/address.ts
```diff
@@ -75,21 +124,87 @@ export function assertIsAddress(putativeAddress: string): asserts putativeAddres
...
+ * const address = addressDecoder.decode(address); // B9Lf9z5BfNPT4d5KMeaBFx8x1G4CULZYR1jA2kmxRDka
+ * ```
+ */
 export function getAddressDecoder(): FixedSizeDecoder<Address, 32> {
     return fixDecoderSize(getMemoizedBase58Decoder(), 32) as FixedSizeDecoder<Address, 32>;
 }
 
+/**
+ * Returns the intersection of the values returned by {@link getAddressDecoder} and
+ * {@link getAddressEncoder}.
```
mcintyre94: I think this wording is a bit confusing

I think I'd prefer to describe this as something like

> Returns a codec that you can use to encode from or decode to a base-58 encoded address

And keep the `@see` links that explain the encoder and decoder separately

mcintyre94 (dismissed):
Looks good!

steveluscher:
> FYI the preview isn't working correctly, I'm getting a 404 on all the type/function pages, eg https://peaceful-sorbet-e44996.netlify.app/types/Address.html . Works locally though so maybe just a Netlify thing?

Oh weird. I redeployed the preview site and updated the link in the description.

packages/addresses/src/address.ts
steveluscher: I see this.

image

Anyway, it all is going to depend on what we use to transform the TypeDoc into a website.

steveluscher: The rendering is hot garbage that we're not going to end up using. What we're going to do is to run TypeDoc with `--json`, walk the output of that, and make our own webpages however we like. The `@example` tag in TypeDoc will create us a section in the JSON called `{ "examples": [...] }` that we can iterate over.

steveluscher: Oh, and since this is part of the TIP blockquote, I don't think I should use `@example` here. The whole entire thing is consumed by the `@example` tag above.

mcintyre94: Gotcha!

mcintyre94: That's what I was hoping you were going to say!

mcintyre94 (approved):
LGTM!

PR #63, opened 2025-01-16
fix: address issue in readme on keys example.

chrismaree:
#### Problem

There is a typo in the example code given within the signers package.

#### Summary of Changes
Simply fixes the example code by using the correct function.


mcintyre94 (changes requested):
Thanks for taking a look at this, but I think the typo is the other way round. We need to update the import to `import { createKeyPairSignerFromBytes } from '@solana/signers;`

That is the function we want to use in this example.

lorisleiva:
Thanks for pointing this out! I've actually got a PR (https://github.com/anza-xyz/solana-web3.js/pull/78) that updates all the docblocks of this package already so I've made the change there. Thanks again! 🍺

PR #71, opened 2025-01-20
Always compare the current PR's bundle size to main

### LINKED ISSUE #70: Bundlemon reports incorrect size deltas when using stacked PRs
Bundlemon is a CI step that compiles the code and then compares the size of key compiled bundles with the mergebase. Unfortunately when you stack a PR (ie. with Graphite) the mergebase is _the PR before it_. Bundlemon has no knowledge of such a branch and thinks that every file has been added anew.

The fix will be to get Bundlemon to use the root branch as the mergebase. Or I guess just _always_ `main`.

steveluscher:
Regardless of what the mergebase of the current PR is, this will cause Bundlemon to _always_ compare the size from the perspective of the last known `main` size.

This prevents stacked PRs from trying (and failing) to find the size of the PR branch directly underneath them.

Fixes #70 

# Test Plan

Works on #72.

### DIFF
--- a/.github/workflows/bundlesize.yml
+++ b/.github/workflows/bundlesize.yml
@@ -35,3 +35,7 @@ jobs:
 
       - name: BundleMon
         uses: lironer/bundlemon-action@v1
+        env:
+          # Always compare to the main branch; prevents stacked PRs from being
+          # compared to the wrong mergebase (ie. the PR beneath the current one)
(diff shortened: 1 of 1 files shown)


steveluscher:
Don't review this yet. I want to make extra sure this works next time I have a stack of PRs in Graphite.

PR #84, opened 2025-01-30
Remove RpcApiMethods from README docs

mcintyre94:
#### Problem

We don't export `RpcApiMethods` any more (see  https://github.com/solana-labs/solana-web3.js/pull/3213), but still have it referenced in the README. Using this snippet as-is will give a Typescript error

#### Summary of Changes

- Remove import from code example (it's not actually used in the code example)
- Remove outdated documentation of `RpcApiMethods` type, which is no longer exported


steveluscher (approved):
Added via Giphy

PR #85, opened 2025-01-30
Fix accountKeys type in getTransaction JSON parsed

### LINKED ISSUE #80: Incorrect Type Definition for TransactionJsonParsed
## Overview

I found a bug in the type definition of [TransactionJsonParsed](https://github.com/anza-xyz/solana-web3.js/blob/821b80fe2c282b62fa4ff85f52f17406df568731/packages/rpc-api/src/getTransaction.ts#L109). The `accountKeys` field is typed as a fixed-length tuple instead of an array, causing a TypeScript error when accessing or defining multiple keys.

## Steps to reproduce

1. Use the following code:

```
import { createSolanaRpc, signature } from '@solana/web3.js';

const rpcConnection = createSolanaRpc('https://api.mainnet-beta.solana.com');

async function getTransaction() {
  const transaction = await rpcConnection
    .getTransaction(
      signature(
        'EnU3TPTWnXxmPDjE6GAq8qwc1h734jMkkHKCu72VNQgLLLVhiDsugK2Cw66h39YLxrdcFzVctL6DB962vjirMJX',
      ),
      {
        commitment: 'confirmed',
        maxSupportedTransactionVersion: 0,
        encoding: 'jsonParsed',
      },
    )
    .send();

  return transaction;
}

getTransaction().then((res) => {
  const accKey = res?.transaction.message.accountKeys[1];

  console.log(accKey);
});

```

2. Observe this TypeScript error:

```
Tuple type '[{ pubkey: Address; signer: boolean; source: string; writable: boolean; }]' of length '1' has no element at index '1'.
```

## Description of bug

The `accountKeys` field is typed as:

```
accountKeys: [
    {
        pubkey: Address;
        signer: boolean;
        source: string;
        writable: boolean;
    },
];
```

This limits `accountKeys` to a single element and throws an error when accessing indexes beyond `[0]`. However, `accountKeys` should support multiple objects.

### Suggested Fix

Change the type to:

```
accountKeys: Array<{
    pubkey: Address;
    signer: boolean;
    source: string;
    writable: boolean;
}>;
```

mcintyre94:
#### Problem

`accountKeys` was incorrectly typed as a list containing one account key, instead of a list of arbitrary length of account keys

#### Summary of Changes

- Fix the type
- Change the type of `source` to the enum it comes from (https://github.com/anza-xyz/agave/blob/8889b4ac1ea39121dffcbc95421dbffae737c1fe/transaction-status-client-types/src/lib.rs#L134-L148)

Fixes #80

### DIFF
--- a/.changeset/red-melons-deny.md
+++ b/.changeset/red-melons-deny.md
@@ -1,0 +1,5 @@
+---
+'@solana/rpc-api': patch
+---
+
+Fix accountKeys type in getTransaction JSON parsed
--- a/packages/rpc-api/src/getTransaction.ts
+++ b/packages/rpc-api/src/getTransaction.ts
@@ -104,16 +104,16 @@ type ParsedTransactionInstruction = Readonly<{
     stackHeight?: number;
 }>;
 
+type ParsedAccount = Readonly<{
+    pubkey: Address;
+    signer: boolean;
+    source: 'lookupTable' | 'transaction';
+    writable: boolean;
+}>;
+
 type TransactionJsonParsed = Readonly<{
     message: {
-        accountKeys: [
-            {
-                pubkey: Address;
-                signer: boolean;
-                source: string;
-                writable: boolean;
-            },
-        ];
+        accountKeys: readonly ParsedAccount[];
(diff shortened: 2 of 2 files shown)


steveluscher (approved):
Added via Giphy

PR #95, opened 2025-02-03
fix: fixed Issue with lsof Command Syntax

annwag:
#### Problem

I’ve fixed the issue with the `lsof` command used to exclude the current process and validator process from the results.
The original usage of `^$BASHPID` and `^$validator_pid` was incorrect, as `lsof` doesn’t support this syntax.
Instead, I’ve updated the command to use the correct format with `-p` for process IDs and a more effective way to exclude these processes. The fix ensures the command now works as intended, and the processes are properly excluded from the `lsof` results.

### DIFF
--- a/scripts/start-shared-test-validator.sh
+++ b/scripts/start-shared-test-validator.sh
@@ -34,7 +34,7 @@ mkdir -p $LOCK_DIR
     fi
   ) 200>$EXCLUSIVE_LOCK_FILE
   validator_pid=$(pidof $TEST_VALIDATOR)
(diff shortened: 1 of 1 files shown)


steveluscher (changes requested):
Thanks for working on this!

What platform are you trying to fix for? I applied your change and it _broke_ Ubuntu.

```shell
~/src/solana-web3.js$ ./scripts/start-shared-test-validator.sh 
Started test validator (PID 2015896)
^CLeaving test validator (PID 2015896) running for other lock participants
```

Since that was the last locker of the validator, it was supposed to kill it, not leave it open for other lock participants.

PR #101, opened 2025-02-07
Export `fetchLookupTables` from `library` and add tests

### LINKED ISSUE #93: Can't import fetchLookupTables function
## Overview

**fetchLookupTables** function not exported in file [packages/library/src/decompile-transaction-message-fetching-lookup-tables.ts](https://github.com/anza-xyz/solana-web3.js/blob/f9c98f604a1b28aca8e5e08faf73c32ebb7cb2ca/packages/library/src/decompile-transaction-message-fetching-lookup-tables.ts#L20)

mcintyre94:
#### Problem

`fetchLookupTables` is a useful function that was used as part of decoding transaction messages, but not separately exported. It returns `AddressesByLookupTableAddress` which maps the input lookup tables to their list of addresses. This is useful outside of the context of decoding transaction messages, for example it's used as the input to `compressTransactionMessageUsingAddressLookupTables`.

#### Summary of Changes

- Extract `fetchLookupTables` into its own file
- Export it from `library`
- Add unit tests 

Fixes #93

### DIFF
--- a/packages/library/src/__tests__/fetch-lookup-tables-test.ts
+++ b/packages/library/src/__tests__/fetch-lookup-tables-test.ts
@@ -1,0 +1,137 @@
+import { Address } from '@solana/addresses';
+import {
+    SOLANA_ERROR__ACCOUNTS__EXPECTED_ALL_ACCOUNTS_TO_BE_DECODED,
+    SOLANA_ERROR__ACCOUNTS__ONE_OR_MORE_ACCOUNTS_NOT_FOUND,
+    SolanaError,
+} from '@solana/errors';
+import { GetMultipleAccountsApi, Rpc } from '@solana/rpc';
+
+import { fetchAddressesForLookupTables } from '../fetch-lookup-tables';
+
+describe('fetchAddressesForLookupTables', () => {
+    describe('when no lookup table addresses are provided', () => {
+        it('returns an empty object', async () => {
+            expect.assertions(1);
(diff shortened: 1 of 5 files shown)


packages/library/src/__tests__/fetch-lookup-tables-test.ts
```diff
@@ -0,0 +1,137 @@
...
+        });
+    });
+
+    describe('when lookup table addresses are provided', () => {
+        it('returns an object with the lookup table addresses as keys and the addresses as values', async () => {
+            expect.assertions(1);
+
+            const rpc: Rpc<GetMultipleAccountsApi> = {
+                getMultipleAccounts: jest.fn().mockReturnValue({
+                    send: jest.fn().mockResolvedValue({
```
steveluscher, line 42: If you _wanted_ to you could make this more resilient to the implementation of `getMultipleAccounts` by making it…

```ts
jest.fn().mockImplementation(addresses => {
    return Promise.resolve({
        send: jest.fn().mockImplementation(() => {
            const response = // actually prepare response for `addresses` in order supplied
            return Promise.resolve(response);
        }),
    });
});
```

…but that's way overkill and this is fine.

steveluscher (dismissed):
Sorry I took so long to get to this! Thanks for fixing it up.

Added via Giphy

steveluscher (approved):

PR #109, opened 2025-02-11
[fix] return signature for send and confirm factories

nickfrosty:
#### Problem

The `sendAndConfirmDurableNonceTransactionFactory` and `sendAndConfirmTransactionFactory` factories return a `Promise<void>` while the underlying `sendTransaction_INTERNAL_ONLY_DO_NOT_EXPORT` returns a `Promise<Signature>`

#### Summary of Changes

Fix the return type of both factories to return the signature that they already have due to just confirming it.

`sendTransaction_INTERNAL_ONLY_DO_NOT_EXPORT` already imports the `Signature` type so this does not add a new dep

### DIFF
--- a/.changeset/silver-pianos-dress.md
+++ b/.changeset/silver-pianos-dress.md
@@ -1,0 +1,5 @@
+---
+'@solana/web3.js': patch
+---
+
+return the transaction signature from the `sendAndConfirmTransactionFactory` and `sendAndConfirmDurableNonceTransactionFactory` since they already have the signature avaialble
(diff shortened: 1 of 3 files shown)


steveluscher (changes requested):
We did this _super_ on purpose to break the connection in people's minds between:

* a transaction being confirmed
* a signature for that transaction being available

You don't have to wait for a transaction to be confirmed to obtain the signature, because the signature is a function of the private key and the message. We encourage folks to obtain the signature early, as in the example here, so that they can make use of it in their app earlier.

https://github.com/anza-xyz/solana-web3.js/blob/1faa88c0a7a5c885cffc7ae60a2af69cdb84ccc1/examples/transfer-lamports/src/example.ts#L177-L180

steveluscher:
Because it makes _waiting_ for the signature ergonomic. We want to discourage waiting for signatures, and encourage computing signatures pre-send, because waiting for anything can make
 apps feel less responsive.

nickfrosty:
okay I guess just close the pr instead of requesting changes

PR #110, opened 2025-02-11
Add RPC-only transaction confirmation support

rroblak:
#### Problem

Transaction confirmation currently requires WebSocket connections, which may not be available or desired in all environments. And there's no documented way to use the transaction confirmation functionality with just RPC (HTTP) polling.

As the [great Rich Hickey said](https://github.com/matthiasn/talk-transcripts/blob/master/Hickey_Rich/ProblemSolving.md):

> So let's say I had a function -- one of those A, B, C -- that already exists and you're already calling, and I wanted to enhance it. Well, if I require more from you than I used to, I've broken you. If I require less from you than I used to, I haven't.

The 2.0 API shouldn't break use cases by requiring a Websocket where only having an RPC connection will suffice.

N.b. I'm not suggesting that RPC-only is technically superior. Just that it should be possible.

#### Summary of Changes

- Add support for RPC-only transaction confirmation through polling
- Make `rpcSubscriptions` parameter optional in `sendAndConfirmTransactionFactory`
- Add configurable polling interval option
- Add new example demonstrating RPC-only transaction confirmation
- Update documentation to reflect RPC-only support
- Add tests for polling-based confirmation strategies


examples/transfer-lamports/src/example-rpc-only.ts
```diff
@@ -0,0 +1,201 @@
+/* eslint-disable sort-keys-fix/sort-keys-fix */
+/**
+ * EXAMPLE
```
rroblak: I copied and minimally changed the existing `example.ts` to demonstrate the new RPC-only functionality.

packages/library/README.md
```diff
@@ -123,17 +123,23 @@ try {
 }
 ```
 
-### `sendAndConfirmTransactionFactory({rpc, rpcSubscriptions})`
+### `sendAndConfirmTransactionFactory({rpc, rpcSubscriptions?})`
 
-Returns a function that you can call to send a blockhash-based transaction to the network and to wait until it has been confirmed.
+Returns a function that you can call to send a blockhash-based transaction to the network and to wait until it has been confirmed. The `rpcSubscriptions` parameter is optional - when provided, it enables WebSocket-based transaction confirmation monitoring for better performance.
```
rroblak: Added this note for clarity.

packages/library/README.md
```diff
@@ -143,7 +149,7 @@ try {
 }
 ```
 
-### `sendTransactionWithoutConfirmingFactory({rpc, rpcSubscriptions})`
```
rroblak, line 146: Corrected this error (the method doesn't actually accept an `rpcSubscriptions` parameter).

packages/library/src/__typetests__/send-and-confirm-transaction-typetests.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
+/* eslint-disable sort-keys-fix/sort-keys-fix */
```
rroblak: Here and elsewhere, `pollingInterval` should be last in the parameter order. The linter disagrees, so I suppressed it as needed.

packages/transaction-confirmation/src/__tests__/confirmation-strategy-blockheight-test.ts
```diff
@@ -25,19 +27,22 @@ describe('createBlockHeightExceedencePromiseFactory', () => {
...
-        const rpc = {
-            getEpochInfo: getEpochInfoMock,
-        };
         getBlockHeightExceedencePromise = createBlockHeightExceedencePromiseFactory({
             rpc,
             rpcSubscriptions,
         });
     });
+    afterEach(() => {
+        jest.useRealTimers();
```
rroblak: Added this to avoid leaking fake timers to other tests.

packages/transaction-confirmation/src/confirmation-strategy-blockheight.ts
```diff
@@ -54,44 +62,79 @@ export function createBlockHeightExceedencePromiseFactory<
             };
         }
         try {
-            const [slotNotifications, { blockHeight: initialBlockHeight, differenceBetweenSlotHeightAndBlockHeight }] =
-                await Promise.all([
+            let currentBlockHeight: bigint;
+
+            if (rpcSubscriptions) {
```
steveluscher: What this does is to create dead code, no matter which way you use this confirmation strategy. If you use subscriptions, all of the polling code is dead. If you use polling, all of the subscriptions code is dead. You get the dead code in your JS bundle unconditionally, which reduces package efficiency.

In short, ‘modes’ are bad, because they create dead code and reduce package efficiency.

steveluscher (changes requested):
tl;dr web3.js 2.0 doesn't do ‘modes.’

`sendAndConfirmTransactionFactory()` is already a _super_ thin wrapper around `createBlockHeightExceedencePromiseFactory()`. I would totally consider including a _new_ confirmer, that describes your _new_ strategy, and implements your polling code.

```ts
const sendAndPollToConfirm = sendAndPollToConfirmTransactionFactory({
    interval: 1000,
    rpc,
});
```

packages/transaction-confirmation/src/confirmation-strategy-blockheight.ts
rroblak: Got it, that's fair. I'll extract these changes into a separate confirmer.

packages/library/src/send-and-poll-to-confirm-transaction.ts
```diff
@@ -0,0 +1,72 @@
...
+}: SendAndPollToConfirmTransactionWithBlockhashLifetimeFactoryConfig<'devnet'>): SendAndPollToConfirmTransactionWithBlockhashLifetimeFunction;
+export function sendAndPollToConfirmTransactionFactory({
+    rpc,
+    pollingInterval,
+}: SendAndPollToConfirmTransactionWithBlockhashLifetimeFactoryConfig<'testnet'>): SendAndPollToConfirmTransactionWithBlockhashLifetimeFunction;
+export function sendAndPollToConfirmTransactionFactory({
+    rpc,
+    pollingInterval,
+}: SendAndPollToConfirmTransactionWithBlockhashLifetimeFactoryConfig<'mainnet'>): SendAndPollToConfirmTransactionWithBlockhashLifetimeFunction;
+export function sendAndPollToConfirmTransactionFactory<
```
rroblak, line 38: The gist of this PR.

packages/transaction-confirmation/src/confirmation-strategy-blockheight.ts
rroblak: @steveluscher Please see the updated changeset.

packages/transaction-confirmation/src/confirmation-strategy-blockheight-polling.ts
```diff
@@ -0,0 +1,84 @@
...
+                currentBlockHeight = await getAbortablePromise(getBlockHeight(), abortController.signal);
+            }
+
+            callerAbortSignal.throwIfAborted();
+            throw new SolanaError(SOLANA_ERROR__BLOCK_HEIGHT_EXCEEDED, {
+                currentBlockHeight,
+                lastValidBlockHeight,
+            });
+        } finally {
+            callerAbortSignal.removeEventListener('abort', handleAbort);
```
steveluscher, line 80: It's not actually necessary to unhook this. When you add the event listener like you have on line 48:

```ts
callerAbortSignal.addEventListener('abort', handleAbort, {
    // This tells the event listener to remove itself when the signal fires.
    signal: abortController.signal,
});
```

Then simply calling `abort()` on the `abortController` is sufficient to unhook it.

packages/transaction-confirmation/src/confirmation-strategy-blockheight-polling.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        async function getBlockHeight() {
+            const { blockHeight } = await rpc
+                .getEpochInfo({ commitment })
+                .send({ abortSignal: abortController.signal });
+            return blockHeight;
+        }
+
+        try {
+            let currentBlockHeight = await getAbortablePromise(getBlockHeight(), abortController.signal);
```
steveluscher, line 58: I know this is a clone of `confirmation-strategy-blockheight.ts`, but none of it is actually necessary.

The reason that it was necessary to create an ‘inner’ abort controller in the existing confirmation strategies was because the _subscription_ would outlast the termination event. For instance, after the blockheight was exceeded, the subscription would still be alive, so you'd have to cancel it.

Your request-response implementation does not have this problem. At the moment you discover a termination event, you _by definition_ have no open requests in flight. There's nothing to abort. You can literally tear all of this code out, place the `callerAbortSignal` on the `rpc.getEpochInfo().send()` call, turn your timeout into an abortable promise, and call it a day. No `try/finally` required.

packages/transaction-confirmation/src/confirmation-strategy-blockheight-polling.ts
```diff
@@ -0,0 +1,84 @@
...
+        const abortController = new AbortController();
+        const handleAbort = () => {
+            abortController.abort();
+        };
+        callerAbortSignal.addEventListener('abort', handleAbort, { signal: abortController.signal });
+
+        async function getBlockHeight() {
+            const { blockHeight } = await rpc
+                .getEpochInfo({ commitment })
+                .send({ abortSignal: abortController.signal });
```
steveluscher, line 53: You need only use `getBlockHeight()` here.

steveluscher (changes requested):
Now that I've had a closer look at the actual implementation, a few things have jumped out that we would need to address in order to continue

1. The blockheight poller and the signature status poller have difference freshness tolerances and will demand different polling intervals. For example, you might want to fetch the signature status very frequently so that you can return control of the app to the user, but you might be quite happy to fetch the blockheight _much_ less frequently. After all, given a single measurement of the blockheight, you can predict with some accuracy when the blockheight will be exceeded (`(expiryHeight - currentHeight) * 400ms`). You could even imagine a poller that polled once, computed the expiry time, and then continued polling rapidly once the expiry time drew near.
2. You will definitely only want to create one blockheight poller, ever. Imagine an application that has 25 confirmations in flight; there's no sense asking for the blockheight repeatedly for each one since it's a global concept. You will want to reference count the confirmers; start a poller when the number of confirmers goes from zero to one, then stop the poller when the number of confirmers falls from one to zero. The singleton poller will (at least) have to use the tightest polling interval of all in-flight confirmations, but preferably _also_ be aware of the exceedence deadlines of all the confirmations, as described above.

All of this machinery is a massive pain in the ass, which is why we built the websocket-based implementation. It's both simpler and more economical. This is why we didn't offer a polling-based implementation in the first place.

Added via Giphy

rroblak:
@steveluscher 

> The blockheight poller and the signature status poller have different freshness tolerances and will demand different polling intervals....

Ok, I'll add separate parameters to allow for different polling intervals.

> ...You could even imagine a poller that polled once, computed the expiry time, and then continued polling rapidly once the expiry time drew near.

I'm going to leave that as an exercise for future contributors..

> You will definitely only want to create one blockheight poller, ever...

Good point, I'll add a singleton blockheight poller and integrate it.

> The singleton poller will (at least) have to use the tightest polling interval of all in-flight confirmations, but preferably also be aware of the exceedence deadlines of all the confirmations, as described above.

I'l just use the tightest polling interval for the time being. Future contributors can improve further.

As a meta-note, it'd be helpful if you can batch all of your feedback in as few comments as possible. Each time I have to circle back it takes significant time to re-familiarize myself with the context.

PR #60, opened 2025-01-13
Fix for key operations in Firefox

steveluscher:
#### Problem

As of version 133.0.3 Firefox will throw a `DOMException` on this native code:

```ts
const kp = await crypto.subtle.generateKey("Ed25519", true, ["sign", "verify"]);
const b = await crypto.subtle.exportKey("raw", kp.publicKey);
const pub = await crypto.subtle.importKey(
  "raw",
  b,
  "Ed25519",
  /* extractable */ true,
  ["verify"]
); // Uncaught (in promise) DOMException: An invalid or illegal string was specified
```

#### Summary of Changes

Call the key operation functions with the object form of the `AlgorithmIdentifier` which Firefox appears to accept.

PR #117, opened 2025-02-12
Add `getPublicKeyFromAddress` to derive public keys from addresses

beeman:
This commit introduces `getPublicKeyFromAddress`, which allows deriving a `CryptoKey` from a given Solana address. This is useful for verifying signatures when only the address is available.

#### Problem
There is no convenient way to derive a `CryptoKey` from an `Address`.

#### Summary of Changes
 - Adds `getPublicKeyFromAddress` to `public-key.ts`
 - Ensures the function properly asserts valid addresses before decoding
 - Updates tests to check valid and invalid address handling
 - Uses `crypto.subtle.importKey` to create an Ed25519 verification key

Fixes #

https://solana.stackexchange.com/questions/18940/how-to-convert-address-to-pubkey-in-solana-web3-js-v2

### DIFF
--- a/packages/addresses/src/public-key.ts
+++ b/packages/addresses/src/public-key.ts
@@ -1,7 +1,7 @@
 import { assertKeyExporterIsAvailable } from '@solana/assertions';
 import { SOLANA_ERROR__ADDRESSES__INVALID_ED25519_PUBLIC_KEY, SolanaError } from '@solana/errors';
 
-import { Address, assertIsAddress, getAddressDecoder, getAddressEncoder } from './address';
+import { Address, getAddressDecoder, getAddressEncoder } from './address';
 
 /**
  * Given a public {@link CryptoKey}, this method will return its associated {@link Address}.
@@ -32,9 +32,7 @@ export async function getAddressFromPublicKey(publicKey: CryptoKey): Promise<Add
  * const publicKey = await getPublicKeyFromAddress(address);
  * ```
  */
-export async function getPublicKeyFromAddress(address: Address, extractable: boolean = false) {
-    assertIsAddress(address);
+export async function getPublicKeyFromAddress(address: Address) {
     const addressBytes = getAddressEncoder().encode(address);
-
-    return await crypto.subtle.importKey('raw', addressBytes, { name: 'Ed25519' }, extractable, ['verify']);
+    return await crypto.subtle.importKey('raw', addressBytes, { name: 'Ed25519' }, true /* extractable */, ['verify']);
 }
--- a/packages/addresses/src/public-key.ts
+++ b/packages/addresses/src/public-key.ts
@@ -1,7 +1,7 @@
-import {assertKeyExporterIsAvailable} from '@solana/assertions';
-import {SOLANA_ERROR__ADDRESSES__INVALID_ED25519_PUBLIC_KEY, SolanaError} from '@solana/errors';
+import { assertKeyExporterIsAvailable } from '@solana/assertions';
+import { SOLANA_ERROR__ADDRESSES__INVALID_ED25519_PUBLIC_KEY, SolanaError } from '@solana/errors';
 
-import {Address, assertIsAddress, getAddressDecoder, getAddressEncoder} from './address';
+import { Address, assertIsAddress, getAddressDecoder, getAddressEncoder } from './address';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/addresses/src/public-key.ts
```diff
@@ -21,3 +21,20 @@ export async function getAddressFromPublicKey(publicKey: CryptoKey): Promise<Add
...
+ * import { getAddressFromPublicKey } from '@solana/addresses';
+ *
+ * const publicKey = await getPublicKeyFromAddress(address);
+ * ```
+ */
+export async function getPublicKeyFromAddress(address: Address, extractable: boolean = false) {
+    assertIsAddress(address);
+    const addressBytes = getAddressEncoder().encode(address);
+
+    return await crypto.subtle.importKey('raw', addressBytes, 'Ed25519', extractable, ['verify']);
```
steveluscher: This will break in Firefox. See #60.

steveluscher (changes requested):
I wonder if the fact that your key operations were so far away from the Firefox fix (ie. `ED25519_ALGORITHM_IDENTIFIER`) is an indication that these functions should _actually_ live in `@solana/keys`.

It's a bit late for that, but… maybe not too late?

cc/ @lorisleiva

packages/addresses/src/public-key.ts
beeman: Thanks, I updated this.

It looks like there's another one [here](https://github.com/anza-xyz/solana-web3.js/blob/main/packages/keys/src/public-key.ts#L27). If so, do you want me to create a PR for that or fix it in this one?

beeman:
> I wonder if the fact that your key operations were so far away from the Firefox fix (ie. `ED25519_ALGORITHM_IDENTIFIER`) is an indication that these functions should _actually_ live in `@solana/keys`.
> 
> It's a bit late for that, but… maybe not too late?
> 
> cc/ @lorisleiva

Happy to move it to `@solana/keys`, or do it in a separate PR. It does mean that `@solana/keys` will have to depend on `@solana/addresses`, which it currently doesn't.

packages/addresses/src/public-key.ts
steveluscher: Believe it or not, that one _works_ in Firefox.

steveluscher (approved):
I'll stamp this puppy, but wait for thoughts from @lorisleiva about the location of these functions.

lorisleiva:
Sorry I thought I responded to this PR. I'm happy with these functions to be in either locations. The fact that one of these locations would add a new dependency makes me think the current location is more appropriate.

packages/addresses/src/public-key.ts
```diff
@@ -21,3 +21,20 @@ export async function getAddressFromPublicKey(publicKey: CryptoKey): Promise<Add
...
+ * Given an {@link Address}, return a {@link CryptoKey} that can be used to verify signatures.
+ *
+ * @example
+ * ```ts
+ * import { getAddressFromPublicKey } from '@solana/addresses';
+ *
+ * const publicKey = await getPublicKeyFromAddress(address);
+ * ```
+ */
+export async function getPublicKeyFromAddress(address: Address, extractable: boolean = false) {
```
mcintyre94: I think it'd make sense to make this extractable by default, or (my preference) make it non-configurable and always extractable.

When we create a keypair we always make the public key extractable: https://github.com/anza-xyz/solana-web3.js/blob/997cfe0900e68151ebe79eeccb8fea3c026eac53/packages/keys/src/key-pair.ts#L30-L35

If it's not extractable then you can't get back to the `Address` and that's going to be annoying if you end up passing the `CryptoKey` around your app and need to do that later. I don't think there should ever be an issue with having an extractable public key.

packages/addresses/src/public-key.ts
```diff
@@ -21,3 +21,20 @@ export async function getAddressFromPublicKey(publicKey: CryptoKey): Promise<Add
...
+    assertIsAddress(address);
```
mcintyre94: Nit: we shouldn't need this because you're already correctly using `Address` as the parameter type.

steveluscher: Facts, but now you're making me think we need to assert that Ed25519 importing is supported (ie. we need to add something to `@solana/assertions` to do that)

PR #130, opened 2025-02-21
Handle code-less JSON RPC errors

lorisleiva:
#### Problem

I've had reports that some RPC errors do not include a `code` attribute in their response. For instance:

```json
{
    "id": "25",
    "jsonrpc": "2.0",
    "error": {
        "message": "No RPC endpoints available"
    }
}
```

Because our types strongly expect a `bigint | number` code to be provided in the response, the `getSolanaErrorFromJsonRpcError` function of the `@solana/errors` package doesn't gracefully handle this use-case and throws the following unhelpful error stack.

```
error: TypeError: Cannot read properties of undefined (reading 'length')
        at getHumanReadableErrorMessage (webpack-internal:///../../../node_modules/.pnpm/@solana+errors@2.0.0_typescript@5.7.2/node_modules/@solana/errors/dist/index.node.cjs:493:27)
        at getErrorMessage (webpack-internal:///../../../node_modules/.pnpm/@solana+errors@2.0.0_typescript@5.7.2/node_modules/@solana/errors/dist/index.node.cjs:550:12)
        at new SolanaError (webpack-internal:///../../../node_modules/.pnpm/@solana+errors@2.0.0_typescript@5.7.2/node_modules/@solana/errors/dist/index.node.cjs:580:21)
        at Object.getSolanaErrorFromJsonRpcError (webpack-internal:///../../../node_modules/.pnpm/@solana+errors@2.0.0_typescript@5.7.2/node_modules/@solana/errors/dist/index.node.cjs:817:11)
        at eval (webpack-internal:///../../../node_modules/.pnpm/@solana+rpc-transformers@2.0.0_fastestsmallesttextencoderdecoder@1.0.22_typescript@5.7.2/node_modules/@solana/rpc-transformers/dist/index.node.cjs:254:20)
        at eval (webpack-internal:///../../../node_modules/.pnpm/@solana+rpc-transformers@2.0.0_fastestsmallesttextencoderdecoder@1.0.22_typescript@5.7.2/node_modules/@solana/rpc-transformers/dist/index.node.cjs:267:54)
        at eval (webpack-internal:///../../../node_modules/.pnpm/@solana+functional@2.0.0_typescript@5.7.2/node_modules/@solana/functional/dist/index.node.cjs:5:34)
        at Array.reduce (<anonymous>)
        at Object.pipe (webpack-internal:///../../../node_modules/.pnpm/@solana+functional@2.0.0_typescript@5.7.2/node_modules/@solana/functional/dist/index.node.cjs:5:14)
        at Object.eval [as responseTransformer] (webpack-internal:///../../../node_modules/.pnpm/@solana+rpc-transformers@2.0.0_fastestsmallesttextencoderdecoder@1.0.22_typescript@5.7.2/node_modules/@solana/rpc-transformers/dist/index.node.cjs:265:23)
```

#### Summary of Changes

This PR fixes this by introducing a new `SOLANA_ERROR__UNRECOGNIZED_JSON_RPC_ERROR` error which gracefully tackles the case where a `code` isn't part of the RPC response.


lorisleiva:
* **#130** <a href="https://app.graphite.dev/github/pr/anza-xyz/solana-web3.js/130?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/solana-web3.js/130?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Jac0xb:
appreciate u!

packages/errors/src/__tests__/json-rpc-error-test.ts
```diff
@@ -40,6 +41,13 @@ describe('getSolanaErrorFromJsonRpcError', () => {
         const error = getSolanaErrorFromJsonRpcError({ code, message: 'o no' });
         expect(error).toHaveProperty('context.__code', 123);
     });
+    it('produces a `SOLANA_ERROR__UNRECOGNIZED_JSON_RPC_ERROR` when no code is given', () => {
```
Jac0xb, line 44: I think the actual conditional that invokes getSolanaErrorFromJsonRpcError looks like `"error" in data` and if it is it calls this.

packages/errors/src/json-rpc-error.ts
```diff
@@ -93,7 +94,7 @@ export interface RpcSimulateTransactionResult {
 
 export function getSolanaErrorFromJsonRpcError({ code: rawCode, data, message }: RpcErrorResponse): SolanaError {
     let out: SolanaError;
-    const code = Number(rawCode);
+    const code = rawCode !== undefined ? Number(rawCode) : SOLANA_ERROR__UNRECOGNIZED_JSON_RPC_ERROR;
```
Jac0xb: In the rare chance should we check that NaN / are there possibly custom error codes an RPC could use that the client error map won't have?

mcintyre94: We do have a default but it's only capturing the data: https://github.com/anza-xyz/solana-web3.js/blob/eba7948b8c8622b1c20e9a59ab7821aa5275c51a/packages/errors/src/json-rpc-error.ts#L131-L134

It might be useful to include code/message in that case too, since data is optional and those are both required (if the server is JSON RPC compliant). Probably makes sense as a separate issue/PR to this one though

packages/errors/src/json-rpc-error.ts
```diff
@@ -124,6 +125,9 @@ export function getSolanaErrorFromJsonRpcError({ code: rawCode, data, message }:
                 // messages in `@solana/errors` to be actual format strings.
                 errorContext = { __serverMessage: message };
                 break;
+            case SOLANA_ERROR__UNRECOGNIZED_JSON_RPC_ERROR:
+                errorContext = { data, message };
+                break;
```
mcintyre94: This makes sense, but I noticed in the default case we're quite defensive before setting `errorContext = data`: `if (typeof data === 'object' && !Array.isArray(data))`

I guess given you have `data: unknown` we probably don't need that?
Should we maybe just add a string check for message to make sure that type holds? Or we could loosen that type to `unknown` too. Not sure if it's worth worrying about though.

mcintyre94 (approved):

packages/errors/src/__tests__/json-rpc-error-test.ts
lorisleiva: Hmm I think that according to the specs if the JSON RPC returns a response whose payload does not include an `"error"` attribute, it should not be treated as an error.

Therefore, the only thing we need to handle is a fallback for an `"error"` object we do not recognise — i.e. no error code provided.

packages/errors/src/json-rpc-error.ts
lorisleiva: Yeah I guess if someone provides `{ code: "123" }` in the payload, this will be NaN so probably worth handling this case as a separate PR.

EDIT: Actually it might make more sense to just be more strict with the expected input and make `"123"` fallback to the `SOLANA_ERROR__UNRECOGNIZED_JSON_RPC_ERROR` error code. Which I can do in this PR.

packages/errors/src/json-rpc-error.ts
lorisleiva: Yeah I'm not sure if `message + data` is in the RPC spec but it seems to be the expectation we make on the error response at the moment.

I could just have `message: string` and `errorPayload: unknown` in the error context such that:
- `message` would be equal to `error.message` if it is a there and it's a valid string, otherwise we would fill it with something generic like: `"Unknown RPC error"`.
- `errorPayload` is just the whole `error` object with no assumptions whatsoever on what's inside.

packages/errors/src/__tests__/json-rpc-error-test.ts
```diff
@@ -20,6 +20,7 @@ import {
     SOLANA_ERROR__JSON_RPC__SERVER_ERROR_TRANSACTION_SIGNATURE_LEN_MISMATCH,
     SOLANA_ERROR__JSON_RPC__SERVER_ERROR_TRANSACTION_SIGNATURE_VERIFICATION_FAILURE,
     SOLANA_ERROR__JSON_RPC__SERVER_ERROR_UNSUPPORTED_TRANSACTION_VERSION,
+    SOLANA_ERROR__UNRECOGNIZED_JSON_RPC_ERROR,
```
steveluscher: _Mayyyyybe_ `_MALFORMED_` to really land the accusation that these JSON-RPC servers are non-conforming.

packages/errors/src/json-rpc-error.ts
```diff
@@ -22,7 +23,7 @@ import { safeCaptureStackTrace } from './stack-trace';
 import { getSolanaErrorFromTransactionError } from './transaction-error';
 
 interface RpcErrorResponse {
-    code: bigint | number;
+    code?: bigint | number;
```
steveluscher: I think we should probably leave the type alone (ie. it should conform to the JSON-RPC spec) but then jam a bunch of assertions into the `get*` method that produce a `_MALFORMED_` error when they fail.

steveluscher (approved):
Just to be completely clear for future readers, this is a patch to gracefully handle JSON-RPC that **do not comply with the spec**. https://www.jsonrpc.org/specification#error_object

mcintyre94 (approved):

PR #151, opened 2025-02-27
Upgrade `feePayer` to a signer when using `addSignersToTransactionMessage`

### LINKED ISSUE #82: `addSignersToTransactionMessage` should upgrade `feePayer` to a signer
## Motivation

Imagine a transaction message with a non-signer fee payer, having address 'abc'. Imagine a signer also having address 'abc'.

If you call `addSignersToTransactionMessage()` on that message, it should notice that the fee payer address matches the signer address, and should ‘upgrade’ the `feePayer` to a signer.

## Example use case

```ts
const message = pipe(
      m => setTransactionMessageFeePayer(signer.address, m), // non-signer version of this function
      m => addSignersToTransactionMessage([signer], m),
);
message.feePayer satisfies TransactionSigner<'abc'>;
```

lorisleiva:
This PR upgrades the `feePayer` attribute of a transaction message if and only if:
- The current `feePayer` attribute is set.
- The current `feePayer` attribute is not already a signer (i.e. address only).
- The current `feePayer` address matches one of the provided signers.

Note that the `addSignersToTransactionMessage` does not narrow the returned transaction message with a `ITransactionMessageWithFeePayerSigner` because this may or may not happen based on the provided signers. If desired, we could iterate over the explicit `TransactionSigner<TAddress>` array to figure out at compile time if we need to return a `ITransactionMessageWithFeePayerSigner` type. This felt like an overkill and over-complication of the types to me but happy to disagree and commit.

Fixes #82

### DIFF
--- a/packages/signers/src/__tests__/add-signers-test.ts
+++ b/packages/signers/src/__tests__/add-signers-test.ts
@@ -3,10 +3,11 @@ import '@solana/test-matchers/toBeFrozenObject';
 import { Address } from '@solana/addresses';
 import { SOLANA_ERROR__SIGNER__ADDRESS_CANNOT_HAVE_MULTIPLE_SIGNERS, SolanaError } from '@solana/errors';
 import { AccountRole, IInstruction } from '@solana/instructions';
-import { BaseTransactionMessage } from '@solana/transaction-messages';
+import { BaseTransactionMessage, ITransactionMessageWithFeePayer } from '@solana/transaction-messages';
 
 import { IAccountSignerMeta, IInstructionWithSigners } from '../account-signer-meta';
 import { addSignersToInstruction, addSignersToTransactionMessage } from '../add-signers';
+import { ITransactionMessageWithFeePayerSigner } from '../fee-payer-signer';
 import { createMockTransactionModifyingSigner, createMockTransactionPartialSigner } from './__setup__';
 
 describe('addSignersToInstruction', () => {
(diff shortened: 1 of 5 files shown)


lorisleiva:
* **#151** <a href="https://app.graphite.dev/github/pr/anza-xyz/solana-web3.js/151?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/solana-web3.js/151?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/signers/src/__tests__/add-signers-test.ts
```diff
@@ -190,6 +191,43 @@ describe('addSignersToTransactionMessage', () => {
         ]);
     });
 
+    it('add updates the fee payer if a matching signer is provided', () => {
```
mcintyre94: Nit: remove the `add` (also line 212)

```suggestion
    it('updates the fee payer if a matching signer is provided', () => {
```

packages/signers/src/add-signers.ts
```diff
@@ -106,12 +108,29 @@ export function addSignersToTransactionMessage<TTransactionMessage extends BaseT
...
+        ? signers.find(signer => signer.address === transactionMessage.feePayer.address)
+        : undefined;
+
+    if (!feePayerSigner && transactionMessage.instructions.length === 0) {
         return transactionMessage as ITransactionMessageWithSigners & TTransactionMessage;
     }
 
     return Object.freeze({
         ...transactionMessage,
+        ...(feePayerSigner ? { feePayer: feePayerSigner } : {}),
```
steveluscher: You can always use `null` in the latter case instead of an empty object.

steveluscher (approved):
🖊️

mcintyre94 (approved):
Nice, LGTM!

steveluscher (approved):

PR #160, opened 2025-02-28
Fix the script that pushes git tags after publish

steveluscher:
#### Problem

`pnpm pkg get version` no longer produces two-line output, so the `sed -n "2p"` step was producing no output.

#### Summary of Changes

- Remove `sed`
- Have the script listen to `steps.changesets.outputs.published` instead.

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -52,9 +52,9 @@ jobs:
           PUBLISH_TAG: latest
 
       - name: Push Git Tag
-        if: steps.changesets.outputs.hasChangesets == 'false'
+        if: steps.changesets.outputs.published == 'true'
         run: |
-          VERSION_TAG=v$(cd packages/kit/ && pnpm pkg get version | sed -n '2p' | grep -o '"\([^"]\)\+"$' | tr -d \")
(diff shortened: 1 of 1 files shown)


steveluscher:
* **#160** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/160?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/160?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/publish-packages.yml
```diff
@@ -52,9 +52,9 @@ jobs:
           PUBLISH_TAG: latest
 
       - name: Push Git Tag
-        if: steps.changesets.outputs.hasChangesets == 'false'
+        if: steps.changesets.outputs.published == 'true'
```
alannza, line 55: If you want to get rid of all the string manipulation and not just `sed`, you could simplify this to `pnpm pkg --json get version | jq -r`.

alannza (approved):
Makes sense to me.

.github/workflows/publish-packages.yml
steveluscher: Totally; I just wanted to avoid having to install `jq` on the GitHub runner.

PR #183, opened 2025-03-05
Document `@solana/kit` with TypeDoc

steveluscher:
Preview at https://spectacular-praline-bbcd37.netlify.app/

### DIFF
--- a/packages/kit/README.md
+++ b/packages/kit/README.md
@@ -15,7 +15,7 @@ This is the JavaScript SDK for building Solana apps for Node, web, and React Nat
 
 ## Functions
 
-In addition to reexporting functions from packages in the `@solana/*` namespace, this package offers additional helpers for building Solana applications, with sensible defaults.
+In addition to re-exporting functions from packages in the `@solana/*` namespace, this package offers additional helpers for building Solana applications, with sensible defaults.
 
 ### `airdropFactory({rpc, rpcSubscriptions})`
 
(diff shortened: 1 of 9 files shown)


steveluscher:
* **#183** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/183?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/183?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/kit/README.md
```diff
@@ -116,6 +116,36 @@ try {
 }
 ```
 
+### `sendAndConfirmDurableNonceTransactionFactory({rpc, rpcSubscriptions})`
```
steveluscher, line 113: Added this, since it was missing.

packages/kit/src/send-and-confirm-durable-nonce-transaction.ts
```diff
@@ -18,10 +18,45 @@ type SendAndConfirmDurableNonceTransactionFunction = (
...
     rpc: Rpc<GetAccountInfoApi & GetSignatureStatusesApi & SendTransactionApi> & { '~cluster'?: TCluster };
+    /** An object that supports the {@link AccountNotificationsApi} and the {@link SignatureNotificationsApi} of the Solana RPC Subscriptions API */
     rpcSubscriptions: RpcSubscriptions<AccountNotificationsApi & SignatureNotificationsApi> & { '~cluster'?: TCluster };
 };
 
+/**
+ * Returns a function that you can call to send a durable-nonce-based transaction to the network and
+ * to wait until it has been confirmed.
+ *
+ * @param config
```
steveluscher, line 31: These just make the param appear as `config` in the docs instead of `__namedParameters`.

packages/kit/README.md
lorisleiva: nit: probably best if the `sendAndConfirmDurableNonceTransactionFactory` is documented _after_ the `sendAndConfirmTransactionFactory` function on the README. It feels more natural to first read about the "default" send and confirm factory before learning there's also one for durable nonces.

packages/kit/src/compute-limit.ts
```diff
@@ -18,6 +19,62 @@ type ComputeUnitEstimateForTransactionMessageFunction = (
...
+ * });
+ *
+ * // Create your transaction message.
+ * const transactionMessage = pipe(
+ *     createTransactionMessage({ version: 'legacy' }),
+ *     /* ... *\/
+ * );
+ *
+ * // Request an estimate of the actual compute units this message will consume.
+ * const computeUnitsEstimate = await getComputeUnitEstimateForTransactionMessage(transactionMessage);
```
lorisleiva: Oh shit I didn't know we had this helper in the main library that's amazing!

Two little bits of information I was 100% clear on when reading the docblock and had to check the code to answer my own questions:

1. It's not super clear that we're estimating the CU by **simulating** the transaction. There is the word "simulation" in the warning note but it is as-if we already talked about that before which is not the case. I think adding a small second paragraph like "This is done by simulating the transaction and grabbing the estimated compute units from the result" could go a long way to make this clear.

2. My main feedback from this example was: "wait, if I don't add the `SetComputeUnitLimit` instruction to the transaction message before estimating, my estimated CU won't be accurate since it will be missing one instruction". However, looking at the code I can see that if and only if that instruction is missing, it will be added to the transaction message for us. I think this should be documented too.

packages/kit/README.md
steveluscher: Where my A-Z gang at?!

steveluscher: Oh my. The most amazing part of this is that this documentation wasn't actually missing. It was there the whole time.

Issue #187, opened 2025-03-06
Propose updated error message for signing error on durable nonce transaction with prepended instruction


MichaelAJay:
## Motivation
The error message is unclear when attempting to sign a transaction with a durable nonce lifetime & a prepended instruction.

Ref: https://solana.stackexchange.com/questions/20682/issue-with-prepending-compute-unit-limit-to-durable-nonce-transactions-in-solan/20688#20688

## Example use case
When someone prepends an instruction onto a transaction message with a durable nonce lifetime, the error message on `signTransactionMessageWithSigners` should be clearer.

## Details
Error message:
`TypeError: Cannot read properties of undefined (reading '0')`

Reproducible code (all set up until the very end at "// Target transaction"):
```
const {
    createSolanaRpc,
    createTransactionMessage,
    generateKeyPairSigner,
    setTransactionMessageFeePayerSigner,
    createKeyPairSignerFromPrivateKeyBytes,
    getBase58Encoder,
    setTransactionMessageLifetimeUsingBlockhash,
    appendTransactionMessageInstructions,
    signTransactionMessageWithSigners,
    sendAndConfirmTransactionFactory,
    createSolanaRpcSubscriptions,
    setTransactionMessageLifetimeUsingDurableNonce,
    prependTransactionMessageInstruction,
  } = require('@solana/web3.js');
  const { getSetComputeUnitLimitInstruction } = require('@solana-program/compute-budget');
  const { pipe } = require('@solana/functional');
  const {
    getCreateAccountInstruction,
    getInitializeNonceAccountInstruction,
    SYSTEM_PROGRAM_ADDRESS,
  } = require('@solana-program/system');
  
  (async () => {
    const rpc = createSolanaRpc('https://api.devnet.solana.com');
    const rpcSubscriptions = createSolanaRpcSubscriptions(
      'wss://api.devnet.solana.com'
    );
  
    const fromPrivateBytes = getBase58Encoder().encode(
      '<your-key-here>'
    );
    const fromKeypairSigner = await createKeyPairSignerFromPrivateKeyBytes(
      fromPrivateBytes
    );
    const nonceKeypairSigner = await generateKeyPairSigner();
    const toKeypairSigner = await generateKeyPairSigner();
  
    // Create nonce account
    const space = 80n;
    const lamportsForRent = await rpc
      .getMinimumBalanceForRentExemption(space)
      .send();
  
    const createAccountInstruction = getCreateAccountInstruction({
[...]

PR #193, opened 2025-03-07
Strip the TransactionMessageWithDurableNonceLifetime type when prepending instructions to a TransactionMessage

mcintyre94:
#### Problem

It's possible to create a `CompilableTransactionMessage` that actually is not. Take a valid durable nonce transaction message, and prepend any instruction with no accounts. This will remain a `CompilableTransactionMessage`, since prepending doesn't change the type, but when you attempt to compile it it'll throw an error trying to read accounts from the first instruction, which it expects to be a durable nonce advance instruction but is not.

See compile code: https://github.com/anza-xyz/kit/blob/0ffbfb63ac9898ce50a1d39da9ab360fde83b967/packages/transactions/src/compile-transaction.ts#L42-L52

#### Summary of Changes

While we could add error checking to that compile code, the root cause is that `prependTransactionMessageInstruction(s)` should never return a `TransactionMessageWithDurableNonceLifetime`. This PR modifies the types so that the `TransactionMessageWithDurableNonceLifetime` is stripped if present on the input type. 

This means that the returned transaction is no longer a `CompilableTransactionMessage` if it was previously a `TransactionMessageWithDurableNonceLifetime`, and you'll correctly get a type error if you attempt to compile it.

This does not fix the bad error message if you eg cast such an invalid message to `TransactionMessageWithDurableNonceLifetime`, so we might still decide to add runtime checks and errors. But since we've generally preferred type errors when possible, I think this is the best fix for now. 

Addresses #187

### DIFF
--- a/.changeset/rich-taxis-ring.md
+++ b/.changeset/rich-taxis-ring.md
@@ -1,0 +1,5 @@
+---
+'@solana/transaction-messages': patch
+---
+
+Strip the TransactionMessageWithDurableNonceLifetime type when prepending instructions to a TransactionMessage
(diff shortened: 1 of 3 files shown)


steveluscher (approved):
Added via Giphy

PR #240, opened 2025-03-17
Make example code in `Rpc*<...>` types runnable

steveluscher:
#### Problem

These are missing type parameters.

#### Summary of Changes

It's enough to match on `unknown`: https://codesandbox.io/p/sandbox/ydjvcr

### DIFF
--- a/packages/rpc/README.md
+++ b/packages/rpc/README.md
@@ -48,13 +48,13 @@ This is useful in cases where you need to make assertions about the suitability 
 ```ts
 async function getSpecialAccountInfo(
     address: Address<'ReAL1111111111111111111111111111'>,
-    rpc: RpcMainnet,
+    rpc: RpcMainnet<unknown>,
 ): Promise<SpecialAccountInfo>;
 async function getSpecialAccountInfo(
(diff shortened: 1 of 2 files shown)


packages/rpc/README.md
```diff
@@ -48,13 +48,13 @@ This is useful in cases where you need to make assertions about the suitability
 ```ts
 async function getSpecialAccountInfo(
     address: Address<'ReAL1111111111111111111111111111'>,
-    rpc: RpcMainnet,
+    rpc: RpcMainnet<unknown>,
```
lorisleiva, line 51: Would it not be best to use a generic type `T` here?

mcintyre94 (approved):

steveluscher:
### Merge activity

* **Mar 19, 1:59 PM PDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/240).

packages/rpc/README.md
steveluscher: `unknown` matches any more specific type, and a type parameter `T` would go unused in the return type, so it would sort of all be for naught.

PR #318, opened 2025-03-31
Harmonize program filter type between RPC and RPC Subscriptions

steveluscher:
#### Problem

The RPC API specifies a `Base58EncodedData | Base64EncodedData` where the RPC Subscriptions API specifies only `string`. Granted the more restrictive type is less ergonomic, but the friction it adds can prevent – at build time – someone from mixing up their encodings.

#### Summary of Changes

Use the common type definition for both.

### DIFF
--- a/packages/rpc-api/src/__tests__/get-program-accounts-test.ts
+++ b/packages/rpc-api/src/__tests__/get-program-accounts-test.ts
@@ -4,7 +4,7 @@ import type { Address } from '@solana/addresses';
 import { getBase58Decoder } from '@solana/codecs-strings';
 import { SOLANA_ERROR__JSON_RPC__SERVER_ERROR_MIN_CONTEXT_SLOT_NOT_REACHED, SolanaError } from '@solana/errors';
 import type { Rpc } from '@solana/rpc-spec';
-import type { Commitment } from '@solana/rpc-types';
+import type { Base58EncodedBytes, Commitment } from '@solana/rpc-types';
 import path from 'path';
 
 import { GetProgramAccountsApi } from '../index';
@@ -2339,7 +2339,7 @@ describe('getProgramAccounts', () => {
                     filters: [
                         {
                             memcmp: {
-                                bytes: mint,
+                                bytes: mint as unknown as Base58EncodedBytes,
                                 encoding: 'base58',
                                 offset: 0n,
                             },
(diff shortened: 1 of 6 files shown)


packages/rpc-api/src/__tests__/get-program-accounts-test.ts
```diff
@@ -2339,7 +2339,7 @@ describe('getProgramAccounts', () => {
                     filters: [
                         {
                             memcmp: {
-                                bytes: mint,
+                                bytes: mint as unknown as Base58EncodedBytes,
```
steveluscher, line 2342: Case in point.

packages/rpc-types/src/account-filters.ts
```diff
@@ -1,26 +1,43 @@
...
+    encoding: 'base58';
+    /** The byte offset into the account data from which to start the comparison */
+    offset: bigint;
+}>;
+
+type ProgramNotificationsMemcmpFilterBase64 = Readonly<{
+    /**
+     * The bytes to match, as a base-64 encoded string.
+     *
+     * Data is limited to a maximum of 128 decoded bytes.
```
steveluscher, line 27: Intellisense will now change the comment on this field depending on what you supplied for `encoding`

packages/rpc-api/src/__tests__/get-program-accounts-test.ts
mcintyre94: I suspect we shouldn't do this because of package dependencies, but it would kinda make sense for `Address` to be a branded `Base58EncodedBytes` instead of string. Then I don't think you'd need the cast to `unknown` here

packages/rpc-types/src/account-filters.ts
mcintyre94: Is base64 limited to 128 bytes as well here? Usually that's just base58 AFAIK

mcintyre94 (approved):
Nice, this is way cleaner!

packages/rpc-types/src/account-filters.ts
steveluscher: It is! https://github.com/steveluscher/agave/blob/dc03d30163c39a235216244f1dcc17db5aab769e/rpc-client-types/src/filter.rs#L45-L49

packages/rpc-api/src/__tests__/get-program-accounts-test.ts
steveluscher: Oooooh, we should. I don't even think this would break anything more than usual, since if you have two different definitions of the _current_ `Address` type in the same project you're already in trouble.

PR #409, opened 2025-04-24
Loosen lifetime constraint on sendAndConfirmTransaction to only require lastValidBlockHeight

mcintyre94:
#### Problem

If you receive a serialized blockhash transaction and want to send it using the Kit `sendAndConfirmTransaction` then you currently need to produce a `TransactionWithBlockhashLifetime` type to attach to the transaction, ie `{blockhash, lastValidBlockHeight}`. But the confirm logic only cares about the `lastValidBlockHeight`, so the blockhash part is unnecessary.

This encourages decompiling the transaction message unnecessarily to obtain the blockhash, which is ignored but required to satisfy the type.

#### Summary of Changes

You can now send and confirm a fully signed transaction with this shape:

```ts
{
  messageBytes: TransactionMessageBytes;
  signatures: SignaturesMap;
  lifetimeConstraint: {
    lastValidBlockHeight: Slot
  }
} 
```

In the case where you're using Kit to sign and we return `FullySignedTransaction & TransactionWithBlockhashLifetime` this is still satisfied.

But in the case where you have a deserialized Transaction `{messageBytes, signatures}` you now only need to attach a `lastValidBlockHeight`. As this is unavailable on the transaction message, this makes it clear that you don't need to parse the transaction in any way to confirm it, and nor should you for instance make an RPC call to fetch a new blockhash.

### DIFF
--- a/.changeset/nine-weeks-shop.md
+++ b/.changeset/nine-weeks-shop.md
@@ -1,0 +1,6 @@
+---
+'@solana/transaction-confirmation': patch
+'@solana/kit': patch
+---
+
+Loosen lifetime constraint on sendAndConfirmTransaction to only require lastValidBlockHeight
--- a/packages/kit/src/__tests__/send-transaction-internal-test.ts
+++ b/packages/kit/src/__tests__/send-transaction-internal-test.ts
@@ -11,7 +11,7 @@ import {
 
 import {
     sendAndConfirmDurableNonceTransaction_INTERNAL_ONLY_DO_NOT_EXPORT,
-    sendAndConfirmTransactionWithBlockhashLifetime_INTERNAL_ONLY_DO_NOT_EXPORT,
+    sendAndConfirmTransactionWithLastValidBlockHeight_INTERNAL_ONLY_DO_NOT_EXPORT,
 } from '../send-transaction-internal';
 
 jest.mock('@solana/transactions');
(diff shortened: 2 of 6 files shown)


lorisleiva (dismissed):
I think that makes sense!

steveluscher (dismissed):
Makes sense, but if it were me I'd leave everything named the way it is now (it's still semantically a blockhash-based lifetime) and just surgically loosen the type of the transaction to `Omit<TransactionWithBlockhashLifetime, 'blockhash'>` wherever needed.

steveluscher:
Yeah, I realized that right after ([playground](https://www.typescriptlang.org/play/?pretty=false#code/C4TwDgpgBAKgTgQwHYGcEGNgEsD2SDqWwAFgEIA2O6A1sQisQDJYBmE2AttALxQDeAKChRyrdli4BhPCmCIsSYAC5+Q4VABGlGnQYrZcBQHMANGuHl6wAGoJRAEwpVqACQhYjxZZo8LgZ4QBfAWCBUEgoAFU0LQgYYmMoXgB5DiIAHnhkNExcAiIybVp6JjFOCBMoAHJRNnLpVDkEPyqAPigAMlULMokIBoNmxRVUjKzUDGw8QhInHRLmOr6AbRreqRkmloBdSqqtZ11iNpCBAXRNqBYcHCSoAAokAFdycih6KCekaiQcAHckO8UFEYuQ4gkkEYAJQAOlq4g2jXkigA3AIAPTo9RQAB6AH4BEA)).

steveluscher:
I'd still keep the method name the same (eg. `sendAndConfirmTransactionWithBlockhashLifetime_INTERNAL_ONLY_DO_NOT_EXPORT` et al.)

mcintyre94:
Makes sense. I've updated the type to use `Omit` like in your playground example, and renamed the functions back. Less changes overall which is nice!

steveluscher (approved):

PR #425, opened 2025-05-01
Add transaction size helpers

lorisleiva:
This PR adds various types and functions to help get the size of a transaction (or its message) and ensure its size is valid (under the transaction size limit).

Additionally, it provides two new types — `TransactionUnderSizeLimit` and `TransactionMessageUnderSizeLimit` — to provide additional type safety before sending a transaction.

In this PR, `TransactionUnderSizeLimit` and `TransactionMessageUnderSizeLimit` are only used by their respective type guards but this stack of PR aims to use these types to offer more type safety in other parts of the code. For instance: adding an instruction to or changing the lifetime of a transaction message should remove the `TransactionMessageUnderSizeLimit` type if any; sending a transaction should require the `TransactionUnderSizeLimit`; compiling a `TransactionMessageUnderSizeLimit` into a transaction should return a `TransactionUnderSizeLimit`; etc.

The next few PRs don't do any of that just yet but prepare the field for such changes. Mainly, it aims to bring more consistency amongst the type flags of transactions and transaction messages so adding yet another one does not affect the developer experience.


lorisleiva:
* **#430** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/430?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#429** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/429?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#428** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/428?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#427** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/427?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#426** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/426?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#425** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/425?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/425?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/errors/src/context.ts
```diff
@@ -575,6 +576,10 @@ export type SolanaErrorContext = DefaultUnspecifiedErrorContextToUndefined<
         [SOLANA_ERROR__TRANSACTION__ADDRESS_MISSING]: {
             index: number;
         };
+        [SOLANA_ERROR__TRANSACTION__EXCEEDS_SIZE_LIMIT]: {
+            transactionSize: number;
+            transactionSizeLimit: number;
```
mcintyre94: Nit: Maybe we could put `Bytes` on these field names?

packages/errors/src/messages.ts
```diff
@@ -578,6 +579,8 @@ export const SolanaErrorMessages: Readonly<{
     [SOLANA_ERROR__TRANSACTION__ADDRESS_MISSING]: 'Transaction is missing an address at index: $index.',
     [SOLANA_ERROR__TRANSACTION__CANNOT_ENCODE_WITH_EMPTY_SIGNATURES]:
         'Transaction has no expected signers therefore it cannot be encoded',
+    [SOLANA_ERROR__TRANSACTION__EXCEEDS_SIZE_LIMIT]:
+        'Transaction size $transactionSize exceeds limit of $transactionSizeLimit',
```
mcintyre94:
```suggestion
        'Transaction size $transactionSize exceeds limit of $transactionSizeLimit bytes',
```

packages/errors/src/context.ts
lorisleiva: Good shout! How do you feel about a `number` type alias called `Bytes` on this file so TypeScript can provide that extra information? Or do you still think a more verbose name would be better for non-TypeScript users?

mcintyre94: That sounds good to me. A type alias is probably more helpful for devs, and if we add bytes to the message itself then we don't need it on the fields too IMO - unlikely they'll be displayed anywhere without either types or that message.

packages/transactions/src/transaction-size.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @example
+ * ```ts
+ * if (isValidTransactionSize(transaction)) {
+ *    transaction satisfies TransactionUnderSizeLimit;
+ * }
+ * ```
+ */
+export function isValidTransactionSize<TTransaction extends Transaction>(
+    transaction: TTransaction,
+): transaction is TransactionUnderSizeLimit & TTransaction {
```
steveluscher: Technically not ‘under’ because of that `<=`

steveluscher (approved):
> For instance: adding an instruction to or changing the lifetime of a transaction message should remove the `TransactionMessageUnderSizeLimit` type

I think this is only true when you're going from not-nonce to nonce. Otherwise the size of the transaction should stay the same or decrease.

| Lifetime before | Lifetime after | Should remove `TransactionMessageUnderSizeLimit`|
|---|---|---|
| None | Blockhash | ❌ |
| None | Nonce | ✅ |
| Blockhash | Nonce | ✅ |
| Blockhash | Blockhash | ❌ |
| Blockhash | None | ❌ |
| Nonce | Nonce | ❌ |
| Nonce | Blockhash | ❌ |
| Nonce | None | ❌ |

packages/transactions/src/transaction-size.ts
lorisleiva: `TransactionUnderSizeLimitPlusOne` haha. Any suggestions for something more accurate? Maybe `TransactionWithinSizeLimit`?

steveluscher: Within would work. Or `TransactionOfAcceptableSize` or something like that with ‘of.’

steveluscher:
I don't think so! Essentially all blockhash-based transactions will be the same size, even one that has ‘no’ blockhash, because all that changes is the choice of 32 lifetime specifier bytes in the message, but never the number of bytes.

If a nonce transaction used to have an acceptable size, it will always have an acceptable size if it changes to a blockhash-based one, because the only thing that could have happened is that the `AdvanceNonce` instruction might have been deleted, or not.

If a nonce transaction used to have an acceptable size, it will still have an acceptable size given a new choice of nonce; while the choice of lifetime specifier and `AdvanceNonce` instruction bytes changes, the number of those bytes will stay the same.

lorisleiva:
Oh so sorry, my brain could categorically not see the difference between "Nonce" and "None" on that table. Makes a lot more sense now.

packages/transactions/src/transaction-size.ts
lorisleiva: I have a slight preference for within since `isTransactionWithinSizeLimit` reads quite well IMO but happy to go with something else. ☺️

lorisleiva:
### Merge activity

* **May 14, 5:56 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/425).

PR #429, opened 2025-05-01
Simplify `isTransactionMessageWithBlockhashLifetime` implementation

lorisleiva:
#### Summary of Changes

Fixes #

PR #428, opened 2025-05-01
Refactor transaction message type tests

lorisleiva:
This PR mainly splits the type tests of the `transaction-message-typetests.ts` files into more granular files — some already existing.

Additionally it extracts a `durable-nonce-instruction.ts` file from `durable-nonce.ts`.

PR #427, opened 2025-05-01
Rename DurableNonceLifetime helpers

lorisleiva:
#### Context

This library uses the following naming convention for types and helpers:

```ts
type MyType;
function isMyType;
function assertIsMyType;
```

For instance, the `TransactionMessageWithBlockhashLifetime` follows this convention:

```ts
type TransactionMessageWithBlockhashLifetime;
function isTransactionMessageWithBlockhashLifetime;
function assertIsTransactionMessageWithBlockhashLifetime;
```

#### Problem

The durable nonce helpers do not follow this convention:

```ts
type TransactionMessageWithDurableNonceLifetime;
function isDurableNonceTransaction;
function assertIsDurableNonceTransactionMessage;
```

#### Summary of Changes

Rename the durable nonce helpers so they follow the same naming convention.

```ts
type TransactionMessageWithDurableNonceLifetime;
function isTransactionMessageWithDurableNonceLifetime;
function assertIsTransactionMessageWithDurableNonceLifetime;
```

PR #426, opened 2025-05-01
Remove I prefixes before Transaction types

lorisleiva:
Most of the types in the `transactions` and `transaction-messages` are not prefixed by the letter `I` apart from four of them. This PR makes them more consistent with the rest of the library by removing that prefix.

PR #430, opened 2025-05-01
Add transaction message Exclude type helpers

lorisleiva:
This PR introduce a `TransactionMessageWithLifetime` type that describes a transaction message with any lifetime set.

It also introduce `ExcludeTransactionMessageX<TTransactionMessage>` types for most `TransactionMessageWithX` types. This means we can simplify the return type of most of our setters by `ExcludeTransactionMessageX<T> & TransactionMessageWithX` to ensure the old state is not kept.

For instance, see how this PR makes the `setTransactionMessageLifetimeUsingBlockhash` function use a single function signature. This will be useful when we'll need to chain more type exclusions to setters such as `ExcludeTransactionMessageUnderSizeLimit`.

### DIFF
--- a/packages/transaction-messages/src/durable-nonce.ts
+++ b/packages/transaction-messages/src/durable-nonce.ts
@@ -8,6 +8,7 @@ import {
     createAdvanceNonceAccountInstruction,
     isAdvanceNonceAccountInstruction,
 } from './durable-nonce-instruction';
+import { ExcludeTransactionMessageLifetime } from './lifetime';
 import { BaseTransactionMessage } from './transaction-message';
 
 type DurableNonceConfig<
@@ -59,6 +60,14 @@ export interface TransactionMessageWithDurableNonceLifetime<
     ];
     readonly lifetimeConstraint: NonceLifetimeConstraint<TNonceValue>;
 }
+
+/**
+ * A helper type to exclude the durable nonce lifetime constraint from a transaction message.
+ */
+export type ExcludeTransactionMessageDurableNonceLifetime<TTransactionMessage extends BaseTransactionMessage> =
+    TTransactionMessage extends TransactionMessageWithDurableNonceLifetime
+        ? ExcludeTransactionMessageLifetime<TTransactionMessage>
+        : TTransactionMessage;
 
 /**
  * A type guard that returns `true` if the transaction message conforms to the
--- a/.changeset/tall-boxes-kick.md
+++ b/.changeset/tall-boxes-kick.md
@@ -1,0 +1,5 @@
+---
+'@solana/transaction-messages': patch
+---
+
+Introduce a new `TransactionMessageWithLifetime` type and special Exclude type helpers to remove information from transaction messages
--- a/packages/kit/src/compute-limit-internal.ts
+++ b/packages/kit/src/compute-limit-internal.ts
@@ -143,7 +143,7 @@ export async function getComputeUnitEstimateForTransactionMessage_INTERNAL_ONLY_
      *         - either a recent blockhash lifetime or a nonce.
      */
     const isDurableNonceTransactionMessage = isTransactionMessageWithDurableNonceLifetime(transactionMessage);
-    let compilableTransactionMessage;
+    let compilableTransactionMessage: CompilableTransactionMessage;
     if (isDurableNonceTransactionMessage || isTransactionMessageWithBlockhashLifetime(transactionMessage)) {
         compilableTransactionMessage = transactionMessage;
     } else {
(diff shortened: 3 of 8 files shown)


lorisleiva:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/430?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#430** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/430?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/430?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#429** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/429?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#428** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/428?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#427** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/427?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#426** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/426?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#425** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/425?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/transaction-messages/src/durable-nonce.ts
```diff
@@ -59,6 +59,15 @@ export interface TransactionMessageWithDurableNonceLifetime<
     readonly lifetimeConstraint: NonceLifetimeConstraint<TNonceValue>;
 }
 
+/**
+ * A helper type to exclude the durable nonce lifetime constraint from a transaction message.
+ */
+export type ExcludeTransactionMessageDurableNonceLifetime<TTransactionMessage extends BaseTransactionMessage> =
+    TTransactionMessage extends TransactionMessageWithDurableNonceLifetime
+        ? Omit<TTransactionMessage, 'instructions' | 'lifetimeConstraint'> &
+              Pick<BaseTransactionMessage, 'instructions'>
```
mcintyre94: If I understand this correctly, it's stripping type knowledge about the instructions? I'm guessing that's to do with stripping the durable nonce instruction type, but I think it's a bit overkill. There are plenty of other things you might want type-awareness of in instructions and I don't think we should strip all of them if we remove the lifetime type. Or is this unavoidable because we're prepending an unknown (at the type level) number of instructions to the start of the array?

mcintyre94 (commented):
This looks good! Just a question about the durable nonce exclude type.

mcintyre94 (approved):

packages/transaction-messages/src/durable-nonce.ts
lorisleiva: I thought about this too and it is technically possible to keep this information. However, currently, all the instruction helpers — both appending and prepending — don't actually update the `TTransactionMessage` to include the new set of instructions. I think it is possible — by concatenating arrays in TypeScript — but it's a fairly big refactoring so perhaps a suggestion for another PR.

Right now though, if I don't remove that information, the prepend/append instruction helpers won't help remove it. So this would cause the prepend helper not to adjust the fact that the durable nonce instruction is no longer the first one.

But maybe we just need a separate type helper that clears instruction info just for these functions?

mcintyre94: Hmm yep I think it would make sense to have a separate type helper for resetting the instructions type like this. Then if we do try to make it more clever in the future (don't think we need to/likely should do it now) it'll be a bit clearer what's going on I think

Also append should maybe be resetting instructions, but shouldn't be removing the durable nonce lifetime -so again I think makes sense to separate them

lorisleiva: Sounds good! I'll make the updates asap. 🙂

lorisleiva: Let me know what you think of the changes. 🙏

mcintyre94: Nice, LGTM now!

mcintyre94 (approved):

steveluscher (approved):

lorisleiva:
### Merge activity

* **May 14, 5:56 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/430).

PR #432, opened 2025-05-02
Improve type safety when adding instructions

lorisleiva:
#### Problem

When appending or prepending instructions to existing transaction messages, we simply return the provided `TTransactionMessage` as-is without including the types for the newly added instructions.

#### Summary of Changes

Adjust the return type of appending and prepending functions such that they concatenate instruction types.

### DIFF
--- a/packages/transaction-messages/src/decompile-message.ts
+++ b/packages/transaction-messages/src/decompile-message.ts
@@ -18,9 +18,9 @@ import type { getCompiledAddressTableLookups } from './compile/address-table-loo
 import { createTransactionMessage } from './create-transaction-message';
 import { Nonce, setTransactionMessageLifetimeUsingDurableNonce } from './durable-nonce';
 import { isAdvanceNonceAccountInstruction } from './durable-nonce-instruction';
-import { setTransactionMessageFeePayer } from './fee-payer';
+import { setTransactionMessageFeePayer, TransactionMessageWithFeePayer } from './fee-payer';
 import { appendTransactionMessageInstruction } from './instructions';
-import { TransactionVersion } from './transaction-message';
+import { BaseTransactionMessage, TransactionVersion } from './transaction-message';
 
 function getAccountMetas(message: CompiledTransactionMessage): IAccountMeta[] {
     const { header } = message;
@@ -249,9 +249,10 @@ export function decompileTransactionMessage(
         createTransactionMessage({ version: compiledTransactionMessage.version as TransactionVersion }),
         m => setTransactionMessageFeePayer(feePayer, m),
         m =>
-            instructions.reduce((acc, instruction) => {
-                return appendTransactionMessageInstruction(instruction, acc);
-            }, m),
+            instructions.reduce(
+                (acc, instruction) => appendTransactionMessageInstruction(instruction, acc),
+                m as BaseTransactionMessage & TransactionMessageWithFeePayer,
+            ),
         m =>
             'blockhash' in lifetimeConstraint
                 ? setTransactionMessageLifetimeUsingBlockhash(lifetimeConstraint, m)
--- a/packages/transaction-messages/src/durable-nonce.ts
+++ b/packages/transaction-messages/src/durable-nonce.ts
@@ -197,9 +197,20 @@ export function setTransactionMessageLifetimeUsingDurableNonce<
         nonceAccountAddress,
         nonceAuthorityAddress,
     }: DurableNonceConfig<TNonceAccountAddress, TNonceAuthorityAddress, TNonceValue>,
-    transactionMessage: TTransactionMessage | (TransactionMessageWithDurableNonceLifetime & TTransactionMessage),
-): TransactionMessageWithDurableNonceLifetime<TNonceAccountAddress, TNonceAuthorityAddress, TNonceValue> &
-    TTransactionMessage {
+    transactionMessage: TTransactionMessage,
+): SetTransactionMessageWithDurableNonceLifetime<
+    TTransactionMessage,
+    TNonceAccountAddress,
+    TNonceAuthorityAddress,
+    TNonceValue
+> {
+    type ReturnType = SetTransactionMessageWithDurableNonceLifetime<
+        TTransactionMessage,
+        TNonceAccountAddress,
+        TNonceAuthorityAddress,
+        TNonceValue
+    >;
+
     let newInstructions: [
         AdvanceNonceAccountInstruction<TNonceAccountAddress, TNonceAuthorityAddress>,
         ...IInstruction[],
@@ -212,12 +223,7 @@ export function setTransactionMessageLifetimeUsingDurableNonce<
                 isTransactionMessageWithDurableNonceLifetime(transactionMessage) &&
                 transactionMessage.lifetimeConstraint.nonce === nonce
             ) {
-                return transactionMessage as TransactionMessageWithDurableNonceLifetime<
-                    TNonceAccountAddress,
-                    TNonceAuthorityAddress,
-                    TNonceValue
-                > &
-                    TTransactionMessage;
+                return transactionMessage as unknown as ReturnType;
             } else {
                 // we already have the right first instruction, leave it as-is
                 newInstructions = [firstInstruction, ...transactionMessage.instructions.slice(1)];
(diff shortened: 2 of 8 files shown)


packages/transaction-messages/src/__typetests__/instructions-typetests.ts
```diff
@@ -18,6 +21,24 @@ type IInstruction = BaseTransactionMessage['instructions'][number];
...
+        newMessage.instructions satisfies readonly [InstructionB, InstructionA];
+        // @ts-expect-error Not readonly.
+        newMessage.instructions satisfies [InstructionA, InstructionB];
+    }
+
+    // It adds instruction types to base transaction messages
+    {
+        const message = null as unknown as BaseTransactionMessage;
+        const newMessage = appendTransactionMessageInstruction(null as unknown as InstructionA, message);
+        newMessage.instructions satisfies readonly [...IInstruction[], InstructionA];
```
lorisleiva: Just wanted to flag this interesting test result. Since `BaseTransactionMessage` already contains a generic instruction array as `IInstruction[]`, we get results like `[...IInstruction[], InstructionA]` when we append. I think this is probably the expected behaviour anyway but there could be a case for having an empty array on the `BaseTransactionMessage` since it starts with no instructions? Not a trivial change though.

mcintyre94 (approved):
Nice, this looks great!

lorisleiva:
Okay we have a bit of an issue because of that `BaseTransactionMessage` type containing a too generic starting point for the `instructions` attribute.

Consider the following TS example:

```ts
type T1 = { readonly _: unique symbol };
type T2 = { readonly _: unique symbol };
type T3 = { readonly _: unique symbol };

type A1 = [T1, T2];
type A2 = [...A1, T3];
null as unknown as A2 satisfies [T1, T2, T3];
```

Since `A1` is clearly defined as `[T1, T2]`, we get the expected result and everyone is happy.

Now imagine A1 was composed of a more generic type first (`TX`) and augmented with an additional `[T1, T2]` type.

```ts
type T1 = { readonly _: unique symbol };
type T2 = { readonly _: unique symbol };
type T3 = { readonly _: unique symbol };
type TX = T1 | T2 | T3; // <- New generic type as the "base" type.

type A1 = TX[] & [T1, T2]; // A1 is the "base" type with an additional type restriction.
type A2 = [...A1, T3];
```

Now suddenly, we get the following result:

```ts
null as unknown as A2 satisfies [T1, T2, T3]; // ❌ Fails
null as unknown as A2 satisfies [...(T1 | T2)[], T3]; // ✅ Succeeds.
```

This is because, when destructuring `A1`, we loose the order with `T1` and `T2` and get `(T1 | T2)[]` instead.

This exact behaviour is happening when dealing with transaction messages that start with a `BaseTransactionMessage` (as they should) which provides a default generic definition of these instructions (i.e. `IInstruction[]`) and _then_ augment their type with a `TransactionMessageWithDurableNonceLifetime` type which adds additional type restrictions on the `instructions` property.

Aside from changing the `BaseTransactionMessage` type so it does not provide a generic type for instructions (which I wouldn't recommend we do), I'm not sure how we can get around that.

Currently, this problem only manifests when we are trying to append instructions to a transaction message already marked as a durable nonce transaction. Because we loose order information of the existing instructions, TypeScript can no longer assert that the durable nonce instruction is the very first one.

I think this niche use-case is probably still an okay price to pay to have better type safety when adding instructions to messages but I'll see if I can find a better alternative.

packages/transaction-messages/src/instructions.ts
```diff
@@ -1,14 +1,29 @@
+import { IInstruction } from '@solana/instructions';
+
 import { ExcludeTransactionMessageDurableNonceLifetime } from './durable-nonce';
 import { BaseTransactionMessage } from './transaction-message';
 
 /**
- * A helper type to reset the instructions of a transaction message to a new array of instructions.
+ * A helper type to append instructions to a transaction message
+ * without loosing type information about the current instructions.
```
steveluscher:
```suggestion
 * without losing type information about the current instructions.
```

steveluscher:
> Currently, this problem only manifests when we are trying to append instructions to a transaction message already marked as a durable nonce transaction. Because we loose order information of the existing instructions, TypeScript can no longer assert that the durable nonce instruction is the very first one.

I wonder.

```ts
TInstructions extends [AdvanceNonceInstruction, ...infer TailInstructions extends IInstruction[]]
  ? [AdvanceNonceInstruction, ...TailInstructions, TInstructionToAppend]
  : [AdvanceNonceInstruction, TInstructionToAppend]
```

lorisleiva:
I tried to generalise Steven's solution using types like this:

```ts
type InferTupleItems4<T extends readonly unknown[]> = T extends readonly [infer T1, infer T2, infer T3, infer T4]
    ? [T1, T2, T3, T4]
    : InferTupleItems3<T>;
type InferTupleItems3<T extends readonly unknown[]> = T extends readonly [infer T1, infer T2, infer T3]
    ? [T1, T2, T3]
    : InferTupleItems2<T>;
type InferTupleItems2<T extends readonly unknown[]> = T extends readonly [infer T1, infer T2]
    ? [T1, T2]
    : InferTupleItems1<T>;
type InferTupleItems1<T extends readonly unknown[]> = T extends readonly [infer T1] ? [T1] : T;
```

(and variations of that including inferring `...TTail` as a fallback as well)

It worked with the simple `TX` example I used but when applying to instructions, the inferred items were generic `IInstruction` instead of the actual instruction type we're trying to retrieve.

So now I'm gonna try to go on a slightly different direction and adjust the function signatures of things like `createTransactionMessage` so that they _satisfy_ `BaseTransactionMessage` but without any instructions to start with. That way, when using the helper functions, you're gradually growing a type that always complies to the base type.

Wish me luck.

packages/transaction-messages/src/durable-nonce.ts
```diff
@@ -196,9 +196,20 @@ export function setTransactionMessageLifetimeUsingDurableNonce<
...
+    TNonceAccountAddress,
+    TNonceAuthorityAddress,
+    TNonceValue
+> {
+    type ReturnType = SetTransactionMessageWithDurableNonceLifetime<
+        TTransactionMessage,
+        TNonceAccountAddress,
+        TNonceAuthorityAddress,
+        TNonceValue
+    >;
```
steveluscher, line 212: I didn't know you could define types in a method body like this.

packages/transaction-messages/src/decompile-message.ts
```diff
@@ -241,9 +241,10 @@ export function decompileTransactionMessage(
         createTransactionMessage({ version: compiledTransactionMessage.version as TransactionVersion }),
         m => setTransactionMessageFeePayer(feePayer, m),
         m =>
-            instructions.reduce((acc, instruction) => {
-                return appendTransactionMessageInstruction(instruction, acc);
-            }, m),
+            instructions.reduce(
+                (acc, instruction) => appendTransactionMessageInstruction(instruction, acc),
+                m as BaseTransactionMessage & TransactionMessageWithFeePayer,
```
steveluscher, line 254: Hunh. Was this not already the case? Why did you have to cast it to `BaseTransactionMessage & TransactionMessageWithFeePayer`?

Also, is there a need to `TransactionMessageWithFeePayer<typeof feePayer>` or no?

packages/transaction-messages/src/__typetests__/durable-nonce-typetests.ts
```diff
@@ -72,4 +85,43 @@ type V0TransactionMessage = Extract<TransactionMessage, { version: 0 }>;
...
+    {
+        const feePayer = null as unknown as Address;
+        const message = pipe(
+            createTransactionMessage({ version: 0 }),
+            m => setTransactionMessageFeePayer(feePayer, m),
+            m => appendTransactionMessageInstruction(null as unknown as InstructionA, m),
+            m => setTransactionMessageLifetimeUsingDurableNonce(mockNonceConfig, m),
+        );
+
+        message satisfies CompilableTransactionMessage;
```
steveluscher: I was going to say maybe wherever you have `Address` make it `Address<'abc'>` to make sure that the specific address flows through, but looks like `CompilableTransactionMessage` doesn't have a type parameter to pass on to `TransactionMessageWithFeePayer`. Grumble.

steveluscher (approved):
Added via Giphy

Such attention to detail!

packages/transaction-messages/src/decompile-message.ts
lorisleiva: This is because the accumulator of the `reduce` function cannot change its type. Before this wasn't an issue because the append instruction helper was wrongly returning `TTransactionMessage` as provided. This means the reduce function had no issues updating the accumulator with the exact same (wrong) type.

Now that the append helper augments the type, we must explicitly tell the reduce function which wider type we will end up with since `createTransactionMessage` returns a transaction message with no instructions.

steveluscher: Ahhh. Got it.

lorisleiva:
### Merge activity

* **May 14, 5:56 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/432).

PR #472, opened 2025-05-06
A package for creating nominal TypeScript types

steveluscher:
#### Problem

Two versions of any given package may introduce two distinct instances of our branded types into a project. That is to say you might get the `Address` type from v2.1.0 and v3.0.0 in the same project.

When this happens, you will end up with values that are runtime identical, but whose types mismatch. This is because the way we brand the types involves `unique symbol` which is locally unique to the particular instance of the package being imported.

#### Summary of Changes

Create a package with utilities that will brand types with shared names across the entire project, instead of unique symbols that are module-local.

Addresses #459

PR #471, opened 2025-05-06
Collect all of the types for the build-time constants in a common place

steveluscher:
#### Problem

Since these get erased at build time there's no harm in keeping them outside of the package directory.

#### Summary of Changes

Move the build-time constants' types into a commmon place and refer to them everywhere they're needed.

PR #473, opened 2025-05-06
Modify all branded types such that they will match with any structurally similar type from any package version, going forward

### LINKED ISSUE #459: Maybe the opaque types are too opaque
## Motivation

We're about to take this library to v3 and beyond with some breaking changes. It occurs to me that despite being _structurally_ identical between v2 and v3, types based on `unique symbol` like `Address` will start to mismatch between the two.

## Example use case

Imagine someone trying to use an `Address` between different versions of `@solana/kit`. They're _absolutely_ runtime compatible, but the type incompatibility between two versions of `unique symbol` is demonstrated by this playground ([link](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAggJnAThAzigagJgDwBUoQAewEAdnClCsIgJakDmUAvFTfQwHwtT4BkUAN5RkAQzgB7UgBsQUAPryARolHkAXFACupWgEct0FCAC2SidKgBfANwAoUJFgJkadAGY8BYmQps6jDzUAVw8-EIiEOJSsgrKqhraugZGpuaWtg7g0PBIqBgALF5EJOSUwRxB7IzcrOHCYpIycooqanCaAOS5riid1vaOOS756ACsxT5l-pWsFTVhUAINUU2xrQkdUN0jaP2ZdgDGUtRQ4nlumDydAIyY7v2ilD2jOLf3nZz2x6Sn570ea53B5nZ67DCed4PL5HE7AM7g9AFIEfUHOC6FbBQz7fOEIjHjFEgp7ogETbEwuz-V5UUTAWgoABmtFQpNe9mpl1p9KZLLBBLewJxVMRVxQdIZzNZLzc7g5ou5kr5bNlWKFMM5GDFEt50sRBXlAsVuv5AKKFMNAO1PKlptGY0tNPFNuVMow5PV9hFBPcxttKq1jtlftdorVHw1iN9zqVep9QYhIbjAMhnu9KaTdrcBvTo2jOv9bqR4ehCcBMZNAfGZfzLuT9pLwtz2czVcwZeRFcLYYtzcKraLcr7SIHUcbkYJnYLocnHdHk-Hc67M7JS+n9bcHojXs1hOXG8Dw7G86ti6PJ7zZeP+6zELPu+v69vSKvF+z98Rj7rz4d55vVa3UsgA)).

## Proposal

In the next major version (3.0.0) modify all of the branded types to fit this pattern:

```ts
type Foo = T & { ['__@solana/kit;brand']: 'Foo' };
```

Playground ([link](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAggJnAThAzigagJgDwBUoQAewEAdnClCsIgJakDmUAvFTfQwHwtT4BkUAN5RkAQzgB7UgBsQUAPryARolHkAXFACupWgEct0FCAC2SidKgBfANwAoUJFgJkadAGY8BYmQps6jDzUAVw8-EIiEOJSslAA2gDkigACKBZqogD0ANa0wDYqanAJALqaCfBIqCgJ1vaO0JWuGAAsXkQk5JTBHEHsjNys4cJikjJyiSlp0hk5eQWq5KXlTdW1tg7gjS7V6ACs7T5d-r2sPQNhUAIjUWOxk-Kp6aRZufmFS2VQFTto6-Z2ADGUmoUHEVTcmB4CQAjJh3LVRJRVpDsLD4QlOPZgaRQeDmh5oXCEWDkb8MJ50QisUCQcAweT0C0iRjSc4Ia00cTMdi6QyOfsWSSkeyCQcqTy7HZMpkoFgoAADFEoBVQYEmMCiYC0JTSaAAdzyAAsoKQJFAJMAjRBEHZ8bsoSgtbQUAAzWioUUO+z2yFUZ1uj1kgU4CU030YR0B92e5UeH2MqPawOxxmU7nhxP+5Mx4MEloJkPZl25r1uNphwsEpMloNljB7KsO4spvO7cUZgEy2DSSxK8mq12ICQmOXuC2kfWiRBwNUjzXa3XQHTayyifxaQHALSqSyAo1qBjQF1QEziCAAOjtaZbpbjmCbbnHTpzdfvXIxmYFz+jb7Tj4pW8-2-D9qQAwkX1rVMBQLa9vyA6D81AyUIwg39EPbcCf1fDC3A7T8AVQ5lINbessHA4j0LbVFKzg-MEOoikKIYsj0wIujdkonDGKZZiSLvRkK07DjyxYuNGxE1oxMZfCwKlVC9mkkNwMU-jgOrZCvzFJSCXcFSdN2Ni5IUgzy30tTcM5WiTIsniJJsqiyNknkgA))

steveluscher:
#### Problem

Two versions of any given package may introduce two distinct instances of our branded types into a project. That is to say you might get the `Address` type from v2.0.0 and v2.1.0 in the same project.

#### Summary of Changes

Here we use the type utilities in the previous stacked PR to reimagine these types as structurally identical, rather than having an identity that's tied to one particular instance of a module.

This will make it so that `Address` and other branded types like it are compatible from version v2.1.1 and onward.

Fixes #459

### DIFF
--- a/.changeset/legal-cases-lose.md
+++ b/.changeset/legal-cases-lose.md
@@ -1,0 +1,58 @@
+---
+'@solana/rpc-subscriptions-channel-websocket': patch
+'@solana/webcrypto-ed25519-polyfill': patch
+'@solana/transaction-confirmation': patch
+'@solana/codecs-data-structures': patch
+'@solana/rpc-subscriptions-spec': patch
+'@solana/fast-stable-stringify': patch
+'@solana/rpc-subscriptions-api': patch
+'@solana/transaction-messages': patch
+'@solana/rpc-transport-http': patch
+'@solana/rpc-subscriptions': patch
+'@solana/rpc-parsed-types': patch
+'@solana/rpc-transformers': patch
+'@solana/codecs-numbers': patch
+'@solana/codecs-strings': patch
+'@solana/rpc-spec-types': patch
+'@solana/nominal-types': patch
+'@solana/instructions': patch
+'@solana/subscribable': patch
+'@solana/transactions': patch
+'@solana/codecs-core': patch
+'@solana/rpc-graphql': patch
+'@solana/assertions': patch
+'@solana/functional': patch
+'@solana/addresses': patch
+'@solana/rpc-types': patch
+'@solana/accounts': patch
+'@solana/programs': patch
+'@solana/promises': patch
+'@solana/rpc-spec': patch
+'@solana/options': patch
+'@solana/rpc-api': patch
+'@solana/signers': patch
+'@solana/sysvars': patch
+'@solana/codecs': patch
+'@solana/compat': patch
+'@solana/errors': patch
+'@solana/react': patch
+'@solana/keys': patch
+'@solana/kit': patch
+'@solana/rpc': patch
+---
+
+The identity of all branded types has changed in such a way that the types from v2.1.1 will be compatible with any other version going forward, which is not the case for versions v2.1.0 and before.
+
+If you end up with a mix of versions in your project prior to v2.1.1 (eg. `@solana/addresses@2.0.0` and `@solana/addresses@2.1.0`) you may discover that branded types like `Address` raise a type error, even though they are runtime compatible. Your options are:
+
+1. Always make sure that you have exactly one instance of each `@solana/*` dependency in your project at any given time
+2. Upgrade all of your `@solana/*` dependencies to v2.1.1 at minimum, even if their minor or patch versions differ.
+3. Suppress the type errors using a comment like the following:
+    ```ts
+    const myAddress = address('1234..5678');     // from @solana/addresses@2.0.0
+    const myAccount = await fetchEncodedAccount( // imports @solana/addresses@2.1.0
+        rpc,
+        // @ts-expect-error Address types mismatch between installed versions of @solana/addresses
+        myAddress,
+    );
+    ```
--- a/packages/rpc-transformers/src/tree-traversal.ts
+++ b/packages/rpc-transformers/src/tree-traversal.ts
@@ -1,6 +1,6 @@
 import { RpcRequest, RpcRequestTransformer, RpcResponseTransformer } from '@solana/rpc-spec-types';
 
-export type KeyPathWildcard = { readonly __brand: unique symbol };
+export type KeyPathWildcard = { readonly ['__keyPathWildcard:@solana/kit']: unique symbol };
 export type KeyPath = ReadonlyArray<KeyPath | KeyPathWildcard | number | string>;
 
 export const KEYPATH_WILDCARD = {} as KeyPathWildcard;
--- a/packages/rpc-types/src/encoded-bytes.ts
+++ b/packages/rpc-types/src/encoded-bytes.ts
@@ -1,6 +1,11 @@
-export type Base58EncodedBytes = string & { readonly __brand: unique symbol };
-export type Base64EncodedBytes = string & { readonly __brand: unique symbol };
-export type Base64EncodedZStdCompressedBytes = string & { readonly __brand: unique symbol };
+import { Brand, CompressedData, EncodedString } from '@solana/nominal-types';
+
+export type Base58EncodedBytes = Brand<EncodedString<string, 'base58'>, 'Base58EncodedBytes'>;
+export type Base64EncodedBytes = Brand<EncodedString<string, 'base64'>, 'Base64EncodedBytes'>;
+export type Base64EncodedZStdCompressedBytes = Brand<
+    EncodedString<CompressedData<string, 'zstd'>, 'base64'>,
+    'Base64EncodedZStdCompressedBytes'
+>;
 
 export type Base58EncodedDataResponse = [Base58EncodedBytes, 'base58'];
 export type Base64EncodedDataResponse = [Base64EncodedBytes, 'base64'];
--- a/packages/rpc-types/src/typed-numbers.ts
+++ b/packages/rpc-types/src/typed-numbers.ts
@@ -1,8 +1,10 @@
+import { Brand } from '@solana/nominal-types';
+
 export type Slot = bigint;
 export type Epoch = bigint;
 
 // Specifically being used to denote micro-lamports, which are 0.000001 lamports.
-export type MicroLamports = bigint & { readonly __brand: unique symbol };
+export type MicroLamports = Brand<bigint, 'MicroLamports'>;
 export type SignedLamports = bigint;
 
 // FIXME(solana-labs/solana/issues/30341)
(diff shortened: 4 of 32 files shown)


steveluscher:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/473?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#473** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/473?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/473?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#472** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/472?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#471** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/471?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/rpc-transformers/src/tree-traversal.ts
```diff
@@ -1,6 +1,6 @@
 import { RpcRequest, RpcRequestTransformer, RpcResponseTransformer } from '@solana/rpc-spec-types';
 
-export type KeyPathWildcard = { readonly __brand: unique symbol };
+export type KeyPathWildcard = { readonly ['__keyPathWildcard:@solana/kit']: unique symbol };
```
steveluscher, line 3: I'm pretty sure this one should stay the same, since the `KEYPATH_WILDCARD` depends on the literal identity of the object being the expected one, in order for the tree traversal to identify it. It's being used as a symbol here.

packages/rpc-types/src/encoded-bytes.ts
```diff
@@ -1,6 +1,11 @@
...
-export type Base64EncodedBytes = string & { readonly __brand: unique symbol };
-export type Base64EncodedZStdCompressedBytes = string & { readonly __brand: unique symbol };
+import { Brand, CompressedData, EncodedString } from '@solana/nominal-types';
+
+export type Base58EncodedBytes = Brand<EncodedString<string, 'base58'>, 'Base58EncodedBytes'>;
+export type Base64EncodedBytes = Brand<EncodedString<string, 'base64'>, 'Base64EncodedBytes'>;
+export type Base64EncodedZStdCompressedBytes = Brand<
+    EncodedString<CompressedData<string, 'zstd'>, 'base64'>,
+    'Base64EncodedZStdCompressedBytes'
+>;
```
steveluscher, line 8: Ooh fancy.

packages/rpc-transformers/src/tree-traversal.ts
lorisleiva: If it isn't exported on the package then I think it's fine but if we offer that type to the outside world, we should probably use our new nominal type system.

packages/rpc-types/src/encoded-bytes.ts
lorisleiva: 🤤

lorisleiva (approved):
Beautiful! There are other `unique symbols` we should consider cleaning such as transaction type flags (e.g. `FullySignedTransaction`). Refer to [my previous comment](https://github.com/anza-xyz/kit/pull/472#discussion_r2077108498) for this.

packages/rpc-transformers/src/tree-traversal.ts
steveluscher: > If it isn't exported on the package then I think it's fine but if we offer that type to the outside world, we should probably use our new nominal type system.

Our new system makes it _worse_. `KEYPATH_WILDCARD` is a token that the tree walker detects via `===`. If you use the `KEYPATH_WILDCARD` from version 2.1.0 with the tree walker from 2.1.1, it _won't detect it as the wildcard_. Keeping this `unique symbol` will make that a type error, rather than confusing runtime behaviour.

The only thing we could do to change this is to make the detection of the keypath wildcard dependent _not_ on the identity of the `{}`, but on the special values of some string (eg. `keyPathWildcard_DO_NO_FORGE_OR_YOU_WILL_BE_FIRED`).

.changeset/legal-cases-lose.md
```diff
@@ -0,0 +1,61 @@
...
+If you end up with a mix of versions in your project that includes v2.x.x (eg. `@solana/addresses@2.1.0` and `@solana/addresses@2.1.1`) you may discover that branded types like `Address` raise a type error, even though they are runtime compatible. Your options are:
+
+1. Always make sure that you have exactly one instance of each `@solana/*` dependency in your project at any given time
+2. Upgrade all of your `@solana/*` dependencies v3 at minimum, even if their minor or patch versions differ.
+3. Suppress the type errors using a comment like the following:
+    ```ts
+    const myAddress = address('1234..5678');     // from @solana/addresses@2.1.0
+    const myAccount = await fetchEncodedAccount( // imports @solana/addresses@2.1.1
+        rpc,
+        // @ts-expected-error Address types mismatch between installed versions of @solana/addresses
```
jordaaash: nit: should be `@ts-expect-error`

packages/rpc-types/src/typed-numbers.ts
```diff
@@ -1,8 +1,10 @@
+import { Brand } from '@solana/nominal-types';
+
 export type Slot = bigint;
 export type Epoch = bigint;
 
 // Specifically being used to denote micro-lamports, which are 0.000001 lamports.
-export type MicroLamports = bigint & { readonly __brand: unique symbol };
+export type MicroLamports = Brand<bigint, 'MicroLamports'>;
 export type SignedLamports = bigint;
```
jordaaash, line 8: (I don't know what SignedLamports are) -- why does it not use the Lamports type above?

Should Slot and Epoch also have branded types?

jordaaash (approved):
❤️

packages/rpc-types/src/typed-numbers.ts
steveluscher: We created this type because the value of lamports in [`Rewards`](https://github.com/anza-xyz/kit/blob/2f4b5115adf7ff4c2669685a343d9b41f67d7571/packages/rpc-types/src/transaction.ts#L376-L383) could be negative. Wanted to drive home that this value was not unsigned.

I think it's important to make things like `Lamports` branded types so that you can't accidentally supply a number denominated in SOL when the API requires it in `Lamports` (as folks have been known to do). Slots and epochs are perhaps less important to brand, because they're not ‘denominations’ that are easily confused in the same way that SOL and `Lamports` both represent the same store of value, denominated in different ways.

steveluscher (commented):
I'm dumb. This isn't a major version bump. Opaque/branded types between patch versions of v2 are already incompatible, so improving this is not a breaking change.

steveluscher:
### Merge activity

* **May 8, 10:25 AM PDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/473).

PR #508, opened 2025-05-30
Make rpc non-thenable

calintje:
#### Problem
RPC objects created by Solana Kit use a proxy that converts all property access into methods. When JavaScript's Promise machinery checks for a `.then` property (which happens when you `async await` an RPC object), the proxy intercepts this and returns a method `{ send: [AsyncFunction: send] }` instead of `undefined` or the proper Promise interface, making the RPC object appear thenable when it isn't. This causes JavaScript/TypeScript to silently terminate the program when you try to `await` the RPC object as if it were a Promise.

#### Summary of Changes
- Modified `makeProxy` function to specifically intercept `then` property access and return `undefined`
- Prevents the proxy from treating `.then` like any other RPC method
- Added test case to verify RPC objects are not thenable

### DIFF
--- a/packages/rpc-spec/src/__tests__/rpc-test.ts
+++ b/packages/rpc-spec/src/__tests__/rpc-test.ts
@@ -75,6 +75,10 @@ describe('JSON-RPC 2.0', () => {
             const sendPromise = rpc.someMethod().send();
             await expect(sendPromise).rejects.toThrow(transportError);
         });
+        it('should not be thenable', () => {
+            expect.assertions(1);
+            expect((rpc as { then?: unknown }).then).toBeUndefined();
+        });
     });
     describe('when calling a method having a concrete implementation', () => {
         let rpc: Rpc<TestRpcMethods>;
--- a/packages/rpc-spec/src/__tests__/rpc-test.ts
+++ b/packages/rpc-spec/src/__tests__/rpc-test.ts
@@ -77,7 +77,7 @@ describe('JSON-RPC 2.0', () => {
         });
         it('should not be thenable', () => {
             expect.assertions(1);
-            expect((rpc as { then?: unknown }).then).toBeUndefined();
+            expect(rpc).not.toHaveProperty('then');
         });
     });
     describe('when calling a method having a concrete implementation', () => {
--- a/packages/rpc-spec/src/rpc.ts
+++ b/packages/rpc-spec/src/rpc.ts
@@ -72,6 +72,9 @@ function makeProxy<TRpcMethods, TRpcTransport extends RpcTransport>(
             return false;
         },
         get(target, p, receiver) {
+            if (p === 'then') {
+                return undefined;
+            }
             return function (...rawParams: unknown[]) {
                 const methodName = p.toString();
                 const getApiPlan = Reflect.get(target, methodName, receiver);
--- a/packages/rpc-subscriptions-spec/src/__tests__/rpc-subscriptions-test.ts
+++ b/packages/rpc-subscriptions-spec/src/__tests__/rpc-subscriptions-test.ts
@@ -1,18 +1,23 @@
 import { SOLANA_ERROR__RPC_SUBSCRIPTIONS__CANNOT_CREATE_SUBSCRIPTION_PLAN, SolanaError } from '@solana/errors';
 
-import { createSubscriptionRpc } from '../rpc-subscriptions';
+import { createSubscriptionRpc, type RpcSubscriptions } from '../rpc-subscriptions';
 
 interface TestRpcSubscriptionNotifications {
     thingNotifications(...args: unknown[]): unknown;
 }
 
 describe('createSubscriptionRpc', () => {
-    it('throws when the API produces no subscription plan', () => {
-        const rpcSubscriptions = createSubscriptionRpc<TestRpcSubscriptionNotifications>({
+    let rpcSubscriptions: RpcSubscriptions<TestRpcSubscriptionNotifications>;
+
+    beforeEach(() => {
+        rpcSubscriptions = createSubscriptionRpc<TestRpcSubscriptionNotifications>({
             // @ts-expect-error Does not implement API on purpose
             api: {},
             transport: jest.fn(),
         });
+    });
+
+    it('throws when the API produces no subscription plan', () => {
         expect(() => {
             rpcSubscriptions.thingNotifications();
         }).toThrow(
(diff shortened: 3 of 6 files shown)


calintje:
That sounds good. As an additional safeguard, we might consider restricting the proxy to only expose known Solana RPC method names. That way, any unexpected property access (like then or others) would default to undefined, avoiding misinterpretation by JavaScript and keeping the surface of the proxy more predictable. I can look into implementing that if it seems worthwhile.

mcintyre94:
For now let's push anything else to that task. We should definitely do `then`, but when we understand what else (if anything) might cause issues we can figure out what makes the most sense. There would be some tradeoffs here, like not being able to add new methods without a custom proxy, and increasing the bundle size a bit (and for every new method).

packages/rpc-subscriptions-spec/src/rpc-subscriptions.ts
```diff
@@ -35,10 +35,10 @@ type PendingRpcSubscriptionsRequestBuilder<TSubscriptionMethodImplementations> =
...
     // Check that this property of the TRpcSubscriptionMethods interface is, in fact, a function.
     TSubscriptionMethodImplementation extends Callable
-        ? (
-              ...args: Parameters<TSubscriptionMethodImplementation>
-          ) => PendingRpcSubscriptionsRequest<ReturnType<TSubscriptionMethodImplementation>>
-        : never;
+    ? (
+        ...args: Parameters<TSubscriptionMethodImplementation>
+    ) => PendingRpcSubscriptionsRequest<ReturnType<TSubscriptionMethodImplementation>>
+    : never;
```
mcintyre94: It looks like something has tweaked the formatting here. If you run `pnpm style:fix` at the root of the repo it'll apply our formatting/prettier and should make the CI happy!

calintje: Thanks, fixed!

packages/rpc-subscriptions-spec/src/rpc-subscriptions.ts
```diff
@@ -57,6 +57,9 @@ export function createSubscriptionRpc<TRpcSubscriptionsApiMethods>(
         },
         get(target, p, receiver) {
             return function (...rawParams: unknown[]) {
+                if (p === 'then') {
+                    return undefined;
+                }
```
mcintyre94: This needs to be outside the `return function` 

```suggestion
        get(target, p, receiver) {
            if (p === 'then') {
                return undefined;
            }
            return function (...rawParams: unknown[]) {
                
```

steveluscher:
> …we might consider restricting the proxy to only expose known Solana RPC method names.

We don't do this on purpose, because to do so would require that we _enumerate_ every method name, which would require that we _ship_ all of those strings in the JavaScript bundle.

packages/rpc-spec/src/__tests__/rpc-test.ts
```diff
@@ -75,6 +75,10 @@ describe('JSON-RPC 2.0', () => {
             const sendPromise = rpc.someMethod().send();
             await expect(sendPromise).rejects.toThrow(transportError);
         });
+        it('should not be thenable', () => {
+            expect.assertions(1);
+            expect((rpc as { then?: unknown }).then).toBeUndefined();
```
steveluscher:
```suggestion
            expect(rpc).not.toHaveProperty('then');
```

packages/rpc-spec/src/rpc.ts
```diff
@@ -72,6 +72,9 @@ function makeProxy<TRpcMethods, TRpcTransport extends RpcTransport>(
             return false;
         },
         get(target, p, receiver) {
+            if (p === 'then') {
```
steveluscher, line 75: Note for myself: this doesn't have to be `p.toString()` because there's no way of calling the `then` method of a `Promise` using a `Symbol`.

packages/rpc-subscriptions-spec/src/__tests__/rpc-subscriptions-test.ts
```diff
@@ -21,4 +26,8 @@ describe('createSubscriptionRpc', () => {
             }),
         );
     });
+
+    it('should not be thenable', () => {
+        expect((rpcSubscriptions as { then?: unknown }).then).toBeUndefined();
```
steveluscher:
```suggestion
        expect(rpcSubscriptions).not.toHaveProperty('then');
```

steveluscher (changes requested):
Thanks for this! Super interesting.

steveluscher:
One last question: are we convinced that we don't also need to blocklist `catch` and `finally`, and why not?

mcintyre94:
@steveluscher I don't think so - the MDN docs call out `thenable` as the thing the language allows in place of promises: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables

calintje:
Hmm, I didn't think about that when I encountered this issue.

After looking a bit online, it seems that both `catch()` and `finallly()` internally call `then()`

1. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch#description

    > catch() internally calls then() on the object upon which it was called, passing undefined and onRejected as arguments.

2. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

    > Like `catch()`, finally() internally calls the then method on the object upon which it was called
    
However, catch would be called on the rpc object and go through the `get` trap first. I wrote a test to verify the behavior, and it passed (ie. it doesn't reach the internal implementation of `catch()`):

```ts
it('treats catch as an RPC method name and fails appropriately', () => {
    const getPropertySpy = jest.spyOn(Reflect, 'get');

    // @ts-expect-error We're intentionally testing runtime behavior that TypeScript prevents
    const catchFunction = rpcSubscriptions.catch;

    // The function should throw when called
    expect(() => {
        catchFunction(() => { });
    }).toThrow(new SolanaError(SOLANA_ERROR__RPC_SUBSCRIPTIONS__CANNOT_CREATE_SUBSCRIPTION_PLAN, {
        notificationName: 'catch',
    }));

    // Verify it tried to look up 'catch' as an RPC method name
    const spyCalls = getPropertySpy.mock.calls;
    expect(spyCalls.some(call => call[1] === 'catch')).toBe(true);

    getPropertySpy.mockRestore();
});
```

Given this behavior, I can make the following changes:
1. Extend the proxy's get trap to explicitly handle all Promise-related methods:
```ts
   if (p === 'then' || p === 'catch' || p === 'finally') {
       return undefined;
   }
```

2. Update the tests to verify this behavior for all Promise methods:

calintje:
Oh @mcintyre94, good find. So JavaScript only treats an object as Promise-like if it has a .then() method, that's what makes it "thenable". My test was actually just confirming that catch is treated like any other invalid RPC method name, not testing any Promise-related behavior.

steveluscher (approved):
Added via Giphy

PR #519, opened 2025-06-03
Add provisory blockhash lifetime helpers

lorisleiva:
This PR adds the following items to the `transaction-messages` package:
- The `PROVISORY_BLOCKHASH_LIFETIME_CONSTRAINT` constant. It defines an invalid but temporary blockhash lifetime constraint that can be used for a variety of applications such as: simulating a transaction before setting its actual blockhash, calculating the size of the transaction message, etc.
- The `setTransactionMessageLifetimeUsingProvisoryBlockhash` function that simply sets the blockhash lifetime constraint using the `PROVISORY_BLOCKHASH_LIFETIME_CONSTRAINT`.
- The `fillMissingTransactionMessageLifetimeUsingProvisoryBlockhash` function that only sets the provisory blockhash if and only if the provided transaction message does not already have a set lifetime constraint.

Note that this is currently [used by the Compute Budget client](https://github.com/solana-program/compute-budget/blob/238034bbe03e7be358a6a6b7427de23c9e358104/clients/js/src/internalMoveToKit.ts) and will be required for Instruction Plans.

### DIFF
--- a/.changeset/every-boxes-tell.md
+++ b/.changeset/every-boxes-tell.md
@@ -1,0 +1,5 @@
+---
+'@solana/transaction-messages': patch
+---
+
+Add provisory blockhash lifetime helpers to use as placeholders when constructing transaction messages. Namely, it adds the `PROVISORY_BLOCKHASH_LIFETIME_CONSTRAINT` constant and the `fillMissingTransactionMessageLifetimeUsingProvisoryBlockhash` and `setTransactionMessageLifetimeUsingProvisoryBlockhash` functions.
--- a/packages/transaction-messages/src/__tests__/blockhash-test.ts
+++ b/packages/transaction-messages/src/__tests__/blockhash-test.ts
@@ -5,7 +5,10 @@ import type { Blockhash } from '@solana/rpc-types';
 
 import {
     assertIsTransactionMessageWithBlockhashLifetime,
+    fillMissingTransactionMessageLifetimeUsingProvisoryBlockhash,
+    PROVISORY_BLOCKHASH_LIFETIME_CONSTRAINT,
     setTransactionMessageLifetimeUsingBlockhash,
+    setTransactionMessageLifetimeUsingProvisoryBlockhash,
     TransactionMessageWithBlockhashLifetime,
 } from '../blockhash';
 import { BaseTransactionMessage } from '../transaction-message';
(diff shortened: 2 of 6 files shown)


lorisleiva:
* **#519** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/519?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/519?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@steveluscher That is a good point. The only thing is the whole instruction plan machinery relies on the `CompilableTransactionMessage` type both as input to figure out the base transaction sizes and as output for the `TransactionPlans`. So if we were to relax some of these rules, we'd need another type definition such as `SizableTransactionMessage` that can be used instead of `CompilableTransactionMessage` by the transaction planners. What do you think?

steveluscher:
Show me a short example; I bet I can make it work.

lorisleiva:
Hmm extracting a small example of this API is tricky but essentially `CompilableTransactionMessage` is the glue that holds all the different plans together.

- In the [iterable instruction plan](https://github.com/solana-program/program-metadata/blob/b9daf783c708ce220ab6f4de84da86d8e86cbe26/clients/js/src/instructionPlans/instructionPlan.ts#L52).
- In the [single transaction plan](https://github.com/solana-program/program-metadata/blob/b9daf783c708ce220ab6f4de84da86d8e86cbe26/clients/js/src/instructionPlans/transactionPlan.ts#L21).
- In the [transaction planner](https://github.com/solana-program/program-metadata/blob/b9daf783c708ce220ab6f4de84da86d8e86cbe26/clients/js/src/instructionPlans/transactionPlannerBase.ts#L29).
- In the [base transaction plan executor](https://github.com/solana-program/program-metadata/blob/b9daf783c708ce220ab6f4de84da86d8e86cbe26/clients/js/src/instructionPlans/transactionPlanExecutorBase.ts#L16).
- In the [single transaction plan result](https://github.com/solana-program/program-metadata/blob/b9daf783c708ce220ab6f4de84da86d8e86cbe26/clients/js/src/instructionPlans/transactionPlanResult.ts#L28).

I believe we can change all of these to use a looser `SizableTransactionMessage` type instead and just let the `TransactionPlanExecutor` take care of the transaction from a `SizableTransactionMessage` to a compiled transaction. But we would have to assume the latest blockhash is required by default instead of failing if one is not explicitly provided.

PR #521, opened 2025-06-03
add loadedAccountsDataSize to simulate transaction response

tao-stones:
#### Problem

https://github.com/anza-xyz/agave/pull/6023 adds loaded-accounts-data-size to simulate transaction result.

#### Summary of Changes

- add new field to RPC response

Implements #541.

### DIFF
--- a/packages/rpc-api/src/index.ts
+++ b/packages/rpc-api/src/index.ts
@@ -343,7 +343,7 @@ function getAllowedNumericKeypaths(): AllowedNumericKeypaths<RpcApi<SolanaRpcApi
                 ['delinquent', KEYPATH_WILDCARD, 'commission'],
             ],
             simulateTransaction: [
-		[KEYPATH_WILDCARD, 'loadedAccountsDataSize'],
+                ['value', 'loadedAccountsDataSize'],
                 ...jsonParsedAccountsConfigs.map(c => ['value', 'accounts', KEYPATH_WILDCARD, ...c]),
                 ...innerInstructionsConfigs.map(c => ['value', 'innerInstructions', KEYPATH_WILDCARD, ...c]),
             ],


packages/rpc-api/src/simulateTransaction.ts
```diff
@@ -117,6 +117,8 @@ type SimulateTransactionApiResponseBase = Readonly<{
     }> | null;
     /** The number of compute budget units consumed during the processing of this transaction */
     unitsConsumed?: bigint;
+    /** The number of bytes of all accounts loaded by this transaction */
+    loadedAccountsDataSize?: bigint;
```
steveluscher: Since this is a u32 on the RPC side, we'll need to add this to ‘[the list of properties that are supposed to be `Number` instead of `BigInt`](https://github.com/anza-xyz/kit/blob/39662f2b2499742080616082ea527d5bc96dbbff/packages/rpc-api/src/index.ts#L345-L348),’ and then type this as `number`.

steveluscher (approved):
Awesome! Just alpha-order those properties, and then add an allowlist entry to tell the system that it's allowed to be a `number` instead of a `bigint`.

You'll almost certainly need to update `simulate-transaction.test.ts` also.

You can get tests running by doing:

```
pnpm install
cd packages/rpc-api/
pnpm turbo compile:typedefs compile:js
pnpm dev
```

The last step starts up a persistent server that runs lint & tests whenever you save a file.

packages/rpc-api/src/simulateTransaction.ts
tao-stones: great, thanks for the pointer! Was doubting myself if `bigint` is proper here. Will make change.

.changeset/tidy-baths-rule.md
```diff
@@ -0,0 +1,5 @@
+---
```
tao-stones, line 1: Since it changes response, thought a changeset is needed?

packages/rpc-api/src/index.ts
```diff
@@ -343,6 +343,7 @@ function getAllowedNumericKeypaths(): AllowedNumericKeypaths<RpcApi<SolanaRpcApi
                 ['delinquent', KEYPATH_WILDCARD, 'commission'],
             ],
             simulateTransaction: [
+		[KEYPATH_WILDCARD, 'loadedAccountsDataSize'],
```
tao-stones: not too sure if `value' is needed here, hope `KEYPATH_WILDCARD` covers all the standard path, does it?

steveluscher: Yeah, it does! The shape of the response is:

```ts
{
    "context": /* ... */,
    "value": {
        /* ... */
        loadedAccountsDataSize: number,
    }
}
```

…so the path is `['value', 'loadedAccountsDataSize']`.

steveluscher:
Looks good! I fixed up lint and pushed a change to the keypath.

Ironic thing is that your tests won't pass until the validator starts to emit this value, so we'll have to wait to land this PR until then.

tao-stones:
> Looks good! I fixed up lint and pushed a change to the keypath.

Thank you! 

> Ironic thing is that your tests won't pass until the validator starts to emit this value, so we'll have to wait to land this PR until then.

Ah, that explains why my local test keeping error out! I'll change this PR to "ready for review" (a post fact :D) for now.

PR #532, opened 2025-06-06
Add empty `@solana/instruction-plans` package

lorisleiva:
This PR adds an empty `@solana/instruction-plans` package to remove the noise from other PRs that add features to this package.

PR #533, opened 2025-06-06
Add `InstructionPlan` type and helpers

lorisleiva:
This PR adds the `InstructionPlan` type and a variety of helpers to create `InstructionPlans` of different kinds.

### DIFF
--- a/packages/instruction-plans/src/__tests__/instruction-plan-test.ts
+++ b/packages/instruction-plans/src/__tests__/instruction-plan-test.ts
@@ -1,7 +1,301 @@
 import '@solana/test-matchers/toBeFrozenObject';
 
-describe('InstructionPlan', () => {
-    it('will have tests #trust', () => {
-        expect(true).toBe(true);
+import { Address } from '@solana/addresses';
+import { pipe } from '@solana/functional';
+import { Instruction } from '@solana/instructions';
+import {
+    BaseTransactionMessage,
+    createTransactionMessage,
+    setTransactionMessageFeePayer,
+    TransactionMessageWithFeePayer,
+} from '@solana/transaction-messages';
+import { getTransactionMessageSize, TRANSACTION_SIZE_LIMIT } from '@solana/transactions';
+
+import {
+    getLinearMessagePackerInstructionPlan,
+    getMessagePackerInstructionPlanFromInstructions,
+    getReallocMessagePackerInstructionPlan,
+    nonDivisibleSequentialInstructionPlan,
+    parallelInstructionPlan,
+    sequentialInstructionPlan,
+    singleInstructionPlan,
+} from '../instruction-plan';
+
+jest.mock('@solana/transactions', () => ({
+    ...jest.requireActual('@solana/transactions'),
+    getTransactionMessageSize: jest.fn(),
+}));
+
+function createInstruction<TId extends string>(id: TId): Instruction & { id: TId } {
+    return { id, programAddress: '11111111111111111111111111111111' as Address };
+}
+
+describe('singleInstructionPlan', () => {
+    it('creates SingleInstructionPlan objects', () => {
+        const instruction = createInstruction('A');
+        const plan = singleInstructionPlan(instruction);
+        expect(plan).toStrictEqual({ instruction, kind: 'single' });
+    });
+    it('freezes created SingleInstructionPlan objects', () => {
+        const instruction = createInstruction('A');
+        const plan = singleInstructionPlan(instruction);
+        expect(plan).toBeFrozenObject();
+    });
+});
+
+describe('parallelInstructionPlan', () => {
+    it('creates ParallelInstructionPlan objects from other plans', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = parallelInstructionPlan([
+            singleInstructionPlan(instructionA),
+            singleInstructionPlan(instructionB),
+        ]);
+        expect(plan).toStrictEqual({
+            kind: 'parallel',
+            plans: [singleInstructionPlan(instructionA), singleInstructionPlan(instructionB)],
+        });
+    });
+    it('accepts instructions directly and wrap them in single plans', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = parallelInstructionPlan([instructionA, instructionB]);
+        expect(plan).toStrictEqual({
+            kind: 'parallel',
+            plans: [singleInstructionPlan(instructionA), singleInstructionPlan(instructionB)],
+        });
+    });
+    it('can nest other parallel plans', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const instructionC = createInstruction('C');
+        const plan = parallelInstructionPlan([instructionA, parallelInstructionPlan([instructionB, instructionC])]);
+        expect(plan).toStrictEqual({
+            kind: 'parallel',
+            plans: [
+                singleInstructionPlan(instructionA),
+                { kind: 'parallel', plans: [singleInstructionPlan(instructionB), singleInstructionPlan(instructionC)] },
+            ],
+        });
+    });
+    it('freezes created ParallelInstructionPlan objects', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = parallelInstructionPlan([instructionA, instructionB]);
+        expect(plan).toBeFrozenObject();
+    });
+});
+
+describe('sequentialInstructionPlan', () => {
+    it('creates divisible SequentialInstructionPlan objects from other plans', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = sequentialInstructionPlan([
+            singleInstructionPlan(instructionA),
+            singleInstructionPlan(instructionB),
+        ]);
+        expect(plan).toStrictEqual({
+            divisible: true,
+            kind: 'sequential',
+            plans: [singleInstructionPlan(instructionA), singleInstructionPlan(instructionB)],
+        });
+    });
+    it('accepts instructions directly and wrap them in single plans', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = sequentialInstructionPlan([instructionA, instructionB]);
+        expect(plan).toStrictEqual({
+            divisible: true,
+            kind: 'sequential',
+            plans: [singleInstructionPlan(instructionA), singleInstructionPlan(instructionB)],
+        });
+    });
+    it('can nest other sequential plans', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const instructionC = createInstruction('C');
+        const plan = sequentialInstructionPlan([instructionA, sequentialInstructionPlan([instructionB, instructionC])]);
+        expect(plan).toStrictEqual({
+            divisible: true,
+            kind: 'sequential',
+            plans: [
+                singleInstructionPlan(instructionA),
+                {
+                    divisible: true,
+                    kind: 'sequential',
+                    plans: [singleInstructionPlan(instructionB), singleInstructionPlan(instructionC)],
+                },
+            ],
+        });
+    });
+    it('freezes created SequentialInstructionPlan objects', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = sequentialInstructionPlan([instructionA, instructionB]);
+        expect(plan).toBeFrozenObject();
+    });
+});
+
+describe('nonDivisibleSequentialInstructionPlan', () => {
+    it('creates non-divisible SequentialInstructionPlan objects from other plans', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = nonDivisibleSequentialInstructionPlan([
+            singleInstructionPlan(instructionA),
+            singleInstructionPlan(instructionB),
+        ]);
+        expect(plan).toStrictEqual({
+            divisible: false,
+            kind: 'sequential',
+            plans: [singleInstructionPlan(instructionA), singleInstructionPlan(instructionB)],
+        });
+    });
+    it('accepts instructions directly and wrap them in single plans', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = nonDivisibleSequentialInstructionPlan([instructionA, instructionB]);
+        expect(plan).toStrictEqual({
+            divisible: false,
+            kind: 'sequential',
+            plans: [singleInstructionPlan(instructionA), singleInstructionPlan(instructionB)],
+        });
+    });
+    it('can nest other non-divisible sequential plans', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const instructionC = createInstruction('C');
+        const plan = nonDivisibleSequentialInstructionPlan([
+            instructionA,
+            nonDivisibleSequentialInstructionPlan([instructionB, instructionC]),
+        ]);
+        expect(plan).toStrictEqual({
+            divisible: false,
+            kind: 'sequential',
+            plans: [
+                singleInstructionPlan(instructionA),
+                {
+                    divisible: false,
+                    kind: 'sequential',
+                    plans: [singleInstructionPlan(instructionB), singleInstructionPlan(instructionC)],
+                },
+            ],
+        });
+    });
+    it('freezes created SequentialInstructionPlan objects', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = nonDivisibleSequentialInstructionPlan([instructionA, instructionB]);
+        expect(plan).toBeFrozenObject();
+    });
+});
+
+describe('getLinearMessagePackerInstructionPlan', () => {
+    let message: BaseTransactionMessage & TransactionMessageWithFeePayer;
+    beforeEach(() => {
+        message = pipe(createTransactionMessage({ version: 0 }), m =>
+            setTransactionMessageFeePayer('E9Nykp3rSdza2moQutaJ3K3RSC8E5iFERX2SqLTsQfjJ' as Address, m),
+        );
+    });
+    it('creates MessagePackerInstructionPlan objects by splitting instructions until we reach the total bytes required', () => {
+        jest.mocked(getTransactionMessageSize).mockReturnValue(100);
+        const expectedLength = TRANSACTION_SIZE_LIMIT - 100 - 1;
+
+        const plan = getLinearMessagePackerInstructionPlan({
+            getInstruction: (offset: number, length: number) => createInstruction(`[${offset},${offset + length})`),
+            totalLength: 2000,
+        });
+
+        const messagePacker = plan.getMessagePacker();
+        expect(messagePacker.done()).toBe(false);
+        expect(messagePacker.packMessageToCapacity(message).instructions[0]).toStrictEqual(
+            createInstruction(`[0,${expectedLength})`),
+        );
+        expect(messagePacker.done()).toBe(false);
+        expect(messagePacker.packMessageToCapacity(message).instructions[0]).toStrictEqual(
+            createInstruction(`[${expectedLength},2000)`),
+        );
+        expect(messagePacker.done()).toBe(true);
+        expect(() => messagePacker.packMessageToCapacity(message)).toThrow();
+    });
+    it('freezes created MessagePackerInstructionPlan objects', () => {
+        const plan = getLinearMessagePackerInstructionPlan({
+            getInstruction: (offset: number, length: number) => createInstruction(`[${offset},${offset + length})`),
+            totalLength: 2000,
+        });
+        expect(plan).toBeFrozenObject();
+    });
+    it('freezes the messagePacker returned by getMessagePacker', () => {
+        const plan = getLinearMessagePackerInstructionPlan({
+            getInstruction: (offset: number, length: number) => createInstruction(`[${offset},${offset + length})`),
+            totalLength: 2000,
+        });
+        expect(plan.getMessagePacker()).toBeFrozenObject();
+    });
+});
+
+describe('getMessagePackerInstructionPlanFromInstructions', () => {
+    let message: BaseTransactionMessage & TransactionMessageWithFeePayer;
+    beforeEach(() => {
+        message = pipe(createTransactionMessage({ version: 0 }), m =>
+            setTransactionMessageFeePayer('E9Nykp3rSdza2moQutaJ3K3RSC8E5iFERX2SqLTsQfjJ' as Address, m),
+        );
+    });
+    it('creates MessagePackerInstructionPlan objects by providing the iterated instruction in advance', () => {
+        const plan = getMessagePackerInstructionPlanFromInstructions([createInstruction('A'), createInstruction('B')]);
+
+        const messagePacker = plan.getMessagePacker();
+        expect(messagePacker.done()).toBe(false);
+        expect(messagePacker.packMessageToCapacity(message).instructions).toStrictEqual([
+            createInstruction('A'),
+            createInstruction('B'),
+        ]);
+        expect(messagePacker.done()).toBe(true);
+        expect(() => messagePacker.packMessageToCapacity(message)).toThrow();
+    });
+    it('freezes created MessagePackerInstructionPlan objects', () => {
+        const plan = getMessagePackerInstructionPlanFromInstructions([createInstruction('A'), createInstruction('B')]);
+        expect(plan).toBeFrozenObject();
+    });
+    it('freezes the messagePacker returned by getMessagePacker', () => {
+        const plan = getMessagePackerInstructionPlanFromInstructions([createInstruction('A'), createInstruction('B')]);
+        expect(plan.getMessagePacker()).toBeFrozenObject();
+    });
+});
+
+describe('getReallocMessagePackerInstructionPlan', () => {
+    let message: BaseTransactionMessage & TransactionMessageWithFeePayer;
+    beforeEach(() => {
+        message = pipe(createTransactionMessage({ version: 0 }), m =>
+            setTransactionMessageFeePayer('E9Nykp3rSdza2moQutaJ3K3RSC8E5iFERX2SqLTsQfjJ' as Address, m),
+        );
+    });
+    it('creates MessagePackerInstructionPlan objects by chunking instruction using the `REALLOC_LIMIT`', () => {
+        const plan = getReallocMessagePackerInstructionPlan({
+            getInstruction: (size: number) => createInstruction(`Size: ${size}`),
+            totalSize: 15_000,
+        });
+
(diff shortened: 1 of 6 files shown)


lorisleiva:
* **#648** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#647** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/647?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#533** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/533?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/533?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/instruction-plans/src/__tests__/instruction-plan-test.ts
```diff
@@ -1,7 +1,340 @@
...
+            createTransactionMessage({ version: 0 }),
+            m => setTransactionMessageFeePayer('E9Nykp3rSdza2moQutaJ3K3RSC8E5iFERX2SqLTsQfjJ' as Address, m),
+            // TODO(loris): Either remove lifetime constraint or use the new
+            // `fillMissingTransactionMessageLifetimeUsingProvisoryBlockhash`
+            // function from https://github.com/anza-xyz/kit/pull/519.
+            m =>
+                setTransactionMessageLifetimeUsingBlockhash(
+                    {} as Parameters<typeof setTransactionMessageLifetimeUsingBlockhash>[0],
+                    m,
+                ),
```
lorisleiva: @steveluscher Here's an example showing how coupled the `CompilableTransactionMessage` type is to this instruction plan system.

packages/instruction-plans/src/__tests__/instruction-plan-test.ts
```diff
@@ -1,7 +1,340 @@
...
+
+        const iterator = plan.getIterator();
+        expect(iterator.hasNext()).toBe(true);
+        expect(iterator.next(message)).toEqual(createInstruction(`[0,${expectedLength})`));
+        expect(iterator.hasNext()).toBe(true);
+        expect(iterator.next(message)).toEqual(createInstruction(`[${expectedLength},2000)`));
+        expect(iterator.hasNext()).toBe(false);
+        expect(iterator.next(message)).toBeNull();
+    });
+    it('offers a `getAll` method that packs everything into a single instruction', () => {
```
steveluscher: Is this necessarily true, or is it just that the one in this test happens to pack it into a single instruction?

steveluscher (changes requested):
Let's talk about this iterator design. Check some assumptions for me.

1. **The iterator produces instructions**. Given a transaction message, it will produce _the next_ instruction or `null` if it doesn't fit in the supplied message.
2. **The `getAll()` method doesn't consider any one transaction message**; it just presumes infinite space for instructions and gives them to you without complaining.
3. As far as I can tell, the only use case for obtaining an instruction is to _append_ it to _the same transaction message you passed to `next()`_.

If those are even close to true, could the API instead take in a transaction message, and do one of two things:

1. Pack all the instructions into a copy of the message then return that.
2. Throw an error with:
    * A new message with as many instructions packed into it as would fit.
    * A new `InstructionPlan` you can use to pack the remaining ones.

```ts
// https://codesandbox.io/p/sandbox/generator-for-instruction-plan-24rpp8
function getWhateverWeCallThis<TInstruction extends Instruction = Instruction>(
  instructions: TInstruction[]
) {
  return {
    packThatMessage(message: BaseTransactionMessage) {
      let out = [message];
      for (let ii = 0; ii < instructions.length; ii++) {
        out.push(`instruction-${instructions[ii]}`);
        if (ii === 2) {
          throw new ItDidNotFitError(out.join(","), instructions.slice(ii + 1));
        }
      }
      return out.join(",");
    },
    [Symbol.iterator](): Iterator<string> {
      let ii = 0;
      return {
        next() {
          if (ii == instructions.length) {
            return {
              value: undefined,
              done: true,
            };
          }
          return {
            value: `instruction-${ii++}`,
            done: false,
          };
        },
      };
    },
  };
}

let thing = getWhateverWeCallThis([
  "doThis",
  "doThat",
  "doTheOtherThing",
  "doOneTooManyThings",
]);

console.log("--- Get all the instructions");
const allInstructions = Array.from(thing);
console.log({ allInstructions });

console.log("--- Pack the instructions into a message");
let message = "message";
while (true) {
  try {
    const packedMessage = thing.packThatMessage(message);
    console.log("packedMessage", packedMessage);
    break;
  } catch (e) {
    if (e instanceof ItDidNotFitError) {
      console.log(
        "At least we managed to pack this so far",
        e.whateverWePackedSoFar
      );
      thing = e.continuationThing;
      message = "newMessage";
      continue;
    }
    throw e;
  }
}
```

Also, the ‘thing’ can implement the Iterator protocol, so that you can get all the instructions in the normal way (ie. passing the `Iterator` to `Array.from()`).

Am I too far off?

lorisleiva:
Let me just step back a bit and add more context on the `IterableInstructionPlan`.

Without the `IterableInstructionPlan`, we can only produce plans whose instruction size is known regardless of the transaction message being packed. That is, creating a plan with sequential, parallel and single nodes, will always produce the same plan which may not be the most optimal way to pack a set of instructions.

To illustrate this, let’s look at the `write` instruction of the Loader V3 program. This instruction is used to gradually write bytes into a buffer such that this buffer can then be used to deploy a new Solana program or update an existing one.

Now, imagine we want to create a helper function that returns an `InstructionPlan` for the operation “Deploy a new program”. Let’s call it `getDeployProgramInstructionPlan(input)`. This will include instructions such as: create buffer, write, deploy (using that buffer).

Since the user will provide the program binary to deploy — in the `input` argument — we know exactly how many bytes we must gradually `write` to the buffer. Since this will likely span multiple transactions, we must provide as many `write` instructions as necessary to fill the entire buffer. We could “guess” the right amount of bytes per message and simply split these instructions using this heuristic. For instance, we could say 900 bytes per transaction is a safe bet. However, this doesn’t take into account that these `write` instructions:
- could be used to fill existing transaction messages with fewer space (e.g. you have 400 bytes free in a previous message being packed).
- may have less space than 900 bytes to work with (e.g. the base transaction message uses lots of meta instructions such as setting compute budget units, prices, etc.).
- may have more than 900 bytes of space to use per transaction which significantly increases the total amount of transactions to land for large binaries.

So instead of providing a static array of `write` instructions guessing how much space each of them will be able to use, we introduce a dynamic `InstructionPlan` that can adapt to the current transaction message being packed.

Let’s see how this works with a concrete example using the current architecture.

First of all, imagine we are in the middle of planning a `TransactionPlan` from a given `InstructionPlan` and the current state of the `TransactionPlan` being built looks like this:

Instruction Plans@2x1

That is, we are inside a `ParallelTransactionPlan` that currently contains two transaction messages. One with 400 bytes left and one with very little space left (10 bytes is likely not going to be enough to add one more instruction since the program address already takes 32 bytes of space unless there is already an instruction from that program in there). To simplify this example a bit, let’s just say that each instruction requires **a base of 50 bytes** to be able to be added to a message even without any instruction data.

Now say that the next `InstructionPlan` we have to pack into this `TransactionPlan` is an `InterableInstructionPlan` for the `write` instruction we discussed before.

For this example, we’ll assume there is a **total of 1000 bytes** we need to write to our buffer account. Looking at the current state of our transaction messages, we can see that the ideal scenario would be:
- To use up the 400 bytes available in Message A.
- To skip Message B since 10 bytes isn’t enough to add a `write` instruction to this message.
- To spin up a new Message C to use up the rest of the bytes.

This is exactly how the base `TransactionPlanner` works (PR in progress but you can [see its prototype here](https://github.com/solana-program/program-metadata/blob/b9daf783c708ce220ab6f4de84da86d8e86cbe26/clients/js/src/instructionPlans/transactionPlannerBase.ts)).

First, we call `getIterator()` so we can start a stateful object keeping track of how many bytes have been successfully packed into transaction messages. Then we call `next(message)` until `hasMessage()` is false. Here’s a diagram showcasing how this algorithm works:

Instruction Plans@2x2

As you can see, before trying a new message to put our `write` instructions to, we first evaluate candidates to consider in order to fill up the spaces of valid transaction messages. We evaluate Message A and B because we are inside a `ParallelInstructionPlan` meaning the order of these instructions does not matter.
- **Message A has 400 bytes left** so — including the base 50 bytes — it means we can use 350 bytes of the total 1000 bytes we need to write to the buffer account. So only 650 bytes left.
- **Message B has 10 bytes left**. That’s not enough to add an instruction so the iterator returns `null` (or could throw an error instead) to let us know that we must try another candidate.
- Since we have no other candidates to evaluate, we create a new one — using a factory function provided by the user. Let’s assume that new messages have 1200 bytes available after they have been set up however the user wants them to be. Therefore, our new candidate, **Message C has 1200 bytes left** which is enough to fit the rest of our iterator. We include the 650 bytes left in an instruction making a total of 700 bytes (with the base 50 bytes).

We are left with the following `TransactionPlan` which matches our previous expectations exactly.

Instruction Plans@2x3

Now, regarding the `getAll()` method of the `IterableInstructionPlan`. This one could probably be removed but it allows some `SequentialInstructionPlans` to be flatten into a parent candidate when applicable. This is only useful for non-divisible sequential plans inside a parent or for sequential plans inside parallel plans. For `write` instructions, the `getAll()` function returns a single instruction using all the bytes it must write — even if that was to overflow the `TRANSACTION_SIZE_LIMIT`. The planner then gathers all the instructions within that branch and decides whether or not it can be flattened into the parent candidate.

Instruction Plans@2x4

As mentioned though, I think there is a way to remove that `getAll()` method and still make the planner work as expected. I will try to remove it in my upcoming PR that offers an implementation for the `TransactionPlanner` function. If I am able to, then I will update this PR accordingly.

Now, to go back to your message. First of all, your assumptions were all correct. However, the example you used to reason with `IterableInstructionPlans` doesn’t truly reflect why it is useful because it simply wraps an array of static instructions. This PR does offer a helper function that does this (especially for realloc instructions) but this was more an exercise to ensure the iterator is flexible enough to only be an array wrapper.

The proposed solution essentially makes the following changes:
- The `getAll()` method is replaced by a native `Iterator` such that you can use `Array.from` on it. However, this is not the most valuable part of the `IterableInstructionPlan` and — as mentioned — could potentially even be removed from the API.
- The `getIterator()` method is replaced by a `packThatMessage` function that must catch an error every time a candidate has been exhausted and return another object with a nested `packThatMessage` to continue the journey. This API feels more complex to me that this homemade `getIterator()` API that simply asks you to call `next(message)` until `hasMessage()` is false.

There is likely a better design out there but I’m not sure that the `packThatMessage` design really is an improvement here. Do let me know if the additional context sparks new design ideas though. 🙏

steveluscher:
Thanks for the detailed explainer! It helps a ton.

I think these are my lingering concerns:

1. I don't love `null` as a sigil for ‘this didn't fit.’ The reason is that an application can ignore a `null`, whereas it can't as easily ignore a thrown `SolanaError`. Perhaps, as you suggested, changing it to throw a `SolanaError` would do.
2. I think the ‘iterator’ language is throwing me off. It's not, really, an iterator because iterators' `next()` functions don't take arguments. I think this API is closer to `WritableStream::getWriter()` or something, without being mutative. If I were to name this function based on what it's literally doing it would be something like `getNextInstructionThatFitsInTransactionMessage(message)`.
3. The only way I've seen this API used so far is to add the ‘next’ instruction to the exact same message that you supplied to `next()`. The gap between getting the instruction and adding it to the instruction makes it possible for the developer to (a) forget to add it, or (b) add it to the wrong message. If the API instead returned a new `message` with the instruction added (eg. `.packNextInstruction(message)`) this could not happen.
4. Right now, if a _bunch_ of instructions fit in one message, the only way to make that so is to call `next()` a bunch of times and mutate it a bunch of times. If there were separate APIs for `packNextInstruction(message)` that packed the next instruction, and `.packNextInstructions(message)` that packed as many as would fit, maybe we could avoid this.
5. Right now, there's no way to distinguish ‘this didn't fit in the message you gave me’ vs. ‘there's no way in hell this will ever fit.’ We can end up in infinite loops if we create an API that encourages developers to keep trying and trying.

All of my suggestions above make the API more restrictive and less flexible, but with maybe fewer ways to screw it up and a little more efficiency. If we adopted those changes, can you think of any use cases that would prohibit?

I think [your example here](https://github.com/solana-program/program-metadata/blob/b9daf783c708ce220ab6f4de84da86d8e86cbe26/clients/js/src/instructionPlans/transactionPlannerBase.ts) would become:

```ts
while (!messagePacker.done()) {
  for (const candidate in candidates) {
    try {
      (candidate as Mutable<SingleTransactionPlan>).message =
        messagePacker.packNextInstructions(candidate.message);
    } catch (e) {
      if (
        isSolanaError(e, SOLANA_ERROR__MESSAGE_PACKER__NO_ROOM_LEFT_AT_THE_INN)
      ) {
        const newPlan = await context.createSingleTransactionPlan(
          [],
          context.abortSignal
        );
        transactionPlans.push(newPlan);
        candidates.push(newPlan);
        continue;
      }
      throw e;
    }
  }
}
```

lorisleiva:
Ooh I like that. Let me tackle your points one by one.

1. ✅ **`SolanaError` over `null`**. Yes, you’re absolutely right that throwing an error here is a better design.
2. ✅ **Confusing name**. I agree the “Iterator” language does not work here. I’ll come back to this after answering the other points as they affect the naming.
3. ✅ **Message package over Instruction getter**. This makes a lot of sense since we’re only ever getting an instruction to immediately pack it. This also gives more power to this dynamic instruction plan since they can decide to transform the provided message however they see fit — e.g. adding multiple instructions, enforcing a lifetime constraint, providing extra context, etc.
4. ✅ **Packing one versus multiple instructions**. Here I don’t think we need to have two API such as `packNextInstruction` and `packNextInstructions`. A simple transform function (naming TBD) that accepts a message and return a new one is enough to allow the dynamic instruction plan to be the one deciding if one or more instructions should be added to the provided candidate.
5. 🤔 **Invalid candidate errors**. Putting aside the `getAll()` function that we’re aiming to remove, there should technically never be a “there's no way in hell this will ever fit” error because that would mean the dynamic instruction plan itself is faulty and trying to fit instructions that are above the transaction size limit. That being said, if the user’s default message (prior to adding instructions from plans) never has enough space for any new instruction, then it is the responsibility of the planner to catch this. Currently, it does this by trying all candidate, then a brand new one and if that one fails, it throws. By design, the dynamic instruction plan (Program land) is decoupled from that default message (Client land) and therefore only the planner can handle this.

I think the provided example is perfect and I’ll work on updating the prototype first to see if there are any gotchas we’re not seeing before updating this stack.

Going back to the naming convention (2), I think `MessagePacker { done(), packNextInstructions() }` works but maybe it is not generic enough to describe how powerful the dynamic instruction plan is — since it can technically do whatever it wants with that message.

What about something like:

```ts
type MessageAssemblerInstructionPlan = {
  getMessageAssembler: () => MessageAssembler;
}

type MessageAssembler = {
  done: () => boolean;
  assembleNext(message: TransactionMessage) => TransactionMessage
}
```

steveluscher:
‘Assemble’ sort of gives the impression that something is being made from scratch, so I'm not sure.

* `MessagePacker`/`packMessageToCapacity()`
* `MessageFiller`/`fillMessage()`

‘Pack’ maybe has precedent 'round these parts, as in ‘pack a block,’ so maybe people understand what ‘packing’ is in this context.

lorisleiva:
@steveluscher I have managed to remove the `getAll()` function ([PR](https://github.com/solana-program/program-metadata/pull/17/files)) and update the Program Metadata prototype ([PR](https://github.com/solana-program/program-metadata/pull/18), [PR](https://github.com/solana-program/program-metadata/pull/19)) using the design changes we discussed.

Before I update this PR stack, let me know that you think! Here are some helpful links:
- The new [`MessagePackerInstructionPlan`](https://github.com/solana-program/program-metadata/blob/30f46760440304d6224525ba71836f40633142be/clients/js/src/instructionPlans/instructionPlan.ts#L35-L47) type definition.
- [The updated implementation of the base planner](https://github.com/solana-program/program-metadata/blob/main/clients/js/src/instructionPlans/transactionPlannerBase.ts#L46-L68) (simplified as much as I could).
- [Catching errors to select the next candidate](https://github.com/solana-program/program-metadata/blob/main/clients/js/src/instructionPlans/transactionPlannerBase.ts#L308-L317).
- [Throwing errors immediately after not being able to pack a new transaction message](https://github.com/solana-program/program-metadata/blob/main/clients/js/src/instructionPlans/transactionPlannerBase.ts#L340-L342).

steveluscher:
Looks great.

1. I think you might have a bug in `getLinearMessagePackerInstructionPlan()` in that it doesn't throw an error if you call `packMessageToCapacity()` when it's met the `done` criteria.
2. We'll definitely want to do `SolanaErrors` sooner than later, because those `instanceof` checks won't endure the dual package hazard.
3. I think you can raw `await Promise<T> | T` without having to wrap it in `Promise.resolve` ([link](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAKlC8UCuA7A1ig9gdxQbgCgCAKAQwGcQUBjKYgSgQD4oBvAqKazFc4KUggHZSAS37EUSADbSB5ZOiy55dAAoAnTAFtR5CAB4YLAD6x69QgF96DPEA))

lorisleiva:
Thanks!

1. Oh good to know, I'll debug this.
2. Yeah the plan is definitely to use `SolanaErrors` here. These are just placeholders for the prototype before we add them to Kit.
3. Good to know as well, I'll update accordingly. 🙏

packages/instruction-plans/src/__tests__/instruction-plan-test.ts
```diff
@@ -1,7 +1,301 @@
...
+
+describe('parallelInstructionPlan', () => {
+    it('creates ParallelInstructionPlan objects from other plans', () => {
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = parallelInstructionPlan([
+            singleInstructionPlan(instructionA),
+            singleInstructionPlan(instructionB),
+        ]);
+        expect(plan).toEqual({
```
steveluscher: All of these tests appear to pass using `toStrictEqual()` as well. Is it important to leave them open to _extra_ properties appearing without breaking a test, or is it important to defend against extra properties sneaking in?

packages/instruction-plans/src/__tests__/instruction-plan-test.ts
```diff
@@ -1,7 +1,301 @@
...
+        expect(messagePacker.done()).toBe(false);
+        expect(messagePacker.packMessageToCapacity(message).instructions[0]).toEqual(
+            createInstruction(`[0,${expectedLength})`),
+        );
+        expect(messagePacker.done()).toBe(false);
+        expect(messagePacker.packMessageToCapacity(message).instructions[0]).toEqual(
+            createInstruction(`[${expectedLength},2000)`),
+        );
+        expect(messagePacker.done()).toBe(true);
+        expect(() => messagePacker.packMessageToCapacity(message)).toThrow();
```
steveluscher, line 218: Verified: these matchers get more specific in the next PR in the stack.

packages/instruction-plans/src/instruction-plan.ts
```diff
@@ -1 +1,526 @@
...
+ * Namely the following plans are supported:
+ * - {@link SingleInstructionPlan} - A plan that contains a single instruction.
+ *   This is a simple instruction wrapper and the simplest leaf in this tree.
+ * - {@link ParallelInstructionPlan} - A plan that contains other plans that
+ *   can be executed in parallel.
+ * - {@link SequentialInstructionPlan} - A plan that contains other plans that
+ *   must be executed sequentially. It also defines whether the plan is divisible
+ *   meaning that instructions inside it can be split into separate transactions.
+ * - {@link MessagePackerInstructionPlan} - A plan that can dynamically fill
+ *  instructions into transaction messages.
```
steveluscher: Pack instructions? No big deal, I know updating PRs is a pain.
```suggestion
 * - {@link MessagePackerInstructionPlan} - A plan that can dynamically pack
 *   instructions into transaction messages.
```

packages/instruction-plans/src/instruction-plan.ts
```diff
@@ -1 +1,526 @@
...
+ *
+ * @example
+ * Non-divisible sequential plan with two instructions.
+ * ```ts
+ * const plan = nonDivisibleSequentialInstructionPlan([instructionA, instructionB]);
+ * plan satisfies SequentialInstructionPlan & { divisible: false };
+ * ```
+ *
+ * @example
+ * Sequential plan with nested parallel plans.
```
steveluscher: I'm pretty sure the way TypeDoc works is that you can name `@examples`.

This nit, everywhere it happens.

```suggestion
 * @example Simple sequential plan with two instructions.
 * ```ts
 * const plan = sequentialInstructionPlan([instructionA, instructionB]);
 * plan satisfies SequentialInstructionPlan;
 * ```
 *
 * @example Non-divisible sequential plan with two instructions.
 * ```ts
 * const plan = nonDivisibleSequentialInstructionPlan([instructionA, instructionB]);
 * plan satisfies SequentialInstructionPlan & { divisible: false };
 * ```
 *
 * @example Sequential plan with nested parallel plans.
```

packages/instruction-plans/src/instruction-plan.ts
```diff
@@ -1 +1,526 @@
...
+    divisible: boolean;
+    kind: 'sequential';
+    plans: InstructionPlan[];
+}>;
+
+/**
+ * A plan wrapping other plans that can be executed in parallel.
+ *
+ * This means direct children of this plan can be executed in separate
+ * parallel transactions without causing any side effects.
```
steveluscher: It's not so much without side effects as it is without consequences. I can still run a bunch of transactions in parallel that produce side effects, but so long as the ordering of those side effects are not consequential to each other, it's safe to run them in parallel.
```suggestion
 * parallel transactions without consequence.
```

packages/instruction-plans/src/instruction-plan.ts
```diff
@@ -1 +1,526 @@
...
+ */
+export type SingleInstructionPlan<TInstruction extends Instruction = Instruction> = Readonly<{
+    instruction: TInstruction;
+    kind: 'single';
+}>;
+
+/**
+ * A plan that can dynamically fill instructions into transaction messages.
+ *
+ * This plan provide a {@link MessagePacker} via the `getMessagePacker`
```
steveluscher:
```suggestion
 * This plan provides a {@link MessagePacker} via the `getMessagePacker`
```

packages/instruction-plans/src/instruction-plan.ts
```diff
@@ -1 +1,526 @@
...
+ * method, which enables instructions to be dynamically filled into the
+ * provided transaction message until there are no more instructions to fill.
+ * The returned {@link MessagePacker} offers a `packMessageToCapacity(message)`
+ * method that fills the provided message — when possible — and a `done()` method
+ * that checks whether there are more instructions to fill.
```
steveluscher: This suggestion, times all the places where ‘fill’ appears.

```suggestion
 * method, which enables instructions to be dynamically packed into the
 * provided transaction message until there are no more instructions to pack.
 * The returned {@link MessagePacker} offers a `packMessageToCapacity(message)`
 * method that packs the provided message — when possible — and a `done()` method
 * that checks whether there are more instructions to pack.
```

packages/instruction-plans/src/instruction-plan.ts
```diff
@@ -1 +1,526 @@
...
+    kind: 'messagePacker';
+}>;
+
+/**
+ * The message packer returned by the {@link MessagePackerInstructionPlan}.
+ *
+ * It offers a `packMessageToCapacity(transactionMessage)` method that fills as many instructions
+ * as possible into the provided transaction message, while still being able to fit into the
+ * transaction size limit. It returns the updated transaction message with the packed instructions
+ * or throws an error if the current transaction message cannot be used to pack this plan.
```
steveluscher:
```suggestion
 * or throws an error if the current transaction message cannot accommodate this plan.
```

packages/instruction-plans/src/instruction-plan.ts
```diff
@@ -1 +1,526 @@
...
+export function singleInstructionPlan(instruction: Instruction): SingleInstructionPlan {
+    return Object.freeze({ instruction, kind: 'single' });
+}
+
+function parseSingleInstructionPlans(plans: (Instruction | InstructionPlan)[]): InstructionPlan[] {
+    return plans.map(plan => ('kind' in plan ? plan : singleInstructionPlan(plan)));
+}
+
+/**
+ * Creates an {@link MessagePackerInstructionPlan} that packs instructions
```
steveluscher:
```suggestion
 * Creates a {@link MessagePackerInstructionPlan} that packs instructions
```

packages/instruction-plans/src/instruction-plan.ts
```diff
@@ -1 +1,526 @@
...
+            });
+        },
+        kind: 'messagePacker',
+    });
+}
+
+const REALLOC_LIMIT = 10_240;
+
+/**
+ * Creates an {@link MessagePackerInstructionPlan} that packs a list of realloc instructions.
```
steveluscher:
```suggestion
 * Creates a {@link MessagePackerInstructionPlan} that packs a list of realloc instructions.
```

steveluscher (approved):
Awesome! 🎉 

Added via Giphy

You don't have to take any of my nits; all optional.

Nits like: I think in the same way that you don't fill water into a bucket (rather you pour it) you don't fill instructions into a message (you pack them).

packages/instruction-plans/src/instruction-plan.ts
```diff
@@ -1 +1,526 @@
...
+                packMessageToCapacity: (message: BaseTransactionMessage & TransactionMessageWithFeePayer) => {
+                    if (offset >= totalBytes) {
+                        // TODO: Coded error (next PR).
+                        throw new Error('message packer is already done');
+                    }
+
+                    const baseTransactionSize = getTransactionMessageSize(
+                        appendTransactionMessageInstruction(getInstruction(offset, 0), message),
+                    );
+                    const maxLength =
```
steveluscher: `freeSpace`?

packages/instruction-plans/src/__tests__/instruction-plan-test.ts
lorisleiva: Good catch, this is just a mistake on my side. This should be `toStrictEqual`. 🙏

lorisleiva:
Thank you for the thorough review, this PR is much better as result! ☺️

lorisleiva:
### Merge activity

* **Jul 15, 11:09 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/533).
* **Jul 15, 11:11 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/533) rebased this pull request as part of a merge.
* **Jul 15, 11:14 AM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/533).

PR #544, opened 2025-06-13
Build with Bun

steveluscher:
#### Problem

Bun is faster than `tsup`/`esbuild`

#### Summary of Changes

Rewrite build scripts to use `bun build`

#### TODO

1. Bun does not yet do `iife` builds with exported globals
2. Bun has a bug where despite setting `"env"` to `"disable"` it will inline `process.env.NODE_ENV` anyway. https://github.com/oven-sh/bun/issues/20183
3. Bun adds a bunch of superstructure to CommonJS bundles that the current build system does not, which makes the CJS bundles fail to treeshake.
4. Bun doesn't have a way to extern all packages _except_ for a few on an allowlist. https://github.com/oven-sh/bun/issues/20380


steveluscher:
* **#544** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/544?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/544?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #574, opened 2025-06-25
Add `TransactionWithLifetime` requirement to signing functions

lorisleiva:
#### Summary of Changes

Fixes #


lorisleiva:
* **#581** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/581?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#587](https://github.com/anza-xyz/kit/pull/587) <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/587?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#594](https://github.com/anza-xyz/kit/pull/594) <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/594?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#574** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/574?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/574?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/signers/src/sign-transaction.ts
```diff
@@ -30,7 +31,10 @@ import {
 import { isTransactionSigner, TransactionSigner } from './transaction-signer';
 import { assertIsTransactionMessageWithSingleSendingSigner } from './transaction-with-single-sending-signer';
 
-type CompilableTransactionMessageWithSigners = CompilableTransactionMessage & TransactionMessageWithSigners;
+type SignableTransactionMessageWithSigners = TransactionMessage &
+    TransactionMessageWithFeePayer &
+    TransactionMessageWithLifetime &
+    TransactionMessageWithSigners;
```
lorisleiva: Note this is just a cosmetic refactoring to remove the `Compilable` internal keyword. In subsequent PRs of this stack, I intend to remove the `TransactionMessageWithLifetime` constraint because there could be a `TransactionSendingSigner` inside your message that will take care of adding a new lifetime to the transaction.

mcintyre94: This is more a question for the future PR, but are we changing the API of signers? Since they take `Transaction[]` as input currently I don't think you'd be able to define an input that doesn't have a lifetime at all. Or would you just put a placeholder there?

lorisleiva: This PR does this for the `TransactionPartialSigner` only since the others can modify the transaction. What do you think?

mcintyre94: The distinction you're drawing makes sense, I'm just a bit unclear how you'd ever construct a `Transaction` that doesn't have a lifetime. We can discuss it in the future PR though, it'll probably be obvious what you're doing when I can see the code.

lorisleiva: Because we're planning on removing the `TransactionMessageWithLifetime` constraint from the `compileTransaction` function. Therefore, we need additional type guards further down the line first so that _something_ fails statically at some point before you send a transaction that needs a lifetime.

The reason we're removing `TransactionMessageWithLifetime` from compilable messages is because it is not uncommon to want to compile a transaction message before we know which lifetime it's going to get. For instance if you want to know its size or if you want to send it to a wallet / off-chain signing that will take care of setting the lifetime for you.

My first attempt at going around this was to add "[provisory blockhash lifetime helpers](https://github.com/anza-xyz/kit/pull/519)" to fill the gap in the type. However, as Steven pointed out, we're probably better off loosening the function signatures slightly so that it better reflects how people use them.

As such, the `CompilableTransactionMessage` type will simply become `TransactionMessage & TransactionMessageWithFeePayer` which is why I'm planning on deprecating it altogether so we can remove it in v3.

That being said, this PR alone introduces a breaking change so it might need to be part of the v3 wagon as well? If so, the entire Instruction Plan feature will also need to wait for v3.

mcintyre94: I'm good with making it so you can compile a transaction message without a lifetime, but is the output of that always going to have a lifetime?

IIRC `transactionWithLifetime` was added so that code that needs to query the lifetime of a transaction can do so without decoding the message. Every `Transaction` has a lifetime in the compiled message, it's just fiddly to get to. Whereas `transactionMessageWithLifetime` distinguishes between a transaction message that does/does not have a lifetime at all. Will this stack make `transactionWithLifetime` more like the transaction message version, and mean you can have a valid `Transaction` that does not have a valid lifetime?

lorisleiva: Hmm I'm not sure I fully understand the question but the `compileTransaction` function already forwards the lifetime from the transaction message to the transaction being created. So we'll just need to have another case where, if the transaction message has no lifetime constraint, neither will the transaction returned. Is that what you meant?

mcintyre94: Yep that's what I was getting at. So the bytes that are compiled in that `Transaction` where the lifetime constraint goes, would they just be a 32-byte placeholder? And then you'd need to use `TransactionWithLifetime` to distinguish between a `Transaction` with a valid lifetime, and one with the invalid placeholder?

lorisleiva: That's right, the default behaviour when compiling a transaction message with no lifetime would essentially be an empty blockhash lifetime (32 zero bytes).

mcintyre94: Got it. Do you have any thoughts on whether we should consider that an invalid blockhash in eg `isBlockhash`? We'd currently allow it because it's a valid address. 

My concern is where you receive a `Transaction` over the wire without a `TransactionWithLifetime`, for example if it's been serialized. If we're using an all-zero blockhash as a placeholder for an invalid value, it'd be useful for apps to have an easy way to identify that in cases where they might come across it.

Annoyingly it probably is technically a possible blockhash, but we don't have any better options there.

lorisleiva: My concern with taking that stance is: where do we draw the line? There are plenty of possible invalid blockhash lifetime constraints. Do we only warn against empty ones?

But you are right, this essentially opens a can of worms where if we allow compilable transaction messages to have no explicit lifetime, then we probably shouldn't expect the reverse operation (decompiling a message) no to have one.

mcintyre94: Yea to be fair this isn't really something we're introducing as a new problem - you can already set the system address/some random address as a blockhash and not get any errors until you send. It just becomes more likely that you accidentally do that if you're using Kit and serializing transactions in your code.

mcintyre94 (approved):
LGTM!

packages/signers/src/sign-transaction.ts
lorisleiva: Maybe the PR on explicitly setting a provisory blockhash lifetime should be reconsidered?

Cc @steveluscher

steveluscher: My comments on some of the questions raised above are:

* Deciding what to write into the lifetime slice of the complied transaction is up to the transaction compiler/codec. No lifetime, all zeros. Lifetime, go ahead and write it.
* Theoretically when you give a `TransactionMessageWithLifetime` to `compileTransaction()` it should return a `TransactionWithLifetime`, and when you don't it should not.
* The type guards/assertions are misunderstood little functions. A function like `isBlockhash()` is only meant to refine types, but people rightfully presume that it's going to like contact the network or something and scan the entire blockchain for evidence of a block with that hash. Same with `isProgramDerivedAddress()`. All that does it make sure that the tuple is of the structure expected, but folks rightfully presume that it's going to check that the address is off curve. This is just a misunderstanding of what these functions are for, and at some higher level, a misunderstanding of TypeScript. tl;dr I'm totally fine with `11111111111111111111111111111111` returning `true` from `isBlockhash()`. It could be a blockhash, and is of the correct form.
* Lastly and most crucially, it's not our fault that encoding something into transaction `legacy`/`v0` wire format is lossy. Presented with only a wire transaction, you indeed have no way of knowing whether it has a lifetime, and certainly not what its `lastValidBlockHeight` is, and I would tend to draw the line at inventing sigils like ‘all zeros’ to represent ‘no lifetime’ since that's not at all a part of the message spec. Developers can always layer that kind of inference over top of their app and cast objects to `TransactionWithLifetime` as they see fit. tl;dr the way that I think about these parts of Kit is that they're designed to move common cases of developer error from runtime to build time, to the degree they succeed it's a win, but they're not magic and the constraints of the protocol make them imperfect.

steveluscher (approved):
Added via Giphy

Looks great. `Signable*` gives me a bit of pause, because of course any old garbage slop is _technically_ ’signable,’ but I don't have a better suggestion. Maybe something about ‘completeness’ or ‘conformance’ or ‘compliance.’ I don't know.

lorisleiva:
Yeah that `Signable*` type is just an internal alias but I guess they will show up in people's IDEs. Maybe I just shouldn't alias them and be explicit so we avoid introducing yet another transaction message flavour?

EDIT: I dissolved the internal `Signable*` type.

lorisleiva:
Btw, are we happy with this being a `minor` bump?

mcintyre94 (approved):

lorisleiva:
### Merge activity

* **Jul 9, 8:32 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/574).
* **Jul 9, 8:32 AM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/574).

PR #575, opened 2025-06-25
Allow Address types as PDA seeds

amilz:
Extended the Seed type to explicitly accept Address values and updated seed processing logic to handle Address seeds by encoding them as base58. Added a test to verify program derived address generation with an Address seed.

#### Problem
Currently, when users pass an unencoded `Address` to `getProgramDerivedAddress` as a seed, they encounter an unexpected error:
`SolanaError: The seed at index X with length 44 exceeds the maximum length of 32 bytes.`

This happens because:
1. The Seed type accepts string, and Address extends string, so TypeScript allows it
2. However, the runtime logic treats Address as a regular string and UTF-8 encodes it (44 bytes)
3. Instead of properly base58 encoding it to get the underlying 32-byte public key

IMO it's a little clunky that we handle encoding for text but not addresses which are very commonly used as seeds. This can create a confusing devex for someone new to Kit.

#### Summary of Changes

This PR updates the seed processing logic in getProgramDerivedAddress to:
- First checks if the `typeof seed === 'string'`. 
- If it is, check `isAddress(seed)` first - Properly detect `Address` type and base58 encode them to 32 bytes
- Then handle regular strings with UTF-8 encoding
- Else assume Uint8Array - Handle byte arrays directly

Explicitly added `Address` as a seed type.

Example Usage (now works correctly)

```typescript
const [pda, bump] = await getProgramDerivedAddress({
    programAddress: 'ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL' as Address,
    seeds: [
        ownerAddress,        // Address type - now properly decoded to 32 bytes
        tokenProgramAddress, // Address type - now properly decoded to 32 bytes  
        mintAddress,         // Address type - now properly decoded to 32 bytes
    ],
});
```

Testing
 - Added test cases for Address types in seeds
 - Verified existing string and Uint8Array functionality unchanged
 - Confirmed proper error handling for invalid inputs

#### Alternates
- Explicitly prevent Address from being a seed to ensure Address gets encoded before passing it as a seed.
- Add more explicit error messaging if an address is detected, e.g.: 

```ts
        if (typeof seed === 'string') {
            bytes = isAddress(seed)
                ? throw new SolanaError(SOLANA_ERROR__ADDRESSES__PDA_ADDRESS_SEED_NOT_ENCODED)
                : (textEncoder ||= new TextEncoder()).encode(seed);
        }
```

### DIFF
--- a/packages/addresses/src/program-derived-address.ts
+++ b/packages/addresses/src/program-derived-address.ts
@@ -88,7 +88,7 @@ type SeedInput = Readonly<{
     seed: Seed;
 }>;
 
-type Seed = ReadonlyUint8Array | string;
+type Seed = ReadonlyUint8Array | string | Address;
 
 const MAX_SEED_LENGTH = 32;
 const MAX_SEEDS = 16;
@@ -106,8 +106,16 @@ async function createProgramDerivedAddress({ programAddress, seeds }: ProgramDer
         });
     }
     let textEncoder: TextEncoder;
+    const base58EncodedAddressCodec = getAddressCodec();
     const seedBytes = seeds.reduce((acc, seed, ii) => {
-        const bytes = typeof seed === 'string' ? (textEncoder ||= new TextEncoder()).encode(seed) : seed;
+        let bytes: ReadonlyUint8Array;
+        if (typeof seed === 'string') {
+            bytes = isAddress(seed)
+                ? base58EncodedAddressCodec.encode(seed)
+                : (textEncoder ||= new TextEncoder()).encode(seed);
+        } else {
+            bytes = seed;
+        }
         if (bytes.byteLength > MAX_SEED_LENGTH) {
             throw new SolanaError(SOLANA_ERROR__ADDRESSES__MAX_PDA_SEED_LENGTH_EXCEEDED, {
                 actual: bytes.byteLength,
@@ -118,7 +126,6 @@ async function createProgramDerivedAddress({ programAddress, seeds }: ProgramDer
         acc.push(...bytes);
         return acc;
     }, [] as number[]);
-    const base58EncodedAddressCodec = getAddressCodec();
     const programAddressBytes = base58EncodedAddressCodec.encode(programAddress);
     const addressBytesBuffer = await crypto.subtle.digest(
         'SHA-256',
--- a/packages/addresses/src/__tests__/program-derived-address-test.ts
+++ b/packages/addresses/src/__tests__/program-derived-address-test.ts
@@ -86,6 +86,15 @@ describe('getProgramDerivedAddress()', () => {
             }),
         ).resolves.toStrictEqual(['6V76gtKMCmVVjrx4sxR9uB868HtZbL3piKEmadC7rSgf', 255]);
     });
+    it('returns a program derived address given a program address and a address seed', async () => {
+        expect.assertions(1);
+        await expect(
+            getProgramDerivedAddress({
+                programAddress: 'EKaNRGA37uiGRyRPMap5EZg9cmbT5mt7KWrGwKwAQ3rK' as Address,
+                seeds: ['9Lk19gWJhcS5dXXuPbDJ1A3QfNWXzxXv4EnFyGHxgEu7' as Address],
+            }),
+        ).resolves.toStrictEqual(['CsPaeTk8tn5B3guQt8yuok6uxrGEg8GFyDtHhRQkj71M', 253]);
+    });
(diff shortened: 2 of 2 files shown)


lorisleiva:
Hmm my gut feeling here tells me we should remove any string encoding altogether to avoid confusion and just request seeds as bytes.

I don't think it makes sense to use `isAddress` to distinguish between addresses and strings because theoretically, these types are not mutually exclusive.

packages/addresses/src/program-derived-address.ts
```diff
@@ -106,8 +106,16 @@ async function createProgramDerivedAddress({ programAddress, seeds }: ProgramDer
...
     }
     let textEncoder: TextEncoder;
+    const base58EncodedAddressCodec = getAddressCodec();
     const seedBytes = seeds.reduce((acc, seed, ii) => {
-        const bytes = typeof seed === 'string' ? (textEncoder ||= new TextEncoder()).encode(seed) : seed;
+        let bytes: ReadonlyUint8Array;
+        if (typeof seed === 'string') {
+            bytes = isAddress(seed)
+                ? base58EncodedAddressCodec.encode(seed)
+                : (textEncoder ||= new TextEncoder()).encode(seed);
```
steveluscher, line 115: So, generally, Kit says no to these sorts of branches in functions. The problem this creates is that if you _only_ ever pass non-addresses to this function, you'll hit the `.encode()` 0% of the time but bundle 100% of the `getAddressCodec()` code. This reduces bundle efficiency.

Now, ironically, this function _already_ bundles `getAddressCodec()` for a different purpose, so the branch you added gets this code for ‘free,’ but still.

packages/addresses/src/program-derived-address.ts
```diff
@@ -106,8 +106,16 @@ async function createProgramDerivedAddress({ programAddress, seeds }: ProgramDer
         });
     }
     let textEncoder: TextEncoder;
+    const base58EncodedAddressCodec = getAddressCodec();
     const seedBytes = seeds.reduce((acc, seed, ii) => {
-        const bytes = typeof seed === 'string' ? (textEncoder ||= new TextEncoder()).encode(seed) : seed;
+        let bytes: ReadonlyUint8Array;
+        if (typeof seed === 'string') {
+            bytes = isAddress(seed)
```
steveluscher, line 113: I agree with @lorisleiva that we probably can't do this. This makes it literally impossible to encode the string `"11111111111111111111111111111111"` to `Uint8Array(49, 49, /* ... */, 49)` if that was your _actual_ intent for some reason.

…which is actually pretty funny, because: https://github.com/helius-labs/kite-og/blob/119c5b8c4067615ba948f39473db77e38e8fa6f5/src/lib/pdas.ts#L30-L39

steveluscher (changes requested):
I think we should probably just leave this the way it is.

We could _remove_ `string` from `Seed` to make it absolutely clear that you need to supply bytes, but people seem to like passing strings:

https://github.com/Solstacker/whirlpools/blob/97e2409b7972746ff2ebea226158bb997751298b/ts-sdk/client/src/pda/position.ts#L10

On the other hand, you have folks who convert their own strings:

https://github.com/solana-foundation/gill/blob/e974f3177ef227faeba2cd1a8cc13d6d1d8438b0/packages/gill/src/programs/token-metadata/addresses.ts#L14-L18

We could do something extreme to make `Address` inadmissible:

```ts
type SeedString = Brand<string, 'SeedString'>;
type Seed = ReadonlyUint8Array | SeedString;
function seedString(putativeSeedString: string): SeedString {
    // Oof.
    if (new TextEncoder().encode(putativeSeedString, 'utf8').length > 32) {
        throw new SolanaError(/* ... */);
    }
    return putativeSeedString as SeedString;
}
getProgramDerivedAddress({
    programAddress,
    seeds: [seedString('config')],
});
```

Or just straight up:

```ts
type Seed = ReadonlyUint8Array;
function seedString(putativeSeedString: string): ReadonlyUint8Array {
    const bytes = new TextEncoder().encode(putativeSeedString, 'utf8');
    if (bytes.length > 32) {
        throw new SolanaError(/* ... */);
    }
    return bytes as ReadonlyUint8Array;
}
getProgramDerivedAddress({
    programAddress,
    seeds: [seedString('config')],
});
```

amilz:
Thanks for the analysis @steveluscher @lorisleiva.
Closing this to leave as is.

PR #576, opened 2025-06-25
Add an `OffCurveAddress` type with guard & assertion functions

### LINKED ISSUE #555: add an "is address on curve" function (and assertion)
## Motivation

There is currently no exported function to check if an address is on curve or not. The `isProgramDerivedAddress` is a type guard that checks if the structure of its input conforms to the `ProgramDerivedAddress` type. 

## Example use case

When validating address being input from various sources, especially for ATAs, it is useful to check if a provided address is on curve or not.

## Details

There is an existing `compressedPointBytesAreOnCurve` function that seems to accomplish this, but is not exported from the `@solana/addresses` package nor does it directly accept an `Address`.

nickfrosty:
#### Problem

Within Kit, there is currently no utilities for a developer to check if a given address is on curve or not.

#### Summary of Changes

- relocated the existing `curve.ts` file to be `curve-internal.ts` (per @steveluscher)
- added `isAddressOnCurve` function with TS doc comments and tests
- added `assertIsAddressOnCurve` and `assertIsAddressOffCurve`

Fixes #555


nickfrosty:
@steveluscher this PR is generally ready for review with the exception of the specific error code thrown by the assertion functions. I was wondering if you have specific thoughts on the error that should be thrown for each

packages/addresses/src/curve.ts
```diff
@@ -1,24 +1,96 @@
...
+ *         const balanceInLamports = await rpc.getBalance(address).send();
+ *     } catch (e) {
+ *         // `address` turned out to NOT be a base58-encoded off-curve address
+ *     }
+ * }
+ * ```
+ */
+export function assertIsAddressOffCurve(
+    putativeAddress: string,
+): asserts putativeAddress is Address<typeof putativeAddress> {
```
steveluscher: If this assertion _passes_ I don't actually get any more type information (ie. the curvedness of the address is immediately lost to TypeScript).

It would be more useful if there was a type that represented an _off-curve address_ which was a strict subset of Address.

```ts
function isAddressOffCurve<TAddress extends string = string>(putativeOffCurveAddress: Address<TAddress>): putativeOffCurveAddress is OffCurveAddress<TAddress>;
function assertAddressOffCurve<TAddress extends string = string>(putativeOffCurveAddress: Address<TAddress>): asserts putativeOffCurveAddress is OffCurveAddress<TAddress>;

const randomString = fetch('https://whatever/');
assertIsAddress(randomString);
assertIsOffCurveAddress(randomString);
randomString satisfies OffCurveAddress; // OK
```

…where I think `OffCurveAddress` is something like this:

```ts
export type OffCurveAddress<TAddress extends string = string> = Brand<Address<TAddress>, 'OffCurveAddress'>;
```

So that it will be accepted anywhere `Address` is.

packages/addresses/src/curve.ts
```diff
@@ -1,24 +1,96 @@
...
+ *         assertIsAddressOnCurve(address);
+ *         // At this point, `address` is an `Address` that can be used with the RPC.
+ *         const balanceInLamports = await rpc.getBalance(address).send();
+ *     } catch (e) {
+ *         // `address` turned out to NOT be a base58-encoded on-curve address
+ *     }
+ * }
+ * ```
+ */
+export function assertIsAddressOnCurve(
```
steveluscher: Do we need this, or can we just be secure in knowing that if `isAddressOffCurve()` is false, it's on curve?

packages/addresses/src/curve.ts
```diff
@@ -1,24 +1,96 @@
...
+ * // Imagine a function that fetches an account's balance when a user submits a form.
+ * function handleSubmit() {
+ *     // We know only that the input conforms to the `string` type.
+ *     const address: string = accountAddressInput.value;
+ *     try {
+ *         // If this type assertion function doesn't throw, then
+ *         // Typescript will upcast `address` to `Address`.
+ *         assertIsAddressOffCurve(address);
+ *         // At this point, `address` is an `Address` that can be used with the RPC.
+ *         const balanceInLamports = await rpc.getBalance(address).send();
```
steveluscher: I wonder if there's a better example that strikes to the heart of something that legit needs an address to be off-curve. Maybe something with creating token account instructions? I dunno.

packages/addresses/src/curve.ts
```diff
@@ -1,24 +1,96 @@
...
+ *         // `address` turned out to NOT be a base58-encoded on-curve address
+ *     }
+ * }
+ * ```
+ */
+export function assertIsAddressOnCurve(
+    putativeAddress: string,
+): asserts putativeAddress is Address<typeof putativeAddress> {
+    if (!isAddressOnCurve(putativeAddress)) {
+        throw new SolanaError(SOLANA_ERROR__ADDRESSES__INVALID_SEEDS_POINT_ON_CURVE);
```
steveluscher: We'll need to create a new error. The message for this one doesn't quite match the intent here.

https://github.com/anza-xyz/kit/blob/main/packages/errors/src/messages.ts#L259

steveluscher:
Should we get even more extreme here? Really what this _means_ is that you have two kinds of addresses on Solana; some that have private keys associated with them, and some that don't. Curvedness is a distraction from that fundamental fact.

Should we have a type that's like `AddressWithAssociatedPrivateKey` or `AddressYouCanSignFor` or `AddressOnlyProgramCanSignFor` or something that really cuts to the heart of the matter?

packages/addresses/src/curve.ts
nickfrosty: makes sense. I will add this new `OffCurveAddress` type and refactor the functions to accommodate. including adding a `offCurveAddress()` functions that mimics `address()` for completeness

packages/addresses/src/curve.ts
nickfrosty: will remove in favor of `assertIsOffCurve()` only

packages/addresses/src/curve.ts
nickfrosty: will think through and add a more useful example. i think some basic token account validation is a great use and likely the most common thing devs will use this logic for

lorisleiva:
Just to comment on the proposed naming, if we wanted to have `Address` being the "super-type" we could have `KeyPairAddress` and `ProgramDerivedAddress` both satisfying `Address`. That way we're more explicit about how that address was created.


The only issue is the current `ProgramDerivedAddress` type includes the bump so maybe we should rename that type to `ProgramDerivedAddressWithBump`?

packages/addresses/src/curve.ts
nickfrosty: new error added and used

packages/addresses/src/curve.ts
nickfrosty: updated in latest commits

nickfrosty:
> Just to comment on the proposed naming, if we wanted to have `Address` being the "super-type" we could have `KeyPairAddress` and `ProgramDerivedAddress` both satisfying `Address`. That way we're more explicit about how that address was created.
> 
> The only issue is the current `ProgramDerivedAddress` type includes the bump so maybe we should rename that type to `ProgramDerivedAddressWithBump`?

To me, this makes the most sense. imo a good candidate to consider for v3

steveluscher:
1. So it turns out that you can't stack `Brand` types. I wrote some typetests and they _all_ passed, even when invalid. To this end, I went and created a new _dimension_ type that you can add to `Address` to mark it as ‘representing an affine point that is either valid or not.’
2. I feel pretty strongly that this function shouldn't do the work of validating addresses from scratch, but rather that you should have to pass it something that is already an `Address`. I changed the input types as such.

What do you all think?

packages/addresses/src/curve.ts
```diff
@@ -33,10 +33,11 @@ export type OffCurveAddress<TAddress extends string = string> = Brand<Address<TA
 export function isOffCurveAddress(
     putativeOffCurveAddress: string,
 ): putativeOffCurveAddress is OffCurveAddress<typeof putativeOffCurveAddress> {
-    if (!isAddress(putativeOffCurveAddress)) return false;
```
steveluscher: I generally avoid this style, because when you…

1. Add a statement to the `if` block
2. Add an else block

The diff in the PR ends up being 5 lines instead of 1.

Compare:

```diff
- if (!isAddress(putativeOffCurveAddress)) return false;
+ if (!isAddress(putativeOffCurveAddress)) {
+     // <- PR adds this line
+     return false;
+ }
```

To this:
```diff
  if (!isAddress(putativeOffCurveAddress)) {
+     // <- PR adds this line
      return false;
  }
```

packages/addresses/src/curve.ts
```diff
@@ -36,7 +36,7 @@ export function isOffCurveAddress(
     if (!isAddress(putativeOffCurveAddress)) {
         return false;
     }
-    const addressBytes = new Uint8Array(getAddressCodec().encode(putativeOffCurveAddress));
```
steveluscher: Now you don't need to re-cast this.

packages/addresses/src/curve.ts
nickfrosty: Either preference works for me. I would suggest just making this a lint rule to enforce your preference here

nickfrosty:
> 1. So it turns out that you can't stack `Brand` types. I wrote some typetests and they _all_ passed, even when invalid. To this end, I went and created a new _dimension_ type that you can add to `Address` to mark it as ‘representing an affine point that is either valid or not.’
> 2. I feel pretty strongly that this function shouldn't do the work of validating addresses from scratch, but rather that you should have to pass it something that is already an `Address`. I changed the input types as such.
> 
> What do you all think?

This works for me :)

I just pushed a few small grammar fixes and fix the lint error of `@typescript-eslint/no-unnecessary-type-assertion`

steveluscher:
Thank you!

nickfrosty:
@steveluscher is there a specific estimated timeline for the next version of the library to be published?

steveluscher:
Yeah! @lorisleiva wants to finish up the InstructionPlan API and then release 3.0.

PR #581, opened 2025-06-26
Allow messages with no lifetime to be compiled

lorisleiva:
#### Summary of Changes

Fixes #


lorisleiva:
* **#581** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/581?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#587](https://github.com/anza-xyz/kit/pull/587) <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/587?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#594](https://github.com/anza-xyz/kit/pull/594) <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/594?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/581?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#574** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/574?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/transaction-messages/src/decompile-message.ts
```diff
@@ -216,7 +216,7 @@ export type DecompileTransactionMessageConfig = {
 export function decompileTransactionMessage(
     compiledTransactionMessage: CompiledTransactionMessage,
     config?: DecompileTransactionMessageConfig,
-): CompilableTransactionMessage {
+): BaseTransactionMessage & TransactionMessageWithFeePayer & TransactionMessageWithLifetime {
```
mcintyre94, line 219: I'm not sure about returning `TransactionMessageWithLifetime` here. My concern is that if you compile a message without a lifetime, and then decompile it, should we say it now has a lifetime?

Maybe, it's going to have an all-zero blockhash in the `lifetimeConstraint` field. But I wonder if it'd be better to make apps narrow to `TransactionMessageWithLifetime` themselves after doing their own checks.

mcintyre94 (approved):
Nice, this looks great!

packages/transaction-messages/src/decompile-message.ts
lorisleiva: Yeah this is the part where I don't think either solution is ideal.

The reality is, whenever we decompile, we always fill in the `lifetimeConstraint` property of the transaction message, even if it's all zeros. So structurally we do _always_ return a `TransactionMessageWithLifetime`.

However, you are right that it feels weird that compiling/decompiling a message magically introduces the `TransactionMessageWithLifetime` flag.

_But_, decompiling a message will typically be used to find information about a transaction that landed and therefore it would be a shame to make apps always use `isTransactionMessageWithBlockhashLifetime` and `isTransactionMessageWithDurableNonceLifetime` when the data is technically always there.

My interpretation of this is kinda like the `From` and `To` type parameters of `Codecs`. `From` is allowed to be looser than `To` because the encoder can fill some gaps. However, decoding is stricter because we cannot fill any gaps if information is missing from the provided bytes.

I'm happy with either solution though.

mcintyre94: I think you're probably right. There are use cases like wallets and relayers where you're dealing with serialized transactions that haven't been landed, but that is a minority and those apps need to be more sophisticated anyway. The comparison to codecs makes sense too.

packages/transaction-messages/src/compile/message.ts
```diff
@@ -46,6 +49,11 @@ type VersionedCompiledTransactionMessage = BaseCompiledTransactionMessage &
         version: number;
     }>;
 
+const EMPTY_BLOCKHASH_LIFETIME_CONSTRAINT = {
+    blockhash: '11111111111111111111111111111111',
+    lastValidBlockHeight: 0n,
```
steveluscher, line 54: I know that `getCompiledLifetimeToken()` ignores this value so it can be literally anything, but this is maybe confusing to the reader (ie. that it's here).

We could obviate the need for this by doing…

```ts
const LIFETIME_TOKEN_WHEN_LIFETIME_ABSENT = '11111111111111111111111111111111';

/* ... */
    lifetimeToken: transactionMessage.lifetimeConstraint
        ? getCompiledLifetimeToken(transactionMessage.lifetimeConstraint)
        : LIFETIME_TOKEN_WHEN_LIFETIME_ABSENT,
```

steveluscher (approved):
Added via Giphy

packages/transaction-messages/src/compile/message.ts
```diff
@@ -58,26 +66,29 @@ type VersionedCompiledTransactionMessage = BaseCompiledTransactionMessage &
...
+        (transactionMessage as Partial<TransactionMessageWithLifetime>).lifetimeConstraint ??
+        EMPTY_BLOCKHASH_LIFETIME_CONSTRAINT;
     return {
         ...(transactionMessage.version !== 'legacy'
             ? { addressTableLookups: getCompiledAddressTableLookups(orderedAccounts) }
             : null),
         header: getCompiledMessageHeader(orderedAccounts),
         instructions: getCompiledInstructions(transactionMessage.instructions, orderedAccounts),
-        lifetimeToken: getCompiledLifetimeToken(transactionMessage.lifetimeConstraint),
+        lifetimeToken: getCompiledLifetimeToken(lifetimeConstraint),
```
steveluscher, line 91: Honestly, I thought we'd just make this optional on `CompiledTransactionMessage` and then make the _compiler_ (ie. `getCompiledTransactionMessageEncoder()`) encapsulate the behaviour of filling it in when absent.

Actually I think I like that better, because then we're not passing a value with a `"11111111111111111111111111111111"` in it around your app.

steveluscher (changes requested):
Yeah, actually maybe I take that back.

There's two levels of compiling here:

1. From a nicely formatted message to the unreadable garbage with account indexes and stuff (ie. `compileTransactionMessage()`)
2. From the garbage format to wire format (ie. `getCompiledTransactionMessageEncoder()`)

I think the behaviour of ‘fill it in with zeros’ should be done at the very latest opportunity – the one where we no longer have a choice. The higher level data structure should just _not have a `lifetimeToken` property_ if there's no lifetime, rather than to fill it in with a placeholder.

lorisleiva:
@steveluscher Hmm implementing this is making me doubt this is the right decision. It makes the whole "what happens when we decompile" problem even deeper — unless we recognise "zeroes" as an empty sigil which you said you'd rather avoid.

Consider the following encoding (A) and decoding (B) paths.

```
A) TransactionMessage → CompiledTransactionMessage → Bytes
B) Bytes → CompiledTransactionMessage → TransactionMessage
```

When we don't have a a lifetime, we get the following back and forth:
- A) `TransactionMessage` with no lifetime → `CompiledTransactionMessage` with no lifetime → Bytes filled with zeroes
- B) Bytes filled with zeroes → `CompiledTransactionMessage` **always with a blockhash lifetime even though the attribute is optional!**  → `TransactionMessage` with blockhash lifetime.

I know we currently have this issue anyway, that is:
- A) `TransactionMessage` with no lifetime → `CompiledTransactionMessage` with blockhash lifetime → Bytes filled with zeroes
- B) Bytes filled with zeroes → `CompiledTransactionMessage` with blockhash lifetime  → `TransactionMessage` with blockhash lifetime.

But I found there is something weird about optionalising the lifetime attribute of the `CompiledTransactionMessage` — which is the low-level / close-to-the-actual-bytes format. It breaks away from the [`From`/`To` analogy of codecs](https://github.com/anza-xyz/kit/pull/581#discussion_r2169059664) I was mentioning to Callum. We could introduce another type for path (B) but the only difference would be that the `lifetimeConstraint` attribute will be required.

What do you think?

P.S.: I submitted [a PR that tries this strategy](https://github.com/anza-xyz/kit/pull/587) but with a zero sigil to avoid the issue above. Lmk what you think.

mcintyre94:
Taking a step back, do we actually need to be able to get to `CompiledTransactionMessage` without a lifetime?

The use case mentioned was if you have a transaction modifying signer, it might add a lifetime. So you want to be able to compile without having one so the signer can take care of it. We also know there are cases where an app might use specific logic to 'fake' its lifetime, but because of the limitation of the bytecode format it does have a lifetime when compiled.

Can we differentiate the signer case further here?

If I'm just using `compileTransactionMessage` then it still requires a lifetime, because it's just a compile step. Whatever lifetime the transaction message has ends up in the `CompiledTransactionMessage`. If the app wants to special-case the lifetime because it's going to do something else after serializing, that's fine, but it's just app logic. It wants to compile so it sets some lifetime.

But if I'm using signers then I can enter the compile flow without a lifetime. The transaction modifying signers get their chance to modify the transaction, and one of them might add a lifetime. Maybe it's an app-specific special case, maybe it's a real one. But if they add a lifetime then great, that ends up in the `CompiledTransactionMessage`. If none of them do, then we throw an error because you can't actually go to `CompiledTransactionMessage` without some lifetime. Either your app or a signer must add one.

The design challenge here is the signer interface operates on `Transaction`, but I wonder if that's where we should be looking to address this use case.

lorisleiva:
Hey @mcintyre94, I'm not sure what you are suggesting about the Signer API, but I believe this was tackled [on this PR](https://github.com/anza-xyz/kit/pull/582). Unless you had something different in mind?

Regarding the lack of `lifetimeToken` in `CompiledTransactionMessage`, I also personally don't find this change ideal because a compiled message always has a dedicated 32-bytes slot for that token. It feels weird to me to make this optional only to push the "fill with zeroes" logic at the Codec level. I did however [implement this suggestion here](https://github.com/anza-xyz/kit/pull/587) so we can see the consequences of that change explicitly rather than speculate on it. ☺️

mcintyre94:
I think my suggestion would be a larger change to `signTransactionMessageWithSigners`, but nothing else would change. `CompiledTransactionMessage` still requires lifetime token, Kit has no logic to default it anywhere, `compileTransactionMessage` still requires a lifetime.

I think the only case where this is really necessary is when you're using signers, and a signer is going to modify the transaction. Anything else, like excluding the lifetime for Solana Pay, is app-specific logic IMO.

So I think you could make a signer method something like:

- For each modifying signer, modify the transaction message
- Assert it has a lifetime after this point, throw an error if it does not
- Compile the resulting transaction message using an unchanged `compileTransactionMessage`
- Sign the transaction for all signers

The concrete API change would be to signers, a modifying signer would need to have a function with the signature `T extends TransactionMessage ->  Promise<U extends TransactionMessage>` at this step. Instead of only `T extends Transaction -> Promise<T>`.

lorisleiva:
@mcintyre94 Oof this is a very different direction haha.

I don't particularly agree with your premise that this is only something only the Signer API must handle. There are valid use cases that require apps to compile transaction messages before knowing which lifetime they are going to use. For instance, if they want to figure out the size of that transaction message to figure out if they can fit more instructions into it. Or anything that they could do with the Signer API — e.g. sending one to a wallet — but directly with the wallet's API.

I do believe there is a strong need for `compileTransactionMessage` to accept a `TransactionMessage` with no lifetime. This is what started this whole journey and is a blocker to `InstructionPlans`. The main question here is: how far do we want to push that lifeless border. Personally, I'd recommend stopping at `CompiledTransactionMessage` since it clearly requires 32 bytes for a `lifetimeToken`.

Regarding the suggested `TransactionModifyingSigner` change, I don't think it is viable to use a signature such as `T extends TransactionMessage ->  Promise<U extends TransactionMessage>` for them. `TransactionModifyingSigners` could technically be stacked one after the other meaning we must use the `Transaction` API in order to avoid having to re-compile each time we sign. More importantly, the `TransactionMessage` API doesn't store any signature dictionary which is needed for `TransactionSigners` to do their job. They ultimately just fill that signature directory with one or more signature based on the type of wallet they are.

lorisleiva:
@steveluscher I'm not sure if you noticed but [the next PR](https://github.com/anza-xyz/kit/pull/594) of this stack now tackles the changes we discussed. Lmk if that's also what you expected. 🙏

lorisleiva:
### Merge activity

* **Jul 9, 8:32 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/581).
* **Jul 9, 8:33 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/581) rebased this pull request as part of a merge.
* **Jul 9, 8:36 AM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/581).

PR #605, opened 2025-07-04
Add details about creating and verifying signatures to the ‘Key pairs’ docs

steveluscher:
#### Summary of Changes

Added:

- Example of using a private key to sign a message
- Example of using a private key to sign a transaction
- Example of using a public key to verify a signature
- Example of using a public key to derive a Solana address
- Instructions on how to use the `Signature` type guards/assertions

### DIFF
--- a/docs/content/docs/concepts/keypairs.mdx
+++ b/docs/content/docs/concepts/keypairs.mdx
@@ -167,9 +167,9 @@ const loadedKeyPair = await new Promise<CryptoKeyPair>((resolve, reject) => {
 You can obtain the 32 bytes of any public key like this:
 
 ```ts twoslash
-const publicKey = null as unknown as CryptoKey;
+const keyPair = null as unknown as CryptoKeyPair;
 // ---cut-before---
-const publicKeyBytes = new Uint8Array(await crypto.subtle.exportKey("raw", publicKey));
+const publicKeyBytes = new Uint8Array(await crypto.subtle.exportKey("raw", keyPair.publicKey));
 ```
 
 Exporting private key material requires a specially constructed `CryptoKey` with its extractable property set to `true`.
@@ -186,9 +186,9 @@ const keyPair = await crypto.subtle.generateKey(
 You can then use that `CryptoKey` with the [`SubtleCrypto#exportKey`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/exportKey) API. The last 32 bytes of a PKCS#8 export of the key are the private key bytes.
 
 ```ts twoslash
-const privateKey = null as unknown as CryptoKey;
+const keyPair = null as unknown as CryptoKeyPair;
 // ---cut-before---
-const exportedPrivateKey = await crypto.subtle.exportKey("pkcs8", privateKey);
+const exportedPrivateKey = await crypto.subtle.exportKey("pkcs8", keyPair.privateKey);
 const privateKeyBytes = new Uint8Array(exportedPrivateKey, exportedPrivateKey.byteLength - 32, 32);
 ```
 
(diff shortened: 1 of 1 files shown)


steveluscher:
* **#605** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/605?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/605?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/content/docs/concepts/keypairs.mdx
```diff
@@ -206,29 +206,174 @@ Private keys can be used by their owner to produce a digital signature of a spec
...
+```
+
+## Signatures
+
+The [`SignatureBytes`](/api/type-aliases/SignatureBytes) type represents a 64-byte Ed25519 signature, and the [`Signature`](/api/type-aliases/Signature) type represents such a signature as a base58-encoded string.
+
+When you acquire a string that you expect to be a base58-encoded signature (eg. of a transaction) from an untrusted network API or user input you can assert it is in fact a base58-encoded byte array of sufficient length using the [`assertIsSignature()`](/api/functions/assertIsSignature) function.
+
+```ts twoslash
+// @noErrors: 1308
```
steveluscher, line 318: This is to suppress the ‘top level await not allowed’ error. If anyone knows a better way to do this let me know.

docs/content/docs/concepts/keypairs.mdx
```diff
@@ -206,29 +206,174 @@ Private keys can be used by their owner to produce a digital signature of a spec
...
+The Solana address associated with any public key can be derived using the [`getAddressFromPublicKey()`](/api/functions/getAddressFromPublicKey) function.
+
+```ts twoslash
+const publicKey = null as unknown as CryptoKey;
+// ---cut-before---
+import { getAddressFromPublicKey } from "@solana/kit";
+const address = await getAddressFromPublicKey(publicKey);
+```
+
+## Signatures
```
steveluscher, line 311: Let me know if you think this entire section is necessary. I didn't know whether this was too much detail or if it's actually important to go over the guard/assertion functions for every type in the library.

docs/content/docs/concepts/keypairs.mdx
```diff
@@ -206,29 +206,174 @@ Private keys can be used by their owner to produce a digital signature of a spec
...
 ### Signing messages
 
-TODO
+Any data that you can serialize as a `Uint8Array` can be signed with a `CryptoKey`.
+
+```ts twoslash
+import { ReadonlyUint8Array } from "@solana/kit";
+// ---cut-before---
+import { getU32Encoder } from "@solana/kit";
+const messageFromText = new TextEncoder().encode("🎉");
```
lorisleiva: Should we try and use codecs in the documentation wherever possible for consistency?

```suggestion
const messageFromText = getUtf8Encoder().encode("🎉");
```

P.S.: Same comment for every other usage of `TextEncoder` in this file.

docs/content/docs/concepts/keypairs.mdx
```diff
@@ -206,29 +206,174 @@ Private keys can be used by their owner to produce a digital signature of a spec
...
+```ts twoslash
+import { Transaction } from "@solana/kit";
+const transaction = null as unknown as Transaction;
+const bobsKeyPair = null as unknown as CryptoKeyPair;
+// ---cut-before---
+import { signTransaction } from "@solana/kit";
+const signedTransaction = await signTransaction([bobsKeyPair], transaction);
+```
+
+In practice, it's rare to sign transactions like this. Typically, you create a transaction message in your application, specify which accounts are required to sign it using the [Signers API](/concepts/signers), and then call [`signTransactionMessageWithSigners`](/api/functions/signTransactionMessageWithSigners) to turn it into a signed `Transaction`.
```
lorisleiva, line 252: 🫶

docs/content/docs/concepts/keypairs.mdx
```diff
@@ -206,29 +206,174 @@ Private keys can be used by their owner to produce a digital signature of a spec
...
+const mallorysSignature = null as unknown as SignatureBytes;
+const signature = null as unknown as SignatureBytes;
+const textEncoder = new TextEncoder();
+const encode = textEncoder.encode.bind(textEncoder);
+// ---cut-before---
+// False claim that Bob signed the message when it was in fact Mallory
+await verifySignature(bobsPublicKey, mallorysSignature, encode("The meeting is at 6:00pm"));
+//                                   ^^^^^^^^^^^^^^^^^
+// False claim about the contents of the message
+await verifySignature(bobsPublicKey, bobsSignature, encode("The meeting is at 6:00am"));
```
lorisleiva: Yeah it would be nice if these were throwing on v3.

docs/content/docs/concepts/keypairs.mdx
```diff
@@ -206,29 +206,174 @@ Private keys can be used by their owner to produce a digital signature of a spec
...
 ### Deriving addresses
 
-TODO
+The Solana address associated with any public key can be derived using the [`getAddressFromPublicKey()`](/api/functions/getAddressFromPublicKey) function.
+
+```ts twoslash
+const publicKey = null as unknown as CryptoKey;
+// ---cut-before---
+import { getAddressFromPublicKey } from "@solana/kit";
+const address = await getAddressFromPublicKey(publicKey);
```
lorisleiva: super nit: Since people most likely have `CryptoKeyPair` instances to work with in their app, it could be useful to use `keyPair.publicKey` in the example? 🤔

docs/content/docs/concepts/keypairs.mdx
lorisleiva: I think this is useful.

docs/content/docs/concepts/keypairs.mdx
steveluscher: Ooh smart. Then I don't have to answer questions about why `TextEncoder` is not available in React Native.

docs/content/docs/concepts/keypairs.mdx
steveluscher: I hate myself.

PR #500, opened 2025-05-15
Trick Changesets into pushing tags and publishing Github releases after our custom publish script completes

steveluscher:
Fixes #499

PR #639, opened 2025-07-09
Actually fix the publish script

steveluscher:
#### Problem

OK, so turns out #500 was never tested until now. The reason this doesn't work is because of some [aggressively adversarial developer tooling](https://github.com/changesets/changesets/blob/b57c77f8648daa11357547dda8bb85643049cc63/packages/release-utils/src/run.ts#L33). We can't presume that this is just a shell script here.

#### Summary of Changes

- Create a `pnpm` script just for Changesets that encapsulates all the commands that need to be run.
- Specify that one command to the changesets action

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -52,16 +52,13 @@ jobs:
         id: changesets
         uses: changesets/action@06245a4e0a36c064a573d4150030f5ec548e4fcc # v1.4.10
         with:
-          # * Some tasks slow down considerably on GitHub Actions runners when concurrency is high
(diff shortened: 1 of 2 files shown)


steveluscher:
* **#639** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/639?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/639?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #646, opened 2025-07-09
Fixed reading 'length' of undefined

### LINKED ISSUE #317: undefined in getHumanReadableErrorMessage()
https://github.com/anza-xyz/kit/blob/06c6fe0111b8f07e4c2dcb181d2d988f222cb9ac/packages/errors/src/message-formatter.ts#L22

Sorry, first time reporter, IDK how to format this properly :)

In case code variable is not one of the recognised errors, this fails with

 undefined is not an object (evaluating 'messageFormatString.length')

gcangussu:
Fixed `TypeError: Cannot read properties of undefined (reading 'length')` at getHumanReadableErrorMessage() (packages/errors/src/message-formatter.ts)

#### Problem

When the code is not found on the `SolanaErrorMessages` object, we'll have `messageFormatString = undefined`. Then we'll crash at `messageFormatString.length` on the following line.

#### Summary of Changes

Fallback to an empty string in case the code is not found on the object, as this is what the code assumes it'll have.

Fixes #317

### DIFF
--- a/packages/errors/src/message-formatter.ts
+++ b/packages/errors/src/message-formatter.ts
@@ -18,7 +18,7 @@ export function getHumanReadableErrorMessage<TErrorCode extends SolanaErrorCode>
     code: TErrorCode,
     context: object = {},
 ): string {
-    const messageFormatString = SolanaErrorMessages[code];
+    const messageFormatString = SolanaErrorMessages[code] ?? "";
     if (messageFormatString.length === 0) {
         return '';
     }


gcangussu:
> What did you do to end up in a situation where the code was not found in the `SolanaErrorMessages` object?

Thank you for the question Steven. I was calling `getSignaturesForAddress` on a loop to consume all the signatures for a constant token mint address. Simplifying my code, what I was doing is basically the below code:

```js
import { createSolanaRpc, mainnet } from "@solana/kit";

const rpc = createSolanaRpc(
  mainnet(`https://mainnet.helius-rpc.com/?api-key=${HELIUS_API_KEY}`)
);

let oldestSignature;

while (true) {
  const signaturesForAddress = await rpc
    .getSignaturesForAddress(TOKEN_MINT, {
      before: oldestSignature,
      commitment: "finalized",
      limit: 1000,
    })
    .send();

  // [...] (save to database)

  oldestSignature = signaturesForAddress.at(-1)?.signature;
  if (!oldestSignature) {
    break;
  }
}
```

The error code I got from the RPC node was `-32019`. The `@solana/kit` was version `2.2.1`.

steveluscher (commented):
Thanks for the pointer. The real bug here is that we're missing a bunch of codes that were added:

https://github.com/anza-xyz/agave/blob/4807a7a0e51148acd1b0dd0f3f52a12c40378ed3/rpc-client-api/src/custom_error.rs#L27-L29

I'll do that now.

steveluscher:
https://github.com/anza-xyz/kit/pull/658

gcangussu:
@steveluscher Wouldn't be best to avoid a hard crash in the case the library is outdated (either because the user didn't update in a while or in this case that the library code is behind RPC code)? It is even more serious because it is a TypeError, it is common to not retry TypeErrors as it indicates an issue on the code.

I think most people would expect the underlying @solana/kit code to not throw TypeErrors, unless they are directly related to parameters passed by the caller.

steveluscher:
You're right that we need _some_ sort of strategy here.

I don't want the strategy to be ‘quietly bail on the error text’ though.

Can you think of a solution that:

1. Has minimal impact on bundle size
2. Gives the developer enough information to recover from the _actual_ error
3. Gives the developer further information on what to do to have the _actual_ error message materialize on the console _and_ in the `npx @solana/errors decode` CLI?

gcangussu:
We could instead of using an empty string, default to some sort of fallback message, like

_Unknown error code '{code}'_

To check number 3, we could include more information, like

_Unknown error code '{code}'. Please update your Solana packages or try `npx @solana/errors@latest decode -- {code}` to retrieve the actual error message_

What do you think?

PR #647, opened 2025-07-09
Add `SolanaErrors` for `MessagePacker`

lorisleiva:
This PR adds two new `SolanaError` codes that can be used by `MessagePackers` to indicate that:
- The provided transaction message cannot be filled with any instructions.
- The message packer is already done and the `packMessageToCapacity` function should no longer be called.


lorisleiva:
* **#648** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#647** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/647?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/647?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#533** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/533?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/errors/src/messages.ts
```diff
@@ -391,6 +393,10 @@ export const SolanaErrorMessages: Readonly<{
     [SOLANA_ERROR__INSTRUCTION_ERROR__UNKNOWN]: '',
     [SOLANA_ERROR__INSTRUCTION_ERROR__UNSUPPORTED_PROGRAM_ID]: 'Unsupported program id',
     [SOLANA_ERROR__INSTRUCTION_ERROR__UNSUPPORTED_SYSVAR]: 'Unsupported sysvar',
+    [SOLANA_ERROR__INSTRUCTION_PLANS__CANNOT_PACK_USING_PROVIDED_MESSAGE]:
+        'Cannot pack the next instructions using the provided message.',
+    [SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_PACKER_IS_ALREADY_DONE]:
+        'Failed to pack the next message because the message packer is already done.',
```
steveluscher: I wonder if we can make it even more clear what ‘done’ means in this context.
```suggestion
    [SOLANA_ERROR__INSTRUCTION_PLANS__PLAN_ALREADY_COMPLETE]:
        'No more instructions to pack; the message packer has completed the instruction plan',
```

packages/errors/src/messages.ts
```diff
@@ -391,6 +393,10 @@ export const SolanaErrorMessages: Readonly<{
     [SOLANA_ERROR__INSTRUCTION_ERROR__UNKNOWN]: '',
     [SOLANA_ERROR__INSTRUCTION_ERROR__UNSUPPORTED_PROGRAM_ID]: 'Unsupported program id',
     [SOLANA_ERROR__INSTRUCTION_ERROR__UNSUPPORTED_SYSVAR]: 'Unsupported sysvar',
+    [SOLANA_ERROR__INSTRUCTION_PLANS__CANNOT_PACK_USING_PROVIDED_MESSAGE]:
+        'Cannot pack the next instructions using the provided message.',
```
steveluscher: In all cases, this is because it won't fit, right?
```suggestion
    [SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN]:
        'The provided message has insufficient capacity to accommodate the next instruction(s) in this plan',
```

steveluscher (approved):
Added via Giphy

packages/errors/src/context.ts
```diff
@@ -394,6 +395,9 @@ export type SolanaErrorContext = DefaultUnspecifiedErrorContextToUndefined<
             index: number;
             instructionErrorContext?: unknown;
         };
+        [SOLANA_ERROR__INSTRUCTION_PLANS__CANNOT_PACK_USING_PROVIDED_MESSAGE]: {
+            transactionMessage: object;
+        };
```
steveluscher: We have this problem a bit already because some of the context properties contain awkward objects like `Uint8Array`, but this is the first instance of `object` types in the error context. A few thoughts here.

1. We have to presume that all of this data will be logged, since it gets crapped out on the console and in our error decoder. Have a think on whether this is OK, given all of the stuff that messages and plans contain.
2. The error decoder (eg. `npx @solana/errors decode ...`) link for this thing will be massive.
3. A lot of this data won't stringify very well. `TransactionMessage` objects will have functions in them if they contain signers, etc.

Now of course, there are two uses for context.

1. For apps that catch the errors
2. For the error message decoder/printer.

I wonder if there's a trick we could do here, which is to make complex data like this _non-enumerable_ so that you can read it in app, but the message printer (which uses [`Object.entries()`](https://github.com/anza-xyz/kit/blob/efc90c990a03de8f5ec8214ac0621e6a2576545e/packages/errors/src/context.ts#L649)) will skip over it.


```ts
const context = {};
Object.defineProperty(context, 'transactionMessage', {
    configurable: false,
    enumerable: false,
    value: transactionMessage,
    writable: false,
});
```

lorisleiva: Oh sorry I wanted to flag this in the PR before requesting review but forgot. It was mostly about the fact that we don't even have the type definition of `TransactionMessage` inside the error context so perhaps it makes it even less useful.

I think that in all these cases we _could_ get away with not providing any `object` in the context whatsoever because that information is likely not going to be used dynamically by the app. It's more additional information to help developers debug planning failures than information that can help the app land on its feet.

So a few options here:
- We remove these `object` types in the context of errors.
- We keep them but add them as non-writable as suggested.
- We try to encode/stringify the information that helps debugging. For instance, we could provide the transaction message as bytes and, when passing instruction plans, we could instead provide a path to the plan that caused the issue (e.g. `sequential[0].parallel[3].single`).

What do you think?

packages/errors/src/messages.ts
lorisleiva: This error is very specific to the `MessagePackerInstructionPlan` since aside from that, there isn't another plan that has the concept of being complete. How about `SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_PACKER_ALREADY_COMPLETE`?

steveluscher: Sure! In any case, the  error message bothered me more than the name of the error code.

packages/errors/src/context.ts
steveluscher: I think removing it for now, as you've done, is the right move – for now. Error messages should lay out what the developer needs to do to remediate the problem, and just dumping the entire transaction message in their lap and being like ‘you figure it out’ probably doesn't help as much as we think it does.

Now that I'm taking a second look at it, we should just do this, right?

```ts
        [SOLANA_ERROR__INSTRUCTION_PLANS__CANNOT_PACK_USING_PROVIDED_MESSAGE]: {
            numBytesRequired: number;
            numFreeBytes: number
        };
```

packages/errors/src/messages.ts
```diff
@@ -391,6 +393,10 @@ export const SolanaErrorMessages: Readonly<{
     [SOLANA_ERROR__INSTRUCTION_ERROR__UNKNOWN]: '',
     [SOLANA_ERROR__INSTRUCTION_ERROR__UNSUPPORTED_PROGRAM_ID]: 'Unsupported program id',
     [SOLANA_ERROR__INSTRUCTION_ERROR__UNSUPPORTED_SYSVAR]: 'Unsupported sysvar',
+    [SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN]:
+        'The provided message has insufficient capacity to accommodate the next instruction(s) in this plan.',
```
steveluscher: Yeah. Totally, we should like, say how many bytes were required and how many were found to be free.

lorisleiva:
### Merge activity

* **Jul 15, 11:09 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/647).
* **Jul 15, 11:15 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/647) rebased this pull request as part of a merge.
* **Jul 15, 11:18 AM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/647).

PR #648, opened 2025-07-09
Add `createBaseTransactionPlanner` function

lorisleiva:
This PR provides a base implementation for the `TransactionPlanner` type that tries to pack transactions in the most optimal way possible.

This function accepts a `createTransactionMessage` callback that dictates how each new transaction message should be created before filling them with instructions. Here's an example:

```ts
const planner = createBaseTransactionPlanner({
  createTransactionMessage: () =>
    pipe(
      (message) => setTransactionMessageFeePayerSigner(payer, message),
      (message) => setTransactionMessageComputeUnitPrice(cuPrice, message),
      // ...
    ),
});

const transactionPlan = await planner(instructionPlan);
```

An optional `onTransactionMessageUpdated` callback can also be provided and will be executed any time something changes in a transaction message being planned (for instance, when new instructions are being added to the message). This allows users to hook logic into the planning process.

Note, the function is called `createBaseTransactionPlanner` because I was initially planning on having a high-level `createDefaultTransactionPlanner` that is all battery included but I'm thinking this interface is already good enough to be used as a high-level. So **maybe I should rename this to `createTransactionPlanner`? Wdyt?**


lorisleiva:
* **#648** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#647** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/647?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#533** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/533?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/errors/src/codes.ts
```diff
@@ -257,6 +257,12 @@ export const SOLANA_ERROR__TRANSACTION_ERROR__UNBALANCED_TRANSACTION = 7050036;
 // Reserve error codes in the range [7618000-7618999].
 export const SOLANA_ERROR__INSTRUCTION_PLANS__CANNOT_PACK_USING_PROVIDED_MESSAGE = 7618000;
 export const SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_PACKER_IS_ALREADY_DONE = 7618001;
+export const SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_FIT_INSTRUCTION_PLAN_IN_NEW_MESSAGE = 7618002;
```
steveluscher: Isn't this semantically the same as `SOLANA_ERROR__INSTRUCTION_PLANS__CANNOT_PACK_USING_PROVIDED_MESSAGE` (which I suggested to be renamed to `SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN ` in #647)?

packages/instruction-plans/src/transaction-planner-base.ts
```diff
@@ -0,0 +1,416 @@
...
+    context: TraverseContext,
+): Promise<MutableTransactionPlan | null> {
+    const candidates: MutableSingleTransactionPlan[] = [...context.parentCandidates];
+    const transactionPlans: TransactionPlan[] = [];
+
+    // Reorder children so message packer plans are last.
+    const sortedChildren = [
+        ...instructionPlan.plans.filter(plan => plan.kind !== 'messagePacker'),
+        ...instructionPlan.plans.filter(plan => plan.kind === 'messagePacker'),
+    ];
```
steveluscher: I think the algorithmic complexity drops if you do this in one pass.


```suggestion
    const sortedChildren = instructionPlan.plans.toSorted((a, b) =>
        (a.kind === 'messagePacker') - (b.kind === 'messagePacker')
    );
```

If we're concerned for how widely `toSorted()` is supported, `Array.from(instructionPlan.plans).sort(...)`.

packages/errors/src/codes.ts
```diff
@@ -257,6 +257,12 @@ export const SOLANA_ERROR__TRANSACTION_ERROR__UNBALANCED_TRANSACTION = 7050036;
...
+export const SOLANA_ERROR__INSTRUCTION_PLANS__MISSING_INSTRUCTIONS_IN_INSTRUCTION_PLAN = 7618003;
```
steveluscher: This error is thrown any time `traverse()` returns `null`. Are there reasons _other_ than ‘there being no instructions’ that `traverse()` would return `null`? Would it suffice for the error to communicate something more along the lines of ‘planner produced empty plan’ or ‘planning failed’ or something?

packages/instruction-plans/src/transaction-planner-base.ts
```diff
@@ -0,0 +1,416 @@
...
+            const messagePacker = instructionPlan.getMessagePacker();
+            while (!messagePacker.done()) {
+                try {
+                    newMessage = messagePacker.packMessageToCapacity(message);
+                    if (getTransactionMessageSize(newMessage) > TRANSACTION_SIZE_LIMIT) {
+                        throw new SolanaError(
+                            SOLANA_ERROR__INSTRUCTION_PLANS__CANNOT_FIT_ENTIRE_INSTRUCTION_PLAN_INSIDE_MESSAGE,
+                            { instructionPlan, transactionMessage: newMessage },
+                        );
+                    }
```
steveluscher: Doesn't `packMessageToCapacity()` already assert this internally and throw `7618000`?

packages/errors/src/codes.ts
```diff
@@ -257,6 +257,12 @@ export const SOLANA_ERROR__TRANSACTION_ERROR__UNBALANCED_TRANSACTION = 7050036;
...
+export const SOLANA_ERROR__INSTRUCTION_PLANS__CANNOT_FIT_ENTIRE_INSTRUCTION_PLAN_INSIDE_MESSAGE = 7618004;
```
steveluscher: It's not clear to me how this error is semantically different enough from `7618000`, especially if we rename `7618000` to `SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN`. I think a ‘failure to accommodate plan’ describes all of the cases where `7618000` and `7618004` are presently thrown.

packages/errors/src/codes.ts
```diff
@@ -257,6 +257,12 @@ export const SOLANA_ERROR__TRANSACTION_ERROR__UNBALANCED_TRANSACTION = 7050036;
...
+export const SOLANA_ERROR__INSTRUCTION_PLANS__INVALID_INSTRUCTION_PLAN_KIND = 7618005;
+export const SOLANA_ERROR__INSTRUCTION_PLANS__INVALID_TRANSACTION_PLAN_KIND = 7618006;
+export const SOLANA_ERROR__INSTRUCTION_PLANS__INVALID_TRANSACTION_PLAN_RESULT_KIND = 7618007;
```
steveluscher: I think all of these belong in the `INVARIANT_VIOLATION` section:

https://github.com/anza-xyz/kit/blob/efc90c990a03de8f5ec8214ac0621e6a2576545e/packages/errors/src/codes.ts#L297-L305

steveluscher: This appears unused in this PR.

packages/errors/src/context.ts
```diff
@@ -395,9 +401,29 @@ export type SolanaErrorContext = DefaultUnspecifiedErrorContextToUndefined<
...
+        };
+        [SOLANA_ERROR__INSTRUCTION_PLANS__INVALID_TRANSACTION_PLAN_KIND]: {
+            kind: string;
+        };
+        [SOLANA_ERROR__INSTRUCTION_PLANS__INVALID_TRANSACTION_PLAN_RESULT_KIND]: {
+            kind: string;
+        };
+        [SOLANA_ERROR__INSTRUCTION_PLANS__MISSING_INSTRUCTIONS_IN_INSTRUCTION_PLAN]: {
+            instructionPlan: object;
+        };
```
steveluscher: See this comment: https://app.graphite.dev/github/pr/anza-xyz/kit/647/Add-%60SolanaErrors%60-for-%60MessagePacker%60#comment-PRRC_kwDONc0xOc6DEF07

packages/instruction-plans/src/__tests__/transaction-planner-base-test.ts
```diff
@@ -0,0 +1,1652 @@
...
+            );
+        });
+
+        /**
+         *  [NonDivSeq] ──────────────▶ [NonDivSeq]
+         *   │                           │
+         *   └── [A(x, 250%)]            ├── [Tx: A(1, 100%)]
+         *                               ├── [Tx: A(2, 100%)]
+         *                               └── [Tx: A(3, 50%)]
+         */
```
steveluscher: What does it mean for something to be non-divisible-sequential at the transaction level like this? Don't the transaction boundaries themselves imply that the plan has been divided? Does this imply they _must_ be landed in a bundle.

packages/instruction-plans/src/__tests__/transaction-planner-base-test.ts
```diff
@@ -0,0 +1,1652 @@
...
+         *   │    └── [B: 40%]                      ├── [Tx: C + D + E + G]
+         *   ├── [NonDivSeq]                        └── [Tx: F]
+         *   │    ├── [Par]
+         *   │    │    ├── [C: 25%]
+         *   │    │    └── [D: 25%]
+         *   │    └── [Par]
+         *   │         ├── [E: 25%]
+         *   │         └── [F: 50%]
+         *   └── [G: 25%]
+         */
```
steveluscher: It's notable that adding `G` to the non-divisible-sequential plan increases its chance of failing to the degree that `G` can fail. Efficiency goes up, but P(success) of the `NonDivSeq` goes down.

packages/instruction-plans/src/transaction-planner-base.ts
```diff
@@ -0,0 +1,416 @@
...
+    transactionMessage: TTransactionMessage,
+    config?: { abortSignal?: AbortSignal },
+) => Promise<TTransactionMessage> | TTransactionMessage;
+
+/**
+ * Configuration object for creating a new transaction planner.
+ *
+ * The `createTransactionMessage` function is used to create new transaction messages
+ * whenver needed and the `onTransactionMessageUpdated` function is an optional callback
+ * that can be used to update transaction messages during the planning process.
```
steveluscher: This kinda just restates the inline docs on the properties themselves.

packages/instruction-plans/src/transaction-planner-base.ts
```diff
@@ -0,0 +1,416 @@
...
+>(
+    transactionMessage: TTransactionMessage,
+    config?: { abortSignal?: AbortSignal },
+) => Promise<TTransactionMessage> | TTransactionMessage;
+
+/**
+ * Configuration object for creating a new transaction planner.
+ *
+ * The `createTransactionMessage` function is used to create new transaction messages
+ * whenver needed and the `onTransactionMessageUpdated` function is an optional callback
```
steveluscher:
```suggestion
 * whenever needed and the `onTransactionMessageUpdated` function is an optional callback
```

packages/instruction-plans/src/transaction-planner-base.ts
```diff
@@ -0,0 +1,416 @@
...
+    candidates: MutableSingleTransactionPlan[],
+    predicate: (
+        message: BaseTransactionMessage & TransactionMessageWithFeePayer,
+    ) => BaseTransactionMessage & TransactionMessageWithFeePayer,
+): Promise<MutableSingleTransactionPlan | null> {
+    for (const candidate of candidates) {
+        try {
+            const message = await context.onTransactionMessageUpdated(predicate(candidate.message), {
+                abortSignal: context.abortSignal,
+            });
```
steveluscher: The user-supplied `onTransactionMessageUpdated` function could fail to make use of the `AbortSignal` here. I think we need to double check them. It's their funeral if they don't cancel _their_ work, but in any case we should cancel _ours_.
```suggestion
            const message = await getAbortablePromise(
                context.onTransactionMessageUpdated(predicate(candidate.message), {
                    abortSignal: context.abortSignal,
                },
                context.abortSignal,
            );
```

packages/instruction-plans/src/transaction-planner-base.ts
```diff
@@ -0,0 +1,416 @@
...
+/**
+ * Creates a new transaction planner based on the provided configuration.
+ *
+ * At the very least, the `createTransactionMessage` function must be provided.
+ * This function is used to create new transaction messages whenever needed.
+ *
+ * Additionally, the `onTransactionMessageUpdated` function can be provided
+ * to update transaction messages during the planning process. This function will
+ * be called whenever a transaction message is updated, e.g. when new instructions
+ * are added to a transaction message.
```
steveluscher: I think it's important to drive home that the `onTransactionMessageUpdated` receives the candidate transaction message as an argument, but in all cases _must_ return a transaction message, modified or not.

packages/instruction-plans/src/transaction-planner-base.ts
```diff
@@ -0,0 +1,416 @@
...
+    predicate: (
+        message: BaseTransactionMessage & TransactionMessageWithFeePayer,
+    ) => BaseTransactionMessage & TransactionMessageWithFeePayer,
+): Promise<BaseTransactionMessage & TransactionMessageWithFeePayer> {
+    const newMessage = await context.createTransactionMessage({
+        abortSignal: context.abortSignal,
+    });
+    const updatedMessage = await context.onTransactionMessageUpdated(predicate(newMessage), {
+        abortSignal: context.abortSignal,
+    });
```
steveluscher: Same here. If the user-supplied functions punk out and don't bother to implement abort-signal-listening code then firing the signal will do nothing.

steveluscher (changes requested):
Added via Giphy

What a marvel of test engineering. Made me not want to bother reading the actual code.

I did anyway.

packages/errors/src/codes.ts
lorisleiva: Hmm that's a good point. The difference between the two is:
- One tells us the `MessagePacker` could not do its job with the provided message. Essentially telling us to try a different candidate.
- The other tells us the `InstructionPlan` provided (could be any kind of plan) did not work even with a brand new message.

That being said, the `MESSAGE_CANNOT_ACCOMMODATE_PLAN` error code could semantically tackle both of these cases. I'm happy to merge them into one error.

packages/instruction-plans/src/__tests__/transaction-planner-base-test.ts
lorisleiva: That's right, it implies they must be using a `TransactionPlanExecutor` that is able to tackle that wider atomicity. By having a non-divisible transaction plan, we're able to push that decision to the `TransactionPlanExecutor` instead of failing when transaction bundles could be used.

packages/instruction-plans/src/__tests__/transaction-planner-base-test.ts
lorisleiva: True but I still think most of the time people will want fewer transaction when planning instruction, right? Otherwise we could have an option that forbids the planner to merge parallel branches together.

packages/errors/src/codes.ts
lorisleiva: Actually, working on that refactoring now something feels wrong.

I like `MESSAGE_CANNOT_ACCOMMODATE_PLAN` for signalling that a specific message could not accommodate a given plan. However, `FAILED_TO_FIT_INSTRUCTION_PLAN_IN_NEW_MESSAGE` goes further than this and does not complain about a specific message but about the **compatibility between a plan and a message factory**.

Therefore, how about something like `MESSAGE_FACTORY_CANNOT_ACCOMMODATE_PLAN`?

packages/instruction-plans/src/transaction-planner-base.ts
lorisleiva: Nice! If `toSorted` is too modern for our browser compatibility settings, will our build step take care of it?

EDIT: Never mind, we'd need to change our target for `toSorted` anyway so I'll just use `sort`.

packages/errors/src/codes.ts
lorisleiva: `traverse()` produces `null` anytime a plan should dissolve itself. For instance, if you have a sequential plan that can fit in a parent candidate message, then the children of that sequential plan don't need to exist anymore as a previously create message is taking care of them. Returning `null` signal to the parent "hey I've injected myself in a candidate somewhere, don't worry about me anymore".

So if we ever get this on the root `traverse()` call, this means there aren't any instructions to plan. Examples of this could be: `seq([])`, `par([])` or `seq([seq([seq([]), seq([])]), par([])])`.

packages/instruction-plans/src/transaction-planner-base.ts
lorisleiva: You're right, the message packer didn't use to always assert this invariant (back in the iterator days) but now it does so I can remove this! Thank you!

lorisleiva: Oh no wait actually this whole section needs to change now because we should only call `packMessageToCapacity` once inside the `fitEntirePlanInsideMessage` function. After that, is `done()` is false, we should fail.

packages/instruction-plans/src/transaction-planner-base.ts
lorisleiva: Oh good to know thanks! I had to wrap the first argument in a `Promise.resolve` otherwise typescript was not happy since `getAbortablePromise` require a `Promise<T>`.

packages/errors/src/codes.ts
lorisleiva: You're 100% right, the naming was confusing me into thinking they were different things but `MESSAGE_CANNOT_ACCOMODATE_PLAN` made it clear to me they are the same. I refactored the PR.

lorisleiva:
> Note, the function is called `createBaseTransactionPlanner` because I was initially planning on having a high-level `createDefaultTransactionPlanner` that is all battery included but I'm thinking this interface is already good enough to be used as a high-level. So **maybe I should rename this to `createTransactionPlanner`? Wdyt?**

@steveluscher In case you missed that.

packages/errors/src/codes.ts
steveluscher: I dunno man; I'm not convinced. Right now both errors get thrown when `size(message) > TRANSACTION_SIZE_LIMIT` so they feel identical to me :). I would hope that the _location_ from which this (same) message is thrown would be enough context to tell someone that one is a problem with the packer and the other is a problem with the planner.

I'm fine with whatever you decide though.

packages/errors/src/codes.ts
steveluscher: Got it. To me this feels like the plan being ‘empty’ rather than it ‘missing instructions.’ I guess it depends if you're a ‘my glass is empty’ or a ‘my glass is missing beer’ kinda guy.

packages/instruction-plans/src/transaction-planner-base.ts
```diff
@@ -0,0 +1,404 @@
...
+    for (const candidate of candidates) {
+        try {
+            const message = await getAbortablePromise(
+                Promise.resolve(
+                    context.onTransactionMessageUpdated(predicate(candidate.message), {
+                        abortSignal: context.abortSignal,
+                    }),
+                ),
+                context.abortSignal,
+            );
```
steveluscher: Oh darn. I made this more difficult by not letting `getAbortablePromise()` take in a non-promise. Sorry.

You _could_ do the whole thenable thing here, but you'd have to think through the implications.
```suggestion
            const message = 'then' in context.onTransactionMessageUpdated
                ? await getAbortablePromise(
                       context.onTransactionMessageUpdated(predicate(candidate.message), {
                           abortSignal: context.abortSignal,
                       }),
                       context.abortSignal,
                   ),
                : context.onTransactionMessageUpdated(predicate(candidate.message), {
                      abortSignal: context.abortSignal,
                  });
```

Nah, what you've done here is better.

steveluscher (approved):
Added via Giphy

Nice. You probably _really_ want some tests that exercise that cancellation stuff, but other than that this looks amazing.

packages/errors/src/context.ts
```diff
@@ -394,6 +399,16 @@ export type SolanaErrorContext = DefaultUnspecifiedErrorContextToUndefined<
...
+        [SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN]: {
+            transactionMessage: object;
+        };
+        [SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_FACTORY_CANNOT_ACCOMMODATE_PLAN]: {
+            instructionPlan: object;
+            transactionMessage: object;
+        };
+        [SOLANA_ERROR__INSTRUCTION_PLANS__MISSING_INSTRUCTIONS_IN_INSTRUCTION_PLAN]: {
+            instructionPlan: object;
+        };
```
steveluscher: We probably still want to get this stuff outta here, or replace it with something along the lines of which I suggested [here](https://app.graphite.dev/github/pr/anza-xyz/kit/647/Add-%60SolanaErrors%60-for-%60MessagePacker%60#comment-PRRC_kwDONc0xOc6DN1ym).

packages/errors/src/codes.ts
lorisleiva: That's fair, you convinced me! ☺️

packages/errors/src/codes.ts
lorisleiva: Gotcha, `MY_INSTRUCTION_PLAN_IS_MISSING_BEER` it is then! 🫡 😅

lorisleiva:
### Merge activity

* **Jul 15, 11:09 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/648).
* **Jul 15, 11:36 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/648) rebased this pull request as part of a merge.
* **Jul 15, 11:38 AM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/648).

PR #659, opened 2025-07-15
prohibit adding nonce account to address lookup table

### LINKED ISSUE #309: Defend against transactions that source nonce account addresses from lookup tables
## Motivation

https://github.com/solana-foundation/solana-improvement-documents/pull/242 will make it impossible to load a nonce account from an address lookup table, so we must adjust the ergonomics of the JS SDK to prohibit creating such a transaction message.

## Details

1. Make sure that `createAdvanceNonceAccountInstruction()` does not produce an `IAccountLookupMeta` (already the case, I think)
2. `isAdvanceNonceAccountInstructionForNonce()` must assert that the nonce account in an `AdvanceNonce` instruction is not an `IAccountLookupMeta`
3. `compressTransactionMessageUsingAddressLookupTables()` should refuse to convert an `AdvanceNonce` instruction's nonce account from an `IAccountMeta` to an `IAccountLookupMeta`. **Note**: this will be bad for package efficiency, because this will mean putting nonce-specific logic in a place where it will almost never be used. Think hard about whether this is worth it or not.
4.  `getAddressMapFromInstructions()` should always [upgrade lookup table entries to static entries](https://github.com/anza-xyz/kit/blob/b97855f04ccb44d7dce873942b3fa1ed9437b9c5/packages/transaction-messages/src/compile/accounts.ts#L120-L124) if they belong to a nonce account that is subject to an `AdvanceNonce` instruction. **Note**: like 3, this will be bad for package efficiency, because this will mean putting nonce-specific logic into the message compiler where it will almost never be used. Think hard about whether this is worth it or not.

shubhiscoding:
#### Problem

https://github.com/solana-foundation/solana-improvement-documents/pull/242 will make it impossible to load a nonce account from an address lookup table, so we must adjust the ergonomics of the JS SDK to prohibit creating such a transaction message.

#### Summary of Changes

- made ```createAdvanceNonceAccountInstruction()``` produces an AdvanceNonce instruction having an ```StaticAccount<WritableAccount<TNonceAccountAddress>>```.
- made ```isAdvanceNonceAccountInstructionForNonce()``` to assert that the nonce account in an AdvanceNonce instruction has static set to true.
- made ```compressTransactionMessageUsingAddressLookupTables()``` to refuse to convert any account to a lookup account, if its incompressible property is true.
- made ```getAccountMetas()``` to mark every address in staticAccounts array with property ```static: true``` (discussed [here](https://github.com/anza-xyz/kit/issues/309#issuecomment-3068614056)).
- made changes to tests to acknowledge new StaticAccount type.

Fixes #309

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "eslint-plugin-typescript-sort-keys": "^3.3.0",
         "jest": "^30.0.0-alpha.6",
         "jest-environment-jsdom": "^30.0.0-alpha.6",
-        "jest-runner-eslint": "^2.2.1",
+        "jest-runner-eslint": "^2.3.0",
         "jest-runner-prettier": "^1.0.0",
         "jest-watch-master": "^1.0.0",
         "jest-watch-select-projects": "^2.0.0",
--- a/packages/instructions/src/accounts.ts
+++ b/packages/instructions/src/accounts.ts
@@ -23,6 +23,7 @@ import { AccountRole } from './roles';
 export interface AccountMeta<TAddress extends string = string> {
     readonly address: Address<TAddress>;
     readonly role: AccountRole;
+    readonly static?: boolean;
 }
 
 /**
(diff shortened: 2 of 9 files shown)


lorisleiva (changes requested):
Thanks for this! However, before I review this in more details, I think we should first figure out if this change is really worth it ([as discussed here](https://github.com/anza-xyz/kit/issues/309#issuecomment-3083574288)).

PR #664, opened 2025-07-15
Add `createTransactionPlanExecutor` function

lorisleiva:
This PR provides a base implementation for the `TransactionPlanExecutor` type.

It does so by requiring a `executeTransactionMessage` function that dictates how a message in a `TransactionPlan` should be executed.

Additionally, a `parallelChunkSize` option can be provided to limit the parallelism of `ParallelTransactionPlans`.


lorisleiva:
* **#674** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/674?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#664** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/664?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/664?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/errors/src/context.ts
```diff
@@ -399,6 +400,9 @@ export type SolanaErrorContext = DefaultUnspecifiedErrorContextToUndefined<
             index: number;
             instructionErrorContext?: unknown;
         };
+        [SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN]: {
+            transactionPlanResult: unknown;
+        };
```
lorisleiva, line 405: @steveluscher This problem is back. However, this time we cannot omit this attribute or reduce it to scalar values. Returning the plan is crucial for the app to land on its feet since it can then grab the part of the plan that did not succeed at try again. Perhaps here we should try the non-writable object property trick?

steveluscher: Non-enumerable, yeah. Meaning that it's invisible to `Object.keys()` et al.

steveluscher: LOL. https://github.com/microsoft/TypeScript/issues/9726#issuecomment-240785269

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
```diff
@@ -0,0 +1,870 @@
...
+    error: SolanaError<typeof SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN>,
+): Promise<void> {
+    const transactionPlanResult = error.context.transactionPlanResult;
+    await expect(promise).rejects.toThrow(error);
+    // This second expectation is necessary since `toThrow` will only check the
+    // error message and `transactionPlanResult` is not part of the message.
+    await expect(promise).rejects.toThrow(
+        expect.objectContaining({ context: expect.objectContaining({ transactionPlanResult }) }),
+    );
+}
```
steveluscher, line 40: When you invent new custom assertion functions, you have to add it here:

https://github.com/anza-xyz/kit/blob/731d3ea0c6e357a281477f06d4ec2c4c656a8cad/eslint.config.mjs#L21-L26

Also, I know this is used in 15 places, but I don't think this suite would be worse if you inlined it.

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
```diff
@@ -0,0 +1,870 @@
...
+            await expect(promise).resolves.toStrictEqual(
+                sequentialTransactionPlanResult([
+                    successfulSingleTransactionPlanResult(messageA, createTransaction('A')),
+                    successfulSingleTransactionPlanResult(messageB, createTransaction('B')),
+                ]),
+            );
+
+            expect(executeTransactionMessage).toHaveBeenCalledTimes(2);
+            expect(executeTransactionMessage).toHaveBeenCalledWith(messageA, { abortSignal: undefined });
+            expect(executeTransactionMessage).toHaveBeenCalledWith(messageB, { abortSignal: undefined });
```
steveluscher: If order matters here, you can use `toHaveBeenNthCalledWith()`.
```suggestion
            expect(executeTransactionMessage).toHaveBeenNthCalledWith(1, messageA, { abortSignal: undefined });
            expect(executeTransactionMessage).toHaveBeenNthCalledWith(2, messageB, { abortSignal: undefined });
```

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
```diff
@@ -0,0 +1,870 @@
...
+
+        it('fails to execute a sequential transaction plan', async () => {
+            expect.assertions(2);
+            const messageA = createMessage('A');
+            const messageB = createMessage('B');
+            const cause = new SolanaError(SOLANA_ERROR__INSTRUCTION_ERROR__INVALID_ARGUMENT, { index: 0 });
+            const executeTransactionMessage = jest.fn().mockImplementation((message: { id: string }) => {
+                // eslint-disable-next-line jest/no-conditional-in-test
+                return message.id === 'B' ? Promise.reject(cause) : forwardId(message);
+            });
```
steveluscher: If you want to avoid logic like this, and make the test more readable at the same time, you can write a little story like this:
```suggestion
            const executeTransactionMessage = jest.fn().mockImplementationOnce(forwardId).mockRejectedValueOnce(cause);
```

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
```diff
@@ -0,0 +1,870 @@
...
+            const promise = executor(sequentialTransactionPlan([messageA, messageB]));
+            await expect(promise).resolves.toStrictEqual(
+                sequentialTransactionPlanResult([
+                    successfulSingleTransactionPlanResult(messageA, createTransaction('A'), { custom: 'context' }),
+                    successfulSingleTransactionPlanResult(messageB, createTransaction('B'), { custom: 'context' }),
+                ]),
+            );
+        });
+
+        it('fails to execute a sequential transaction plan', async () => {
```
steveluscher: This description needs a ‘when’ somewhere in here.
```suggestion
        it('fails to execute a sequential transaction plan when the executor function rejects', async () => {
```

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -1,7 +1,207 @@
...
+            if (lastChunk && lastChunk.length < chunkSize) {
+                lastChunk.push(subPlan);
+            } else {
+                chunks.push([subPlan]);
+            }
+            return chunks;
+        },
+        [[]] as TransactionPlan[][],
+    );
+}
```
steveluscher: Let's do even better. I think that concurrency is a better model than parallelism. This design waits for every operation in the chunk to finish before starting the next chunk. Let's implement a design that _always_ has at most _n_ in flight at once.

* rename `parallelChunkSize` to `maxConcurrentOperations` or something
* always have _n_ inflight promises and resolve a single promise when done

This suggested implementation isn't perfect, but it's probably close.

```suggestion
async function traverseParallel(
    transactionPlan: ParallelTransactionPlan,
    context: TraverseContext,
): Promise<TransactionPlanResult> {
    const results: TransactionPlanResult[] = [];
    let thread!: { fail: (_e: unknown) => void; unblock: () => void };
    function okToContinue() {
        return new Promise<void>((resolve: () => void, reject: (_e: unknown) => void) => {
            thread.fail = reject;
            thread.unblock = resolve;
        });
    }
    let numInFlightOperations = 0;
    for (let ii = 0; ii < transactionPlan.plans.length; ii++) {
        if (numInFlightOperations === context.maxConcurrentOperations) {
            await okToContinue();
        }
        numInFlightOperations++;
        traverse(transactionPlan.plans[ii], context)
            .then(res => {
                results[ii] = res;
                thread.unblock();
            })
            .catch(thread.fail)
            .finally(() => {
                numInFlightOperations--;
            });
    }
    while (numInFlightOperations > 0) {
        await okToContinue();
    }
    return parallelTransactionPlanResult(results);
}
```

steveluscher (changes requested):
Added via Giphy

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
lorisleiva: Oh okay. Any reason the lint tests aren't failing despite the fact that the function name isn't whitelisted?

I was inlining them at first but then wanted to explain why I had to provide a second assertion just to compare the context. Tbh, I think it is very misleading that `toThrow(someError)` doesn't actually compare `someError` with the thrown error aside from comparing their messages. Perhaps we could have a generic function like `toThrowStrictEqual()`?

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
lorisleiva: Oh I didn't think of that. That's much nicer thanks! I've not done this for parallel scenarios though since technically there are no requirements for starting transaction A before B in this case.

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
lorisleiva: Thanks, I've done this on most places aside from a few parallel tests where it would be valid if things were called the other way around.

packages/errors/src/context.ts
lorisleiva: This required changes to the `SolanaError` constructor so I did the work on [a separate PR](https://github.com/anza-xyz/kit/pull/674).

packages/instruction-plans/src/transaction-plan-executor.ts
steveluscher: Darn it, I meant to press send on this before you woke up, but didn’t.

---

I’m actually wondering if we should offer this feature at all. If we took all of this away, you could still limit concurrency by:

* using an executor function that limited concurrency
* using a network transport that limited concurrency

That would offer more flexibility too, since a user-supplied executor function could have absurd rules like ‘limit the concurrency of messages that end with the letter ‘h’ to 11 at a time. Thoughts?

lorisleiva: Ah I didn't see this in time and it's now implemented haha. I actually really like the `maxConcurrentOperations` implementation. My brain can never come up with `Promise` things like this haha so this was a fun exercise regardless.

I don't know that it would be trivial to add concurrency to the `executeTransactionMessage` function since that function's job is to send only one message and at this point, you're not sure if you come from a sequential or a parallel parent. I think at this point, someone could just write their own `TransactionPlanExecutor` function since we're not doing anything crazy here. Looping for sequential plans, concurrency for parallel plan, call given function for single plans. 🤷‍♂️ 

So with that in mind, I think it's nice to have an opt-in `maxConcurrentOperations` option within this simple vanilla implementation. That helps people get a better idea of the type of things they can do if they want to implement their own `TransactionPlanExecutor` function.

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
steveluscher: It's subtle. This lint succeeds _because_ you've supplied `expect.assertions(n)`. If you remove that, it fails with ‘test has no assertions’.

> I think it is very misleading that toThrow(someError) doesn't actually compare someError with the thrown error aside from comparing their messages.

I mean hell, until recently, it didn't even compare the `cause`.

* https://github.com/jestjs/jest/issues/15316
* https://github.com/jestjs/jest/issues/15697

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
```diff
@@ -0,0 +1,876 @@
...
+            executor(parallelTransactionPlan([messageA, messageB, messageC, messageD])).catch(() => {});
+
+            // The first two operations are executed immediately.
+            await jest.runAllTimersAsync();
+            expect(executeTransactionMessage).toHaveBeenCalledTimes(2);
+            expect(executeTransactionMessage).toHaveBeenCalledWith(messageA, { abortSignal: undefined });
+            expect(executeTransactionMessage).toHaveBeenCalledWith(messageB, { abortSignal: undefined });
+
+            // When operation B resolves, the next one is executed.
+            resolvers['B']();
```
steveluscher: Nice.

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
```diff
@@ -0,0 +1,876 @@
...
+            const abortController = new AbortController();
+            const abortSignal = abortController.signal;
+            const cause = new Error('Aborted during execution') as SolanaError;
+            let resolveA = () => {};
+            const executeTransactionMessage = jest.fn().mockImplementation((message: { id: string }) => {
+                // eslint-disable-next-line jest/no-conditional-in-test
+                if (message.id === 'A') {
+                    return new Promise(resolve => {
+                        resolveA = () => resolve({ transaction: createTransaction('A') });
+                    });
```
steveluscher: You might be interested in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers.

packages/instruction-plans/src/transaction-plan-executor.ts
steveluscher: > I don't know that it would be trivial to add concurrency to the `executeTransactionMessage` function…

It's doable. It would look something like:

```ts
const scheduler = new Scheduler({ maxConcurrency: 2 });
function executeTransactionMessage(message) {
    // ...
}
```

> … at this point, you're not sure if you come from a sequential or a parallel parent.

That's fair. I didn't think you'd care though – the only thing you're concerned about is saturating your network connection, and that's a concern that's global to your app.

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -1,7 +1,228 @@
...
+            })
+            .catch(e => {
+                numInFlightOperations--;
+                thread.fail(e);
+            });
+    }
+
+    while (numInFlightOperations > 0) {
+        await okToContinue();
+    }
```
steveluscher: This part (of my original design) sort of sucks because it creates a new promise every time one of these puppies resolves even though there are no more operations to schedule, instead of continuing to wait on the same one until `numInFlightOperations` falls to zero.

Doesn't matter, just leaving a note here for posterity.

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -1,7 +1,228 @@
...
+    function okToContinue() {
+        return new Promise<void>((resolve, reject) => {
+            thread.fail = reject;
+            thread.unblock = resolve;
+        });
+    }
+
+    let numInFlightOperations = 0;
+    for (let index = 0; index < operations.length; index++) {
+        if (numInFlightOperations === maxConcurrentOperations) {
```
steveluscher: We should probably throw an invariant error if `maxConcurrentOperations` is `<1`. If someone sets it to zero the program will just halt.

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -1,7 +1,228 @@
...
+}
+
+async function traverseParallel(
+    transactionPlan: ParallelTransactionPlan,
+    context: TraverseContext,
+): Promise<TransactionPlanResult> {
+    const results = await concurrentOperations(
+        transactionPlan.plans.map(plan => () => traverse(plan, context)),
+        context.maxConcurrentOperations,
+    );
```
steveluscher: We alright calling this unconditionally, even if no concurrency config is set?

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -1,7 +1,228 @@
...
+        ? sequentialTransactionPlanResult(results)
+        : nonDivisibleSequentialTransactionPlanResult(results);
+}
+
+async function traverseParallel(
+    transactionPlan: ParallelTransactionPlan,
+    context: TraverseContext,
+): Promise<TransactionPlanResult> {
+    const results = await concurrentOperations(
+        transactionPlan.plans.map(plan => () => traverse(plan, context)),
```
steveluscher: Not saying you have to do this, but just some prior art to look at: `p-limit's` `limitFunction()`. You could wrap `traverse` in `limitFunction()` and then just blast calls into it.

https://github.com/sindresorhus/p-limit/tree/main?tab=readme-ov-file#limitfunctionfn-options-named-export

steveluscher: More prior art: https://github.com/sindresorhus/p-map

packages/instruction-plans/src/transaction-plan-executor.ts
steveluscher: Another design is like this:

```ts
const actualTraverse = context.maxConcurrentOperations
    ? getConcurrencyLimitedFunction(traverse, context.maxConcurrentOperations)
    : traverse;
const results = transactionPlan.plans.map(plan => actualTraverse(plan, context));
```

packages/instruction-plans/src/transaction-plan-executor.ts
lorisleiva: As discussed offline, let's rip this concurrency stuff out since it's much better to implement a scheduler at a higher level anyway — e.g. on the RPC client.

lorisleiva:
@steveluscher I cleaned all the concurrency stuff as discussed. Lmk what you think. 🙏

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -1,7 +1,179 @@
...
+    return transactionPlan.divisible
+        ? sequentialTransactionPlanResult(results)
+        : nonDivisibleSequentialTransactionPlanResult(results);
+}
+
+async function traverseParallel(
+    transactionPlan: ParallelTransactionPlan,
+    context: TraverseContext,
+): Promise<TransactionPlanResult> {
+    const results = await Promise.all(transactionPlan.plans.map(plan => traverse(plan, context)));
```
steveluscher, line 145: Added via Giphy

steveluscher (approved):
There's no code like no code.

lorisleiva:
### Merge activity

* **Jul 22, 8:57 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/664).
* **Jul 22, 8:58 AM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/664).

PR #701, opened 2025-07-24
Make Union codecs FixedSize if all their variants are FixedSize 

mcintyre94:
#### Problem

Currently codec functions like `getUnionEncoder` return an `Encoder<something>`, but at runtime they may return a `FixedSizeEncoder<something>` depending on their variants.

This means that `getUnionEncoder(fixedSizeThings)` returns a fixed size encoder, but not the type `FixedSizeEncoder<something>`

This causes difficulties for Codama, eg https://github.com/codama-idl/codama/pull/730

#### Summary of Changes

We add some type gymnastics to `getUnion{Encoder}` and `getDiscriminatedUnion{Encoder}` (which calls Union)

If all variants are `FixedSize{Encoder}` then we return a `FixedSize{Encoder}<something>`, else we continue to return `{Encoder}<something>` as we currently do.

This means that when Codama generates `getDiscriminatedUnionEncoder([['V1', getStructEncoder([['amount', getU64Encoder()]])]])` it will correctly be typed as a `FixedSizeEncoder<something>`. 

Note that I also tried to do this using overloads instead of a conditional type, but I got type errors when the codec function calls the encoder/decoder. I think it's because the variable codec would be incompatible with calls to the fixed encoder/decoder overload. It'd be a type-only issue but I figured this is cleaner than suppressing confusing errors in that approach.

### DIFF
--- a/packages/codecs-data-structures/src/discriminated-union.ts
+++ b/packages/codecs-data-structures/src/discriminated-union.ts
@@ -1,5 +1,15 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { Codec, combineCodec, Decoder, Encoder, transformDecoder, transformEncoder } from '@solana/codecs-core';
+import {
+    Codec,
+    combineCodec,
+    Decoder,
+    Encoder,
+    FixedSizeCodec,
+    FixedSizeDecoder,
+    FixedSizeEncoder,
+    transformDecoder,
+    transformEncoder,
+} from '@solana/codecs-core';
 import { getU8Decoder, getU8Encoder, NumberCodec, NumberDecoder, NumberEncoder } from '@solana/codecs-numbers';
 import { SOLANA_ERROR__CODECS__INVALID_DISCRIMINATED_UNION_VARIANT, SolanaError } from '@solana/errors';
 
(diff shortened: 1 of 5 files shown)


packages/codecs-data-structures/src/discriminated-union.ts
```diff
@@ -129,6 +139,29 @@ type GetDecoderTypeFromVariants<
...
+        ? FixedSizeCodec<
+              GetEncoderTypeFromVariants<TVariants, TDiscriminatorProperty>,
+              GetDecoderTypeFromVariants<TVariants, TDiscriminatorProperty> &
+                  GetEncoderTypeFromVariants<TVariants, TDiscriminatorProperty>
+          >
+        : Codec<
+              GetEncoderTypeFromVariants<TVariants, TDiscriminatorProperty>,
+              GetDecoderTypeFromVariants<TVariants, TDiscriminatorProperty> &
+                  GetEncoderTypeFromVariants<TVariants, TDiscriminatorProperty>
+          >;
```
mcintyre94: The beauty brings tears to my eyes.

lorisleiva (approved):
Thank you! Well done for taming the TS beast.

giphy

packages/codecs-data-structures/src/discriminated-union.ts
```diff
@@ -129,6 +139,29 @@ type GetDecoderTypeFromVariants<
         : never) & { [P in TDiscriminatorProperty]: TVariants[I][0] };
 }>[ArrayIndices<TVariants>];
 
+type UnionEncoder<TVariants extends Variants<Encoder<any>>, TDiscriminatorProperty extends string> =
```
steveluscher: I’m not at a computer right now to try it out for myself; does `unknown` function as well as `any` for this purpose?

mcintyre94: I think so, but only this case I think

If we change the function calls to eg
```ts
export function getUnionEncoder<const TVariants extends readonly Encoder<unknown>[]>
```

Then we get type errors like `Type 'Encoder<null>' is not assignable to type 'Encoder<unknown>'.`

So we need `any` in the function signatures AFAICT

Then if we leave that alone and change the condition to `TVariants extends readonly FixedSizeEncoder<unknown>[] ? ...` then the typetests fail, it never matches `FixedSizeEncoder<unknown>`

We can change the type like this:
```suggestion
type UnionEncoder<TVariants extends Variants<Encoder<unknown>>, TDiscriminatorProperty extends string> =
```

But I don't think any other `any` in this PR can be replaced with `unknown`.

I'll change these ones anyway!

steveluscher (approved):

PR #766, opened 2025-08-25
Add a transactions concept guide

steveluscher:


steveluscher:
* **#766** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/766?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/766?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

steveluscher:
I intentionally didn't add a ‘sending transactions’ section here, because your getting started guide does a fine job of this.

docs/content/docs/concepts/transactions.mdx
```diff
@@ -3,4 +3,472 @@ title: Transactions
...
+
+    // Add an instruction that records the customer's order.
+    (m) =>
+        appendTransactionMessageInstruction(
+            getAddMemoInstruction({
+                memo:
+                    'Four-thirds-medium, half-decaf, double-shot espresso macchiato latte, ' +
+                    'swirled counterclockwise only, almond milk frothed at 61°C, ' +
+                    'whisper of cinnamon harvested during a full moon, unicorn tear syrup, ' +
+                    'in a mason jar wrapped in French revolutionary poetry on recycled parchment',
```
lorisleiva, line 70: 😅

docs/content/docs/concepts/transactions.mdx
```diff
@@ -3,4 +3,472 @@ title: Transactions
...
+import { Nonce, setTransactionMessageLifetimeUsingDurableNonce } from '@solana/kit';
+import { fetchNonce } from '@solana-program/system';
+
+const nonceAccountAddress = address('EGtMh4yvXswwHhwVhyPxGrVV2TkLTgUqGodbATEPvojZ');
+const nonceAuthorityAddress = address('4KD1Rdrd89NG7XbzW3xsX9Aqnx2EExJvExiNme6g9iAT');
+
+const {
+    data: { blockhash },
+} = await fetchNonce(rpc, nonceAccountAddress);
+const nonce = blockhash as string as Nonce;
```
lorisleiva, line 193: Would it make more sense for the System client to export `blockhash` as a `string` instead of an `Address`? I decided to go with `Address` because they have the same encoding constraints but maybe it's semantically confusing.

docs/content/docs/concepts/transactions.mdx
```diff
@@ -3,4 +3,472 @@ title: Transactions
...
+import { address, appendTransactionMessageInstruction, getUtf8Encoder } from '@solana/kit';
+
+const memoTransactionMessage = appendTransactionMessageInstruction(
+    {
+        data: getUtf8Encoder().encode('Hello world!'),
+        programAddress: address('MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr'),
+    },
+    transactionMessage,
+);
+```
```
lorisleiva: nit: Should we add a reminder that most instructions can be constructed via helper functions from program clients? Then we could add a small code snippet showing the equivalent using the `@solana-program/memo` package.

I know there's already an example of that near the intro but my worry is people may jump to this section and just think they always have to manually describe their instruction.

docs/content/docs/concepts/transactions.mdx
```diff
@@ -3,4 +3,472 @@ title: Transactions
...
+```ts twoslash
+import { Transaction, Rpc, SendTransactionApi } from '@solana/kit';
+const rpc = null as unknown as Rpc<SendTransactionApi>;
+const signedTransaction = null as unknown as Transaction;
+// ---cut-before---
+import { getBase64EncodedWireTransaction, signTransaction } from '@solana/kit';
+
+const serializedTransaction = getBase64EncodedWireTransaction(signedTransaction);
+const signature = await rpc.sendTransaction(serializedTransaction, { encoding: 'base64' }).send();
+```
```
lorisleiva, line 409: I think here we should link to the send and confirm helpers for discoverability. If a reader starts with this transaction guide, they will go back to their code thinking `rpc.sendTransaction` is the recommended way to send them using Kit. A simple callout with a link to `/docs/getting-started/send-transaction` should suffice.

lorisleiva (approved):
This is soooo good! IMO you've hit the right spot between not assuming too much prior knowledge from the reader and still being relevant for power users. Thank you! ☕

docs/content/docs/concepts/transactions.mdx
mcintyre94: Could it be typed as `Nonce` from Kit? It'd be a bit confusing because it's a field in a type called `Nonce` in the client, but would probably be the easiest thing to work with.

lorisleiva: It could but we'd have to override the generated code. That's because Codama has the concept of `PublicKeyTypeNodes` but doesn't have something like a `NonceTypeNode` (which is too system-program specific IMO to be part of the standard).

docs/content/docs/concepts/transactions.mdx
```diff
@@ -3,4 +3,472 @@ title: Transactions
...
+import { decompileTransactionMessage } from '@solana/kit';
+const transactionMessage = decompileTransactionMessage(compiledTransactionMessage);
+```
+
+<Callout type="warn">
+    You can not fully reconstruct a source message from a compiled message without extra
+    information. In particular, supporting details about the lifetime constraint and the concrete
+    addresses of accounts sourced from account lookup tables are lost to compilation, but can be
+    supplied in the `config` argument of `decompileTransactionMessage`.
+</Callout>
```
mcintyre94: Is it worth mentioning `decompileTransactionMessageFetchingLookupTables` here?

mcintyre94 (approved):
This is a great guide IMO!

docs/content/docs/concepts/transactions.mdx
steveluscher: I mean, it's actually semantically incorrect to type it `Address` so we have to do something. `blockhash` is annoyingly technically correct because that's where we _source_ new nonces, but it's again semantically incorrect, sort of.

docs/content/docs/concepts/transactions.mdx
steveluscher: Yeah, this bothered me too. I didn't want to throw more concepts (program clients) into this example, but I also don't want to make people read disclaimers. I'll just use the memo program client.

docs/content/docs/concepts/transactions.mdx
steveluscher: Thanks for the reminder!

PR #787, opened 2025-08-27
Fix API links with hash fragments

lorisleiva:
This PR fixes an issue in the way URLs are generated in the "API References" section of our documentation website.

Prior to this PR, if a URL contained a hash fragment — e.g. `"/api/enumerations/AccountRole.mdx#readonly"` — it would not be recognised by our custom regex and therefore the `.mdx` extension would not be removed as expected.

This PR fixes this by adjusting the regex so it removes the `.mdx` extension even if the URL contains something afterwards.

### DIFF
--- a/typedoc.plugin.mjs
+++ b/typedoc.plugin.mjs
@@ -74,9 +74,9 @@ export function load(app) {
     // - strip the .mdx extension
     app.renderer.on(td.MarkdownPageEvent.END, page => {
         if (!page.contents) return;
-        page.contents = page.contents.replace(/\(((?:[^\/\)]+\/)*[^\/\)]+)\.mdx\)/gm, (_, path) => {
+        page.contents = page.contents.replace(/\(((?:[^\/\)]+\/)*[^\/\)]+)\.mdx([^)]*)?\)/gm, (_, path, suffix) => {
             const rootRelativeUrl = resolve('/api', dirname(page.url), path);
-            return `(${rootRelativeUrl})`;
+            return `(${rootRelativeUrl}${suffix ?? ''})`;
         });
     });
 }


lorisleiva:
* **#787** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/787?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/787?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

typedoc.plugin.mjs
```diff
@@ -74,9 +74,9 @@ export function load(app) {
     // - strip the .mdx extension
     app.renderer.on(td.MarkdownPageEvent.END, page => {
         if (!page.contents) return;
-        page.contents = page.contents.replace(/\(((?:[^\/\)]+\/)*[^\/\)]+)\.mdx\)/gm, (_, path) => {
+        page.contents = page.contents.replace(/\(((?:[^\/\)]+\/)*[^\/\)]+)\.mdx([^)]*)?\)/gm, (_, path, suffix) => {
```
lorisleiva, line 77: nit: I'm matching everything that isn't a closing parenthesis after `.mdx` to match a wider spectrum of suffixes but there's a case for only matching hash fragments to be explicit and ensure we don't match an unwanted `.mdx`.

```suggestion
        page.contents = page.contents.replace(/\(((?:[^\/\)]+\/)*[^\/\)]+)\.mdx(#[^)]*)?\)/gm, (_, path, suffix) => {
```

steveluscher: What you've done here is probably fine for now.

PR #793, opened 2025-08-28
The recent transaction confirmation strategy now throws transaction errors from the one-shot test

### LINKED ISSUE #711: createRecentSignatureConfirmationPromiseFactory does not throw error if initial status request returns with transaction error
## Overview

If you wait on a transaction using `getRecentSignatureConfirmationPromise` returned from `createRecentSignatureConfirmationPromiseFactory` (which is used by `waitForRecentTransactionConfirmation`) most of the time you will have the promise reject with an error if the transaction is confirmed but has an error. The issue is in the case where the transaction has already confirmed before you start checking and the initial signature status request returns that the transaction is confirmed but doesn't check that there is an error like the subscription does.

## Steps to reproduce

It's difficult to reliably reproduce.

## Description of bug

https://github.com/anza-xyz/kit/blob/9205484d33af9426fc9de9594bab204b8f954faf/packages/transaction-confirmation/src/confirmation-strategy-recent-signature.ts#L78-L129

It looks like this needs to be added to check for an error in the initial status request:
```patch
         /**
         * STEP 2: Having subscribed for updates, make a one-shot request for the current status.
         *         This will only yield a result if the signature is still in the status cache.
         */
        const signatureStatusLookupPromise = (async () => {
            const { value: signatureStatusResults } = await rpc
                .getSignatureStatuses([signature])
                .send({ abortSignal: abortController.signal });
            const signatureStatus = signatureStatusResults[0];
            if (
                signatureStatus &&
                signatureStatus.confirmationStatus &&
                commitmentComparator(signatureStatus.confirmationStatus, commitment) >= 0
            ) {
-                return;
+                if (signatureStatus.err) {
+                    throw getSolanaErrorFromTransactionError(signatureStatus.err)
+                } else {
+                    return
+                }
            } else {
                await new Promise(() => {
                    /* never resolve */
                });
            }
        })();
```

steveluscher:
#### Problem

The recent signature transaction strategy does two things:

- Subscribes for changes to a signature's status
- Does a point lookup of the signature's status

It was reported that if the signature's status is an error, the one-shot lookup would not throw it.

#### Summary of Changes

Throw any transaction errors discovered during the one-shot lookup.

Fixes #711.

### DIFF
--- a/.changeset/rotten-weeks-enjoy.md
+++ b/.changeset/rotten-weeks-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@solana/transaction-confirmation': patch
+---
+
+Fixed a bug where transaction errors discovered during recent transaction confirmation might not be thrown
--- a/packages/transaction-confirmation/src/__tests__/confirmation-strategy-signature-test.ts
+++ b/packages/transaction-confirmation/src/__tests__/confirmation-strategy-signature-test.ts
@@ -111,6 +111,20 @@ describe('createSignatureConfirmationPromiseFactory', () => {
         });
         await expect(signatureConfirmationPromise).resolves.toBeUndefined();
     });
+    it('fatals when the signature status returned by the one-shot query is an error', async () => {
+        expect.assertions(1);
+        getSignatureStatusesMock.mockResolvedValue({
+            value: [{ err: 'o no' }],
+        });
+        const signatureConfirmationPromise = getSignatureConfirmationPromise({
+            abortSignal: new AbortController().signal,
+            commitment: 'finalized',
+            signature: 'abc' as Signature,
+        });
+        await expect(signatureConfirmationPromise).rejects.toThrow(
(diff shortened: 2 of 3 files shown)


steveluscher:
* **#793** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/793?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/793?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva (approved):
Thanks!

steveluscher:
### Merge activity

* **Aug 29, 11:54 AM UTC**: @steveluscher merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/793).

PR #814, opened 2025-09-01
feat: Add clientId field to getClusterNodes response

leafaar:
#### Problem

The Agave RPC now returns a `clientId` field in the getClusterNodes response (see anza-xyz/agave#7817), but the JavaScript client doesn't include this field in its type definitions.

#### Summary of Changes

- Add `clientId: number | null` field to the `ClusterNode` type
- Update tests to expect the clientId field in responses
- Add JSDoc documentation for the new field

#### References

- Depends on: anza-xyz/agave#7817

### DIFF
--- a/.changeset/nasty-days-wash.md
+++ b/.changeset/nasty-days-wash.md
@@ -1,0 +1,5 @@
+---
+'@solana/rpc-api': minor
+---
+
+Add clientId field to getClusterNodes response
--- a/packages/rpc-api/src/__tests__/get-cluster-nodes-test.ts
+++ b/packages/rpc-api/src/__tests__/get-cluster-nodes-test.ts
@@ -72,6 +72,7 @@ describe('getClusterNodes', () => {
             const [featureSet, gossip, pubkey, rpc, shredVersion, version] = await getNodeInfoFromLogFile();
             const res = await mockRpc.getClusterNodes().send();
             expect(res[0]).toStrictEqual({
+                clientId: expect.any(Number),
                 featureSet,
                 gossip,
                 pubkey,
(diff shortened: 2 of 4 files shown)


steveluscher (changes requested):
Make sure to update all of the tests (run `pnpm dev` in `packages/rpc-api/` as you work) and to add an exemption for `clientId` so that the client doesn't presume it's supposed to be a `BigInt`.

See #521 for a great example of how to do all of that!

leafaar:
Hey, I updated and fixed a lint, where the `ClusterNode` need to be in alphabetical order, so I moved `clientId` to the top. Also added the exception, so the client doesn't interpret as a BigInt. Sorry for the delay to fix that, I'm not too familiar with this part of the stack. Hope I did everything correct, if not, just let me know, and I'll be happy to fix it again :D

steveluscher:
Cool. Tests are failing merely because the current validator does not return `clientId` in the response. Assigning this to @alannza who's also assigned to the RPC PR that adds `clientId`.

leafaar:
Why was this one closed?

mcintyre94:
Sorry @leafaar - I think we should have marked this do-not-close so the bot doesn't close it. Reopened and added that label, hopefully it'll be left alone now! 

We have a github action that closes stale issues/PRs automatically: https://github.com/anza-xyz/kit/blob/main/.github/workflows/manage-stale-threads.yml

leafaar:
No problem, thank you :grin:

mcintyre94:
Thanks for this @leafaar! Agave 4.0 has been released as the latest stable version, and this change is incorporated in our PR to update Kit accordingly: https://github.com/anza-xyz/kit/pull/1658.

PR #863, opened 2025-09-08
Remove `chalk` dependency from `@solana/errors`

elliotkennedy:
#### Problem

A compromised version of chalk was recently published to npm, it makes sense to remove and to keep the zero-dep ethos of kit

https://github.com/chalk/chalk/issues/656

#### Summary of Changes

Removes `chalk` as a dependency and keeps the functionality. This is possibly overkill but it's trivial to write a minimal version of the compromised lib with AI tools

Fixes # tbd

#### Comparison

Chalk version:

Screenshot 2025-09-08 at 21 56 42

Zero-dep version:

Screenshot 2025-09-08 at 21 58 43

### DIFF
--- a/packages/errors/package.json
+++ b/packages/errors/package.json
@@ -74,7 +74,6 @@
         "maintained node versions"
     ],
     "dependencies": {
-        "chalk": "^5.6.0",
         "commander": "^14.0.0"
     },
     "peerDependencies": {
--- a/packages/errors/src/cli.ts
+++ b/packages/errors/src/cli.ts
@@ -1,4 +1,3 @@
-import chalk from 'chalk';
 import { Command, InvalidArgumentError } from 'commander';
 
 import { version } from '../package.json';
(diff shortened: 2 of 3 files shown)


elliotkennedy:
> Thanks for this idea!
> 
> I don't know how I feel about this. It's a bit like posting a guard at the door of a store that's already been robbed. This work is also only useful if we eliminate the _entire_ threat, and at the moment, `commander` is still depended upon by the errors CLI.
> 
> The good news is that `chalk` and `commander` are both zero-dependency, so once we do 🪄something with them, we're done.
> 
> At the very _minimum_ we must _pin_ their versions to known-good versions. This will prevent `pnpm dlx @solana/errors` from installing _fresh_, potentially compromised versions of `chalk` and `commander`.
> 
> We could _also_ consider vendoring them entirely (copying their source code into our own, and depending on that locally) but all that would accomplish would be to protect against a compromise of the package registry itself. At that point, there's no point, because the vendored code is _in_ the package registry, and could itself be modified.
> 
> Would you like to put up the version pinning PR, or would you like me to do it?

Thanks for the response @steveluscher - I agree it's not my favourite solution. After espousing the benefits of kit to the team for months I could have cried when I checked the dependency tree yesterday :D

Could I ask you to do the pinning PR please? I'm not at all familiar with pnpm.

I also thought maybe it's worth adding a `@solana/errors-cli` package that isn't depended on by any other kit modules and doesn't have to follow the zero-dep ethos. Happy to give it a go if you want just lmk.

Thanks a lot!

PR #871, opened 2025-09-10
Do not allow decoding transactions with an unsupported version

### LINKED ISSUE #866: Transaction / transaction message decoders must fail on unimplemented version
## Overview

Our decoders, today, will happily decode a transaction message that claims to be of version 1 or higher. 

## Steps to reproduce

https://codesandbox.io/p/sandbox/fyq7rf

## Description of bug

This should be a runtime error now, so that when v1 actually gets introduced, old Kit clients don't start reading out garbage and/or fataling.

mcintyre94:
#### Problem

Kit currently supports codecs for legacy transactions, and transactions with version 0. There's currently a SIMD for transactions with version 1, which drastically changes how transactions are encoded. Kit will currently decode transactions claiming to have transaction message 1, but this decoding will be nonsense when such transactions exist.

#### Summary of Changes

We now throw a SolanaError when we decode a transaction and receive a version number greater than 0. 

A future version of Kit will be able to add support for v1 transactions, while continuing to throw for still unsupported versions. 

Fixes #866


packages/transaction-messages/src/codecs/transaction-version.ts
```diff
@@ -31,6 +35,12 @@ export function getTransactionVersionEncoder(): VariableSizeEncoder<TransactionV
                     actualVersion: value,
                 });
             }
+
+            if (value > MAX_SUPPORTED_TRANSACTION_VERSION) {
+                throw new SolanaError(SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_NOT_SUPPORTED, {
+                    unsupportedVersion: value,
+                });
+            }
```
mcintyre94, line 43: I've also disallowed encoding a `TransactionVersion` > 0, but I haven't added other checks for encoding because I don't think there's risk of accidentally encoding the wrong versions from Kit APIs, and we have a Typescript type for valid versions.

I mostly did this because I don't think it makes sense to have different supported/unsupported versions in the encode and decode tests.

packages/errors/src/messages.ts
```diff
@@ -648,4 +649,6 @@ export const SolanaErrorMessages: Readonly<{
     [SOLANA_ERROR__TRANSACTION__SIGNATURES_MISSING]: 'Transaction is missing signatures for addresses: $addresses.',
     [SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_OUT_OF_RANGE]:
         'Transaction version must be in the range [0, 127]. `$actualVersion` given',
+    [SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_NOT_SUPPORTED]:
+        'This version of Kit does not support decoding transactions with version $unsupportedVersion. The current max supported version is 0.',
```
steveluscher, line 653: I like this better than what I wrote in the offchain messages PR. I'm going to use this instead.

packages/transaction-messages/src/decompile-message.ts
```diff
@@ -239,6 +240,15 @@ export function decompileTransactionMessage(
...
 
+    if (
+        compiledTransactionMessage.version !== 'legacy' &&
+        compiledTransactionMessage.version > MAX_SUPPORTED_TRANSACTION_VERSION
+    ) {
+        throw new SolanaError(SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_NOT_SUPPORTED, {
+            actualVersion: compiledTransactionMessage.version,
+        });
+    }
+
```
steveluscher: Does this need to be here? If we:

1. Hardcode `0` into the `VersionedCompiledTransactionMessage` type.
2. Can't decode v1 transactions with the codec after this PR.

…then you should never be able to obtain one of these data structures, right?

mcintyre94: I think you're right, updated. The test I added for this is now a type error (version: 1) so I removed that test. But I think this is cleaner and it makes more sense to restrict the type there.

PR #879, opened 2025-09-11
Error messages for offchain message utilities

steveluscher:

PR #880, opened 2025-09-11
Helpers for encoding and decoding offchain messages

steveluscher:
This PR introduces codecs for version 0 offchain messages according to this spec: https://github.com/anza-xyz/agave/blob/master/docs/src/proposals/off-chain-message-signing.md#application-domain


steveluscher:
* **#991** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/991?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#984** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/984?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#880** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/880?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/880?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

steveluscher:
Removed the signers API stuff that still needs a bit of work.

steveluscher:
I just pushed a change to change `requiredSigners` to `requiredSignatories` in an effort to reduce some ambiguity when later on in the stack I go to create types with names like `OffchainMessageSignatoryWithSigner` (as in: the `Signer` API).

packages/offchain-messages/src/content.ts
```diff
@@ -0,0 +1,189 @@
...
+    format: OffchainMessageContentFormat.UTF8_65535_BYTES_MAX;
+    text: Brand<TContent, 'offchainMessageContentUtf8Of65535BytesMax'>;
+}>;
+
+export type OffchainMessageContent =
+    | OffchainMessageContentRestrictedAsciiOf1232BytesMax
+    | OffchainMessageContentUtf8Of1232BytesMax
+    | OffchainMessageContentUtf8Of65535BytesMax;
+
+export function assertIsOffchainMessageContentRestrictedAsciiOf1232BytesMax(putativeContent: {
```
lorisleiva, line 74: Just a heads up that the exported functions of this file don't have docblocks.

packages/offchain-messages/src/codecs/envelope.ts
```diff
@@ -0,0 +1,193 @@
...
+            getOffchainMessageSigningDomainDecoder(),
+            // Validate the version
+            createDecoder({
+                read(bytes, offset) {
+                    const [_, newOffset] = getOffchainMessageVersionDecoder().read(bytes, offset);
+                    return [undefined, newOffset];
+                },
+            }),
+        ],
+    );
```
lorisleiva: Nice! So if I understand this correctly here you are using the `getHiddenPrefixDecoder` for the signing domain and the version because you want to make sure they don't throw error when being decoded (even though their actual values are ignored) but you're using the `offsetDecoder` for the application domain and the message format because these can be anything and it would be a waste of time to try and decode them.

packages/offchain-messages/src/message-v0.ts
```diff
@@ -0,0 +1,73 @@
...
+export type OffchainMessageWithContent =
+    | OffchainMessageWithRestrictedAsciiOf1232BytesMaxContent
+    | OffchainMessageWithUtf8Of1232BytesMaxContent
+    | OffchainMessageWithUtf8Of65535BytesMaxContent;
+
+export type OffchainMessageV0 = BaseOffchainMessageV0 &
+    OffchainMessageWithContent &
+    OffchainMessageWithRequiredSignatories;
+
+export function assertIsOffchainMessageRestrictedAsciiOf1232BytesMax<TMessage extends OffchainMessageV0>(
```
lorisleiva, line 64: These also don't have any docblocks despite being exposed to the consumer.

packages/offchain-messages/src/codecs/message-v0.ts
```diff
@@ -0,0 +1,111 @@
...
+import { OffchainMessageContentFormat } from '../content';
+import {
+    assertIsOffchainMessageRestrictedAsciiOf1232BytesMax,
+    assertIsOffchainMessageUtf8Of1232BytesMax,
+    assertIsOffchainMessageUtf8Of65535BytesMax,
+    OffchainMessageV0,
+} from '../message-v0';
+import { getOffchainMessageV0PreambleDecoder, getOffchainMessageV0PreambleEncoder } from './preamble-v0';
+
+export function getOffchainMessageV0Decoder(): VariableSizeDecoder<OffchainMessageV0> {
```
lorisleiva, line 43: Same here for the docblocks (and `codecs/message.ts` which is also exposed to the consumer).

packages/offchain-messages/src/codecs/message.ts
```diff
@@ -0,0 +1,67 @@
...
+                case 0:
+                    return getOffchainMessageV0Encoder().write(offchainMessage, bytes, offset);
+                default:
+                    throw new SolanaError(SOLANA_ERROR__OFFCHAIN_MESSAGE__VERSION_NUMBER_NOT_SUPPORTED, {
+                        unsupportedVersion: offchainMessage.version satisfies never,
+                    });
+            }
+        },
+    });
+}
```
lorisleiva, line 86: FWIW I think you could also use a union codec here to avoid to redundancy of `getSizeFromValue`. Something like:

```ts
return getUnionEncoder(
  [
    getOffchainMessageV0Encoder(),
    // getOffchainMessageV1Encoder(), later
  ],
  value => {
    if (value.version != 0) {
      throw new SolanaError(...);
    }
    return value.version; // Returns the index of the array above.
  }
);
```

The equivalent for decoders would be:

```ts
return getUnionDecoder(
  [
    getOffchainMessageV0Decoder(),
    // getOffchainMessageV1Decoder(), later
  ],
  (bytes, offset) => {
    const version = getHiddenPrefixDecoder(getOffchainMessageVersionDecoder(), [
      // Discard the signing domain
      getOffchainMessageSigningDomainDecoder(),
     ]).decode(bytes, offset);
    if (value.version != 0) {
      throw new SolanaError(...);
    }
    return value.version; // Returns the index of the array above.
  }
);
```

packages/offchain-messages/src/codecs/envelope.ts
steveluscher: I can't remember why I did it this way, actually. I think I need to change this anyway, because v1 message don't have an application domain, so I'll need to stop after decoding the version to know what to do next.

packages/offchain-messages/src/codecs/message.ts
steveluscher: I tried this, but ended up with a type error. Do you know why this is a type error? https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgNQIZWKgRgGwKYDKwAXngKIB2AxhACZ5RwC+cAZlBCHAOQACAzhByoKqAPQ16VfgFoaUPNwDcAKFCRYiOAHM8MAKoVgECpUkNmbDlz6DhoiXTzSZtVDFQz+MKAFcqML4K-MoqKjQU3nB41E6MALxwFL44OHCo-HC+FADWFBAA7hTpmWgY2PhEpGZxADzJIFgMAHyqKroGRiY19FAAFADaMeZQALoANHB9AJRw8c1wAAyz-O7A-KzAeKXomLiEJOSxvfW+jS1KQA

steveluscher: Oh! I remember why I did this now. It bothered me that I had to _eagerly_ construct every single codec and then throw them all out but one.

```ts
getUnionCodec([
    createCodec1(),  // Eagerly evaluated
    createCodec2(),  // Eagerly evaluated
    createCodec3(),  // Eagerly evaluated
], () => /* ... */);
```

We might consider changing this API so that `variants` is either:

```ts
TVariants extends (Encoder<any> | () => Encoder<any>)[]
```

or…

```ts
TVariants extends (index: number) => Encoder<any>
```

packages/offchain-messages/src/content.ts
steveluscher: I will do a big documentation pass after everything else comes together!

packages/offchain-messages/src/codecs/message.ts
lorisleiva: Oh yeah that's a good point!

steveluscher:
### Merge activity

* **Nov 19, 10:37 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/880).
* **Nov 19, 10:39 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/880) rebased this pull request as part of a merge.
* **Nov 19, 10:41 PM UTC**: @steveluscher merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/880).

PR #888, opened 2025-09-12
Allow Origin header to be set (for non-browser platforms).

### LINKED ISSUE #887: "Origin" header is Forbidden - createSolanaRpc() not usable in React Native apps when Origin is required
## Overview
`createSolanaRpc()` does not allow configuring an `Origin` header. This prevents us from using Triton's Cascade SWQOS, which enforces specific origins. I'm guessing this was added to make it a little bit harder to spoof and rely on browsers automatically adding the Origin header, **but misses the common use-case of mobile applications - React Native does not automatically set an Origin header.**

from `http-transport-headers.d.ts`:
```
type ForbiddenHeaders = 'Accept-Charset'
/**
 * Though technically forbidden in non-Node environments, we don't have a way to target
 * TypeScript types depending on which platform you are authoring for. `Accept-Encoding` is
 * therefore omitted from the forbidden headers type, but is still a runtime error in dev mode
 * when supplied in a non-Node context.
 */
 | 'Access-Control-Request-Headers' | 'Access-Control-Request-Method' | 'Connection' | 'Content-Length' | 'Cookie' | 'Date' | 'DNT' | 'Expect' | 'Host' | 'Keep-Alive' | 'Origin' | 'Permissions-Policy' | 'Referer' | 'TE' | 'Trailer' | 'Transfer-Encoding' | 'Upgrade' | 'Via' | `Proxy-${string}` | `Sec-${string}`;
export declare function assertIsAllowedHttpRequestHeaders(headers: Record<string, string>): asserts headers is AllowedHttpRequestHeaders;
```

## Steps to reproduce
`createSolanaRpc("your-rpc-url", { headers: "Origin": "https://some.origin" })`

prashanFOMO:
#### Problem
As described in the ticket, React Native clients do not automatically set an Origin header. This is preventing us from using Triton's SWQOS since it restricts access by Origin header. 

#### Summary of Changes
Remove Origin from list of forbidden headers.

Fixes #887

### DIFF
--- a/.changeset/eager-lizards-stand.md
+++ b/.changeset/eager-lizards-stand.md
@@ -2,4 +2,4 @@
 '@solana/rpc-transport-http': minor
 ---
 
-Remove Origin from list of forbidden RPC headers
+The React Native and Node builds now permit you to set the `Origin` header. This header continues to be forbidden in the browser build, as it features on the list of forbidden request headers: https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_request_header
--- a/.changeset/eager-lizards-stand.md
+++ b/.changeset/eager-lizards-stand.md
@@ -1,0 +1,5 @@
+---
+'@solana/rpc-transport-http': minor
+---
+
+Remove Origin from list of forbidden RPC headers
--- a/packages/rpc-transport-http/src/__tests__/http-transport-headers-test.ts
+++ b/packages/rpc-transport-http/src/__tests__/http-transport-headers-test.ts
@@ -27,7 +27,6 @@ describe('assertIsAllowedHttpRequestHeader', () => {
         'Expect',
         'Host',
         'Keep-Alive',
-        'Origin',
         'Permissions-Policy',
         'Proxy-Anything',
         'Proxy-Authenticate',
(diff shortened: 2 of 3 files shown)


steveluscher (changes requested):
Take a look at what we did with `Accept-Encoding` to make it a runtime error when supplied in a non-Node context. Could you do something similar when `Origin` is supplied in a non-React-Native context? Should we actually allow `Origin` in all non-browser contexts, but make it a runtime error in browser contexts?

prashanFOMO:
@steveluscher thanks for the review! IMO it feels reasonable to allow this header across the board: browsers will just ignore it anyways + the use case I cite here could apply to any non-browser environment?

But won't die on this hill, happy to make the change if you'd like.

steveluscher:
> IMO it feels reasonable to allow this header across the board: browsers will just ignore it anyways…

The goal of the type system is to alert the developer of invalid inputs _before_ they run their software, and the goal of runtime invariants is to alert them of invalid inputs the moment the call is made. Now type enforcement is not possible here just for browser builds, since TypeScript doesn't have the concept of ‘use these types for browser builds only,’ but if we take away the runtime enforcement then that will leave the developer on a treasure hunt for why their `Origin` header is having no effect. Failing loudly should shorten the gap between them shipping such a bug, and repairing it.

prashanFOMO:
@steveluscher made the change - lmk if it looks good. will leave it to you to decide in the future whether this should be an error in Node envs, this change should unblock React Native users.

PR #901, opened 2025-09-19
Add `SignatureBytes` assertion, guard, and coercion methods

guibescos:
#### Problem

There is currently no way to create values with the SignatureBytes in packages that depend on @solana/kit (expect typecasting).

#### Summary of Changes

Add `assertIsSignatureBytes`, `isSignatureBytes` and `signatureBytes` which are analogous to the current `assertIsSignature`, `isSignature` and `signature` when tne signature is a byte array.


steveluscher:
I'm down. If you want to write typetests and tests (just copy-paste the `assertIsSignature` et cetera ones from `@solana/keys`) that would be great. If you don't have time and can wait a little longer, one of us can do it; just let us know.

packages/keys/src/__tests__/signatures-test.ts
```diff
@@ -143,6 +143,38 @@ describe('assertIsSignature()', () => {
...
+        expect(() => {
+            assertIsSignatureBytes(new Uint8Array(0));
+        }).toThrow();
+    });
+    it('throws when the byte array has a length other than 64 bytes', () => {
+        expect(() => {
+            assertIsSignatureBytes(new Uint8Array(63));
+        }).toThrow();
+    });
+    it('does not throw when supplied a base-58 encoded signature', () => {
```
steveluscher:
```suggestion
    it('does not throw when supplied a 64-byte signature as a bytearray', () => {
```

packages/keys/src/__tests__/signatures-test.ts
```diff
@@ -143,6 +143,38 @@ describe('assertIsSignature()', () => {
...
+
+            // example signatures
+            assertIsSignatureBytes(MOCK_DATA_SIGNATURE);
+            jest.mocked(getBase58Encoder).mockReturnValue(originalGetBase58Encoder);
+            assertIsSignatureBytes(
+                getBase58Encoder().encode(
+                    '5HkW5GttYoahVHaujuxEyfyq7RwvoKpc94ko5Fq9GuYdyhejg9cHcqm1MjEvHsjaADRe6hVBqB2E4RQgGgxeA2su',
+                ),
+            );
+        }).not.toThrow();
```
steveluscher, line 165: The way this test is written, if it throws you don't know why. Is it because the method doesn't like all zeros? Is it because it erroneously breaks on `MOCK_DATA_SIGNATURE`? Is it because someone else broke the base58 decoder?

This should be three tests, and even then I don't think you should include the base58 encoder here because to do so means that you're testing _two_ parts of the library when you should only be testing the parts that `assertIsSignatureBytes` is responsible for.

packages/keys/src/__typetests__/signatures-typetest.ts
```diff
@@ -18,3 +18,20 @@ import { Signature, signature } from '../signatures';
...
+    }
+}
+
+// [DESCRIBE] SignatureBytes
+{
+    // It satisfies the `Uint8Array` brand
+    {
+        const signatureBytes = null as unknown as SignatureBytes;
+        signatureBytes satisfies Uint8Array;
+    }
```
steveluscher:
```suggestion
    // It satisfies the `ReadonlyUint8Array` type
    {
        const signatureBytes = null as unknown as SignatureBytes;
        signatureBytes satisfies ReadonlyUint8Array;
    }
```

packages/keys/src/__tests__/signatures-test.ts
```diff
@@ -171,7 +203,7 @@ describe('verify', () => {
     });
     it('returns `true` when the correct signature is supplied for a given payload', async () => {
         expect.assertions(1);
-        const result = await verifySignature(mockPublicKey, MOCK_DATA_SIGNATURE, MOCK_DATA);
+        const result = await verifySignature(mockPublicKey, signatureBytes(MOCK_DATA_SIGNATURE), MOCK_DATA);
```
steveluscher: By adding this here, the `signatureBytes()` method comes under test by the `verify()` tests. This means that if someone breaks the `signatureBytes()` implementation, it will break the `verify()` tests, which is misleading, because verify _did not break_.

In general, never stack unrelated implementations in the same test.

```suggestion
        const result = await verifySignature(mockPublicKey, MOCK_DATA_SIGNATURE as SignatureBytes, MOCK_DATA);
```

packages/keys/src/__tests__/signatures-test.ts
```diff
@@ -143,6 +143,38 @@ describe('assertIsSignature()', () => {
     });
 });
 
+describe('assertIsSignatureBytes()', () => {
+    it('throws when supplied an empty byte array', () => {
+        expect(() => {
+            assertIsSignatureBytes(new Uint8Array(0));
+        }).toThrow();
+    });
+    it('throws when the byte array has a length other than 64 bytes', () => {
```
steveluscher: Nit: you could do your same `it.each([63, 65])` trick that you did above, here.

packages/keys/src/__typetests__/signatures-typetest.ts
guibescos: Turns out SignatureBytes is a branded `Uint8Array` and not `ReadonlyUint8Array`

mcintyre94: Would it add a lot of change to this PR to change it to `ReadonlyUint8Array`? It definitely should be, but we can handle that separately if it's a lot of churn.

steveluscher (approved):

PR #904, opened 2025-09-24
Export all of the channel creators from `@solana/rpc-subscriptions`

steveluscher:
#### Problem

In https://github.com/trezor/trezor-suite/issues/19354 I suggested that the maintainers fork the default channel creator so that they can omit the autopinger. It appears as though that isn't even possible, because we don't export everything from `@solana/rpc-subscriptions`

#### Summary of Changes

Exported all of the raw material that one would need to fork `createDefaultSolanaRpcSubscriptionsChannelCreator()`.

### DIFF
--- a/.changeset/three-zoos-act.md
+++ b/.changeset/three-zoos-act.md
@@ -1,0 +1,5 @@
+---
+'@solana/rpc-subscriptions': patch
+---
+
+yExported all of the channel creators that form part of `createDefaultSolanaRpcSubscriptionsChannelCreator()` so that developers can configure their own custom channels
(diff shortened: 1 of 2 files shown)


steveluscher:
* **#904** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/904?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/904?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

steveluscher:
Is this sensible?

lorisleiva (approved):
I think that's a good move. Otherwise we're forcing consumers to re-invent the wheel when all they need to do is tweak the default channel.

mcintyre94 (approved):
Agree with this - it matches how we always intended Kit to be able to be used

PR #906, opened 2025-09-25
fix: fromVersionedTransaction bug

guibescos:
#### Problem

`fromVersionedTransaction` currently fails for `VersionedTransaction` that use address lookup tables.
The reason for this is `fromVersionedTransaction` tries to resolve all the accounts for the input `VersionedTransaction`, but doesn't have access to the lookup tables involved in the transaction. However if we look further in the function we notice we are only interested in the static accounts, which is where the signers are.

#### Summary of Changes

Use `message.staticAccountKeys` instead of `message.getAccountKeys()` which tries to resolve all the accounts.

### DIFF
--- a/.changeset/rare-bears-fix.md
+++ b/.changeset/rare-bears-fix.md
@@ -1,0 +1,5 @@
+---
+"@solana/compat": patch
+---
+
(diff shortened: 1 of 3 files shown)


steveluscher (approved):
giphy-2126420993

PR #907, opened 2025-09-25
Prevent the documentation preview from running on PRs from forks

steveluscher:
#### Problem

The Vercel deploy secrets are not available on pull requests from forks, so this workflow will never work. Right now no third-party PRs will pass CI.

#### Summary of Changes

Disable the docs preview when the PR comes from a fork.


lorisleiva (approved):
Thanks!

PR #918, opened 2025-09-30
Add a function to extract the lifetime from a CompiledTransactionMessage

mcintyre94:
#### Problem

When writing a signer, it is possible that you receive a Transaction from for example an external wallet, and need to determine its lifetime constraint.

#### Summary of Changes

This PR adds a new function `getTransactionLifetimeConstraintFromCompiledTransactionMessage` which enables a best-effort extraction of the lifetime constraint from a `CompiledTransactionMessage`, ie without full information about the transaction.

The check is less robust than the equivalent for a `TransactionMessage`, because the accounts passed to an advance nonce instruction may be in lookup tables - and therefore we can't verify their address or role. However, we are able to check the program address, program data and number of accounts passed in. A transaction that incorrectly passes this check is one that passes incorrect accounts to the advance nonce instruction. The impact of attempting to confirm it with an incorrect lifetime should be minimal (it will fail to land regardless).

There is one error case I've added - if the first account passed to the instruction is not a static address, then we can't use its address in the returned `TransactionDurableNonceLifetime` and we throw. I think this is an edge case and not worth the cost of resolving lookup tables.

This function will be used to enable `useWalletAccountTransactionSigner` to return a lifetime in cases where the wallet modifies the lifetime token of the input transaction. I think it will also be useful to anybody else writing similar signers.


mcintyre94:
* **#927** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/927?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#919** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/919?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#918** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/918?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/918?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/transactions/src/lifetime.ts
```diff
@@ -102,3 +106,61 @@ export type SetTransactionLifetimeFromTransactionMessage<
...
+ */
+export function getTransactionLifetimeConstraintFromCompiledTransactionMessage(
+    compiledTransactionMessage: CompiledTransactionMessage & CompiledTransactionMessageWithLifetime,
+): TransactionBlockhashLifetime | TransactionDurableNonceLifetime {
+    const firstInstruction = compiledTransactionMessage.instructions[0];
+    const { staticAccounts } = compiledTransactionMessage;
+
+    // We need to check if the first instruction is an AdvanceNonceAccount instruction
+    if (firstInstruction && compiledInstructionIsAdvanceNonceInstruction(firstInstruction, staticAccounts)) {
+        const nonceAccountAddress = staticAccounts[firstInstruction.accountIndices![0]];
```
steveluscher: Rather than add a `!` here, could we instead do this?
```suggestion
function compiledInstructionIsAdvanceNonceInstruction(
    instruction: CompiledTransactionMessage['instructions'][number],
    staticAddresses: Address[],
): instruction is typeof instruction & { accountIndices: [number, number, number] } {
    return (
        staticAddresses[instruction.programAddressIndex] === SYSTEM_PROGRAM_ADDRESS &&
        // Test for `AdvanceNonceAccount` instruction data
        instruction.data != null &&
        isAdvanceNonceAccountInstructionData(instruction.data) &&
        // Test for exactly 3 accounts
        instruction.accountIndices?.length === 3
    );
}

function isAdvanceNonceAccountInstructionData(data: ReadonlyUint8Array): boolean {
    // AdvanceNonceAccount is the fifth instruction in the System Program (index 4)
    return data.byteLength === 4 && data[0] === 4 && data[1] === 0 && data[2] === 0 && data[3] === 0;
}

/**
 * Check if a compiled instruction is an AdvanceNonceAccount instruction
 * @param compiledTransactionMessage A compiled transaction message that includes a lifetime token
 * @returns True if the instruction is an AdvanceNonceAccount instruction, false otherwise
 * Note that this is less precise than checking a decompiled instruction, as we can't inspect
 * the address or role of input accounts (which may be in lookup tables). However, this is
 * sufficient for all valid advance durable nonce instructions.
 * Note that the program address must not be in a lookup table, see [this answer on StackExchange](https://solana.stackexchange.com/a/16224/289)
 * @see {@link isAdvanceNonceAccountInstruction}
 */
export function getTransactionLifetimeConstraintFromCompiledTransactionMessage(
    compiledTransactionMessage: CompiledTransactionMessage & CompiledTransactionMessageWithLifetime,
): TransactionBlockhashLifetime | TransactionDurableNonceLifetime {
    const firstInstruction = compiledTransactionMessage.instructions[0];
    const { staticAccounts } = compiledTransactionMessage;

    // We need to check if the first instruction is an AdvanceNonceAccount instruction
    if (firstInstruction && compiledInstructionIsAdvanceNonceInstruction(firstInstruction, staticAccounts)) {
        const nonceAccountAddress = staticAccounts[firstInstruction.accountIndices[0]];
```

packages/transactions/src/lifetime.ts
```diff
@@ -102,3 +106,61 @@ export type SetTransactionLifetimeFromTransactionMessage<
...
+        }
+        return {
+            nonce: compiledTransactionMessage.lifetimeToken as Nonce,
+            nonceAccountAddress,
+        };
+    } else {
+        return {
+            blockhash: compiledTransactionMessage.lifetimeToken as Blockhash,
+            // This is not known from the compiled message, so we set it to the maximum possible value
+            lastValidBlockHeight: 2n ** 64n - 1n,
```
steveluscher: I see. So this is the ‘reconstruct it but actually punt on the last valid block height’ strategy. My only reservation here is that if we release this function as sync, we can't go back on that later and implement actual ‘go get the last valid block height’ code.

Seems we have three options:

1. Don't care. `u64::MAX` it is. Block height based transaction confirmation expiry won't work.
2. Release this as like `createTotallyMadeUpNonExipiringTransactionLifetimeConstraintFromCompiledTransactionMessage()` and then later release a `getTransactionLifetimeConstraintFromCompiledTransactionMessage()` that is async and takes in an `Rpc<GetAccountInfoApi>` or whatever as part of the config
3. Make this `async getTransactionLifetimeConstraintFromCompiledTransactionMessage(compiledTransactionMessage)` now and `getTransactionLifetimeConstraintFromCompiledTransactionMessage(compiledTransactionMessage, rpc?: Rpc<GetAccountInfoApi>)` later as a backward compatible upgrade.

packages/transactions/src/__tests__/lifetime-test.ts
```diff
@@ -0,0 +1,122 @@
...
+        const compiledTransactionMessage = {
+            instructions: [],
+            lifetimeToken: 'abc',
+        } as unknown as CompiledTransactionMessage & CompiledTransactionMessageWithLifetime;
+
+        const lifetime = getTransactionLifetimeConstraintFromCompiledTransactionMessage(compiledTransactionMessage);
+        expect(lifetime).toEqual({ blockhash: 'abc' as Blockhash, lastValidBlockHeight: U64_MAX });
+    });
+
+    describe('returns a blockhash transaction lifetime when the first instruction is not an AdvanceNonceAccount instruction', () => {
```
steveluscher, line 28: Maybe add tests that are like ‘because it has no `accountIndices` property’ and ‘because it has no `data` property.’

packages/transactions/src/lifetime.ts
steveluscher: Can save everyone a bit of math.
```suggestion
            lastValidBlockHeight: 0xFFFFFFFFFFFFFFFFn,
```

steveluscher (commented):
I don't have answers, only questions. Should we take the same approach as with transaction confirmation and split these into one function for blockhash based transactions and one for nonce transactions, presuming that the applications knows which one it prepared, rather than to make one function that autodetects the type of lifetime? I presume the answer is no, because in the next PR it's presumed that the wallet-account-transaction-signer is a `TransactionModifyingSigner`, meaning that just because the transaction went _in_ with a blockhash-based lifetime doesn't mean it's going to come out with one.

steveluscher (approved):

packages/transactions/src/lifetime.ts
mcintyre94: Let's do 3. It's designed for signers, which are async. If we add `rpc` later then we could extend the durable nonce check to resolve lookup tables too, if we think that's worth doing.

packages/transactions/src/__tests__/lifetime-test.ts
```diff
@@ -0,0 +1,151 @@
...
+import {
+    CompiledTransactionMessage,
+    CompiledTransactionMessageWithLifetime,
+    Nonce,
+} from '@solana/transaction-messages';
+
+import { getTransactionLifetimeConstraintFromCompiledTransactionMessage } from '../lifetime';
+
+const SYSTEM_PROGRAM_ADDRESS = '11111111111111111111111111111111' as Address;
+const U64_MAX = 2n ** 64n - 1n;
```
lorisleiva, line 13: nit: Do you want to make the same change here?

packages/transactions/src/lifetime.ts
```diff
@@ -102,3 +106,63 @@ export type SetTransactionLifetimeFromTransactionMessage<
...
+
+function isAdvanceNonceAccountInstructionData(data: ReadonlyUint8Array): boolean {
+    // AdvanceNonceAccount is the fifth instruction in the System Program (index 4)
+    return data.byteLength === 4 && data[0] === 4 && data[1] === 0 && data[2] === 0 && data[3] === 0;
+}
+
+/**
+ * Check if a compiled instruction is an AdvanceNonceAccount instruction
+ * @param compiledTransactionMessage A compiled transaction message that includes a lifetime token
+ * @returns True if the instruction is an AdvanceNonceAccount instruction, false otherwise
```
lorisleiva: Is this comment correct? It doesn't seem to match the signature.

packages/transactions/src/lifetime.ts
```diff
@@ -102,3 +106,63 @@ export type SetTransactionLifetimeFromTransactionMessage<
...
+ * Note that this is less precise than checking a decompiled instruction, as we can't inspect
+ * the address or role of input accounts (which may be in lookup tables). However, this is
+ * sufficient for all valid advance durable nonce instructions.
+ * Note that the program address must not be in a lookup table, see [this answer on StackExchange](https://solana.stackexchange.com/a/16224/289)
+ * @see {@link isAdvanceNonceAccountInstruction}
+ * Note that this function is async to allow for future implementations that may fetch `lastValidBlockHeight` using an RPC
+ */
+// eslint-disable-next-line @typescript-eslint/require-await
+export async function getTransactionLifetimeConstraintFromCompiledTransactionMessage(
```
lorisleiva, line 143: Sorry if I've missed something but it doesn't seem like this function needs to be asynchronous. Is it to future proof it?

lorisleiva (approved):
Thank you for your work on that stack! I'm gradually making my way through your PRs. ☺️

packages/transactions/src/__tests__/lifetime-test.ts
mcintyre94: I don't have a strong opinion either way, but I think this reads a bit clearer about the intention for the tests which was why I didn't change it here.

packages/transactions/src/lifetime.ts
mcintyre94: Yep, this was Steve's suggested option 3 here: https://app.graphite.dev/github/pr/anza-xyz/kit/918/Add-a-function-to-extract-the-lifetime-from-a-CompiledTransactionMessage#comment-PRRC_kwDONc0xOc6OnleT

If we make this async then we can add an optional `Rpc<Something>` parameter later if we get functionality to resolve a `lastValidBlockHeight` for a blockhash, without it being a breaking change. Since making it async doesn't affect the signer use case I think it makes sense.

lorisleiva: Gotcha, thank you for clarifying. 🙏

mcintyre94:
### Merge activity

* **Oct 3, 10:36 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/918).
* **Oct 3, 10:36 AM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/918).

Issue #891, opened 2025-09-15
Typing of `modifyAndSignTransactions` is incorrect


mcintyre94:
## Overview

This is the root cause of #886 

We allow modifying signers, which can modify the input transaction at the point they sign it. These need to return a `Transaction & TransactionWithLifetime` in order for us to return the correct type from `signModifyingAndPartialTransactionSigners`. This is required because we need to carry the lifetime constraint forward for our transaction confirmation logic.

https://github.com/anza-xyz/kit/blob/54a140359a38e0d69c0596adbf6af8c43afd19cb/packages/signers/src/sign-transaction.ts#L290-L300

We cast to this type, but do not enforce that modifying signers actually return it:

https://github.com/anza-xyz/kit/blob/54a140359a38e0d69c0596adbf6af8c43afd19cb/packages/signers/src/transaction-modifying-signer.ts#L56-L62

In addition, the function `signModifyingAndPartialTransactionSigners` is returning a `TransactionFromTransactionMessage<TTransactionMessage>`:

https://github.com/anza-xyz/kit/blob/54a140359a38e0d69c0596adbf6af8c43afd19cb/packages/transactions/src/transaction.ts#L30-L34

As we are allowing the signers to modify the transaction, this type is more restrictive than we can guarantee. A modifying signer could modify the size or lifetime of a transaction. 

## Steps to reproduce

I haven't written a repro showing the problem this causes, but [Playground Link](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgKDmuMCeYCmcAqUAhgHYDORAxjMBCQMrADmJOUANCunAIIAmfKDjJlkAXzgAzKBBBwA5AAEyEADakiAegDWwGPIDcyZJTpl4MYuSo06jFmwBcBKxWq0GzVlDgBeRKjoRAJCIs7yRABGlPJwRGS8IcJk7IFoIBB8wJKYPCR89iSEpG625M7xmCSUGAAeCb4AfAHc3JLQcAAUpuQWtXAQknVkAJQtrRPZnfKq2Tg0IDgAwmaWRMAk+nAbdWMIaROHfDiq83gwtQDaM3MLy6vEG-oAukaHrWIH3J-vQjAArlASMM3uhPp8gA) shows a modifying signer that does not return a lifetime constraint.

[...]

PR #919, opened 2025-09-30
Update useWalletAccountTransactionSigner to return a LifetimeConstraint for the updated transaction

mcintyre94:
#### Problem

Currently this signer only returns what is required for the signer interface, a `Transaction`. However, per #891 this interface needs to be changed to return a `Transaction & TransactionWithLifetime`, ie augmented with a `lifetimeConstraint` field.

#### Summary of Changes

The signer now returns a lifetime, by decoding the returned `messageBytes` and comparing the returned `lifetimeToken` with the existing lifetime of the input transaction.

- If the input transaction does not have a lifetime, then we create one for the signed transaction using #918. Otherwise:
- If the input transaction and signed transaction have identical message bytes, then we return the existing lifetime. Ie if the wallet returns the transaction unchanged.
- If the message bytes differ, but the `lifetimeToken` of the signed transaction matches the existing lifetime (either blockhash or nonce field), then we return the existing lifetime. Ie if the wallet modifies the transaction but not its lifetime.
- If the lifetime token of the signed transaction differs from that of the input transaction, then we create a new lifetime for the signed transaction using #918. 

This pre-emptively makes `useWalletAccountTransactionSigner` comply with a stricter interface for `TransactionModifyingSigner` that will require returning `Transaction & TransactionWithLifetime`.

### DIFF
--- a/packages/react/src/__tests__/useWalletAccountTransactionSigner-test.ts
+++ b/packages/react/src/__tests__/useWalletAccountTransactionSigner-test.ts
@@ -1,8 +1,22 @@
 import { Address } from '@solana/addresses';
-import type { VariableSizeCodec } from '@solana/codecs-core';
-import { SOLANA_ERROR__SIGNER__WALLET_MULTISIGN_UNIMPLEMENTED, SolanaError } from '@solana/errors';
+import type { VariableSizeCodec, VariableSizeDecoder } from '@solana/codecs-core';
+import {
+    SOLANA_ERROR__SIGNER__WALLET_MULTISIGN_UNIMPLEMENTED,
+    SOLANA_ERROR__TRANSACTION__NONCE_ACCOUNT_CANNOT_BE_IN_LOOKUP_TABLE,
+    SolanaError,
+} from '@solana/errors';
 import { SignatureBytes } from '@solana/keys';
-import { Transaction, TransactionMessageBytes } from '@solana/transactions';
+import { Blockhash } from '@solana/rpc-types';
+import {
+    CompiledTransactionMessage,
+    CompiledTransactionMessageWithLifetime,
+    getCompiledTransactionMessageDecoder,
+} from '@solana/transaction-messages';
+import {
+    getTransactionLifetimeConstraintFromCompiledTransactionMessage,
+    Transaction,
+    TransactionMessageBytes,
+} from '@solana/transactions';
 import { getTransactionCodec } from '@solana/transactions';
 import type { UiWalletAccount } from '@wallet-standard/ui';
 
@@ -10,6 +24,7 @@ import { renderHook } from '../test-renderer';
 import { useSignTransaction } from '../useSignTransaction';
 import { useWalletAccountTransactionSigner } from '../useWalletAccountTransactionSigner';
 
+jest.mock('@solana/transaction-messages');
 jest.mock('@solana/transactions');
 jest.mock('../useSignTransaction');
 
@@ -120,8 +135,8 @@ describe('useWalletAccountTransactionSigner', () => {
     });
     it('decodes and returns the signed transaction bytes returned by `signTransactions`', async () => {
         expect.assertions(2);
-        const mockSignedTransaction = new Uint8Array([1, 2, 3]);
-        const mockDecodedTransaction = {} as Transaction;
+        const mockSignedTransaction = new Uint8Array([1, 2, 3, 4, 5, 6]);
+        const mockDecodedTransaction = { messageBytes: [1, 2, 3] } as unknown as Transaction;
         mockSignTransaction.mockResolvedValue({ signedTransaction: mockSignedTransaction });
         mockDecodeTransaction.mockReturnValue(mockDecodedTransaction);
         const { result } = renderHook(() => useWalletAccountTransactionSigner(mockUiWalletAccount, 'solana:danknet'));
@@ -130,7 +145,9 @@ describe('useWalletAccountTransactionSigner', () => {
             throw result.current;
         } else {
             const { modifyAndSignTransactions } = result.current;
+            const lifetimeConstraint = { blockhash: 'abc', lastValidBlockHeight: 123n };
             const inputTransaction = {
+                lifetimeConstraint,
                 messageBytes: new Uint8Array([1, 2, 3]) as unknown as TransactionMessageBytes,
                 signatures: {
                     '11111111111111111111111111111114': new Uint8Array(64).fill(2) as SignatureBytes,
@@ -141,7 +158,7 @@ describe('useWalletAccountTransactionSigner', () => {
             // eslint-disable-next-line jest/no-conditional-expect
             expect(mockDecodeTransaction).toHaveBeenCalledWith(mockSignedTransaction);
             // eslint-disable-next-line jest/no-conditional-expect
-            await expect(signPromise).resolves.toEqual([mockDecodedTransaction]);
+            await expect(signPromise).resolves.toEqual([{ ...mockDecodedTransaction, lifetimeConstraint }]);
         }
     });
     it('calls `signTransaction` with all options except the `abortSignal`', () => {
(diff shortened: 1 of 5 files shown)


mcintyre94:
* **#927** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/927?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#919** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/919?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/919?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#918** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/918?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/react/src/__tests__/useWalletAccountTransactionSigner-test.ts
```diff
@@ -1,15 +1,31 @@
...
+} from '@solana/transaction-messages';
+import {
+    getTransactionLifetimeConstraintFromCompiledTransactionMessage,
+    Transaction,
+    TransactionMessageBytes,
+} from '@solana/transactions';
 import { getTransactionCodec } from '@solana/transactions';
 import type { UiWalletAccount } from '@wallet-standard/ui';
 
+import { Blockhash } from '../../../rpc-types/dist/types';
```
steveluscher: Gotta add this to `package.json` to avoid this.

```suggestion
import { Blockhash } from '@solana/rpc-types';
import {
    CompiledTransactionMessage,
    CompiledTransactionMessageWithLifetime,
    getCompiledTransactionMessageDecoder,
} from '@solana/transaction-messages';
import {
    getTransactionLifetimeConstraintFromCompiledTransactionMessage,
    Transaction,
    TransactionMessageBytes,
} from '@solana/transactions';
import { getTransactionCodec } from '@solana/transactions';
import type { UiWalletAccount } from '@wallet-standard/ui';

```

packages/react/src/useWalletAccountTransactionSigner.ts
```diff
@@ -68,7 +80,7 @@ export function useWalletAccountTransactionSigner<TWalletAccount extends UiWalle
                     throw new SolanaError(SOLANA_ERROR__SIGNER__WALLET_MULTISIGN_UNIMPLEMENTED);
                 }
                 if (transactions.length === 0) {
-                    return transactions;
+                    return [];
```
steveluscher: y tho?

packages/react/src/useWalletAccountTransactionSigner.ts
```diff
@@ -80,9 +92,69 @@ export function useWalletAccountTransactionSigner<TWalletAccount extends UiWalle
...
+                    return Object.freeze([
+                        {
+                            ...decodedSignedTransaction,
+                            lifetimeConstraint,
+                        },
+                    ]);
+                } catch (e) {
+                    if (isSolanaError(e, SOLANA_ERROR__TRANSACTION__NONCE_ACCOUNT_CANNOT_BE_IN_LOOKUP_TABLE)) {
+                        throw new SolanaError(SOLANA_ERROR__SIGNER__NONCE_ACCOUNT_CANNOT_BE_IN_LOOKUP_TABLE, {
+                            signedTransaction,
```
steveluscher: So, I presume that you attach the `signedTransaction` to the error context in the case that the caller wants to say ‘I don't care about the lifetime, just give me the damn signature.’

Tricky. We can't stuff signatures into the error context for fear of logging, but I can't think of another way of getting them out of this callback.

**How about we just punt on this case, throw your error, and offer no remediation**. To say it back to you, the conditions that would have to stack up for this error to ever get hit are:

* A proposer would need to be using nonces, AND
* That nonce account would have to be found only in an address lookup table, AND
* Someone would have to be using a Wallet Standard wallet to sign that transaction, AND
* The wallet would have had to have modified the transaction before signing

Other than being a lot of ducks to line up, all in a row, for things to go wrong, I think it's actually pretty unlikely. The primary use case for nonce transactions is to make it possible for multiple signers to participate in a transactions in cases where it will take longer than the expiry of a blockhash to organize all of them. If you have a wallet _modifying_ the transaction message in such a case, your signature will mismatch with the others unless their wallets also made the exact same modification. Unlikely.

packages/errors/src/context.ts
```diff
@@ -582,6 +583,9 @@ export type SolanaErrorContext = DefaultUnspecifiedErrorContextToUndefined<
         [SOLANA_ERROR__SIGNER__EXPECTED_TRANSACTION_SIGNER]: {
             address: string;
         };
+        [SOLANA_ERROR__SIGNER__NONCE_ACCOUNT_CANNOT_BE_IN_LOOKUP_TABLE]: {
+            signedTransaction: Uint8Array;
```
steveluscher: We have to keep in mind that in production, the error context object gets serialized and logged. We must keep sensitive information out of error context, and this probably qualifies as sensitive enough to exclude.

packages/react/src/__tests__/useWalletAccountTransactionSigner-test.ts
mcintyre94: Do we want to add dependencies that are only used in tests? I've noticed if you use the `dist/types` in published code then you get errors, but it doesn't seem to cause any issue in tests. Happy to add it if that's what we do elsewhere though! Do we use `devDependencies` for this case, or just `dependencies` if it's used by any code? I think we've discussed this before but can't remember where we ended up.

packages/react/src/useWalletAccountTransactionSigner.ts
mcintyre94: Yep it was just to provide a recovery mechanism when you get that `SOLANA_ERROR__TRANSACTION__NONCE_ACCOUNT_CANNOT_BE_IN_LOOKUP_TABLE` edge case. I figured that you'd still get the signed transaction in your app, you just wouldn't have a lifetime - but you might be able to recover depending what your app needs.

I agree it's unlikely and that logging the transaction is too dangerous. I'll just drop the custom error here, not worth the risk and - as you say - very unlikely to happen in practice.

packages/react/src/useWalletAccountTransactionSigner.ts
mcintyre94: When I change the interface to make this return `Transaction & TransactionWithLifetime` in the later PR, Typescript sees that the input transactions don't fulfil that and errors on `return transactions`. It doesn't do anything clever with the knowledge that the list is empty, and thinks it's returning a list of things that don't match the type. This was fine before because it was expected to return the same type as its input.

packages/react/src/__tests__/useWalletAccountTransactionSigner-test.ts
steveluscher: Pretty sure jamming it in `devDependencies` does the job! Crucially, that will cause Turborepo to build it, whereas just indexing into a `dist/` that may or may not exist yet won't.

packages/react/src/useWalletAccountTransactionSigner.ts
steveluscher: I think you should probably still just tell it to stfu via `return transactions as readonly (Transaction & TransactionWithLifetime)[]`. This keeps the property that no change implies that you get the same object back (in the `===` sense).

mcintyre94: Gotcha - I'll leave this as `return transactions` here, and add that cast upstream when I change the type

packages/react/src/useWalletAccountTransactionSigner.ts
mcintyre94: Okay I've removed the specific error and error handling here. It'll just throw `SOLANA_ERROR__TRANSACTION__NONCE_ACCOUNT_CANNOT_BE_IN_LOOKUP_TABLE` in this annoying edge case now.

lorisleiva (approved):
Looks good to me!

mcintyre94:
### Merge activity

* **Oct 3, 10:36 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/919).
* **Oct 3, 10:38 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/919) rebased this pull request as part of a merge.
* **Oct 3, 10:39 AM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/919).

PR #925, opened 2025-10-01
Add functions to narrow a TransactionWithLifetime to a specific lifetime

mcintyre94:
#### Summary of Changes

Fixes #

PR #927, opened 2025-10-01
Update the signer API to return Transaction & TransactionWithLifetime

### LINKED ISSUE #891: Typing of `modifyAndSignTransactions` is incorrect

mcintyre94:
#### Problem

The `TransactionModifyingSigner` was previously typed to return the exact same type `T` as its input transactions. This is incorrect, as we always intended for signers to be able to eg change the lifetime of a transaction. In practice one example of `TransactionModifyingSigner` returns signed transactions from wallet-standard wallets, which can make arbitrary changes to the transaction before signing it.

This also meant that such a signer could not return a `TransactionWithLifetime`, ie a `lifetimeConstraint` field. Casts in signer functions would treat the signer as having returned the expected type, but there was no type safety.

In practice, this meant `useWalletAccountTransactionSigner` and third-party `TransactionModifyingSigner` did not return a `lifetimeConstraint`, but appeared to do so to Typescript. This led to a runtime error when attempting to confirm transactions (#891) 

#### Summary of Changes

This PR changes the type of modifying signers to:
**input**: (Transaction | (Transaction & TransactionWithLifetime))[]
**output**: (Transaction & TransactionWithLifetime & TransactionWithinSizeLimit)[]

Note that an upstream PR (#919) already made `useWalletAccountTransactionSigner` satisfy the `TransactionWithLifetime` part of this interface.

It also changes the return type of `signModifyingAndPartialTransactionSigners` to `Transaction & TransactionWithLifetime & TransactionWithinSizeLimit`, and removes the cast to the expected return type that limited type safety in this function.

Fixes #891

### DIFF
--- a/packages/signers/src/sign-transaction.ts
+++ b/packages/signers/src/sign-transaction.ts
@@ -4,9 +4,9 @@ import { BaseTransactionMessage, TransactionMessageWithFeePayer } from '@solana/
 import {
     assertIsFullySignedTransaction,
     compileTransaction,
-    FullySignedTransaction,
+    SendableTransaction,
     Transaction,
-    TransactionFromTransactionMessage,
+    TransactionWithinSizeLimit,
     TransactionWithLifetime,
 } from '@solana/transactions';
 
@@ -41,8 +41,6 @@ import { assertIsTransactionMessageWithSingleSendingSigner } from './transaction
  * {@link TransactionModifyingSigner} if no other signer implements that interface.
  * Otherwise, it will be used as a {@link TransactionPartialSigner}.
  *
- * @typeParam TTransactionMessage - The inferred type of the transaction message provided.
- *
  * @example
  * ```ts
  * const signedTransaction = await partiallySignTransactionMessageWithSigners(transactionMessage);
@@ -64,12 +62,10 @@ import { assertIsTransactionMessageWithSingleSendingSigner } from './transaction
  * @see {@link signTransactionMessageWithSigners}
  * @see {@link signAndSendTransactionMessageWithSigners}
  */
-export async function partiallySignTransactionMessageWithSigners<
-    TTransactionMessage extends BaseTransactionMessage & TransactionMessageWithFeePayer & TransactionMessageWithSigners,
->(
-    transactionMessage: TTransactionMessage,
+export async function partiallySignTransactionMessageWithSigners(
+    transactionMessage: BaseTransactionMessage & TransactionMessageWithFeePayer & TransactionMessageWithSigners,
     config?: TransactionPartialSignerConfig,
-): Promise<TransactionFromTransactionMessage<TTransactionMessage>> {
+): Promise<Transaction & TransactionWithinSizeLimit & TransactionWithLifetime> {
     const { partialSigners, modifyingSigners } = categorizeTransactionSigners(
         deduplicateSigners(getSignersFromTransactionMessage(transactionMessage).filter(isTransactionSigner)),
         { identifySendingSigner: false },
@@ -91,8 +87,6 @@ export async function partiallySignTransactionMessageWithSigners<
  * This function delegates to the {@link partiallySignTransactionMessageWithSigners} function
  * in order to extract signers from the transaction message and sign the transaction.
  *
- * @typeParam TTransactionMessage - The inferred type of the transaction message provided.
- *
  * @example
  * ```ts
  * const mySignedTransaction = await signTransactionMessageWithSigners(myTransactionMessage);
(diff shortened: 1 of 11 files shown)


mcintyre94:
* **#927** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/927?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/927?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#919** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/919?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#918** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/918?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/signers/src/sign-transaction.ts
```diff
@@ -69,7 +69,7 @@ export async function partiallySignTransactionMessageWithSigners<
 >(
     transactionMessage: TTransactionMessage,
```
lorisleiva: Should we remove the type parameter now that the return type is no longer a function of the parameters? The same applies to the other function signatures of this file.

mcintyre94: Makes sense! I think we just want `transactionMessage: BaseTransactionMessage & TransactionMessageWithFeePayer & TransactionMessageWithSigners` here now.

mcintyre94:
### Merge activity

* **Oct 3, 10:36 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/927).
* **Oct 3, 10:44 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/927) rebased this pull request as part of a merge.
* **Oct 3, 10:45 AM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/anza-xyz/kit/927).

PR #944, opened 2025-10-06
Add a function to create a decoder that checks the size of the input bytes

mcintyre94:
#### Summary of Changes

Fixes #


mcintyre94:
* **#944** <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/944?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/anza-xyz/kit/944?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/content/api/index.mdx
steveluscher, line 1: Sorry for the merge conflict: https://github.com/anza-xyz/kit/pull/945.

packages/codecs-core/src/decoder-exact-byte-array.ts
```diff
@@ -0,0 +1,45 @@
...
+ * decoder.decode(new Uint8Array([0, 0, 0, 0, 0])); // throws
+ * decoder.decode(new Uint8Array([0, 0, 0])); // throws
+ *
+ * // with an offset
+ * decoder.decode(new Uint8Array([0, 0, 0, 0, 0]), 1); // 0
+ * decoder.decode(new Uint8Array([0, 0, 0, 0, 0, 0]), 1); // throws
+ * decoder.decode(new Uint8Array([0, 0, 0, 0]), 1); // throws
+ * ```
+ */
+export function createDecoderThatUsesExactByteArray<T>(decoder: Decoder<T>): Decoder<T> {
```
steveluscher: `ConsumesEntire`?

¯\\_(°ペ)\_/¯

packages/codecs-core/src/decoder-exact-byte-array.ts
```diff
@@ -0,0 +1,45 @@
...
+    return transformDecoder(decoder, (value, bytes, offset, newOffset) => {
+        if (newOffset !== bytes.length) {
+            throw new SolanaError(SOLANA_ERROR__CODECS__BYTES_LENGTH_DECODER_MISMATCH, {
+                bytesLength: bytes.length,
+                offsetAfterDecoding: newOffset,
+                offsetBeforeDecoding: offset,
```
steveluscher: This makes people do math. Just `expectedBytes`/`actualBytes` or just `numExcessBytes`?

packages/codecs-core/src/transform-codec.ts
```diff
@@ -95,25 +95,25 @@ export function transformEncoder<TOldFrom, TNewFrom>(
  */
 export function transformDecoder<TOldTo, TNewTo, TSize extends number>(
     decoder: FixedSizeDecoder<TOldTo, TSize>,
-    map: (value: TOldTo, bytes: ReadonlyUint8Array | Uint8Array, offset: number) => TNewTo,
+    map: (value: TOldTo, bytes: ReadonlyUint8Array | Uint8Array, offset: number, newOffset: number) => TNewTo,
```
steveluscher: Optionally, a new PR, but I can definitely pretend like I didn't see this.

packages/errors/src/messages.ts
```diff
@@ -330,6 +331,8 @@ export const SolanaErrorMessages: Readonly<{
         'Expected sentinel [$hexSentinel] to be present in decoded bytes [$hexDecodedBytes].',
     [SOLANA_ERROR__CODECS__UNION_VARIANT_OUT_OF_RANGE]:
         'Union variant out of range. Expected an index between $minRange and $maxRange, got $variant.',
+    [SOLANA_ERROR__CODECS__BYTES_LENGTH_DECODER_MISMATCH]:
+        'Byte array of length `$bytesLength` was not decoded to the last byte by the provided decoder. End offset was `$offsetAfterDecoding`.',
```
steveluscher: Maybe something with a suggestion of what to do, more like:

> This decoder expected a byte array of exactly `$expectedLength` bytes, but `$numExcessBytes` unexpected excess bytes remained after decoding. Are you sure that you have chosen the correct decoder for this data?

steveluscher (approved):
Added via Giphy

packages/codecs-core/src/transform-codec.ts
steveluscher, line 1: Probably deserves some extra tests in `transform-codec-test.ts`

packages/codecs-core/src/decoder-exact-byte-array.ts
```diff
@@ -0,0 +1,45 @@
...
+ * decoder.decode(new Uint8Array([0, 0, 0])); // throws
+ *
+ * // with an offset
+ * decoder.decode(new Uint8Array([0, 0, 0, 0, 0]), 1); // 0
+ * decoder.decode(new Uint8Array([0, 0, 0, 0, 0, 0]), 1); // throws
+ * decoder.decode(new Uint8Array([0, 0, 0, 0]), 1); // throws
+ * ```
+ */
+export function createDecoderThatUsesExactByteArray<T>(decoder: Decoder<T>): Decoder<T> {
+    return transformDecoder(decoder, (value, bytes, offset, newOffset) => {
```
lorisleiva: Since `transformDecoder` is just:

```ts
return createDecoder({
    ...decoder,
    read: (bytes, offset) => {
        // Some mapping.
    },
});
```

Could we not avoid the extra argument to `transformDecoder` and simply use `createDecoder` for this helper?

mcintyre94: That makes sense to me and should make this PR simpler, will refactor. Thanks for the pointer!

docs/content/api/index.mdx
mcintyre94: Thanks! I thought I'd forgotten to update something on my previous PR stack, but getting rid of this from the repo is even better!

packages/codecs-core/src/decoder-exact-byte-array.ts
mcintyre94: The reason I switched from `Entire` to `Exact` was because I started writing tests for things like "this decoder is going to read 4 bytes, and you give it 3 bytes, this gives you a useful error". So I was checking too long and too short, which `Exact` felt like a better fit for.

But I think this was the wrong approach because eg. our number codecs already error when the input bytes are too short,  and I think it probably makes sense to leave that as the responsibility for the codec and just cover the case where the input bytes are too long here - which we don't want codecs to check by default.

I'll change this to just check for a byte array being too long, which I think your other suggestions fit better with too.

packages/codecs-core/src/decoder-exact-byte-array.ts
mcintyre94: Yep agreed. Again this was because I wanted to cover the case where you provide too few bytes, and figured the offset before/after is the most helpful information there. But I can remove that case and provide a more helpful error that just considers the bytes being too long.

packages/codecs-core/src/transform-codec.ts
mcintyre94: This PR no longer changes `transformDecoder` which should be a lot cleaner

mcintyre94:
Thanks both for the review! I've made a bunch of changes:

- It now only checks if the offset is before the end of the byte array, instead of also throwing if there are too few bytes. The latter makes sense to keep as the responsibility of individual decoders, and restricting to only the first case makes the error message much more useful
- It no longer changes the interface of `transformDecoder`

steveluscher (approved):
Added via Giphy

PR #951, opened 2025-10-07
fix: skip compressing program addresses during transaction message compression

### LINKED ISSUE #917: Don't convert address to `AddressLookupMeta` if there is already an identical program address at the top level
## Overview

Consider a transaction that has the following instructions:

```ts
[
  { addresses: [], programAddress: 'ABC' },
  { addresses: ['ABC'], programAddress: 'DEF' },
]
```

There is no point in ‘compressing’ address `ABC` in the second instruction into a lookup table, because its presence as a program address of a top-level instruction implies that it _has_ to go into the static address array.

## Suggestion

Augment `compressTransactionMessageUsingAddressLookupTables()` to skip creating an `AddressLookupMeta` for any address that is already the program address of a top-level instruction.

tanmay5114:
#### Problem
Program addresses that appear as the program of a top-level instruction 
were being unnecessarily compressed into address lookup tables. 

#### Summary of Changes
- Added a check to skip compression for any account address that is also 
  a program address in the transaction’s top-level instructions.

Fixes #917

### DIFF
--- a/packages/transaction-messages/src/compress-transaction-message.ts
+++ b/packages/transaction-messages/src/compress-transaction-message.ts
@@ -94,6 +94,7 @@ export function compressTransactionMessageUsingAddressLookupTables<
     addressesByLookupTableAddress: AddressesByLookupTableAddress,
 ): TTransactionMessage | WidenTransactionMessageInstructions<TTransactionMessage> {
     const lookupTableAddresses = new Set(Object.values(addressesByLookupTableAddress).flatMap(a => a));
+    const programAddresses = new Set(transactionMessage.instructions.map(ix => ix.programAddress));
 
     const newInstructions: Instruction[] = [];
     let updatedAnyInstructions = false;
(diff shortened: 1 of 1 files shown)


packages/transaction-messages/src/compress-transaction-message.ts
```diff
@@ -106,10 +111,10 @@ export function compressTransactionMessageUsingAddressLookupTables<
         const newAccounts: Mutable<NonNullable<Instruction['accounts']>> = [];
         let updatedAnyAccounts = false;
         for (const account of instruction.accounts) {
-            // If the address is already a lookup, is not in any lookup tables, or is a signer role, return as-is
+            // If the address is already a lookup, is not in any lookup tables, or is a signer role, return as-is
```
steveluscher: I think you meant to update this comment, but didn't get around to it.

steveluscher (changes requested):
Thanks for this! All this PR needs is tests, and we can send it on its way to main.

```shell
cd packages/transaction-messages/
pnpm i
pnpm dev
```

packages/transaction-messages/src/compress-transaction-message.ts
tanmay5114: ya actually i changed it and add one more condition, but later I optimized the code and include the logic in the already present "if condition", so unchanged the comment later

tanmay5114:
> Thanks for this! All this PR needs is tests, and we can send it on its way to main.
> 
> ```shell
> cd packages/transaction-messages/
> pnpm i
> pnpm dev
> ```

Yes, I’ve added a test to cover this condition and confirmed that all existing tests are passing.

PR #974, opened 2025-10-23
errors: Lop off the string part of BorshIoError

joncinque:
#### Problem

The `BorshIoError` variant of `InstructionError` was changed from a string to a unit variant in the Rust SDK in [this PR](https://github.com/anza-xyz/solana-sdk/pull/12), but the change hasn't been reflected in Kit yet.

#### Summary of changes

Change the `BorshIoError` variant to lop off any string that comes back for compatibility with old and new errors emitted by the runtime.

Add the new variant to the RPC error type list.

### DIFF
--- a/packages/errors/src/__tests__/instruction-error-test.ts
+++ b/packages/errors/src/__tests__/instruction-error-test.ts
@@ -104,14 +104,6 @@ describe('getSolanaErrorFromInstructionError', () => {
             }),
         );
     });
-    it('produces the correct `SolanaError` for a `BorshIoError` error (SDK 3.0+ unit style)', () => {
-        const error = getSolanaErrorFromInstructionError(123, 'BorshIoError');
-        expect(error).toEqual(
-            new SolanaError(SOLANA_ERROR__INSTRUCTION_ERROR__BORSH_IO_ERROR, {
-                index: 123,
-            }),
-        );
-    });
     it("returns the unknown error when encountering an enum name that's missing from the map", () => {
         const error = getSolanaErrorFromInstructionError(123, 'ThisDoesNotExist');
         expect(error).toEqual(


packages/errors/src/__tests__/instruction-error-test.ts
```diff
@@ -95,11 +96,10 @@ describe('getSolanaErrorFromInstructionError', () => {
             }),
         );
     });
-    it('produces the correct `SolanaError` for a `BorshIoError` error', () => {
+    it('produces the correct `SolanaError` for a `BorshIoError` error from a string', () => {
         const error = getSolanaErrorFromInstructionError(123, { BorshIoError: 'abc' });
```
steveluscher, line 100: This change means that we need to make two tests, right: one that handles the error being `{ BorshIoError: 'abc' }` and a second test that handles it being `"BorshIoError"`?

joncinque: I added it on line 54 as part of the generic test -- shouldn't that do what you're suggesting?

joncinque:
~~And sorry, how do I change set?~~

Never mind, seems like that only happens with version bumps, so let me know if you want me to include one

steveluscher:
Aagh! I said I'd work on it, and now I can't push because you updated the branch, lol. :)

joncinque:
> Aagh! I said I'd work on it, and now I can't push because you updated the branch, lol. :)

Ah sorry, I took a guess and assumed you preferred atomic commits over adding commits to PRs :pray:

steveluscher:
OK, ready to review @joncinque! Did I get everything right?

packages/rpc-types/src/transaction-error.ts
```diff
@@ -54,8 +55,6 @@ type InstructionError =
     | 'UninitializedAccount'
     | 'UnsupportedProgramId'
     | 'UnsupportedSysvar'
-    | 'BorshIoError' // SDKv3 has a fieldless `BorshIoError`
-    | { BorshIoError: string } // SDK pre-v3 has a newtype for `BorshIoError`
```
joncinque: This is my lack of knowledge of how this is plumbed throughout the system -- if an old block contains `{ BorshIoError: "string" }`, will that cause an issue? Or will it be properly parsed in the errors package?

steveluscher: I guess it, in fact, depends. Command clicking ‘show references’ on `InstructionError` eventually leads me to `TransactionForAccountsMetaBase` which exposes `err: TransactionError` (which contains `InstructionError`, sometimes).

I guess the thing that _I_ don't know, is: after the change to the SDK, can RPC methods that return this type (eg. `getBlock`, `blockSubscribe`) return the `newtype` format anymore, or will they always return the unit type of `BorshIoError`?

steveluscher:
> And sorry, how do I change set?

You can run `pnpm changeset` on the CLI.

packages/rpc-types/src/transaction-error.ts
joncinque: v2 RPCs will always return the newtype, while v3 RPCs will always return the unit.

steveluscher: Yeah, this would require the thing that we don't have yet, which is the ability to build different Kit bundles based on the version of the RPC and/or enabled features (#28). Options are:

1. Scorched Earth, remove the old type, presume that all RPCs will be replace with v3 before anybody who would be affected upgrades to the next version of Kit
2. Keep both types, make a plan to eventually remove the old one when v3 is the network.

I figured in this particular case, option one would be fine, but you tell me – how long is it before v3 likely becomes the network?

joncinque: v3 currently has the majority of stake (including frankendancer), so maybe scorched earth is ok:

```
Stake By Version:
3.1.0   -    6 current validators ( 0.15%)
3.0.8   -    1 current validators ( 0.01%)
3.0.7   -   65 current validators ( 4.20%)
3.0.6   -  359 current validators (38.31%)   5 delinquent validators ( 0.00%)
3.0.2   -    1 current validators ( 0.25%)
3.0.1   -    1 current validators ( 0.01%)
2.3.14  -    2 current validators ( 0.03%)
2.3.13  -   25 current validators ( 3.29%)   1 delinquent validators ( 0.02%)
2.3.11  -   63 current validators (16.21%)   2 delinquent validators ( 0.00%)
2.3.10  -   21 current validators ( 2.17%)
2.3.9   -    6 current validators ( 0.10%)   1 delinquent validators ( 0.00%)
2.3.8   -   98 current validators ( 8.51%)   8 delinquent validators ( 0.03%)
2.3.7   -    1 current validators ( 0.04%)
2.3.6   -   59 current validators ( 2.33%)   1 delinquent validators ( 0.00%)
2.3.5   -    2 current validators ( 0.40%)
0.712.30006 -  163 current validators (16.67%)
0.711.30006 -    2 current validators ( 0.19%)
0.709.30000 -    4 current validators ( 0.46%)
0.708.20306 -   50 current validators ( 6.39%)
0.707.20306 -    2 current validators ( 0.23%)
0.1.1   -    0 current validators ( 0.00%)   1 delinquent validators ( 0.00%)
unknown -    0 current validators ( 0.00%)  21 delinquent validators ( 0.01%)
```

steveluscher: OK. Let's do it. I need a stamp from you since I'm the last pusher.

packages/errors/src/__tests__/instruction-error-test.ts
steveluscher: I dumb.

joncinque:
Thanks for the help!

PR #978, opened 2025-10-24
rpc-api parameters should use consistent readonly modifier

nonergodic:
Made parameters in rpc-api consistently `readonly` so I can pass readonly arrays as expected.

You guys should separately make up your mind whether responses should be readonly or not, because right now it's a wild mix where:
[some arrays are returned mutable](https://github.com/anza-xyz/kit/blob/92455b81277ec181f39dd0629fcc0cf2ae525d08/packages/rpc-api/src/getBlocks.ts#L3)
[while others are readonly](https://github.com/anza-xyz/kit/blob/92455b81277ec181f39dd0629fcc0cf2ae525d08/packages/rpc-api/src/getClusterNodes.ts#L40)
and while all response objects consistently use `Readonly<...>`, arrays within them
[sometimes do](https://github.com/anza-xyz/kit/blob/92455b81277ec181f39dd0629fcc0cf2ae525d08/packages/rpc-api/src/getBlock.ts#L33)
[and sometimes don't](https://github.com/anza-xyz/kit/blob/92455b81277ec181f39dd0629fcc0cf2ae525d08/packages/rpc-api/src/getBlockCommitment.ts#L8)

I think readonly params and mutable returns are likely the cleanest approach.

### DIFF
--- a/packages/rpc-api/src/getInflationReward.ts
+++ b/packages/rpc-api/src/getInflationReward.ts
@@ -46,5 +46,8 @@ export type GetInflationRewardApi = {
      *
      * @see https://solana.com/docs/rpc/http/getinflationreward
      */
-    getInflationReward(addresses: readonly Address[], config?: GetInflationRewardApiConfig): GetInflationRewardApiResponse;
+    getInflationReward(
+        addresses: readonly Address[],
+        config?: GetInflationRewardApiConfig,
+    ): GetInflationRewardApiResponse;
 };


steveluscher:
Love it. I spot checked the API and added a couple more.

> I think readonly params and mutable returns are likely the cleanest approach.

Can you say more about why you would prefer mutable returns?

steveluscher (approved):
Added via Giphy

nonergodic:
> Can you say more about why you would prefer mutable returns?

Something along the lines of Postel's law/robustness principle. E.g. if I get an array of addresses back, why shouldn't I be allowed to mutably concat another array of addresses to it. Essentially, if you're return is const, you are being prescriptive in what your users can do with it. This only makes sense if you're exposing internals for efficiency instead of handing back a fresh copy.

So the more explicit principle would be: Mutable returns because it gives the user more freedom, unless there's a reason not to (e.g. the user has to pay in performance).

nonergodic:
And separately, most TypeScript code is terrible and not correctly readonly aware, so if you return readonly arrays and people have to forward these returns to functions that forgot to make their input arrays readonly, you always force your users to cast away constness, which makes using your SDK more annoying. (I would not have gone to the trouble of opening a PR like this in other TS projects, because for the most part, unlike you guys, nobody cares about writing proper code.)

steveluscher:
Interesting. If you want to continue to pull on this thread, can you open an issue to continue the discussion? In general, I worry that allowing mutations in responses will destroy their integrity, because so many of these responses have cross-field dependencies (eg. just mutating the `AccountRole` of a single account can invalidate the list of static addresses and address lookups, so now your transaction object is a lie).

I also get the point about trying to forward a `readonly` array to a third-party API that hasn't typed its inputs as `readonly`, but that seems to please me and displease me in equal measure.

PR #984, opened 2025-10-24
Add functions that let you sign offchain message envelopes

steveluscher:
This PR introduces `partiallySignOffchainMessageEnvelope`​ and `signOffchainMessageEnvelope`​ methods that you can use to sign an offchain message using a `CryptoKey`​.

See the next PR in this stack for the implementation of the Signers API.

### DIFF
--- a/packages/offchain-messages/src/signatures.ts
+++ b/packages/offchain-messages/src/signatures.ts
@@ -1,7 +1,14 @@
-import { Address } from '@solana/addresses';
-import { SOLANA_ERROR__OFFCHAIN_MESSAGE__SIGNATURES_MISSING, SolanaError } from '@solana/errors';
+import { Address, getAddressFromPublicKey } from '@solana/addresses';
+import { ReadonlyUint8Array } from '@solana/codecs-core';
+import {
+    SOLANA_ERROR__OFFCHAIN_MESSAGE__ADDRESSES_CANNOT_SIGN_OFFCHAIN_MESSAGE,
+    SOLANA_ERROR__OFFCHAIN_MESSAGE__SIGNATURES_MISSING,
+    SolanaError,
+} from '@solana/errors';
+import { SignatureBytes, signBytes } from '@solana/keys';
 import { NominalType } from '@solana/nominal-types';
 
+import { decodeRequiredSignatoryAddresses } from './codecs/preamble-common';
 import { OffchainMessageEnvelope } from './envelope';
 
 /**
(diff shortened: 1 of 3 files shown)


steveluscher:
* **#991** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/991?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#984** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/984?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/984?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#880** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/880?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/offchain-messages/src/signatures.ts
```diff
@@ -9,6 +15,118 @@ import { OffchainMessageEnvelope } from './envelope';
...
+            const address = await getAddressFromPublicKey(keyPair.publicKey);
+            const existingSignature = offchainMessageEnvelope.signatures[address];
+
+            // Check if the address is expected to sign the message
+            if (existingSignature === undefined) {
+                // address is not an expected signer for this message
+                unexpectedSigners ||= new Set();
+                unexpectedSigners.add(address);
+                return;
+            }
```
steveluscher, line 77: This is a copy-paste of the transaction signing code. This code doesn't _actually_ check the message content to read the required signers; it trusts that the message envelope `signatures` map was compiled from the message preamble faithfully. Thoughts on this?

mcintyre94: I think this is fine for signing, I think it makes sense to trust the Envelope values here

packages/offchain-messages/src/signatures.ts
```diff
@@ -9,6 +15,118 @@ import { OffchainMessageEnvelope } from './envelope';
...
+
+            // Return if there are any unexpected signers already since we won't be using signatures
+            if (unexpectedSigners) {
+                return;
+            }
+
+            const newSignature = await signBytes(keyPair.privateKey, offchainMessageEnvelope.content);
```
mcintyre94, line 85: The area where I'm a bit more concerned about trusting the envelope is signing - if `offchainMessageEnvelope.content` was bytes for a `Transaction` rather than a message correctly compiled with the signing domain specifier. I think it's quite unlikely with the API you've got, but maybe still worth protecting against. 

For an additional layer of security, I wonder if it'd be worth adding a `signNonTransactionBytes` function to `@solana/keys`, that would throw if the first byte of the bytes to sign is not `0xff`. It'd probably be better to have `signBytes` do this check and have to explicitly call `signTransactionBytes` when you don't want it, but I suspect that's not worth a breaking change. 

It might also make sense to pair this with a `NonTransactionBytes` type, but I'm most concerned about having a runtime check where you sign bytes that you expect not to be a transaction.

steveluscher: I like this framing. This is probably a good argument for doing validation of the whole message every time.

steveluscher:
Added code that verifies the message starts with the signing domain before signing it.

packages/offchain-messages/src/signatures.ts
steveluscher: OK I didn't validate the entire message, just the signing domain, version, and required signers.

packages/offchain-messages/src/signatures.ts
```diff
@@ -75,3 +216,7 @@ export function assertIsFullySignedOffchainMessageEnvelope<TEnvelope extends Off
         });
     }
 }
+
+function uint8ArraysEqual(arr1: ReadonlyUint8Array, arr2: ReadonlyUint8Array) {
+    return arr1.length === arr2.length && arr1.every((value, index) => value === arr2[index]);
+}
```
lorisleiva, line 212: This would be a nice one to add next to `containsBytes` as `equalBytes` or something on `@solana/codecs-core`.

mcintyre94: Agree with this - we have this copied in a few places now.

steveluscher: #1027.

steveluscher:
### Merge activity

* **Nov 19, 10:37 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/984).
* **Nov 19, 10:43 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/984) rebased this pull request as part of a merge.
* **Nov 19, 10:45 PM UTC**: @steveluscher merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/984).

PR #991, opened 2025-10-27
Utilities and types that let you sign offchain message using the Signers API

steveluscher:
This PR allows you to supply a `MessageSigner`​ as one of an offchain message's signatories. When run through the `partiallySignOffchainMessageWithSigners`​ or `signOffchainMessageWithSigners`​ methods, the signing implementation of those `MessageSigners`​ will be used to sign the message for those addresses.


steveluscher:
* **#991** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/991?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/991?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#984** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/984?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#880** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/880?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/signers/src/offchain-message-signer.ts
```diff
@@ -0,0 +1,111 @@
...
+ *     signer,
+ * };
+ * ```
+ */
+export interface OffchainMessageSignatoryWithSigner<
+    TAddress extends string = string,
+    TSigner extends MessageSigner<TAddress> = MessageSigner<TAddress>,
+> extends OffchainMessageSignatory<TAddress> {
+    readonly signer: TSigner;
+}
```
lorisleiva: Sorry if this is still WIP and going to change but shouldn't this be something like:

```ts
type OffchainMessageSignatoryWithSigner<TAddress> = OffchainMessageSignatory<TAddress> & MessageSigner<TAddress>;
```

So that the signatory _is_ a `Signer` instead of containing one.

steveluscher: Very yes. Thank you for pointing this out.

steveluscher:
I think this is ready to review now.

packages/signers/src/offchain-message-signer.ts
```diff
@@ -0,0 +1,82 @@
...
+import { MessageSigner } from './message-signer';
+
+/**
+ * A union type that supports base offchain message signatories metas as well as
+ * {@link OffchainMessageSignatoryWithSigner | ones that support the Signers API}.
+ */
+export type OffchainMessageSignatoryWithSigner<TAddress extends string = string> = MessageSigner<TAddress> &
+    Readonly<{
+        address: Address<TAddress>;
+    }>;
```
lorisleiva: Isn't this redundant since `MessageSigner<TAddress>` already contains `{ address: Address<TAddress> }`?

packages/signers/src/offchain-message-signer.ts
```diff
@@ -0,0 +1,82 @@
...
+ */
+export function getSignersFromOffchainMessage<
+    TAddress extends string = string,
+    TSigner extends MessageSigner<TAddress> = MessageSigner<TAddress>,
+    TOffchainMessageWithSignatories extends Pick<
+        OffchainMessageWithSigners<TAddress, TSigner>,
+        'requiredSignatories'
+    > = Pick<OffchainMessageWithSigners<TAddress, TSigner>, 'requiredSignatories'>,
+>({ requiredSignatories }: TOffchainMessageWithSignatories): readonly TSigner[] {
+    const signers = requiredSignatories.filter(s => 'signMessages' in s || 'modifyAndSignMessages' in s) as TSigner[];
```
lorisleiva: There is a `isMessageSigner` helper that could be used here instead.

packages/signers/src/sign-offchain-message.ts
```diff
@@ -0,0 +1,165 @@
...
+ * ```
+ *
+ * @see {@link signOffchainMessageWithSigners}
+ */
+export async function partiallySignOffchainMessageWithSigners(
+    offchainMessage: OffchainMessage,
+    config?: MessagePartialSignerConfig,
+): Promise<OffchainMessageEnvelope> {
+    const { partialSigners, modifyingSigners } = categorizeMessageSigners(
+        deduplicateSigners(offchainMessage.requiredSignatories.filter(isMessageSigner)),
```
lorisleiva: Isn't this just `getSignersFromOffchainMessage`?

packages/signers/src/sign-offchain-message.ts
```diff
@@ -0,0 +1,165 @@
...
+    // Ensure there are any MessageModifyingSigner in the first place.
+    const modifyingSigners = signers.filter(isMessageModifyingSigner);
+    if (modifyingSigners.length === 0) return [];
+
+    // Prefer modifying signers that do not offer partial signing.
+    const nonPartialSigners = modifyingSigners.filter(signer => !isMessagePartialSigner(signer));
+    if (nonPartialSigners.length > 0) return nonPartialSigners;
+
+    // Otherwise, choose only one modifying signer (whichever).
+    return [modifyingSigners[0]];
```
lorisleiva, line 123: Perfect!

packages/signers/src/offchain-message-signer.ts
steveluscher: I've simplified these types in the latest revision. I don't feel like I understand the conceptual underpinning of these types. I've just been trying to copy-paste-modify the `account-signer-meta` files and repurpose them for `requiredSignatories` but at most junctures the types don't seem to do what I want them to, let alone something useful.

I think you wrote the original types for account meta signers. Does what I'm doing here make any sense at all?

packages/signers/src/sign-offchain-message.ts
steveluscher: I mean, is it? Is it even desirable to permit modifying signers here? Why would you let a signer modify an offchain message, especially one that is destined for multiple signers? Should we even allow modifying signers in the first place?

lorisleiva: I can't think of any use-cases for modifying these messages when signing but there is likely some that I don't see and I think it would be nice to support it for consistency with the rest of the signer API.

Do you have any concerns with this approach?

steveluscher: Only that it's easy to widen the type later, but harder to narrow it. @jordaaash, do you know of any reason why we would or would not allow modifying signers for offchain messages?

mcintyre94: I guess one example is if your v0 message can be displayed with Restricted ASCII but you gave it the message format ASCII, a software wallet that knows it's using a Ledger account might swap the message format before sending it to Ledger.

Also the SIWS spec ([or at least Phantom's docs for it](https://github.com/phantom/sign-in-with-solana?tab=readme-ov-file#specification)) have some fields the wallet is expected to determine if they're not provided, so if you were using this for SIWS then you'd expect the wallet to modify the nested message in those cases.

packages/signers/src/offchain-message-signer.ts
```diff
@@ -0,0 +1,72 @@
...
+ *     /* ... *\/
+ *     requiredSignatories: [firstSignatory, secondSignatory],
+ * }
+ * ```
+ */
+export interface OffchainMessageWithSigners<
+    TAddress extends string = string,
+    TSigner extends MessageSigner<TAddress> = MessageSigner<TAddress>,
+> {
+    requiredSignatories: readonly OffchainMessageSignatorySigner<TAddress, TSigner>[];
```
lorisleiva, line 39: If I understand correctly, here you want to accept both signers and "address wrappers" as signatories. If so, what you had before makes more sense:

```suggestion
    requiredSignatories: readonly (OffchainMessageSignatory<TAddress> | OffchainMessageSignatorySigner<TAddress, TSigner>)[];
```

packages/signers/src/offchain-message-signer.ts
lorisleiva: I've made another comment but this makes sense to me. Though, you could probably simplify this with:

```ts
export type OffchainMessageSignatorySigner<TAddress extends string = string> = MessageSigner<TAddress>;
```

steveluscher:
### Merge activity

* **Nov 19, 10:37 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/991).
* **Nov 19, 10:47 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/991) rebased this pull request as part of a merge.
* **Nov 19, 10:49 PM UTC**: @steveluscher merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/991).

PR #1001, opened 2025-10-30
Enable docs analytics

ZYJLiu:
Enable analytics on Kit docs to populate Inkeep dashboard

### DIFF
--- a/docs/src/lib/InKeepSearchDialog.tsx
+++ b/docs/src/lib/InKeepSearchDialog.tsx
@@ -40,6 +40,9 @@ export default function InKeepSearchDialog(props: SharedProps) {
                 const rootStyles = getComputedStyle(document.documentElement);
                 return rootStyles.getPropertyValue('--color-coral-400').trim();
             })(),
+            privacyPreferences: {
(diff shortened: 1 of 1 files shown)


docs/src/lib/InKeepSearchDialog.tsx
```diff
@@ -40,10 +40,6 @@ export default function InKeepSearchDialog(props: SharedProps) {
                 const rootStyles = getComputedStyle(document.documentElement);
                 return rootStyles.getPropertyValue('--color-coral-400').trim();
             })(),
-            privacyPreferences: {
-                optOutAnalyticalCookies: true,
```
steveluscher, line 44: Though this PR description doesn't state as much, you previously said that the goal is to see what kinds of questions folks have of the search, to make sure that we're producing the content that people are looking for.

Does this require cookies? I would imagine that cookies are only important for longitudinal analysis, which as far as I know is not something that we're interested in.

ZYJLiu: added back the `optOutAnalyticalCookies` settings for Inkeep config

PR #1003, opened 2025-11-03
createRecentSignatureConfirmationPromiseFactory now properly throws error

damianac:
#### Problem
The issue reported [here](https://github.com/anza-xyz/kit/issues/711) has been somewhat fixed [here](https://github.com/anza-xyz/kit/commit/eb0a122f4bfef16d8ee8df400e7bb3505a6903ee)

However, cause of `signatureStatus?.err` check being down the line in the code it may result with getting no error output even though there is one.

If you were to apply the tests added in this PR, they should fail with current `main` code. 

#### Summary of Changes

Moves error check before all other if/else which makes sure error is thrown in all cases.

### DIFF
--- a/.changeset/clean-steaks-try.md
+++ b/.changeset/clean-steaks-try.md
@@ -1,0 +1,5 @@
+---
+"@solana/transaction-confirmation": patch
+---
+
+Actually fixed a bug where transaction errors discovered during recent transaction confirmation might not be thrown. The fix in #793 was insufficient.


steveluscher (approved):
Added via Giphy

Oh wow, that's embarrassing. Thanks for catching that!

damianac:
no worries, it happens, hard to catch. thanks for quick reaction @steveluscher

PR #1024, opened 2025-11-16
Add a token-airdrop example

mcintyre94:
Adds a new example app using the instruction plans API

The example does the following:
- Create + Initialize a new token mint
- Airdrop 1000 of the new token to each of 100 addresses

The plan therefore does the following:
- Create/fund the mint account, and then initialize it
- In parallel, for each of the 100 addresses, create the ATA and then mint to it

The transaction plan fetches the latest blockhash and uses the same fee payer for each transaction created, sets the CU price to 100 microlamports/CU, and sets a provisory CU limit

The transaction executor estimates the actual CUs and updates the provisory CU limit, before sending + confirming each transaction

In practice this sends a total of 11 transactions, but the example doesn't need to worry about how these are packed.

### DIFF
--- a/examples/token-airdrop/src/example.ts
+++ b/examples/token-airdrop/src/example.ts
@@ -1,0 +1,306 @@
+/**
+ * EXAMPLE
+ * Create a new token mint and airdrop it to 100 recipients.
+ *
+ * Before running any of the examples in this monorepo, make sure to set up a test validator by
+ * running `pnpm test:live-with-test-validator:setup` in the root directory.
+ *
+ * To run this example, execute `pnpm start` in this directory.
+ */
+import { createLogger } from '@solana/example-utils/createLogger.js';
+import pressAnyKeyPrompt from '@solana/example-utils/pressAnyKeyPrompt.js';
+import {
+    appendTransactionMessageInstruction,
+    assertIsTransactionWithBlockhashLifetime,
+    createKeyPairSignerFromBytes,
+    createSolanaRpc,
+    createSolanaRpcSubscriptions,
+    createTransactionMessage,
+    createTransactionPlanExecutor,
+    createTransactionPlanner,
+    generateKeyPairSigner,
+    getSignatureFromTransaction,
+    parallelInstructionPlan,
+    pipe,
+    sendAndConfirmTransactionFactory,
+    sequentialInstructionPlan,
+    setTransactionMessageFeePayer,
+    setTransactionMessageLifetimeUsingBlockhash,
+    signTransactionMessageWithSigners,
+    TransactionPlan,
+} from '@solana/kit';
+import {
+    estimateAndUpdateProvisoryComputeUnitLimitFactory,
+    estimateComputeUnitLimitFactory,
+    fillProvisorySetComputeUnitLimitInstruction,
+    getSetComputeUnitPriceInstruction,
+} from '@solana-program/compute-budget';
+import { getCreateMintInstructionPlan, getMintToATAInstructionPlanAsync } from '@solana-program/token';
+
+const log = createLogger('Token Airdrop');
+
+/**
+ * SETUP: SOURCE ACCOUNT
+ * The account that will pay for the airdrop needs to sign the transaction. We need to
+ * create a `TransactionSigner` for it. You can find the account this key relates to in the test
+ * validator fixtures in `/scripts/fixtures/example-transfer-sol-source-account.json`
+ */
+const SOURCE_ACCOUNT_SIGNER = await createKeyPairSignerFromBytes(
+    /**
+     * These are the bytes that we saved at the time this account's key pair was originally
+     * generated. Here, they are inlined into the source code, but you can also imagine them being
+     * loaded from disk or, better yet, read from an environment variable.
+     */
+    new Uint8Array(
+        // prettier-ignore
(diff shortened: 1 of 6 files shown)


mcintyre94:
* **#1025** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1025?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1024** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1024?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1024?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

examples/token-airdrop/src/example.ts
```diff
@@ -0,0 +1,206 @@
...
+    }),
+);
+log.info('[setup] Generated 100 destination account addresses');
+
+// SETUP: get a blockhash, used for all transactions
+const { value: latestBlockhash } = await rpc.getLatestBlockhash({ commitment: 'confirmed' }).send();
+
+// SETUP: transaction planner
+const transactionPlanner = createTransactionPlanner({
+    createTransactionMessage() {
```
steveluscher, line 103: Maybe really drive this home with a comment that's like ‘every time the planner needs to create a new transaction message, this function determines the form the base message should take, before any instructions are added by the planner etc etc etc etc’

examples/token-airdrop/src/example.ts
```diff
@@ -0,0 +1,206 @@
...
+     * The sender will resolve when the transaction is reported to have been confirmed, or will
+     * reject in the event of an error, or a timeout if the transaction lifetime is thought to have
+     * expired.
+     */
+    rpcSubscriptions,
+});
+
+// SETUP: CU estimator
+const estimateCULimit = estimateComputeUnitLimitFactory({ rpc });
+// TODO: should we add a number => number function param to estimateAndUpdateProvisoryComputeUnitLimitFactory?
```
steveluscher: I don't think so. What you've done here is perfectly… functional.

examples/token-airdrop/src/example.ts
```diff
@@ -0,0 +1,206 @@
...
+    const estimate = await estimateCULimit(...args);
+    return Math.ceil(estimate * 1.1);
+}
+const estimateAndSetCULimit = estimateAndUpdateProvisoryComputeUnitLimitFactory(estimateWithMultiplier);
+
+// SETUP: transaction executor
+const transactionExecutor = createTransactionPlanExecutor({
+    async executeTransactionMessage(message, config) {
+        const abortSignal = config ? config.abortSignal : undefined;
+        const messageWithCUEstimate = await estimateAndSetCULimit(message, { abortSignal });
```
steveluscher: Maybe note that, in general, the executor is not the place to start patching extra crap into transaction messages (lest they grow to exceed the max size) but that this is OK because you're just replacing an existing value with another.

examples/token-airdrop/src/example.ts
```diff
@@ -0,0 +1,206 @@
...
+const destinationAddresses = await Promise.all(
+    Array.from({ length: 100 }, async () => {
+        const signer = await generateKeyPairSigner();
+        return signer.address;
+    }),
+);
+log.info('[setup] Generated 100 destination account addresses');
+
+// SETUP: get a blockhash, used for all transactions
+const { value: latestBlockhash } = await rpc.getLatestBlockhash({ commitment: 'confirmed' }).send();
```
steveluscher: Maybe, for clarity, fetch the blockhash _inside_ `createTransactionMessage()`? By clarity, I mean to demonstrate that `createTransactionMessage()` can be called by the same planner over a long period of time, and it's actually probably appropriate to get the freshest blockhash _at that time_.

I know the performance of this sucks, and delays the message getting off, but it has better correctness (eg. if the same planner is reused after a long delay).

examples/token-airdrop/src/example.ts
mcintyre94: Yep agreed. This example doesn't have a multiplier yet, but I have used that as I've poked closer to prototyping a framework. This will end up something like:

```ts
const estimateCULimit = estimateComputeUnitLimitFactory({ rpc });
async function estimateWithMultiplier(...args: Parameters<typeof estimateCULimit>): Promise<number> {
    const estimate = await estimateCULimit(...args);
    return Math.ceil(estimate * 1.1);
}
const estimateAndSetCULimit = estimateAndUpdateProvisoryComputeUnitLimitFactory(estimateWithMultiplier);
```

examples/token-airdrop/src/example.ts
lorisleiva: I agree with this but would even go further and fetch that blockhash right before executing the transaction instead of when it is being planned. The reason being, if you plan lots of sequential transactions you'll likely end up with an invalid blockhash by the time the last one is ready to be executed.

mcintyre94: Thanks both, makes sense- updated to fetch the blockhash in the executor

examples/token-airdrop/src/example.ts
```diff
@@ -0,0 +1,306 @@
...
+        );
+
+        // Sign this updated transaction message with any signers included in its instructions
+        const signedTransaction = await signTransactionMessageWithSigners(updatedMessage, { abortSignal });
+
+        const signature = getSignatureFromTransaction(signedTransaction);
+        log.info({ signature }, `[transaction executor] Sending transaction`);
+
+        // Send and confirm the transaction using the helper we created earlier
+        assertIsTransactionWithBlockhashLifetime(signedTransaction);
```
lorisleiva, line 207: This may no longer be needed since we're setting the lifetime in the same function.

lorisleiva (approved):
Nice, thank you!

examples/token-airdrop/src/example.ts
mcintyre94: Unfortunately it is because `signTransactionMessageWithSigners`​ only returns `TransactionWithLifetime`​ to account for modifying signers

I wonder if we could type-juggle it into returning the precise lifetime when there's only non-modifying signers though

Edit: Not yet - https://github.com/anza-xyz/kit/issues/1036

PR #1032, opened 2025-11-19
A method to verify an offchain message

steveluscher:
A `verifyOffchainMessage()` function that:

- Reads the required signatories from the message bytes, and
- Verifies their envelope signatures against the entire message

This method throws if a signature is missing or invalid.


steveluscher:
* **#1032** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1032?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1032?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/offchain-messages/src/signatures.ts
```diff
@@ -207,6 +207,37 @@ export function assertIsFullySignedOffchainMessageEnvelope<TEnvelope extends Off
     }
 }
 
+export async function verifyOffchainMessageEnvelope(offchainMessageEnvelope: OffchainMessageEnvelope): Promise<
+    | Readonly<{
+          signatoriesWithInvalidSignatures: readonly Address[];
+          signatoriesWithMissingSignatures: readonly Address[];
+      }>
+    | true
```
steveluscher: I'm not 100% sure about this design, can we discuss it?

Alternatives:

* Throw if verification fails
* Return `false` and don't comment at all on which signatures were invalid or which failed verification

lorisleiva: Yeah I'm not a bit fan of the `true | ReasonsForFailure` design.

How about:
- `assertIsValidOffchainMessageEnvelope(): Promise<void>`: That throws with an error containing the invalid and missing signatures in its context.
- `isValidOffchainMessageEnvelope(): Promise<boolean>`: That doesn't throw (if you don't care about the error context) and just gives you `true` or `false`. This can delegate to `assertIsValidOffchainMessageEnvelope` under the hood.

steveluscher: I hate that there are two kinds of ‘validity’ here. `assertIsFullySignedOffchainMessageEnvelope()`​ just ensures that there _is_ a signature for each required signatory, but doesn't verify those signatures. Someone could easily confuse that with _verification_ and allow a message that is not, in fact, signed by the required signatories. If I introduce `assertIsValidOffchainMessageEnvelope()`​ is someone likely to understand that means ‘signature validity’ or might they expect it to mean ‘message spec adherence?’  
  
The difference here, as compared to `Transaction`​ is that there's no harm in someone confusing a `FullySignedTransaction`​ with one whose signatures verify cryptographically because the chain has the final say. Where offchain messages are concerned, it's conceivable that someone will be using _Kit_ to make the final determination of whether a message is (authentically) signed by all parties before allowing some application to do {something} (eg. authorize the user, mutate some data, et cetera).  
  
Is the right answer to just have one method (`verifyOffchainMessageEnvelope()`​) but to have it return `Promise<void>`​ and throw when there's a signature mismatch or a missing signature?

lorisleiva: Oh I see your point. In this case, yeah I think a `Promise<void>` helper that throws makes the most sense.

PR #1035, opened 2025-11-21
Add a function to flatten transaction plan results

mcintyre94:
#### Problem

When I've been looking at transaction plans, the `TransactionPlanResult` object is a bit too unweildy to be suitable for eg logging or quickly interpreting. 

This is mostly because of its nesting and the fact it includes the transaction message and transaction objects.

#### Summary of Changes

This PR provides a function `flattenTransactionPlanResult` that flattens (depth first) a transaction plan result.

We can build a summary function on top of this.

### DIFF
--- a/.changeset/free-ants-begin.md
+++ b/.changeset/free-ants-begin.md
@@ -1,0 +1,5 @@
+---
+'@solana/instruction-plans': minor
+---
+
+Add a function to flatten a transaction plan result


mcintyre94:
* **#1056** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1056?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1044** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1044?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1043** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1043?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1035** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1035?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1035?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/instruction-plans/src/transaction-plan-result.ts
```diff
@@ -1,6 +1,8 @@
 import { SolanaError } from '@solana/errors';
 import { BaseTransactionMessage, TransactionMessageWithFeePayer } from '@solana/transaction-messages';
-import { Transaction } from '@solana/transactions';
+import { getSignatureFromTransaction, Transaction } from '@solana/transactions';
+
+import { Signature } from '../../keys/dist/types';
```
lorisleiva: nit:


```suggestion
import type { Signature } from '@solana/keys';
```

packages/instruction-plans/src/transaction-plan-result.ts
```diff
@@ -362,3 +364,51 @@ export function canceledSingleTransactionPlanResult<
...
+          status: 'canceled';
+      }>;
+
+/**
+ * Summarizes a {@link TransactionPlanResult} into a flat array of compact single transaction summaries.
+ * @param result The transaction plan result to summarize
+ * @returns An array of compact single transaction summaries
+ */
+export function summarizeTransactionPlanResult(result: TransactionPlanResult): CompactSingleTransactionSummary[] {
+    const transactionResults: CompactSingleTransactionSummary[] = [];
```
lorisleiva: Essentially, this function does two things:

1. It flattens the `TransactionPlanResult` tree into an array of leaves.
2. It maps those leaves slightly so that we return a signature instead of the whole transaction when successful.

I think it may be more beneficial to do the following:
- Offer a `flattenTransactionPlanResult(result: TransactionPlanResult): SingleTransactionPlanResult[]`.
- Don't map. The mapping seems a bit redundant to me because they can just use `getSignatureFromTransaction` on a successful result.
- Offer a `summarizeTransactionPlanResult(result: TransactionPlanResult): TransactionPlanSummary` such that the `TransactionPlanSummary` offer something significantly more useful than a flattening. I'm thinking of something like:

```ts
type TransactionPlanSummary = {
  successful: boolean,
  successfulTransactions: (SingleTransactionPlanResult & { status: 'successful' })[],
  failedTransactions: (SingleTransactionPlanResult & { status: 'failed' })[],
  canceledTransactions: (SingleTransactionPlanResult & { status: 'canceled' })[],
}
```

What do you think?

mcintyre94: Fair enough - the mapping is mostly for display reasons TBH, but we can push that out of the core lib.

Just to check, are you suggesting we would have both `flattenTransactionPlanResult` and `summarizeTransactionPlanResult`?

lorisleiva: Yes that's just a suggestion though. If you feel like lots of people would value having the signature directly, then we can include it as well. I just want to be mindful that we only have one chance with this `TransactionPlanResultSummary` naming so we should make the most of it. ☺️

mcintyre94: That's definitely fair. I'll rework this PR to just flatten, and brainstorm a bit more on the summary.

mcintyre94: Rewrote this PR to just add `flattenTransactionPlanResult`. Will stack a new summary PR on top of that.

packages/instruction-plans/src/transaction-plan-result.ts
steveluscher: We should probably either not care, or add a linter for this.  
  
https://github.com/anza-xyz/kit/issues/1045

steveluscher: (To be clear, about the `import type`​ bit, not about the absolute import)

mcintyre94:
### Merge activity

* **Nov 26, 3:53 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1035).
* **Nov 26, 3:53 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1035).

PR #1040, opened 2025-11-22
feat: add shared byte equality helper and tests

### LINKED ISSUE #1027: Move `uint8ArraysEqual` into `@solana/codecs-core`
Find all the copies of `function uint8ArraysEqual` scattered around the codebase and move them to `@solana/codecs-core`.

OrmEmbaar:
#### Problem

- Multiple packages duplicated the same `uint8ArraysEqual` helper, which made maintenance harder and limited discoverability.
- No shared utility existed in `@solana/codecs-core`, so other packages couldn’t import a canonical byte equality function.

#### Summary of Changes

- Added `areBytesEqual` to `packages/codecs-core/src/bytes.ts` and exported it for consumers.
- Replaced local `uint8ArraysEqual` helpers in react, transactions, and offchain-messages packages with imports of the shared helper.
- Expanded `packages/codecs-core/src/__tests__/bytes-test.ts` to cover equal arrays, empty arrays, and length mismatches.

Fixes #1027

### DIFF
--- a/.changeset/long-pots-build.md
+++ b/.changeset/long-pots-build.md
@@ -1,7 +1,6 @@
 ---
 '@solana/addresses': patch
 '@solana/offchain-messages': patch
-'@solana/transactions': patch
 '@solana/codecs-core': patch
 '@solana/react': patch
 '@solana/transactions': patch


carrithers:
Rebased onto upstream/main to resolve conflicts and force pushed.

packages/codecs-core/src/bytes.ts
```diff
@@ -127,3 +127,23 @@ export function containsBytes(
...
+ * @param bytes2 - The second byte array to compare.
+ *
+ * @example
+ * ```ts
+ * const bytes1 = new Uint8Array([0x01, 0x02, 0x03, 0x04]);
+ * const bytes2 = new Uint8Array([0x01, 0x02, 0x03, 0x04]);
+ * areBytesEqual(bytes1, bytes2); // true
+ * ```
+ */
+export function areBytesEqual(
```
mcintyre94: Nitpick: What do you think of calling it `bytesEqual`? I'm not sure why but `areBytesEqual` reads a bit awkwardly to me, especially in an if statement.

carrithers: Sure, i'm okay with that.

.changeset/long-pots-build.md
```diff
@@ -0,0 +1,8 @@
+---
+'@solana/offchain-messages': patch
+'@solana/transactions': patch
+'@solana/codecs-core': patch
+'@solana/react': patch
+---
+
+feat: add shared byte equality helper and tests
```
steveluscher: Make this more conversational, like ‘Add a function called X that you can use to do Y.’

steveluscher (changes requested):
Thanks for this!

Missed a few though. All of these can now be replaced with your helper.

- https://github.com/anza-xyz/kit/blob/ba03c041fce8fcd5b9ea242f43e6ea0f8f3755d6/packages/addresses/src/program-derived-address.ts#L210-L211
- https://github.com/anza-xyz/kit/blob/ba03c041fce8fcd5b9ea242f43e6ea0f8f3755d6/packages/codecs-core/src/bytes.ts#L127-L128
- https://github.com/anza-xyz/kit/blob/ba03c041fce8fcd5b9ea242f43e6ea0f8f3755d6/packages/react/src/useWalletAccountMessageSigner.ts#L77
- https://github.com/anza-xyz/kit/blob/ba03c041fce8fcd5b9ea242f43e6ea0f8f3755d6/packages/test-matchers/toEqualArrayBuffer.ts#L13

Leave the one in `webcrypto-ed25519-polyfill`​ the way it is.

packages/test-matchers/toEqualArrayBuffer.ts
```diff
@@ -1,3 +1,5 @@
+import { bytesEqual } from '@solana/codecs-core';
```
steveluscher: Dang it. On a fresh install (see the CI run) this fails.  
  
\> Cannot find module '@solana/codecs-core' from '../test-matchers/toEqualArrayBuffer.ts'  
  
I can't remember if our private packages have a restriction that makes it impossible to take on dependencies. Do you have time to troubleshoot this (or just revert it to the way it was)

lorisleiva: Since this is an internal helper for tests that also compares bytes, I'd say it's probably best to avoid the extra dependency here and revert the one-liner.

carrithers:
Should be good now?

Issue #366, opened 2025-04-10
Investigate whether we can remove the `downcastNodeToNumberIfBigint` _request_ transform


steveluscher:
I'm not clear on what Agave's RPC will do if you give it a JSON-RPC request with a number in it that exceeds `Number.MAX_SAFE_INTEGER`. Will it parse it as a u64 just fine? Maybe?

Investigate whether we can remove the _request_ transform `downcastNodeToNumberIfBigint()`.

PR #1041, opened 2025-11-24
solved issue#366

NeelContractor:
This PR removes the getBigIntDowncastRequestTransformer() step from the default Solana RPC request transformer chain.


mcintyre94:
The issue description (https://github.com/anza-xyz/kit/issues/366) is about investigating whether we need this step or not. Have you tested this to see how it works when the RPC sends a large number?

PR #1043, opened 2025-11-25
Make Transaction optional in successful transaction plan result + add signature

mcintyre94:
#### Problem

- The `transaction: Transaction` field being required means that you can't use for example a sending signer in a `TransactionExecutor`, as they only return a signature (mirroring wallet-standard sign+send)
- It is useful to have the signature of successful transactions in the `TransactionPlanResult`

#### Summary of Changes

This PR modifies the successful status to make `transaction` optional, and adds a new mandatory `Signature` field.

The function `successfulSingleTransactionPlanResult` now uses `getSignatureFromTransaction` to extract the signature from the transaction it receives, which it includes in the `SingleTransactionPlanResult`.

A new function `successfulSingleTransactionPlanResultFromSignature` can be used to create a successful `SingleTransactionPlanResult` with a signature but no transaction.

### DIFF
--- a/packages/instruction-plans/src/__tests__/__setup__.ts
+++ b/packages/instruction-plans/src/__tests__/__setup__.ts
@@ -1,8 +1,9 @@
 import { Address, getAddressDecoder } from '@solana/addresses';
-import { fixEncoderSize, getUtf8Encoder } from '@solana/codecs';
+import { fixEncoderSize, getBase58Encoder, getUtf8Encoder } from '@solana/codecs';
 import { SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN, SolanaError } from '@solana/errors';
 import { pipe } from '@solana/functional';
 import type { Instruction } from '@solana/instructions';
+import { SignatureBytes } from '@solana/keys';
 import {
     appendTransactionMessageInstruction,
     type BaseTransactionMessage,
@@ -10,7 +11,7 @@ import {
     setTransactionMessageFeePayer,
     type TransactionMessageWithFeePayer,
 } from '@solana/transaction-messages';
-import { getTransactionMessageSize, Transaction, TRANSACTION_SIZE_LIMIT } from '@solana/transactions';
+import { getTransactionMessageSize, SignaturesMap, Transaction, TRANSACTION_SIZE_LIMIT } from '@solana/transactions';
 
 import { MessagePackerInstructionPlan } from '../instruction-plan';
 
@@ -31,8 +32,12 @@ export function createMessage<TId extends string>(
 }
 
 export function createTransaction<TId extends string>(id: TId): Transaction & { id: TId } {
-    return Object.freeze({ id }) as unknown as Transaction & { id: TId };
+    // We set the id as the signature of the transaction, by setting its bytes in the signatures map
+    const signatures: SignaturesMap = { ['' as Address]: signatureEncoder.encode(id) as SignatureBytes };
+    return Object.freeze({ id, signatures }) as unknown as Transaction & { id: TId };
 }
+
+const signatureEncoder = getBase58Encoder();
 
 export function instructionFactory(baseSeed?: string) {
     const seedPrefix = baseSeed ? `${baseSeed}-` : '';
--- a/packages/instruction-plans/src/transaction-plan-executor.ts
+++ b/packages/instruction-plans/src/transaction-plan-executor.ts
@@ -3,6 +3,7 @@ import {
     SOLANA_ERROR__INVARIANT_VIOLATION__INVALID_TRANSACTION_PLAN_KIND,
     SolanaError,
 } from '@solana/errors';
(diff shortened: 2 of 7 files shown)


mcintyre94:
* **#1056** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1056?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1044** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1044?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1043** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1043?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1043?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1035** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1035?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/instruction-plans/src/__tests__/__setup__.ts
```diff
@@ -31,9 +32,13 @@ export function createMessage<TId extends string>(
 }
 
 export function createTransaction<TId extends string>(id: TId): Transaction & { id: TId } {
-    return Object.freeze({ id }) as unknown as Transaction & { id: TId };
+    // We set the id as the signature of the transaction, by setting its bytes in the signatures map
+    const signatures: SignaturesMap = { ['' as Address]: signatureEncoder.encode(id) as SignatureBytes };
+    return Object.freeze({ id, signatures }) as unknown as Transaction & { id: TId };
```
steveluscher, line 37: I don't see this new property used in this stack. Did I just miss it?

mcintyre94: It's a bit implicit. But many of the tests pass a transaction created using this to `successfulSingleTransactionPlanResult`​, which now extracts the signature from the transaction passed to it. So many tests will now fail if you pass them `{ id }`​ as a transaction.

Since this is strictly making `createTransaction`​ return something that looks more like a transaction, I just updated the existing test helper instead of making a second one that adds the signature.

packages/instruction-plans/src/transaction-plan-result.ts
```diff
@@ -168,7 +169,7 @@ export type SingleTransactionPlanResult<
  * @template TContext - The type of the context object that may be passed along with successful results
  */
 export type TransactionPlanResultStatus<TContext extends TransactionPlanResultContext = TransactionPlanResultContext> =
-    | Readonly<{ context: TContext; kind: 'successful'; transaction: Transaction }>
+    | Readonly<{ context: TContext; kind: 'successful'; signature: Signature; transaction?: Transaction }>
```
lorisleiva, line 172: Technically, this introduces a breaking changes and means we should bump a major version. However, since instruction plans is relatively recent and underused, I doubt anyone is fiddling with `TransactionPlanResultStatuses` yet. Additionally, we can take the stance that this is a bug fix because there are situations where a `transaction` object just isn't available to return — e.g. with `TransactionSendingSigners` as mentioned in the PR description.

TL;DR; is I'm happy with a minor bump but just wanted to flag the potential breaking change.

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -29,10 +31,14 @@ export type TransactionPlanExecutor<TContext extends TransactionPlanResultContex
     config?: { abortSignal?: AbortSignal },
 ) => Promise<TransactionPlanResult<TContext>>;
 
+type ExecuteResult<TContext extends TransactionPlanResultContext> = {
+    context?: TContext;
+} & ({ signature: Signature } | { transaction: Transaction });
```
lorisleiva, line 36: Nice! I love that the executor can keep returning `{ context, transaction }` but if they need something looser they can just return `{ context, signature }` instead. I also like that we don't allow them to offer both the `transaction` and the `signature` to avoid any mismatches.

packages/instruction-plans/src/transaction-plan-result.ts
mcintyre94: Yep agreed with this, this will be a small breaking change for anyone constructing this type themselves. The other reason I think it's ok is that if you're using the helper `successfulSingleTransactionPlanResult` then you won't be affected.

mcintyre94:
### Merge activity

* **Nov 26, 3:53 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1043).
* **Nov 26, 3:55 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1043) rebased this pull request as part of a merge.
* **Nov 26, 3:57 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1043).

PR #1059, opened 2025-11-26
Add to react-app example of signing with wallet and then server signer

### LINKED ISSUE #1034: Example: partial sign on client, then server
I've had a few people ask about how to do this and I think it'd be a good example to demonstrate signers 

The problem people are trying to solve is signing a transaction using a browser wallet, and then their server, perhaps with the latter as a fee payer.

The signers API should actually make this very easy, you just need to implement a `TransactionPartialSigner` that passes the transaction to your fee payer signing API 

Then the signers logic will sign with the wallet (a modifying signer) first, and then call the server second, passing it the partially signed transaction from the wallet. 

I think it'd be worth writing a very simple react example app that demonstrates this pattern. We don't actually need to implement a server to show this, just writing a `TransactionPartialSigner` that calls something else to sign for it should be enough to demonstrate how it works.

mcintyre94:
#### Problem

I've seen a few people asking how to do this, and signers provide a really neat approach. Since I think it makes most sense to demonstrate this using a browser wallet, I've added it to the react-app example

#### Summary of Changes

The new "Partial Sign Transaction" panel is very similar to the existing "Sign Transaction" one.

A `mockApiRequest` takes a serialized `Transaction` as input, and returns the `SignatureBytes` for this address (the keypair is stored in a JSON file). This mocks a server being used to sign transactions, for example as a fee payer.

A custom `TransactionPartialSigner` called `feePayerSigner` signs transactions by encoding them and passing them to `mockApiRequest`, mocking making an API request to a server that can sign a transaction.

The connected wallet continues to be a signer, for the SOL transfer instruction.

When the transaction is signed, it is first sent to the user's wallet to be signed, and then it is signed by the custom `feePayerSigner`. This ordering is because the user's wallet is a modifying signer, and is enforced by `mockApiRequest` using `signTransaction` which means that after it signs the transaction is fully signed.

Fixes #1034


mcintyre94:
* **#1059** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1059?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1059?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1068** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1068?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

examples/react-app/src/components/SolanaPartialSignTransactionFeaturePanel.tsx
```diff
@@ -0,0 +1,315 @@
...
+            assertIsSendableTransaction(transaction);
+            assertIsTransactionWithBlockhashLifetime(transaction);
+            setSignTransactionState({
+                kind: 'ready-to-send',
+                recipientAddress: recipientAccount.address as Address,
+                transaction,
+            });
+        } catch (e) {
+            setLastSignature(undefined);
+            setError(e);
```
steveluscher, line 166: This, and the one below are definitely type errors, because your `useState`​ up there implicitly types the state setter as `SetStateAction<Symbol>`​, which doesn't allow for the type of `e`​ here. Question is, why is this not failing CI.

steveluscher (approved):
I'd do a couple of things here to help make it more obvious what this does.

1. This is a two-stage sign then send thing. Maybe make the Send button green; I had trouble detecting that the button had changed from ‘Sign’ to ‘Send’ and that I needed to click it again to make something happen.
2. When I have insufficient lamports, I just get ‘Transaction simulation failed.’ It would be good to surface the `cause`​ in the dialog too.

examples/react-app/src/components/SolanaPartialSignTransactionFeaturePanel.tsx
```diff
@@ -0,0 +1,317 @@
...
+export function SolanaPartialSignTransactionFeaturePanel({ account }: Props) {
+    const { mutate } = useSWRConfig();
+    const { current: NO_ERROR } = useRef(Symbol());
+    const { rpc, rpcSubscriptions } = useContext(RpcContext);
+    const sendAndConfirmTransaction = useMemo(
+        () => sendAndConfirmTransactionFactory({ rpc, rpcSubscriptions }),
+        [rpc, rpcSubscriptions],
+    );
+    const wallets = useWallets();
+    const [error, setError] = useState<unknown>(NO_ERROR);
```
steveluscher, line 91: Really `unknown | typeof NO_ERROR` but meh.

PR #1061, opened 2025-12-01
fix(packages/react): allow single transaction to wallet modify and sign hook

datasalaryman:


datasalaryman:
@mcintyre94 could I ask for what you'd prefer here? I'm looking for a way to pass a single transaction to the modifyAndSign function. Would you rather the API:

1. allow for both a single transaction and array of transactions? (this PR)
2. have a separate function to pass just one transaction to sign?

mcintyre94:
Can you explain the use case here a bit? I'm not sure i see the benefit because the output is still an array. For the app to make its input an array should be trivial, I'd expect the complexity here comes from handling an output array.

I think if we did this then it would make more sense to mirror the signers API with single transaction versions, that take a single transaction as both input and output. Though that would definitely restrict the API a bit, if it makes sense for some signer to return multiple transactions from a single transaction input for example. I'm not sure it'd be better than what we have now.

It'd be good to understand the use case better first though. And would be curious if @lorisleiva has any thoughts on this too!

sidhys:
Chiming in: like [mcintyre94](https://github.com/mcintyre94) said, dealing with output is the actual issue - have you considered something along the lines of `modifyAndSignTransactions([transaction])`?

Issue #777, opened 2025-08-26
Feature idea: a function to calculate minimum rent required without an RPC call


mcintyre94:
## Motivation

If you'd like to create an account on chain to store some data, you need to fund it with a balance that makes it rent-exempt. To find out the required balance you can use the [getMinimumBalanceForRentExemption](https://solana.com/docs/rpc/http/getminimumbalanceforrentexemption) RPC call.

But the logic for this RPC call is stateless, and can be computed without making an RPC call. We could add a function to Kit to do this, which would save consumers making the RPC call.

## Example use case

```ts
// we want to create an account and store some data structure
const spaceRequired = getMyThingCodec().fixedSize; 
// no need to make an RPC call
const balanceRequired = getMinimumBalanceForRentExemption(spaceRequired);

// Create an account with the required balance
const message = pipe(
  createTransactionMessage({ version: 0}),
  m => appendTransactionMessageInstruction(
    getCreateAccountInstruction({
      space: spaceRequired,
      lamports: balanceRequired,
      // ...
   })
)

## Details

<!--
  Go into detail about how this new feature must behave. If you have
  ideas on how to implement it, go into them here.
-->

Gill has an implementation: https://github.com/DecalLabs/gill/blob/9d30e69d5faaf34e773def8cde17a444ad6fd7f1/packages/gill/src/core/accounts.ts#L4

The values and rent calculation come from: https://github.com/anza-xyz/solana-sdk/blob/c07f692e41d757057c8700211a9300cdcd6d33b1/rent/src/lib.rs#L93-L97

While this could in theory change, it is not just an RPC value since it is enforced when new accounts are created on-chain. So I think the risk of it changing often or quickly is very low, and I don't think this would be a significant maintenance burden for Kit.

PR #1065, opened 2025-12-02
feat: Add local rent exemption calculator

rajgoesout:
#### Problem

Logic for the [getMinimumBalanceForRentExemption function](https://solana.com/docs/rpc/http/getminimumbalanceforrentexemption) is stateless, so there's no need to make an rpc call.

#### Summary of Changes

  - New helper `packages/kit/src/get-minimum-balance-for-rent-exemption.ts`
  - Exported the helper from `packages/kit/src/index.ts`.

  Tests added

  - `packages/kit/src/__tests__/get-minimum-balance-for-rent-exemption-test.ts`

Closes https://github.com/anza-xyz/kit/issues/777

### DIFF
--- a/packages/kit/src/get-minimum-balance-for-rent-exemption.ts
+++ b/packages/kit/src/get-minimum-balance-for-rent-exemption.ts
@@ -7,6 +7,9 @@ import type { Lamports } from '@solana/rpc-types';
  * Values are sourced from the on-chain rent parameters in the Solana runtime:
  * https://github.com/anza-xyz/solana-sdk/blob/c07f692e41d757057c8700211a9300cdcd6d33b1/rent/src/lib.rs#L93-L97
  *
+ * Note that this logic may change, or be incorrect depending on the cluster you are connected to.
+ * You can always use the RPC method `getMinimumBalanceForRentExemption` to get the current value.
+ *
  * @param space The number of bytes of account data.
  */
 export function getMinimumBalanceForRentExemption(space: bigint): Lamports {


packages/kit/src/get-minimum-balance-for-rent-exemption.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Calculates the minimum {@link Lamports | lamports} required to make an account rent exempt for a
+ * given data size, without performing an RPC call.
+ *
+ * Values are sourced from the on-chain rent parameters in the Solana runtime:
+ * https://github.com/anza-xyz/solana-sdk/blob/c07f692e41d757057c8700211a9300cdcd6d33b1/rent/src/lib.rs#L93-L97
+ *
+ * @param space The number of bytes of account data.
+ */
+export function getMinimumBalanceForRentExemption(space: bigint | number = 0): Lamports {
```
mcintyre94: I'd prefer to just use `bigint` as input here, to match the rest of the library. We generally leave type conversion to callers to make the library simpler and avoid branching (in cases where that's relevant).

I'd also remove the default, because I think it'll make the calling code clearer.

mcintyre94:
Re the CI issue, if you run `pnpm style:fix` in the package it should fix the prettier issues for you!

Re the changeset, as this is new functionality we'd want to release a new version with this. If you do `pnpm changeset` in the root of the repo it'll guide you through it. We'd want to release a minor version of `@solana/kit`

rajgoesout:
@mcintyre94 Thank you for the review. I've updated the function signature, fixed prettier issues, and included the changeset.

packages/kit/src/get-minimum-balance-for-rent-exemption.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param space The number of bytes of account data.
+ */
+export function getMinimumBalanceForRentExemption(space: bigint): Lamports {
+    const RENT = {
+        ACCOUNT_STORAGE_OVERHEAD: 128n,
+        DEFAULT_EXEMPTION_THRESHOLD: 2n,
+        DEFAULT_LAMPORTS_PER_BYTE_YEAR: 3_480n,
+    } as const;
+    const dataLength = BigInt(space);
```
mcintyre94: We can remove this line now and just use `space`

packages/kit/src/__tests__/get-minimum-balance-for-rent-exemption-test.ts
```diff
@@ -0,0 +1,24 @@
...
+            1_586_880n as unknown as ReturnType<typeof getMinimumBalanceForRentExemption>,
+        );
+    });
+
+    it('accepts number input for data size', () => {
+        expect.assertions(1);
+        expect(getMinimumBalanceForRentExemption(1_024n)).toBe(
+            8_017_920n as unknown as ReturnType<typeof getMinimumBalanceForRentExemption>,
+        );
+    });
```
mcintyre94: The names of these tests aren't right any more, and I'm not sure we need to separately name them

A trick you can do in Jest is to define a set of input-output pairs, and have it generate a test for each one. You can see an example of how we use this in existing tests here: https://github.com/anza-xyz/kit/blob/e01507e8341f8b5b3e4d2f34928aaebd0e274cc8/packages/kit/src/__tests__/send-transaction-internal-test.ts#L99-L105

I think we could do the same here, and write these tests something like this:

```ts
it.each`
    space    | lamports
    ${0n}    | ${890_880n}
    ${100n}  | ${1_586_880n}
    ${1024n} | ${8_017_920n}
`(
    'calculates the correct rent for an account with $space bytes of space allocated',
    ({ space, lamports }) => {
        expect.assertions(1);
        expect(getMinimumBalanceForRentExemption(space)).toBe(lamports as Lamports);
    }
);
```

This would reduce the boilerplate for the tests, now that they're just asserting on some example input-outputs. And it'd be really easy to extend with other cases if we need to do that in future.

WDYT?

rajgoesout: makes sense!

mcintyre94 (approved):
Looks good now, thankyou!

mcintyre94:
It says merging is blocked with "Code scanning is waiting for results from CodeQL for the commits 05d060c or 8fa6ff0." 
I think this is a github bug, so I'm going to merge

PR #1066, opened 2025-12-02
Stop relying on `NPM_TOKEN` to publish; add id-token permission to publishing workflows

steveluscher:
#### Problem

The secret to getting this Trusted Publisher workflow to work is to add the `id-token` permission.

#### Summary of Changes

- Add `id-token: write` permission
- Merge the ‘Tag Release Manually’ workflow into the publish workflow file so that it can take advantage of the Trusted Publisher permissions.


steveluscher:
* **#1066** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1066?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1066?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/publish-packages.yml
```diff
@@ -1,10 +1,19 @@
-name: Publish Packages
+name: Publish Packages (or Tag Release Manually upon manual dispatch)
```
steveluscher, line 1: This is _really really_ bad, but in order to merge these workflows I had to create ‘tag release if you dispatch this manually’ but ‘actually publish if you don't.’ This means that you can no longer publish manually.

Someone less lazy than me can, in the future, create a manual dispatch menu with a list of actions you can manually dispatch, and use `if` conditions to select the right jobs based on the inputs.

PR #1084, opened 2025-12-03
Try to get ID token in CI

mcintyre94:
Environment variable + test job from https://docs.github.com/en/actions/how-tos/secure-your-work/security-harden-deployments/oidc-in-cloud-providers#adding-permissions-settings

Also see NPM PR that added oidc and shows how these variables are used: https://github.com/npm/cli/pull/8336/files 

Hopefully at least the test job will give some feedback!


mcintyre94:
Merging - will clean up test script/job if this works in case they're useful for debugging...

PR #1101, opened 2025-12-04
Experimental: change tx size to 4096 and add example using large tx in temporary environment

mcintyre94:
Not to merge to main, but I'd like to publish as an experimental version

This is a version of Kit that works with the temporary environment for SIMD-0296 (larger tx sizes). In this environment all transaction versions support a larger tx size, so we don't need v1 transactions to test this

See https://simd-0296.surfnet.dev/

- change the max tx size to 4096
- add an example that creates a memo with a 2000+ character string

Example tx: https://explorer.solana.com/tx/evtKYbpL8ingNBskfrQUvBNohtuLPa3eunDUxYTP2w637ohCexgM6yw4hyMx9EVE8GcfQMNdSVkNxTY3JTm2NtV?cluster=custom&customUrl=https://simd-0296.surfnet.dev:8899

### DIFF
--- a/packages/rpc-api/src/__tests__/send-transaction-test.ts
+++ b/packages/rpc-api/src/__tests__/send-transaction-test.ts
@@ -14,7 +14,7 @@ import {
 } from '@solana/errors';
 import { createPrivateKeyFromBytes } from '@solana/keys';
 import type { Rpc } from '@solana/rpc-spec';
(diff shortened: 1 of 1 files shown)


mcintyre94:
Published as 5.1.0-experimental-20251205104522

Issue #1104, opened 2025-12-05
Export plural versions of `useSignTransaction` and `useSignAndSendTransaction`


mcintyre94:
## Motivation

Wallet-standard allows requesting a wallet to sign multiple transactions. But we only export a hook capable of signing one, eg `useSignTransaction`

The actual implementation of this uses a hook that can accept multiple transactions, eg https://github.com/anza-xyz/kit/blob/52c804d9a9ab61689915b43fd9ae6f71d26a4e06/packages/react/src/useSignTransaction.ts#L81-L85

We should export the plural version so that apps can easily use our hooks to request a wallet to sign multiple transactions

## Example use case

```ts
const tx1: Uint8Array = makeTx1();
const tx2: Uint8Array = makeTx2();

const signTransactions = useSignTransactions(account, 'solana:devnet');
await signTransactions({transaction: tx1}, {transaction: tx2});
// wallet will sign both txs together
```

## Details

The existing signatures should be correct

We will need to add type overloads, docstrings and unit tests

steveluscher:
I did this on purpose because there's no known wallet that has a sensible multi-transaction signing UI _and_ there's no sensible way right now to handle fewer-than-all transactions failing.

cc/ @jordaaash

mcintyre94:
@steveluscher I disagree with not doing this yet. I agree that we're not happy with multi-transaction signing, but wallet-standard supports it, and wallets and apps use it as-is. We should do something better and wallets should support that when we do, but for now this is what we have and people use it. I don't think we should make it hard to use `@solana/react` for this use case.

PR #1105, opened 2025-12-06
Add plural sign transaction hooks

rajgoesout:
Export plural versions of useSignTransaction and useSignAndSendTransaction.

#### Problem

Wallet-standard allows requesting a wallet to sign multiple transactions. But we only export a hook capable of signing one, eg useSignTransaction.

The actual implementation of this uses a hook that can accept multiple transactions.

#### Summary of Changes

- Exposed useSignTransactions and useSignAndSendTransactions as public React hooks.
- Expanded unit/type tests to cover multi-transaction, empty-input, and chain-validation cases.
- Updated docs/examples.

Closes https://github.com/anza-xyz/kit/issues/1104

### DIFF
--- a/packages/react/src/__tests__/useSignAndSendTransaction-test.ts
+++ b/packages/react/src/__tests__/useSignAndSendTransaction-test.ts
@@ -167,21 +167,6 @@ describe('useSignAndSendTransaction', () => {
                 expect(results).toEqual([{ signature: 'abc' }, { signature: 'def' }]);
             }
         });
-        it('returns an empty array without calling the wallet when no inputs are provided', async () => {
-            expect.assertions(2);
-            const { result } = renderHook(() => useSignAndSendTransactions(mockUiWalletAccount, 'solana:danknet'));
-            // eslint-disable-next-line jest/no-conditional-in-test
-            if (result.__type === 'error' || !result.current) {
-                throw result.current;
-            } else {
-                const signAndSendTransactions = result.current;
-                const outputs = await signAndSendTransactions();
-                // eslint-disable-next-line jest/no-conditional-expect
-                expect(mockSignAndSendTransaction).not.toHaveBeenCalled();
-                // eslint-disable-next-line jest/no-conditional-expect
-                expect(outputs).toEqual([]);
-            }
-        });
         it('fatals when passed a wallet account that does not support the specified chain', () => {
             const { result } = renderHook(() =>
                 useSignAndSendTransactions({ ...mockUiWalletAccount, chains: ['solana:basednet'] }, 'solana:danknet'),
--- a/packages/react/src/__tests__/useSignTransaction-test.ts
+++ b/packages/react/src/__tests__/useSignTransaction-test.ts
@@ -167,21 +167,6 @@ describe('useSignTransaction', () => {
                 expect(results).toEqual([{ signature: 'abc' }, { signature: 'def' }]);
             }
         });
-        it('returns an empty array without calling the wallet when no inputs are provided', async () => {
-            expect.assertions(2);
-            const { result } = renderHook(() => useSignTransactions(mockUiWalletAccount, 'solana:danknet'));
-            // eslint-disable-next-line jest/no-conditional-in-test
-            if (result.__type === 'error' || !result.current) {
-                throw result.current;
-            } else {
-                const signTransactions = result.current;
-                const outputs = await signTransactions();
-                // eslint-disable-next-line jest/no-conditional-expect
-                expect(mockSignTransaction).not.toHaveBeenCalled();
-                // eslint-disable-next-line jest/no-conditional-expect
-                expect(outputs).toEqual([]);
-            }
-        });
         it('fatals when passed a wallet account that does not support the specified chain', () => {
             const { result } = renderHook(() =>
                 useSignTransactions({ ...mockUiWalletAccount, chains: ['solana:basednet'] }, 'solana:danknet'),


steveluscher:
Talked to @mcintyre94 about this. I didn't know that some apps were _already_ using the API in this way. I'll review this today.

steveluscher:
OK, maybe today.

steveluscher:
OK, maybe today, but no guarantees.

packages/react/README.md
```diff
@@ -54,6 +54,57 @@ function SignInButton({ wallet }) {
...
+                        { transaction: transactionBytes2 },
+                    );
+                    const [firstSignature, secondSignature] = [first.signature, second.signature].map(signature =>
+                        getBase58Decoder().decode(signature),
+                    );
+                    window.alert(
+                        `View transactions: https://explorer.solana.com/tx/${firstSignature}?cluster=devnet and https://explorer.solana.com/tx/${secondSignature}?cluster=devnet`,
+                    );
+                } catch (e) {
+                    console.error('Failed to send transactions', e);
```
steveluscher: Note for future readers: this is a lie. The underlying API does not let you know which transaction failed to send; some of them might, in fact, have landed.

packages/react/src/useSignTransaction.ts
```diff
@@ -78,15 +78,56 @@ export function useSignTransaction<TWalletAccount extends UiWalletAccount>(
...
+): (...inputs: readonly Input[]) => Promise<readonly Output[]>;
+export function useSignTransactions<TWalletAccount extends UiWalletAccount>(
     uiWalletAccount: TWalletAccount,
     chain: `solana:${string}`,
 ): (...inputs: readonly Input[]) => Promise<readonly Output[]> {
     if (!uiWalletAccount.chains.includes(chain)) {
         throw new WalletStandardError(WALLET_STANDARD_ERROR__FEATURES__WALLET_ACCOUNT_CHAIN_UNSUPPORTED, {
             address: uiWalletAccount.address,
             chain,
-            featureName: SolanaSignAndSendTransaction,
```
steveluscher, line 89: Oh dear. Nice find!

This needs to be a separate PR with an associated `pnpm changeset` note.

packages/react/src/useSignAndSendTransaction.ts
```diff
@@ -101,6 +142,9 @@ function useSignAndSendTransactions<TWalletAccount extends UiWalletAccount>(
     const account = getWalletAccountForUiWalletAccount_DO_NOT_USE_OR_YOU_WILL_BE_FIRED(uiWalletAccount);
     return useCallback(
         async (...inputs) => {
+            if (inputs.length === 0) {
+                return [];
+            }
```
steveluscher: These need to be in a separate PR with its associated ‘does not call the wallet's underlying ... implementation’ test.

steveluscher (changes requested):
Thanks for this work, and sorry it's taken me so long.

I'm _really_ struggling with this API. I understand the need for it now, but I think we should end up with _one_ API and deprecate the old one. I think this looks like:

* Mark the singular versions with `@deprecated Use {plural version}`
* Export the plural versions
* Rewrite all of the tests and examples to use the plural versions instead
* Release 5.{y+1}.0
* Delete the singular implementation
* Release 6.0.0

Would you be OK with this @mcintyre94?

mcintyre94:
I can see the argument for just exporting the plural version, since it covers both use cases and mirrors the wallet-standard feature. It also mirrors the signers APIs which are all multi-transaction.

That said, I think the single transaction use case is going to be used overwhelmingly more common, and it is easier if you don't need to use a plural API and unpack the result from an array. The cost of providing the single API is very small, since it's just a wrapper around the plural one in the current implementation.

Most importantly though, given the limitations with error handling for multiple transactions, there is a semantic difference (in practice) between sending one transaction and sending multiple. If you're dealing with one then you can understand its success/error status, if you send multiple then an error means some might have landed. Apps that do send multiple will need to treat these cases as meaningfully different. So I think it might be worth keeping the single transaction versions, with their more predictable behaviour, for the typical case of an app sending only one transaction. And then we have a clear place to document the issues with multiple transactions on the plural API.

packages/react/README.md
rajgoesout: Should I update the log to something like: `('Send returned an error; some transactions may still have landed. Verify each signature', e)`

rajgoesout:
@steveluscher @mcintyre94 Thank you for the review. I've removed the unrelated changes (will create separate PR for those). Let me know if we should keep the singular version or deprecate it.

packages/react/README.md
mcintyre94: I think we should probably just go with "Error returned by signAndSendTransactions", and leave it at that.

packages/react/src/useSignTransaction.ts
mcintyre94: Opened #1181 to track

packages/react/src/useSignAndSendTransaction.ts
mcintyre94: Opened #1182 to track

packages/react/README.md
```diff
@@ -54,6 +54,57 @@ function SignInButton({ wallet }) {
 }
 ```
 
+### `useSignAndSendTransactions(uiWalletAccount, chain)`
```
mcintyre94: Can we move this to the end of the README? Below where you added `useSignTransactions`

mcintyre94 (approved):
This is looking great, thankyou!

We will need a changeset for this, we can do a minor version with a message like: "Add `useSignTransactions` and `useSignAndSendTransactions` hooks that you can use to send multiple transactions to a connected wallet."

LMK if you want me to pick up these last few changes to get it over the line :) 

Discussed with @lorisleiva and we're going to keep shipping both the single and plural APIs, so no need to deprecate the single ones.

rajgoesout:
@mcintyre94 thanks for the review. I've made the changes

mcintyre94 (approved):

mcintyre94:
Github saying "Code scanning is waiting for results from CodeQL for the commits 51a853f or 3fd0e3f." - seems to be a github bug
Merging manually

PR #1152, opened 2025-12-23
Include program + type when available in fetchJsonParsedAccount

### LINKED ISSUE #1151: Include program + type when available in `fetchJsonParsedAccount`
When `jsonParsed` successfully decodes accounts for core programs, it includes a `program` and `type` field:

```json
{
    "value": {
      "data": {
        "parsed": {
          "info": {
            "decimals": 6,
            "freezeAuthority": "7dGbd2QZcCKcTndnHcTL8q7SMVXAkp688NTQYwrRCrar",
            "isInitialized": true,
            "mintAuthority": "BJE5MMbqXjVwjAF7oxwPYXnTXDyspzZyt4vwenNw5ruG",
            "supply": "11457520777246667"
          },
          "type": "mint" <--
        },
        "program": "spl-token", <--
        "space": 82
      },
      "executable": false,
      "lamports": 427263616843,
      "owner": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
      "rentEpoch": 18446744073709551615,
      "space": 82
    }
  }
}
```

But when we use `fetchJsonParsedAccount` we only include `parsed.info` as data

https://github.com/anza-xyz/kit/blob/84b2f81e76fa9a2675db6743e48941d4d2679ab9/packages/accounts/src/parse-account.ts#L102-L109

I think it'd be helpful to also include `program` and `type`, probably both as `string?` in the return type

Maybe we could return `MaybeAccount<TData & { program?: string, type?: string } , TAddress>`? 

Then callers would have access to these fields if they were returned by `jsonParsed`, which is useful for core programs.

rajgoesout:
#### Problem
When using [fetchJsonParsedAccount](https://github.com/anza-xyz/kit/blob/84b2f81e76fa9a2675db6743e48941d4d2679ab9/packages/accounts/src/fetch-account.ts#L89), we just get `parsed.info` as data; `program` and `type` are not included even if available.

#### Summary of Changes
Added `program`/`type` to jsonParsed account data (optional in types), updated fetch/parse return types, and adjusted tests (including a missing-info edge case) to reflect the new metadata.

Closes #1151

### DIFF
--- a/packages/accounts/src/parse-account.ts
+++ b/packages/accounts/src/parse-account.ts
@@ -106,26 +106,15 @@ export function parseJsonRpcAccount<TData extends object, TAddress extends strin
     rpcAccount: JsonParsedRpcAccount | null,
 ): Account<JsonParsedAccountData<TData>, TAddress> | MaybeAccount<JsonParsedAccountData<TData>, TAddress> {
     if (!rpcAccount) return Object.freeze({ address, exists: false });
-    const parsed = rpcAccount.data.parsed;
-    const info = parsed.info as TData | undefined;
-    const parsedAccountMeta =
-        rpcAccount.data.program || parsed.type
-            ? { program: rpcAccount.data.program, ...(parsed.type ? { type: parsed.type } : {}) }
-            : undefined;
-    let data: JsonParsedAccountData<TData>;
-    if (parsedAccountMeta) {
-        if (Array.isArray(info)) {
-            const infoArray = [...info] as unknown as JsonParsedAccountData<TData>;
-            (infoArray as unknown as { parsedAccountMeta?: ParsedAccountMeta }).parsedAccountMeta = parsedAccountMeta;
-            data = infoArray;
-        } else if (info && typeof info === 'object') {
-            data = { ...(info as object), parsedAccountMeta } as JsonParsedAccountData<TData>;
-        } else {
-            data = { parsedAccountMeta } as JsonParsedAccountData<TData>;
-        }
-    } else {
-        data = (info ?? {}) as JsonParsedAccountData<TData>;
+    const data = (rpcAccount.data.parsed.info || {}) as TData;
+
+    if (rpcAccount.data.program || rpcAccount.data.parsed.type) {
+        (data as JsonParsedAccountData<TData>).parsedAccountMeta = {
+            program: rpcAccount.data.program,
+            type: rpcAccount.data.parsed.type,
+        };
     }
+
     return Object.freeze({ ...parseBaseAccount(rpcAccount), address, data, exists: true });
 }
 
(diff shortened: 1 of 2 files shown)


packages/accounts/src/parse-account.ts
```diff
@@ -76,6 +76,7 @@ export function parseBase58RpcAccount<TAddress extends string = string>(
 }
 
 type JsonParsedRpcAccount = AccountInfoBase & { readonly data: JsonParsedDataResponse<unknown> };
+type JsonParsedAccountData<TData extends object> = TData & { program?: string; type?: string };
```
mcintyre94: I think we should make this type a bit more strict, because you only have a type if there is a program. If we move this into its own optional field, then we could also remove the empty case. I think we could have:

```ts
type ParsedAccountMeta = { program: string, type?: string }
type  JsonParsedAccountData<TData extends object> = TData & { parsedAccountMeta?: ParsedAccountMeta };
```

rajgoesout:
@mcintyre94 makes sense. BTW should we keep the `parsedAccountMeta` field outside the `data` field as well?

mcintyre94:
@rajgoesout If we were doing this from scratch then I might think about using a new `Account` type and taking it outside of data, but I don't think it's worth that big a change here. We're good to add `parsedAccountMeta` to data.

rajgoesout:
@mcintyre94 

CI was failing because adding `parsedAccountMeta` caused JSON‑parsed array data to be turned into objects like `{ "0": ..., "1": ... }`. That broke array‑shaped sysvars (recent blockhashes, slot hashes, stake history) against the toMatchObject assertions. (`sysvar-test.ts`)

Screenshot 2026-01-12 at 10 33 16 PM

lmk if this fix makes sense: https://github.com/anza-xyz/kit/pull/1152/commits/22b9748550125dd388a33eb0bd619ab8d0cac3db

packages/accounts/src/parse-account.ts
```diff
@@ -94,17 +96,36 @@ type JsonParsedRpcAccount = AccountInfoBase & { readonly data: JsonParsedDataRes
...
     if (!rpcAccount) return Object.freeze({ address, exists: false });
-    const data = rpcAccount.data.parsed.info as TData;
+    const parsed = rpcAccount.data.parsed;
+    const info = parsed.info as TData | undefined;
+    const parsedAccountMeta =
+        rpcAccount.data.program || parsed.type
+            ? { program: rpcAccount.data.program, ...(parsed.type ? { type: parsed.type } : {}) }
+            : undefined;
+    let data: JsonParsedAccountData<TData>;
+    if (parsedAccountMeta) {
```
mcintyre94: Hmm good spot, I hadn't thought about how sysvars use this. Could you add some more unit tests in `parse-account-test` to capture the different cases here?

This function is nice to test because we can control the `rpcAccount` input so we should be able to poke at each scenario here in isolation.

rajgoesout: added

PR #1154, opened 2025-12-27
Feat: Persistent selected wallet account context provider

### LINKED ISSUE #1148: Feat: A React context provider to persist the selected wallet
## Motivation

I think the biggest piece of complexity in the example react-app is about persisting the selected wallet. While wallet-standard allows an app to be connected to multiple wallets/accounts per wallet, in practice most apps will want to have some context about the currently active account, ie the `UiWalletAccount` to pass to hooks like `useSignAndSendTransaction`.

In the react-app example this is achieved with a custom context provider: https://github.com/anza-xyz/kit/blob/main/examples/react-app/src/context/SelectedWalletAccountContext.tsx and https://github.com/anza-xyz/kit/blob/main/examples/react-app/src/context/SelectedWalletAccountContextProvider.tsx 

It is consumed by components, for example here: https://github.com/anza-xyz/kit/blob/84b2f81e76fa9a2675db6743e48941d4d2679ab9/examples/react-app/src/components/ConnectWalletMenu.tsx#L22 

I think we should create something similar to this as part of the Kit library.

## Example use case

Just a draft API, but I imagine you would create the provider something like this:

```ts
type SelectedWalletAccountContextProviderProps = {
    // filter the supported wallets, as returned by `useWallets`
    filterWallets: UiWallet => Boolean,
    // state sync
    stateSync: {
        storeSelectedWallet: (string) => void,
        getStoredWallet: () => string,
        deleteStoredWallet: () => void
    }
}

// App includes the context provider at a high level
<SelectedWalletAccountContextProvider
    filterWallets={includesSolanaChain},
    stateSync = localStorageStateSync('my-app-localstorage-key')
>

// API exposed to components
const [selectedWallet, setSelectedWallet] = useContext(SelectedWalletAccountContext); 
if (selectedWallet) { // undefined if none selected
    const { wallet, account } = selectedWallet; // return UiWallet and UiWalletAccount
}
setSelectedWallet({wallet: UiWallet, account: UiWalletAccount}) 
```

## Details

This would mostly be a generalisation of the existing logic in the example react app, just exposed with a more flexible API

- Being able to filter the wallets is important, so that if an app restricts the allowed wallets to exclude one previously selected, then it is not re-selected. We could expose `filteredWallets` from this context to make this easy to keep in sync across an app.
- The current implementation syncs to localstorage, but I think it makes sense to generalise that. The important thing is that the context calls create/get/delete at the right times
- Unlike our hooks, this would branch on whether or not there's a selection. I think this makes sense because it is a context, and is also providing the set function. This is how the react-app example does it too 
- We could use the persistence format used in the example app, which is `walletName:address`, eg `phantom:BvcLNnNZKfUvhs18dMi1ywWNg6p1atUK7RPhD2HmifR6` 

I think this is worth having in `@solana/react`, because any app that needs to use any of the hooks that require a `UiWalletAccount` is going to need something like it. There's quite a bit of complexity/nuance in this code, so I don't think it makes sense for every app to re-implement it.

ningthoujamSwamikumar:
#### Problem
The biggest piece of complexity in the example react-app is about persisting the selected wallet. And while wallet-standard allows an app to be connected to multiple wallets/accounts per wallet, in practice most apps will want to have some context about the currently active account, ie the UiWalletAccount to pass to hooks like useSignAndSendTransaction.

#### Summary of Changes
- Created react context and context provider for selected wallet accounts.
- Expose the context as a hook with valid Value.
- Generalize the provider by taking functions to store, delete, and retrieve persistent account keys.
- User's selection/choice of wallet account precedes auto-restore of persisted wallet account, using a per provider instance flag.

Fixes #1148

### DIFF
--- a/.changeset/lovely-mugs-flash.md
+++ b/.changeset/lovely-mugs-flash.md
@@ -1,0 +1,5 @@
+---
+'@solana/react': minor
+---
+
+Add a context provider `<SelectedWalletAccountContext>` and `useSelectedWalletAccount` to persist a selected wallet account


ningthoujamSwamikumar:
Hi @mcintyre94 I have done the selected wallet context provider. And I am trying to unit test for that but I find the existing tests are being written using react-test-renderer which is deprecated. Should I write the test for this using react-test-renderer or react-testing-library? I think its better to start writing in react-testing-library as its gonna be migrated to this at some point.

mcintyre94:
Thanks @ningthoujamSwamikumar, I'll try to review this soon. In terms of testing I'm happy to use `react-testing-library` as this is what `react-test-renderer` seems to be recommending. Don't worry about updating the existing tests but I'll open an issue for that. Thanks for flagging! 

BTW a useful validation check for this would be using it in the react app :)

ningthoujamSwamikumar:
@mcintyre94 Hi. Can you please review this PR. I have ensured all test pass except for node unit test because of pattern matching issue in test-config/jest-unit.config.node.ts .

https://github.com/anza-xyz/kit/blob/b80b09239762262116cb70b43271ad98a2f716b5/packages/test-config/jest-unit.config.node.ts#L20
the ignore path pattern (above permalink) doesn't match with tsx file. Similar issue in test-config/jest-unit.config.browser.ts . 
We need to update the ignore path patterns to match '.tsx' files as well. It is matching only '.ts' files now. And I think since we now have two types of environment to run test on namely browser and node, we need to name the test files appropriately to match ignore path pattern in these test configs. 
i.e.
A test file name should match the pattern '-test.browser.tsx?$', or '-test.node.tsx?$' respectively for browser, and node tests. This way browser test will ignore node files, and node test will ignore browser test files. Hence we won't get test failures due to wrong environment which I am current having while running test for node. Because the filename ending with browser.tsx does not match the ignore path pattern of node test. But the file name should have extensin tsx as it has jsx syntax.

Let me know of any feedbacks and comments on my thoughts and the changes made in this PR.

packages/react/src/SelectedWalletAccountContextProvider.tsx
```diff
@@ -0,0 +1,144 @@
...
+} from "@wallet-standard/react";
+import { SelectedWalletAccountContext, SelectedWalletAccountState } from "./selectedWalletAccountContext";
+
+export type SelectedWalletAccountContextProviderProps = { children: React.ReactNode } & {
+    filterWallet: (wallet: UiWallet) => Boolean,
+    stateSync: {
+        storeSelectedWallet: (walletId: string) => void,
+        getSelectedWallet: () => string | null,
+        deleteSelectedWallet: () => void,
+    }
```
mcintyre94: We could do this in a follow up PR, but I think we should have a `localstorageStateSync` helper that just takes a localstorage key and creates this.

packages/react/src/__tests__/SelectedWalletAccountContextProvider-test.browser.tsx
```diff
@@ -0,0 +1,311 @@
...
+                Pick B
+            </button>
+            <button data-testid="clear" onClick={() => setSelectedWalletAccount(undefined)}>
+                Clear
+            </button>
+        </div>
+    );
+}
+
+describe("SelectedWalletAccountContextProvider", () => {
```
mcintyre94: These tests are generally great, and seem to cover the cases I'd be thinking about.

It'd be nice to add a test that uses the filter functionality, and makes sure we don't connect to a filtered wallet.

Nitpick: Solana addresses don't start `0x` and it's a bit confusing to see in our tests! I know they don't need to be valid addresses, but we usually just use `aaa` or `123` or something like that :)

packages/react/src/selectedWalletAccountContext.ts
```diff
@@ -0,0 +1,21 @@
...
+export type SelectedWalletAccountState = UiWalletAccount | undefined;
+
+export type SelectedWalletAccountContextValue = readonly [
+    selectedWalletAccount: SelectedWalletAccountState,
+    setSelectedWalletAccount: React.Dispatch<React.SetStateAction<SelectedWalletAccountState>>,
+];
+
+export const SelectedWalletAccountContext = /*#__PURE__*/ createContext<SelectedWalletAccountContextValue | undefined>(
+    undefined,
+);
```
mcintyre94: I don't think we want to have the `undefined` state here. The react-app example does it like this:

```ts
export const SelectedWalletAccountContext = createContext<
    readonly [
        selectedWalletAccount: SelectedWalletAccountState,
        setSelectedWalletAccount: React.Dispatch<React.SetStateAction<SelectedWalletAccountState>>,
    ]
>([
    undefined /* selectedWalletAccount */,
    function setSelectedWalletAccount() {
        /* empty */
    },
]);
```

Having the context be able to be undefined makes this much harder to consume, and I don't think it adds anything over an empty state like the above.

packages/react/src/selectedWalletAccountContext.ts
```diff
@@ -0,0 +1,21 @@
+import { UiWalletAccount } from '@wallet-standard/react';
+import React, { createContext } from 'react';
+
+export type SelectedWalletAccountState = UiWalletAccount | undefined;
```
mcintyre94, line 4: I think on the issue I suggested returning the `UiWallet` as well, but I'm not sure that makes sense thinking about it more. So I think I'm okay with this as the `SelectedWalletAccountState`

packages/react/src/selectedWalletAccountContext.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export type SelectedWalletAccountContextValue = readonly [
+    selectedWalletAccount: SelectedWalletAccountState,
+    setSelectedWalletAccount: React.Dispatch<React.SetStateAction<SelectedWalletAccountState>>,
```
mcintyre94, line 8: I think we should also return `filteredWallets` from this context. Not essential for this PR, but I think it'd be useful and avoid apps needing to filter `useWallets` multiple times and potentially lose sync. So that `filterWallets` in our provider becomes a source of truth.

You'd then also be able to use `uiWalletAccountBelongsToUiWallet` on the wallets in that list to find the connected wallet, which is what the react-app example does.

packages/react/src/SelectedWalletAccountContextProvider.tsx
ningthoujamSwamikumar: Sure, I will pickup this in a follow up PR.

packages/react/src/selectedWalletAccountContext.ts
ningthoujamSwamikumar: Added the filtered wallets in the context return.

packages/react/src/selectedWalletAccountContext.ts
ningthoujamSwamikumar: I was giving undefined as the default context value trying to solve the tree-shakeability  issue, turns out the /** PURE **/ annotation is enough. So, its now updated back to it was, similar to the example.

packages/react/src/__tests__/SelectedWalletAccountContextProvider-test.browser.tsx
ningthoujamSwamikumar: added the test for filtered wallets

ningthoujamSwamikumar:
> Thankyou, this is looking really good! Great tests too :)
> 
> On the Node test issue, I agree with you that we should change that pattern for Node. `'-test.browser.tsx?$'` works for me locally in the jest-unit.config.node.ts file and makes sense. Feel free to add that to this PR or I can. Don't worry about the browser equivalent, we're unlikely to use JSX in a test that can't run in the browser.

I have updated `test-config/jest-unit.config.node.ts` to reflect the same. And all test are passed, atleast in my local.

Thanks for all the feedback.

packages/react/src/SelectedWalletAccountContextProvider.tsx
mcintyre94: Issue created to track: #1184

packages/react/src/__typetests__/selectedWalletAccountContextProvider-typetest.ts
```diff
@@ -0,0 +1,71 @@
...
+// allows UiWalletAccount
+// @ts-expect-error TS6196
+type _AssertAllowsWallet = Assert<UiWalletAccount extends SelectedWallet ? true : false>;
+
+// Positive: selected wallet account can be a UiWalletAccount
+// @ts-expect-error TS6196
+const _selected2: UiWalletAccount = selected!;
+
+type FilteredWallets = CtxValue[2];
+declare const filteredWallets: FilteredWallets;
```
mcintyre94: I'm unclear what these typetests are doing. Is `TS6196` a typescript bug? We don't seem to use these types

mcintyre94: I think what you're trying to do can be done with satisfies:

```ts
type SelectedWallet = CtxValue[0];
undefined satisfies SelectedWallet;
null as unknown as UiWalletAccount satisfies SelectedWallet;
```

I think some tests are missing at the end for `filteredWallets`

I can update these, but LMK if I'm mis-understanding

ningthoujamSwamikumar: > I'm unclear what these typetests are doing. Is `TS6196` a typescript bug? We don't seem to use these types

This is just a unused variable error, I was getting while running typecheck test. Adding '_' before the variable was not working, so I had to add this as expected.

mcintyre94: No worries, I've updated these typetests to use a similar approach to our other ones :)

ningthoujamSwamikumar: > I think what you're trying to do can be done with satisfies:
> 
> ```ts
> type SelectedWallet = CtxValue[0];
> undefined satisfies SelectedWallet;
> null as unknown as UiWalletAccount satisfies SelectedWallet;
> ```
> 
> I think some tests are missing at the end for `filteredWallets`
> 
> I can update these, but LMK if I'm mis-understanding

Yes, I have missed to add the type assertions for filteredWallets. Maybe I was carried away. I will add it.
And about the type assertion, 'satisfies' might work but I am not sure.

mcintyre94: Don't worry, I've updated these :) LMK if you think anything is missing but I think they cover everything we need now.

mcintyre94 (approved):

PR #1155, opened 2025-12-27
Throw on non-divisible transaction plan

lorisleiva:
#### Problem

We currently offer a default implementation of `TransactionPlanExecutors` which requires a function to send a single transaction message. As such, this default implementation cannot handle `NonDivisibleSequentialTransactionPlans` since it would involve sending multiple transactions atomically — e.g. using Jito bundles.

The issue is the default implementation currently ignores the requirement of `NonDivisibleSequentialTransactionPlans` and sends them as if they were regular `SequentialTransactionPlans`. This is a mistake (made by me) as it invalid the atomicity of the `NonDivisibleSequentialTransactionPlans` and could lead to serious issues.

#### Summary of Changes

This PR fixes this issue by throwing a `SolanaError` when we encounter a `NonDivisibleSequentialTransactionPlan` in our default `TransactionPlanExecutor` implementation.

Importantly, it asserts that no `NonDivisibleSequentialTransactionPlan` exists in the potentially nested `TransactionPlan` provided _before_ trying to send any transaction. This avoid having a partially successful plan execution when we could have avoided it from the start.

#### Changeset reasoning

I went with a `patch` update here because, whilst this is a breaking change, it was not working as intended before. Therefore, I see this more as a bug fix than a breaking change but do let me know if you have another view on this.

#### Acknowledgements

Shout out to [Andy](https://linktr.ee/solandy) for flagging this in [one of his latest videos](https://www.youtube.com/watch?v=9S4ybrfiqUs).


lorisleiva:
* **#1155** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1155?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1155?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/errors/src/messages.ts
```diff
@@ -426,6 +427,8 @@ export const SolanaErrorMessages: Readonly<{
     [SOLANA_ERROR__INSTRUCTION_ERROR__UNSUPPORTED_SYSVAR]: 'Unsupported sysvar',
     [SOLANA_ERROR__INVARIANT_VIOLATION__INVALID_INSTRUCTION_PLAN_KIND]: 'Invalid instruction plan kind: $kind.',
     [SOLANA_ERROR__INSTRUCTION_PLANS__EMPTY_INSTRUCTION_PLAN]: 'The provided instruction plan is empty.',
+    [SOLANA_ERROR__INSTRUCTION_PLANS__NON_DIVISIBLE_TRANSACTION_PLANS_NOT_SUPPORTED]:
+        'This transaction plan executor does not support non-divisible sequential plans. To support them, you may create your own executor such that multi-transaction atomicity is preserved — e.g. by targetting RPCs that support transaction bundles.',
```
lorisleiva, line 431: This is an important part of the PR to review. Lmk what you think. 🙏

mcintyre94 (approved):
LGTM! Error message looks good, and agree with a patch update.

PR #1159, opened 2026-01-05
Add a default export for each package, matching browser

### LINKED ISSUE #1158: package.json exports missing default condition breaks Next.js SSR
## Overview

All `@solana/kit` packages are missing a `default` export condition, causing Next.js to resolve to Node.js builds during SSR of client components, which then fail because `ws` is unavailable.

## Steps to reproduce

https://github.com/GuiBibeau/solana-ssr-repro

## Description of bug

**Error:**
```
Module not found: Can't resolve 'ws'
./node_modules/@solana/rpc-subscriptions-channel-websocket/dist/index.node.mjs
```

**Why this happens:**

1. Client components with `"use client"` are [pre-rendered on the server](https://nextjs.org/docs/app/getting-started/server-and-client-components) before hydration
2. During SSR, Turbopack's `browser` condition [only matches client-side code](https://nextjs.org/docs/app/api-reference/config/next-config-js/turbopack), not SSR
3. Without a `default` fallback, resolution falls back to `main`/`module` which point to Node.js builds
4. Node.js builds import `ws`, which isn't installed (and is intentionally optional)

**Current exports:**
```json
"exports": {
  "browser": { "import": "./dist/index.browser.mjs" },
  "node": { "import": "./dist/index.node.mjs" }
}
"main": "./dist/index.node.cjs"
"module": "./dist/index.node.mjs"
```

**Fix:** Add `default` pointing to browser builds. I know webpack is a bit older, but I still like [their guidance](https://webpack.js.org/guides/package-exports/) on this: *"As the future is unknown, webpack assumes an environment similar to browsers for the default fallback."*

```json
"exports": {
  "browser": { "import": "./dist/index.browser.mjs" },
  "node": { "import": "./dist/index.node.mjs" },
  "default": { "import": "./dist/index.browser.mjs" }
}
"main": "./dist/index.browser.cjs"
"module": "./dist/index.browser.mjs"
```

**Environment:** Next.js 16.1.1, `@solana/kit` 5.1.0

mcintyre94:
#### Problem

Good detail in the issue: https://github.com/anza-xyz/kit/issues/1158

TLDR: Without this, NextJS runs the Node bundle for client components in SSR, and then complains that there's no `ws` 

#### Summary of Changes

Add a default export that matches the browser export for each public package

Fixes #1158

### DIFF
--- a/.changeset/easy-cloths-fail.md
+++ b/.changeset/easy-cloths-fail.md
@@ -1,0 +1,46 @@
+---
+'@solana/rpc-subscriptions-channel-websocket': patch
+'@solana/webcrypto-ed25519-polyfill': patch
+'@solana/transaction-confirmation': patch
+'@solana/codecs-data-structures': patch
+'@solana/rpc-subscriptions-spec': patch
+'@solana/fast-stable-stringify': patch
+'@solana/rpc-subscriptions-api': patch
+'@solana/transaction-messages': patch
+'@solana/rpc-transport-http': patch
+'@solana/instruction-plans': patch
+'@solana/offchain-messages': patch
+'@solana/rpc-subscriptions': patch
+'@solana/rpc-parsed-types': patch
+'@solana/rpc-transformers': patch
+'@solana/codecs-numbers': patch
+'@solana/codecs-strings': patch
+'@solana/rpc-spec-types': patch
+'@solana/instructions': patch
+'@solana/subscribable': patch
+'@solana/transactions': patch
+'@solana/codecs-core': patch
+'@solana/plugin-core': patch
+'@solana/rpc-graphql': patch
+'@solana/assertions': patch
+'@solana/functional': patch
+'@solana/addresses': patch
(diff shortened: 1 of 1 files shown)


steveluscher:
More concretely, if the browser build was served to the Vercel Edge Runtime, this line would fatal owing to the absence of `globalThis.navigator`:

https://github.com/anza-xyz/kit/blob/b80b09239762262116cb70b43271ad98a2f716b5/packages/rpc-subscriptions/src/rpc-subscriptions-autopinger.ts#L54-L56

Now I don't actually think you can _get_ to that line in an SSR pass, but my point is that browser builds can fatal the SSR runtime because they presume browser APIs like this one are present.

mcintyre94:
Yea I'm not convinced either. We need to document the `ws` dependency for Node, maybe that *should* include SSR?

mcintyre94:
Closing in favour of https://github.com/anza-xyz/kit/pull/1163 (merged)

PR #1163, opened 2026-01-07
Make `ws` a direct dependency

lorisleiva:
#### Problem

Currently, Node users must explicitly install `ws` as a peer dependency to use RPC Subscription features. This creates onboarding friction as users need to understand what is `ws`, which environment requires it and why it is required just to get started with Kit.

#### Summary of Changes

This PR move `ws` from `peerDependencies` to `dependencies`. Since the packages use conditional exports with separate entry points for Node and browser environments, browser builds will never include `ws` as the browser entry point simply doesn't import it.

This means Node users get `ws` out of the box while browser bundle sizes remain unaffected.

### DIFF
--- a/packages/rpc-subscriptions-channel-websocket/package.json
+++ b/packages/rpc-subscriptions-channel-websocket/package.json
@@ -76,7 +76,8 @@
         "@solana/errors": "workspace:*",
         "@solana/functional": "workspace:*",
         "@solana/rpc-subscriptions-spec": "workspace:*",
-        "@solana/subscribable": "workspace:*"
+        "@solana/subscribable": "workspace:*",
+        "ws": "^8.18.0"
     },
     "devDependencies": {
         "@solana/event-target-impl": "workspace:*",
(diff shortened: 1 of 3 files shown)


lorisleiva:
* **#1163** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1163?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1163?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/rpc-subscriptions-channel-websocket/package.json
```diff
@@ -76,22 +76,14 @@
...
+        "@solana/subscribable": "workspace:*",
+        "ws": "^8.18.0"
     },
     "devDependencies": {
         "@solana/event-target-impl": "workspace:*",
         "@solana/ws-impl": "workspace:*",
         "jest-websocket-mock": "^2.5.0"
     },
-    "peerDependencies": {
-        "typescript": ">=5.9.3",
```
mcintyre94, line 87: I think this change lost the `typescript` peer dependency

lorisleiva: Oops sorry not sure how I missed that. Fixing now.

PR #1168, opened 2026-01-07
Remove mergify from the repo

mcintyre94:
#### Problem

I think this is doing more harm than good

- autolabel PRs doesn't work
- it still doesn't use the merge queue even after https://github.com/anza-xyz/kit/commit/d53301d7a96c3aebfd7bb8a61b344ffa3787e350 
- don't think it's giving us any value really

#### Summary of Changes

- remove mergify config file
- remove automerge label from dependabot PRs for now


lorisleiva (approved):
Thanks!

PR #1174, opened 2026-01-07
Try bumping Jest to 30.2.0

mcintyre94:
We've been stuck on 30.0.0-alpha.6 for ages, and I keep seeing CI jobs fail with what seem to be flaky Jest runtime errors (database locked). I'm hoping moving to 30.2.0 and being able to adopt new Jest versions in future will make our CI more stable/less annoying. 

- Seems to be some weird issues with nested Typescript config file imports. These break between alpha.6 and alpha.7, and stay broken on 30.2.0. I don't think it's worth hacks to make this work, so I've just changed the config files to JS. This is most of the changes, since package.json files reference the config file.
- Fix a test in [error-test.ts](https://github.com/anza-xyz/kit/pull/1174/files#diff-ed54e75e339795d787c57bbafdd5e95bf8203bef25914974bde94a87f409bad4) where we return `undefined` but tested with `expect.not.objectContaining({ [propName]: errorOptionValue }),`. It seems Jest is stricter and `undefined` no longer matches this
- Fix a test in [transaction-plan-executor-test.ts](https://github.com/anza-xyz/kit/pull/1174/files#diff-be5be80ead9e0bbce5aadcf9b07d79928dda347955728ba70c400eb43df82693) where Jest is stricter about the non-enumerated `transactionPlan` field. 

Note: no changeset because this only changes tests/config

### DIFF
--- a/packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
+++ b/packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
@@ -31,9 +31,16 @@ async function expectFailedToExecute(
     error: SolanaError<typeof SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN>,
 ): Promise<void> {
     const transactionPlanResult = error.context.transactionPlanResult;
-    await expect(promise).rejects.toThrow(error);
-    // This second expectation is necessary since `toThrow` will only check the
-    // error message and `transactionPlanResult` is not part of the message.
+    // Check for the error code and message (but not the full context since transactionPlanResult is non-enumerable)
+    await expect(promise).rejects.toThrow(
+        expect.objectContaining({
+            context: expect.objectContaining({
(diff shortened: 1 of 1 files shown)


mcintyre94:
I think the socket security flag is a false positive/bug, because acorn [does not at all try to impersonate JFrog](https://www.npmjs.com/package/acorn) and isn't a VSCode extension. I'll do some more digging before making a decision here though. 

Edit: Confirmed by Socket team that this was a bug their side. A previous socket-security post flagged a major issue with `acorn`.

mcintyre94:
* **#1174** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1174?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1174?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1186** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1186?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1186, opened 2026-01-09
Fix type of error in sendTransaction preflight error

mcintyre94:
#### Problem

I noticed this while working on the Jest upgrade PR. This is an existing bug, but the new version of Jest exposed it in a test that behaved more strictly.

 If we call `sendTransaction` and get a preflight error, this is converted to a `SOLANA_ERROR__JSON_RPC__SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE`. This is done by parsing the `error.data` field, which includes a `RpcSimulateTransactionResult`

The problem is that errors do not use our normal number/bigint logic. The transport layer upcasts all number fields to bigint, but the bigint response transformer (which changes non-bigint fields back to number based on a whitelist) only runs on successful JSON rpc responses.

This means that all fields in `RpcSimulateTransactionResult` were actually returned as bigint, despite our attempt to type it.

I also noticed that some of the fields were typed as `number` incorrectly, when the server actually returns them as `u64` 

#### Summary of Changes

- Fix the types in `RpcSimulateTransactionResult` to mirror the server response
- Add a `getBigIntUpcastVisitor` with an appropriate whitelist, to downcast the appropriate bigint values to number in the case of sendTransaction preflight failure. This is applied to the `error.data` field.
- Add unit tests for `getThrowSolanaErrorResponseTransformer`

### DIFF
--- a/.changeset/quiet-lands-hunt.md
+++ b/.changeset/quiet-lands-hunt.md
@@ -1,0 +1,10 @@
+---
+'@solana/rpc-transformers': patch
+'@solana/errors': patch
+---
+
+Fix type of error in sendTransaction preflight error
+
+Some fields in `RpcSimulateTransactionResult` were incorrectly typed as number when they should have been bigint. At runtime these were bigint because of a bug.
+
+At runtime all numeric fields in `RpcSimulateTransactionResult` were a bigint, but those typed as number are now correct.


.changeset/quiet-lands-hunt.md
```diff
@@ -0,0 +1,6 @@
+---
+'@solana/rpc-transformers': patch
+'@solana/errors': patch
```
mcintyre94, line 3: Arguably this is major because it changes the type of `RpcSimulateTransactionResult`

But due to the bigint upcast, this type was incorrect (in practice all fields are bigint at runtime), so I think we can treat this as a bug.

mcintyre94:
* **#1174** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1174?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1186** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1186?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1186?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #1181, opened 2026-01-08
Incorrect feature name in `useSignTransaction` error case


mcintyre94:
https://github.com/anza-xyz/kit/blob/47c9901b0a4956dbf1a5a3f1969e04e26527e822/packages/react/src/useSignTransaction.ts#L89

Found as part of https://github.com/anza-xyz/kit/pull/1105

PR #1199, opened 2026-01-13
fix(react): correct featureName in `signTransaction` error

rajgoesout:
#### Problem

Incorrect feature name in `useSignTransaction` error case

#### Summary of Changes

correct `featureName` in `signTransaction` error and assert context in tests.

Fixes https://github.com/anza-xyz/kit/issues/1181

### DIFF
--- a/.changeset/lazy-cats-shave.md
+++ b/.changeset/lazy-cats-shave.md
@@ -1,0 +1,5 @@
+---
+'@solana/react': patch
+---
+
+Correct featureName in `signTransaction` error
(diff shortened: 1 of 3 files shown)


.changeset/lazy-cats-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solana/react': patch
+---
+
+Correct signTransaction featureName and assert context in tests
```
mcintyre94: I'd go with something like this:

```suggestion
Correct featureName in `useSignTransaction` error
```

PR #1203, opened 2026-01-14
Do not throw from transaction plan executor if a transaction fails to execute

### LINKED ISSUE #1201: Should we re-think having the executor from `createTransactionPlanExecutor` throw?
## Overview

Motivating example, testing kit-plugins:https://github.com/anza-xyz/kit-plugins/blob/d22551610ff9d4ad2b5deb303c2549e9f31ae466/examples/transfer-lamports/src/index.ts

And the relevant `TransactionPlanExecutor`: https://github.com/anza-xyz/kit-plugins/blob/d22551610ff9d4ad2b5deb303c2549e9f31ae466/packages/kit-plugin-instuction-plan/src/rpc.ts#L134

If we make `TransactionPlanExecutor` throw, then instead of getting a result with `failedTransactions` that we can parse, the function that calls the executor in fact throws. The error it throws includes a `TransactionPlanResult` that does have failed results, eg see the Kit tests: https://github.com/anza-xyz/kit/blob/9bde4d74f8d338495112ff00519177857b78884f/packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts#L267

I *think* the result of this is that `executor` will only actually return without throwing if all messages in the plan were successful. But then we're returning the complex nested `TransactionPlanResult` object, with only successful transactions. And even if we use the `summarizeTransactionPlanResult` helper, this still has a boolean indicating success which I think must always be true if we did not throw, and fields for cancelled/failed transactions which I think will always be empty.

The result of this is that trying to write a simple example for this API with error handling is difficult:

```ts
const transactionExecutor = createTransactionPlanExecutor({...});

try {
  const result = await transactionExecutor(myPlan)
  // result is `TransactionPlanResult` and according to the types some transactions might have failed
} catch(err) {
  // but this is where we actually end up if anything failed
  if (isSolanaError(e, SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN)) {
    const result = err.context.transactionPlanResult; // here the types will help us identify what failed
  }
}
```

I think it might make more sense to have the executor only throw in unexpected cases, eg the divisible transaction plan check. But if the error is from executing a specific transaction message, then we should instead return it without throwing in the `TransactionPlanResult`. Developers are already forced to handle failed/cancelled transactions here because of the types.

mcintyre94:
#### Problem

Currently if a transaction plan fails to execute, this is thrown as a `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN` error which includes the cause and the `TransactionPlanResult` that includes the failed/canceled transactions.

If the transaction plan successfully executes then the result is a `TransactionPlanResult`, which according to the types may include failed/canceled transactions, but in fact will only contain successful transaction results because otherwise the executor will have thrown.

This means that consumers that don't know this implementation detail think they need to handle failed/canceled transactions in multiple places - both in the response, and in a catch narrowed on `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN`.

#### Summary of Changes

This PR changes the behaviour so that if the executor fails while executing a transaction message, ie a specific transaction fails for some reason, then this is returned as the resulting `TransactionPlanResult`. So we can now return a result with failed/canceled transactions.

The exception is now only used for exceptional circumstances which an app is unlikely to attempt to recover from, for example a non-divisible instruction plan with the default executor.

I think most apps/documentation would not need to catch the exception with this change, and would simply process the result in one place.

The downside is that apps will now need to process the result and check if there are failed/canceled transactions, instead of just assuming that no error means all transactions executed successfully. But the `TransactionPlanResult` type made consumers (or at least their type checker) think they already needed to do this.

Fixes #1201

### DIFF
--- a/.changeset/tasty-coats-jog.md
+++ b/.changeset/tasty-coats-jog.md
@@ -1,0 +1,5 @@
+---
+'@solana/instruction-plans': patch
+---
+
+Do not throw from transaction plan executor if a transaction fails to execute


.changeset/tasty-coats-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solana/instruction-plans': patch
```
mcintyre94, line 2: Debatable - could argue this is a major, because if someone wrote code assuming that if this doesn't throw then all transactions succeeded then that will be broken by this change. Counter-argument is that the type points you away from writing this code despite that assumption previously holding.

mcintyre94:
* **#1204** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1204?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1203** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
```diff
@@ -160,14 +120,8 @@ describe('createTransactionPlanExecutor', () => {
...
 
-            await expectFailedToExecute(
-                promise,
-                new SolanaError(SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN, {
-                    cause,
-                    transactionPlanResult: canceledSingleTransactionPlanResult(messageA),
-                }),
-            );
-
+            // TODO: loses the abort cause, add that to `canceledSingleTransactionPlanResult`?
```
mcintyre94, line 123: Updated this in a following PR: #1204

mcintyre94:
Closing - see further discussion on our plan here: https://github.com/anza-xyz/kit/issues/1201#issuecomment-3783940098

PR #1213, opened 2026-01-19
Add a function to get an array of instructions from an InstructionPlan

mcintyre94:
#### Problem

Currently if you are writing an API that works with `TransactionMessage` then you are unable to use `InstructionPlan` objects. In some cases an `InstructionPlan` can be useful even without the transaction plan API, for example the helper we added to the token client for minting a new token. This is only 2 instructions and can fit in a single transaction. I think many useful helpers for clients will fit in this category, and it's worth making them as easy to use as possible.

For example there's a plugin for a fluent transaction builder: https://github.com/amilz/kit-helpers/tree/main/plugins/transaction-builder - I think this API should also be able to handle an instruction plan in cases where it fits within a single transaction.

#### Summary of Changes

Adds a new function `getInstructionsFromInstructionPlan` that gets the instructions from an instruction plan.

These can then be passed to eg `appendTransactionMessageInstructions`, or any other function that expects an array of instructions.

Note that we can't add a function to append an `InstructionPlan` to a `TransactionMessage` in the transaction-messages package because of a circular dependency. But I think `appendTransactionMessageInstructions(getInstructionsFromInstructionPlan(plan))` is a good enough low-level API for Kit. I think this is more likely to be used by libraries that want to handle instruction plans.


mcintyre94:
* **#1213** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1213?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
Discussed offline, we will instead add `flattenInstructionPlan` (and `flattenTransactionPlan` for completeness)

And implement `appendTransactionMessageInstructionPlan` in the `instruction-plans` package.

PR #1215, opened 2026-01-19
allow number in getMinimumBalanceForRentException not just bigint

loopcreativeandy:
reason for this change is developer convenience
especially when using other function to calculate size like getMintSize

realistically account size doest exceed 10MB which can be represented by a number and does not require bigints

the rpc transformer will downcast bigints to numbers anyway so numbers can already be used
this is just to satifsy the type checker

#### Summary of Changes
add number support to getMinimumBalanceForRentException in kit and rpc-api

### DIFF
--- a/packages/kit/src/get-minimum-balance-for-rent-exemption.ts
+++ b/packages/kit/src/get-minimum-balance-for-rent-exemption.ts
@@ -12,14 +12,14 @@ import type { Lamports } from '@solana/rpc-types';
  *
  * @param space The number of bytes of account data.
  */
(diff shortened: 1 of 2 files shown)


mcintyre94:
Thanks for this PR, but we try to avoid having multiple types and type conversions in the library like this, particularly in the rpc. I agree that `space` in practice doesn't need to be a bigint, but it's represented as a `usize` in the server code and we map that to bigint. We want to keep our rpc layer aligned with the server as much as we can.

PR #1229, opened 2026-01-22
Stop setting cause on `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN` errors

mcintyre94:
#### Problem

Currently error handling for transaction plans can be confusing, because there's the `error.cause` providing the first failure (or abort reason), and there's also the `transactionPlanResult` object which contains all the failed/cancelled transactions and provides detail on them all.

Consumers should use the plan, and explicitly choose to only handle the first error etc if they'd like to.

#### Summary of Changes

This PR stops setting `cause` when we throw this error. Consumers will need to check the plan to find out why the transaction failed.

Draft:
- We might not want to merge this until we have helpers to make processing the plan easier
- We might want to consider this a breaking change


mcintyre94:
Closing - superseded by #1260 + not worth the breaking change yet

PR #1230, opened 2026-01-22
Add a function to get the cause from a simulation related error

mcintyre94:
#### Problem

Some higher-level code can become confusing because it throws a simulation error (currently from either preflight, or from CU estimation), which is wrapping an error that describes why the simulation failed. In many cases we don't need the simulation error itself, and the nested error is more useful. 

#### Summary of Changes

Add a new function `unwrapSimulationError`. If the input is an error from simulation, and it has a truthy `cause`, the function returns this cause. Otherwise the input is returned unchanged. 

Relates to https://github.com/anza-xyz/kit-plugins/issues/25

### DIFF
--- a/.changeset/wicked-lizards-dig.md
+++ b/.changeset/wicked-lizards-dig.md
@@ -1,0 +1,5 @@
+---
+'@solana/errors': minor
+---
+
+Add a function `unwrapSimulationError`, which will return the cause of an error if it is a simulation error. Otherwise it is returned unchanged.


packages/errors/src/simulation-errors.ts
```diff
@@ -0,0 +1,49 @@
...
+export function unwrapSimulationError(error: unknown): unknown {
+    const simulationCodes: SolanaErrorCode[] = [
+        SOLANA_ERROR__JSON_RPC__SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE,
+        SOLANA_ERROR__TRANSACTION__FAILED_WHEN_SIMULATING_TO_ESTIMATE_COMPUTE_LIMIT,
+    ];
+    for (const code of simulationCodes) {
+        if (isSolanaError(error, code) && !!error.cause) {
+            return error.cause;
+        }
+    }
```
lorisleiva: super nit (feel free to ignore): just wondering if it's best to call `isSolanaError` once without a code and then check if the code is in the array to avoid re-running the statements that check if error is an instance of `SolanaError` for each code.


```suggestion
    if (isSolanaError(error) && !!error.cause && simulationCodes.includes(error.context.__code)) {
        return error.cause;
    }
```

PR #1233, opened 2026-01-23
Add `every` predicate helpers for instruction plan types

lorisleiva:
#### Problem

`InstructionPlan`, `TransactionPlan` and `TransactionPlanResult` are nested tree structures which makes asserting predicates hold for all nodes inside a plan non-trivial.

#### Summary of Changes

This PR adds `everyInstructionPlan`, `everyTransactionPlan` and `everyTransactionPlanResult` functions. These work similarly to `Array.prototype.every()` but operate on tree structures, performing depth-first traversal and returning true only if the given predicate returns true for every node in the tree (including the root and all nested nodes).

This complements the existing `find*` functions (introduced in #1232) by providing a way to assert conditions across entire plan trees rather than searching for a single matching node.

#### Example usage

```ts
// Check if all instructions are from a given program.
const allFromMyProgram = everyInstructionPlan(
  plan,
  (p) => p.kind !== 'single' || p.instruction.programAddress === myProgramAddress,
);

// Check if all sequential plans are divisible.
const allDivisible = everyTransactionPlan(
  plan,
  (p) => p.kind !== 'sequential' || p.divisible,
);

// Check if all transactions were successful.
const allSuccessful = everyTransactionPlanResult(
  result,
  (r) => r.kind !== 'single' || r.status.kind === 'successful',
);
```

### DIFF
--- a/.changeset/dark-showers-hunt.md
+++ b/.changeset/dark-showers-hunt.md
@@ -1,0 +1,5 @@
+---
+'@solana/instruction-plans': minor
+---
+
+Add `everyInstructionPlan`, `everyTransactionPlan` and `everyTransactionPlanResult` functions that can be used to ensure a given predicate holds for all nodes inside their respective plan structures.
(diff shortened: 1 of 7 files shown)


lorisleiva:
* **#1233** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1233?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1233?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1234, opened 2026-01-23
Fix race condition in durable nonce transaction confirmation

### LINKED ISSUE #1211: Bug: sendAndConfirmDurableNonceTransactionFactory error on successful transactions
## Overview

While testing some durable nonce transactions, I noticed that `sendAndConfirmDurableNonceTransactionFactory` intermittently throws `SOLANA_ERROR__INVALID_NONCE` on transactions that actually succeeded. This occurs when sending multiple durable nonce transactions in sequence.

## Steps to reproduce

1. Create a nonce account
2. Send multiple durable nonce transactions in a loop, fetching a fresh nonce before each transaction
3. Observe that some transactions throw `SOLANA_ERROR__INVALID_NONCE` despite succeeding on-chain

```typescript
import {
    createTransactionMessage,
    pipe,
    setTransactionMessageFeePayerSigner,
    setTransactionMessageLifetimeUsingDurableNonce,
    appendTransactionMessageInstructions,
    signTransactionMessageWithSigners,
    sendAndConfirmDurableNonceTransactionFactory,
} from '@solana/kit';
import { fetchNonce } from '@solana-program/system';
import { getAddMemoInstruction } from '@solana-program/memo';

// Setup rpc, rpcSubscriptions, payer, nonceAccountAddress

for (let i = 0; i < 10; i++) {
    const { data: { blockhash: nonce } } = await fetchNonce(rpc, nonceAccountAddress);

    const message = pipe(
        tx => setTransactionMessageFeePayerSigner(payer, tx),
        tx => setTransactionMessageLifetimeUsingDurableNonce(
            { nonce, nonceAccountAddress, nonceAuthorityAddress: payer.address },
            tx
        ),
        tx => appendTransactionMessageInstructions([getAddMemoInstruction({ memo: `test-${i}` })], tx),
    );

    const signed = await signTransactionMessageWithSigners(message);
    const sendAndConfirm = sendAndConfirmDurableNonceTransactionFactory({ rpc, rpcSubscriptions });

    // This intermittently throws SOLANA_ERROR__INVALID_NONCE even when the transaction succeeds
    await sendAndConfirm(signed, { commitment: 'confirmed' });
    // tested with various timeouts 
}
```

Tested on localnet with `solana-test-validator` using `@solana/kit` 5.4.0.

## Description of bug

The nonce confirmation strategy in `@solana/transaction-confirmation` (`confirmation-strategy-nonce.ts`) races two conditions:

1. Signature confirmation - waiting for the signature to reach the requested commitment level
2. Nonce advancement - watching for the nonce account value to change using account subscribe

When a durable nonce transaction succeeds, the nonce value advances which triggers a subscription update. If the nonce subscription detects this advancement **before** the signature subscription confirms, the strategy throws `SOLANA_ERROR__INVALID_NONCE` - even though the transaction actually succeeded.

https://github.com/anza-xyz/kit/blob/c1a9e73ea10587d83d0b387d08eebf945bd620c2/packages/transaction-confirmation/src/confirmation-strategy-nonce.ts#L112-L122

The error message is misleading because nonce advancement actually indicates the transaction was processed - it's evidence of success, not failure.

**Workaround/fix**: Catch `INVALID_NONCE` and verify signature status before propagating--(I think this is the intent here...if 2 txns attempt to use a nonce, fail fast one 1 when we know it's nonce is stale). Or you could just only run this check 1x at the begining (and not do the subscription)

mcintyre94:
#### Problem

When we attempt to confirm a durable nonce transaction, we race two promises (simplifying both a bit, and both at the desired commitment level):

- Subscribe to account notifications for the durable nonce, reject if the nonce advances
- Subscribe to signature notifications for the transaction, resolve/reject when it lands

When a durable nonce transaction is executed, the durable nonce account value is changed as part of the transaction. If another transaction advances the nonce first then we want to fail, as the transaction sent with the same nonce can no longer confirm.

The problem is that in the case where *our* transaction advances the nonce, the client may receive the account notification *before* the signature one. There is no guaranteed ordering at the RPC layer. In this case we would throw because we see the nonce has advanced, but in fact our signature has simply confirmed.

#### Summary of Changes

We now wrap the call to `getNonceInvalidationPromise` and catch an invalid nonce error. When this happens, we use `rpc.getSignatureStatuses` to check the status of the transaction that we are trying to confirm.

- If the request fails, we throw the invalid nonce error
- If the response is `null`, then our transaction is not responsible for the nonce advancing, so we throw the invalid nonce error
- If a status is returned but with a commitment earlier than the one we are waiting for, we do not throw the error and we suspend this promise indefinitely.  This will allow the signature notification to be received at the correct commitment level, which will determine the result.
- If a status is returned with a commitment greater than or equal to the one we are waiting for, we use this as the result. If `status.err` is set then we throw the error (as a `SolanaError`), otherwise we resolve the promise.

Fixes #1211

### DIFF
--- a/packages/kit/src/send-and-confirm-durable-nonce-transaction.ts
+++ b/packages/kit/src/send-and-confirm-durable-nonce-transaction.ts
@@ -107,12 +107,18 @@ export function sendAndConfirmDurableNonceTransactionFactory<
             } catch (e) {
                 // If nonce became invalid, check if our transaction actually landed
                 if (isSolanaError(e, SOLANA_ERROR__INVALID_NONCE)) {
-                    const { value: statuses } = await rpc
-                        .getSignatureStatuses([signature])
-                        .send({ abortSignal: config.abortSignal });
-                    const status = statuses[0];
+                    let status;
+                    try {
+                        const { value: statuses } = await rpc
+                            .getSignatureStatuses([signature])
+                            .send({ abortSignal: config.abortSignal });
+                        status = statuses[0];
+                    } catch {
+                        // RPC failed - propagate the original nonce error
+                        throw e;
+                    }
 
-                    if (status === null) {
+                    if (status === null || status === undefined) {
                         // Transaction doesn't exist - nonce was truly invalid
                         throw e;
                     }
--- a/packages/kit/src/__tests__/send-and-confirm-durable-nonce-transaction-test.ts
+++ b/packages/kit/src/__tests__/send-and-confirm-durable-nonce-transaction-test.ts
@@ -219,6 +219,35 @@ describe('sendAndConfirmDurableNonceTransactionFactory', () => {
             ).rejects.toThrow(SolanaError);
         });
 
+        it('throws INVALID_NONCE when getSignatureStatuses RPC call fails', async () => {
+            expect.assertions(2);
+
+            // Setup: nonce invalidation throws INVALID_NONCE
+            const invalidNonceError = new SolanaError(SOLANA_ERROR__INVALID_NONCE, {
+                actualNonceValue: 'xyz',
+                expectedNonceValue: 'abc',
+            });
+            mockGetNonceInvalidationPromise.mockRejectedValue(invalidNonceError);
+
+            // Setup: getSignatureStatuses throws
+            getSignatureStatusesMock.mockRejectedValue(new Error('RPC connection failed'));
+
+            const sendAndConfirm = sendAndConfirmDurableNonceTransactionFactory({
+                rpc,
+                rpcSubscriptions,
+            });
+
+            // Should throw INVALID_NONCE (not the RPC error)
+            await expect(
+                sendAndConfirm(MOCK_DURABLE_NONCE_TRANSACTION, {
+                    commitment: 'finalized',
+                }),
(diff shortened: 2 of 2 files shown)


packages/kit/src/send-and-confirm-durable-nonce-transaction.ts
```diff
@@ -82,15 +90,68 @@ export function sendAndConfirmDurableNonceTransactionFactory<
...
+        return async function wrappedGetNonceInvalidationPromise(config) {
+            try {
+                return await getNonceInvalidationPromise(config);
+            } catch (e) {
+                // If nonce became invalid, check if our transaction actually landed
+                if (isSolanaError(e, SOLANA_ERROR__INVALID_NONCE)) {
+                    const { value: statuses } = await rpc
+                        .getSignatureStatuses([signature])
+                        .send({ abortSignal: config.abortSignal });
+                    const status = statuses[0];
```
amilz: should this be in a try/catch? if this call throws, i think we'd want to propogate `e`

mcintyre94: Thanks, agree with this. I've updated it so that if this request fails we propagate the original nonce error.

PR #1236, opened 2026-01-23
Add a function to get the first failed transaction plan result

mcintyre94:
#### Problem

If you have a `TransactionPlanResult` that contains some failed transactions, it is non-trivial to get the first one that failed, which is often useful for error handling.

#### Summary of Changes

Add a new function `getFirstFailedTransactionPlanResult` that finds the first `SingleTransactionPlanResult` that has a failed status. It throws a new `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TRANSACTION_PLAN_RESULT_NOT_FOUND` if there is no such failed transaction plan result


packages/instruction-plans/src/transaction-plan-result.ts
```diff
@@ -533,17 +503,23 @@ export function flattenTransactionPlanResult(result: TransactionPlanResult): Sin
...
+    TContext extends TransactionPlanResultContext = TransactionPlanResultContext,
+> = SingleTransactionPlanResult<TContext> & { status: { kind: 'failed' } };
 
 /**
  * A {@link SingleTransactionPlanResult} with 'canceled' status.
  */
-export type CanceledSingleTransactionPlanResult = SingleTransactionPlanResult & { status: { kind: 'canceled' } };
+export type CanceledSingleTransactionPlanResult<
+    TContext extends TransactionPlanResultContext = TransactionPlanResultContext,
+> = SingleTransactionPlanResult<TContext> & { status: { kind: 'canceled' } };
```
lorisleiva: This is technically more accurate but the context really only matters for successful single results. They have to be provided to parent type like sequential/parallel but here it might be confusing for people to see a `TContext` type parameter that's never used in the type? Happy to keep it for consistency though. Just wanted to flag that failed/canceled single results actually don't make use of the context.

packages/instruction-plans/src/__tests__/transaction-plan-result-test.ts
```diff
@@ -609,3 +499,148 @@ describe('summarizeTransactionPlanResult', () => {
...
+        expect(() => getFirstFailedTransactionPlanResult(mixedResult)).toThrow(
+            new SolanaError(SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TRANSACTION_PLAN_RESULT_NOT_FOUND, {
+                context: {
+                    transactionPlanResult: mixedResult,
+                },
+            }),
+        );
+    });
+
+    it('error context contains transactionPlanResult as non-enumerable', () => {
```
lorisleiva: nit: sentence doesn't start with `it` haha

packages/instruction-plans/src/transaction-plan-result.ts
```diff
@@ -454,58 +453,29 @@ export function findTransactionPlanResult<TContext extends TransactionPlanResult
...
-export function everyTransactionPlanResult(
-    transactionPlanResult: TransactionPlanResult,
-    predicate: (plan: TransactionPlanResult) => boolean,
-): boolean {
-    if (!predicate(transactionPlanResult)) {
-        return false;
-    }
-    if (transactionPlanResult.kind === 'single') {
-        return true;
+export function getFirstFailedTransactionPlanResult<
```
lorisleiva: nit: Just need a docblock on that one.

packages/instruction-plans/src/transaction-plan-result.ts
mcintyre94: Fair point, I'll remove it from failed/canceled because then it's obvious why it's eg. not relevant to this function (which I can also remove it from). We'll end up using some of these types for some of our type assertion functions, so again nice not to pull in a type param we don't need there.

packages/instruction-plans/src/transaction-plan-result.ts
```diff
@@ -454,6 +455,61 @@ export function findTransactionPlanResult<TContext extends TransactionPlanResult
...
+ * const firstFailed = getFirstFailedTransactionPlanResult(result);
+ * // Returns the failed result for messageB.
+ * ```
+ *
+ * @see {@link FailedSingleTransactionPlanResult}
+ * @see {@link findTransactionPlanResult}
+ */
+export function getFirstFailedTransactionPlanResult(
+    transactionPlanResult: TransactionPlanResult,
+): FailedSingleTransactionPlanResult {
```
lorisleiva: I like the name `getFirstFailedTransactionPlanResult` and this is what we decided on when we planned this but now that I see the `FailedSingleTransactionPlanResult` return type next to it I wonder if we should call it `getFirstFailedSingleTransactionPlanResult` instead? It is slightly less verbose and _technically_ there isn't another form of "failed plan result" but maybe it's better to stay consistent with our type names so the formula is simply `getFirstX` where `X` is the type name and not some abbreviation version of it. What to you think?

mcintyre94: Makes sense, I slightly prefer the consistency with the type name too. Will update it

packages/errors/src/codes.ts
```diff
@@ -286,6 +286,7 @@ export const SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_PACKER_ALREADY_COMPLETE =
 export const SOLANA_ERROR__INSTRUCTION_PLANS__EMPTY_INSTRUCTION_PLAN = 7618002;
 export const SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN = 7618003;
 export const SOLANA_ERROR__INSTRUCTION_PLANS__NON_DIVISIBLE_TRANSACTION_PLANS_NOT_SUPPORTED = 7618004;
+export const SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TRANSACTION_PLAN_RESULT_NOT_FOUND = 7618005;
```
lorisleiva: Maybe here as well?


```suggestion
export const SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_SINGLE_TRANSACTION_PLAN_RESULT_NOT_FOUND = 7618005;
```

PR #1245, opened 2026-01-26
Add `flattenInstructionPlan` and `flattenTransactionPlan` functions

mcintyre94:
#### Problem

As the instruction/transaction plan/result structures have a tree structure, it can be difficult to extract the individual leaf nodes from these. We've already seen a need for this - we have a `flattenTransactionPlanResult` and `getAllSingleTransactionPlans` functions.

#### Summary of Changes

- Deprecate `getAllSingleTransactionPlans` in favour of a new `flattenTransactionPlan` with the same behaviour
- Add a new `flattenInstructionPlan` function
- Refactor `flattenTransactionPlanResult` to have the same simpler code structure as the above two functions, improve the docstring and add typetests

PR #1247, opened 2026-01-26
Add appendTransactionMessageInstructionPlan function

mcintyre94:
#### Problem

For APIs that operate on `TransactionMessage`, it is not currently possible to use instruction plans. Often this makes sense, many instruction plans won't fit in a single message and require the transaction planning capabilities. But in some cases, like the simple `SequentialInstructionPlan` that we added to the token client, it is possible to add the instructions from these to a transaction message.

#### Summary of Changes

- Add a new function `appendTransactionMessageInstructionPlan(plan, message)` , which flattens an instruction plan and then adds the instructions to the message
- For a `SingleInstructionPlan`, it appends the instruction to the message
- For a `MessagePackerInstructionPlan`, it attempts to pack the message with its instructions. If it is able to fit within the capacity of the message, then this is successful. Note that we assume a stricter type than `packMessageToCapacity` guarantees, and if this function behaves differently to what is expected (appending instructions) then the returned type from this function will be inaccurate. This is intentional because it's more useful to return a strong type from this function. 
  - Note that I considered enforcing this, ie just extracting the instructions from the returned packed transaction. But I think on balance this would be more tricky to debug in the case where the message packer is changing the message in unexpected ways.

### DIFF
--- a/.changeset/funny-chairs-throw.md
+++ b/.changeset/funny-chairs-throw.md
@@ -1,0 +1,5 @@
+---
+'@solana/instruction-plans': minor
+---
+
+Add a new function `appendTransactionMessageInstructionPlan` that can be used to add the instructions from an instruction plan to a transaction message


mcintyre94:
* **#1247** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1247?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1247?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1245** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1245?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

.changeset/calm-carrots-raise.md
lorisleiva, line 5: I'm not sure if this is a GitHub or Graphite bug but it seems like the previous PR changes are being displayed here as well. Maybe worth doing at `gt sync` or something? I'll take a look at the new changes only anyway.

packages/instruction-plans/src/append-instruction-plan.ts
```diff
@@ -0,0 +1,86 @@
...
+            }
+            if (plan.kind === 'messagePacker') {
+                const messagerPacker = plan.getMessagePacker();
+                let nextMessage: Out = messageSoFar;
+                while (!messagerPacker.done()) {
+                    nextMessage = messagerPacker.packMessageToCapacity(nextMessage) as Out;
+                }
+                return nextMessage;
+            }
+            return messageSoFar;
```
lorisleiva: Maybe we should throw a `SOLANA_ERROR__INVARIANT_VIOLATION__INVALID_INSTRUCTION_PLAN_KIND` here instead since we only expect `single` and `messagePacker` plans?

Imagine there's a typo in a plan kind (e.g. `"singel"`), then we'd be skipping it silently which could cause on-chain side-effects.

packages/instruction-plans/src/__tests__/append-instruction-plan-test.ts
```diff
@@ -0,0 +1,149 @@
...
+        const instructionA = createInstruction('A');
+        const instructionB = createInstruction('B');
+        const plan = sequentialInstructionPlan([instructionA, instructionB]);
+
+        const result = appendTransactionMessageInstructionPlan(plan, message);
+
+        expect(result.instructions).toStrictEqual([existingInstruction, instructionA, instructionB]);
+    });
+
+    it('appends instructions from a message packer instruction plan', () => {
```
lorisleiva, line 109: It would be nice to have a test using a message packer that is too big to fit in a single transaction message to ensure we get an appropriate `SolanaError`.

.changeset/calm-carrots-raise.md
mcintyre94: Thanks for the heads up, should be a much cleaner diff now!

packages/instruction-plans/src/append-instruction-plan.ts
mcintyre94: Good idea, added this

packages/instruction-plans/src/__tests__/append-instruction-plan-test.ts
mcintyre94: Good idea, added a test that verifies it throws with `SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN`.

packages/instruction-plans/src/append-instruction-plan.ts
```diff
@@ -0,0 +1,86 @@
+import { Instruction } from '@solana/instructions';
```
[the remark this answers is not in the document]
mcintyre94: Changed this to a type import, but we don't generally enforce that and should add a lint rule if we want to. Report that it's unused is incorrect.

Issue #1246, opened 2026-01-26
RPC: Include `index` in `getSignaturesForAddress` response


joncinque:
## Problem

With https://github.com/anza-xyz/agave/pull/9683, to be included with Agave v4.0, `getSignaturesForAddress` will include the block index of each transaction returned. Kit should also include this new field.

## Proposed solution

Add the new field to the appropriate response.

PR #1248, opened 2026-01-27
RPC: Include index in getSignaturesForAddress response #1246

sdt03:
#### Problem [Refers to Issue #1246 ]
With https://github.com/anza-xyz/agave/pull/9683, to be included with Agave v4.0, getSignaturesForAddress will include the block index of each transaction returned. Kit should also include this new field.

#### Summary of Changes
- Add index field to GetSignaturesForAddressTransaction type
- Configure index as allowed numeric keypath to prevent bigint conversion

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 node_modules
 .pnp
 .pnp.js
+.pnpm-store/*
 
 # misc
 .DS_Store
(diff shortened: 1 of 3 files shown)


mcintyre94:
Thanks for this! We need to figure out when to align Kit to Agave 4.0 (and if we need a new branch etc), so I've put this as draft for now. But will get this merged once we figure out our approach!

sdt03:
> Thanks for this! We need to figure out when to align Kit to Agave 4.0 (and if we need a new branch etc), so I've put this as draft for now. But will get this merged once we figure out our approach!

Thank you for reviewing this. Would it be possible to assign it back to me if any changes or follow ups are needed while you figure out the approach?

mcintyre94:
Thanks for this! It looks like the field landed in 4.0 as `transactionIndex`, this PR is superseded by #1660.

PR #1260, opened 2026-01-27
Deprecate the `cause` property on SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN

mcintyre94:
#### Problem

The `cause` property on this error includes only the first error in a plan result, and is not how we want developers to handle errors with transaction plans. The `transactionPlanResult` includes all failed/canceled results and has enough information for any recovery approach.

#### Summary of Changes

- Update the message for `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN`
- Add a new type `SolanaErrorWithDeprecatedCause` which replaces `cause` so that it is marked as deprecated
- Return this type from `isSolanaError` when the code is `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN` (or any code we add in future to `SolanaErrorCodeWithDeprecatedCause`) 

This PR does not change how we currently set `cause`, which would be a breaking change, but consumers will see it marked deprecated for this function:

Screenshot 2026-01-27 at 17 02 50

Caveat: we may not want to merge this because there's a bug in... something (VSCode?) that's causing the deprecation text (but not strikethrough decoration) to show on errors where it shouldn't:

Screenshot 2026-01-27 at 17 21 35

### DIFF
--- a/.changeset/pretty-clubs-press.md
+++ b/.changeset/pretty-clubs-press.md
@@ -1,0 +1,5 @@
+---
+'@solana/errors': minor
+---
+
+Mark the `cause` deprecated for `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN` error


mcintyre94:
Refactored to use overloads instead of a conditional type, and the weird bug has gone away!

Screenshot 2026-01-27 at 17 41 41

Screenshot 2026-01-27 at 17 41 53

Issue #1150, opened 2025-12-22
Rewrite base64 + base16 codecs using native APIs


mcintyre94:
TIL, the web platform has built in conversion between base64/hex strings and Uint8array 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toBase64
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromHex
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toHex

All are in Baseline 2025, ie available in the latest version of all browsers

Should make our base64 + base16 codecs much simpler + probably more performant

- We may want to wait until Node 26 releases, since it is only in 25 and that will be the next LTS
- Will need to use `EsNext.TypedArrays` lib in typescript 6.0 https://github.com/microsoft/TypeScript/pull/61696/files

PR #1271, opened 2026-01-30
Rewrite base64 + base16 codecs using native APIs

rajgoesout:
#### Problem

https://github.com/anza-xyz/kit/issues/1150

#### Summary of Changes

refactor codecs-strings: use native base64/hex typedarray helpers

Closes https://github.com/anza-xyz/kit/issues/1150

### DIFF
--- a/.changeset/dirty-carpets-change.md
+++ b/.changeset/dirty-carpets-change.md
@@ -1,0 +1,5 @@
+---
+'@solana/codecs-strings': patch
+---
+
+Rewrite base64 + base16 codecs using native APIs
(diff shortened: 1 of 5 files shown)


lorisleiva:
Hi there, I know this is still a draft but would you mind keeping the encoding/decoding logic in the place it is already in? Currently, it is tricky to see what has actually been changed due to the various new helper functions.

Also, please avoid top level variables such as:
```ts
const uint8ArrayFromHex = (Uint8Array as unknown as { fromHex?: (value: string) => Uint8Array }).fromHex;
const uint8ArrayToHex = (Uint8Array.prototype as unknown as { toHex?: () => string }).toHex;
```

Finally, I'd rather the new tests were not using nested helper functions (e.g. `withNativeHexHelpers`) and were located in the already existing test files to keep the tests consistent.

mcintyre94:
Also please note from the issue that this is only available in Node since v25, so v26 will be the first LTS supporting it, which is probably the earliest we would want to use this. We could consider a polyfill (the implementation we already have) when the native methods are not available though.

Issue #1147, opened 2025-12-19
Remove the export of `BaseTransactionMessage`


mcintyre94:
In [this stack of PRs](https://github.com/anza-xyz/kit/pull/1146) we refactored transaction message functions to operate over `TransactionMessage` instead of `BaseTransactionMessage` which improves typing. We also deprecate `BaseTransactionMessage` but it remains exported.

In a future major version we should stop exporting `BaseTransactionMessage`, and refactor the rest of the library to use `TransactionMessage`. The types are compatible, and shouldn't require any non-type code changes.

mcintyre94:
Fixed by #1289

PR #1274, opened 2026-01-30
Remove BaseTransactionMessage from signers and transactions packages

mcintyre94:
#### Summary of Changes

These are the only packages outside of transaction-messages where we use it. Can remove from both trivially

Part of #1147


mcintyre94:
Turns out this isn't trivial, and also I forgot that `pnpm dev` doesn't run `compile:js` or `compile:typedefs`

Will take a closer look at how signers is using it and see if we can refactor it out

PR #1316, opened 2026-02-04
ci: use github app generated token for dependabot pipeline

yihau:
#### Problem

let's use github app generated token instead of PAT for dependabot's PR

#### Summary of Changes

update

### DIFF
--- a/.github/workflows/update-docs-lockfile.yml
+++ b/.github/workflows/update-docs-lockfile.yml
@@ -11,12 +11,30 @@ permissions:
 jobs:
   update-lockfile:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
     if: github.event.pull_request.user.login == 'dependabot[bot]'
     steps:
-      - name: Checkout PR branch
-        uses: actions/checkout@v4
+      - uses: actions/create-github-app-token@v2
+        id: app-token
         with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.PRIVATE_KEY }}
+
+      - name: Set git config
+        run: |
+          git config --global user.email "github-actions@github.com"
+          git config --global user.name "github-actions"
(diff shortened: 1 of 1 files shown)


.github/workflows/update-docs-lockfile.yml
```diff
@@ -11,12 +11,30 @@ permissions:
...
+      - uses: actions/create-github-app-token@v2
+        id: app-token
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.PRIVATE_KEY }}
+
+      - name: Set git config
+        run: |
+          git config --global user.email "github-actions@github.com"
+          git config --global user.name "github-actions"
```
mcintyre94, line 27: Would it be better to let this just identify its commits as our github app? Might be easier to track down where it's coming from in future!

.github/workflows/update-docs-lockfile.yml
```diff
@@ -11,12 +11,30 @@ permissions:
...
+          git config --global url."https://x-access-token:${GITHUB_TOKEN}@github.com/".insteadOf https://github.com/
```
mcintyre94, line 28: Do you know why this is needed even though we're passing the token in `actions/checkout`?

mcintyre94 (approved):
Awesome thankyou! 🚀

.github/workflows/update-docs-lockfile.yml
yihau: we can use github app's identity if you prefer. personally, I use github actions for all bot-generated commits

.github/workflows/update-docs-lockfile.yml
yihau: tbh, I'm not sure. I didn't set this when I integrated it in other repos. it complained, so I found this small workaround on gitHub forum. unfortunately, I forgot the original post link 🫠

PR #1340, opened 2026-02-09
Replace enum usage with string unions and u8 codecs for Endian and OffchainMessageContentFormat

### LINKED ISSUE #1337: Remove Typescript enums
We'd probably be better off not using typescript enums, due to them compiling to a slightly complex reverse-mapped function type, and also not being helpful when logged (which we generally try to be with eg our `Address` type).

We use enums currently for:

```ts
export enum Endian {
    Little,
    Big,
}
```

- I'd guess that this could just be a union `"Little" | "Big"`

```ts
export enum AccountRole {
    // Bitflag guide: is signer ⌄⌄ is writable
    WRITABLE_SIGNER = /* 3 */ 0b11, // prettier-ignore
    READONLY_SIGNER = /* 2 */ 0b10, // prettier-ignore
    WRITABLE =        /* 1 */ 0b01, // prettier-ignore
    READONLY =        /* 0 */ 0b00, // prettier-ignore
}
```

- This would be more useful with string values so that when we log a transaction message instruction account it is legible. We'd need to convert it when compiled though, and would also need to think about the bitshifting used to convert roles etc

```ts
export enum OffchainMessageContentFormat {
    RESTRICTED_ASCII_1232_BYTES_MAX = 0,
    UTF8_1232_BYTES_MAX = 1,
    UTF8_65535_BYTES_MAX = 2,
}
```

This can probably also just a union of strings.

dpsi9:
#### Problem
The codebase currently relies on TypeScript enums in places where they are
suboptimal: enums compile to reverse-mapped runtime objects, are awkward to work
with in codecs, and do not produce particularly helpful output when logged.

The issue calls out three enums for cleanup. This PR addresses two of them
(Endian and OffchainMessageContentFormat), which are better represented as
string-based APIs with explicit numeric encoding only at the wire boundary. ThisPR addresses a subset of that work.

#### Summary of Changes
- Replace Endian with a const object + string union
- Replace OffchainMessageContentFormat with a string union for better logging
- Add an explicit numeric encoder for OffchainMessageContentFormat at
serialization, preserving existing wire semantics
- Remove enum-based codecs; validation and error behavior remain unchanged

Cleanup of AccountRole is intentionally left out and can be handled in a
follow-up PR, as it involves additional bitflag semantics and encoding concerns.
Fixes #1337

### DIFF
--- a/packages/offchain-messages/src/__tests__/content-test.ts
+++ b/packages/offchain-messages/src/__tests__/content-test.ts
@@ -11,6 +11,7 @@ import {
     assertIsOffchainMessageContentRestrictedAsciiOf1232BytesMax,
     assertIsOffchainMessageContentUtf8Of1232BytesMax,
     assertIsOffchainMessageContentUtf8Of65535BytesMax,
+    encodeOffchainMessageContentFormat,
     isOffchainMessageContentRestrictedAsciiOf1232BytesMax,
     isOffchainMessageContentUtf8Of1232BytesMax,
     isOffchainMessageContentUtf8Of65535BytesMax,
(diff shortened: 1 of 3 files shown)


mcintyre94:
Thanks but this will probably be a breaking change and we'll need to do it carefully and are not planning to do it for a while. I'm going to close this for now because it needs careful consideration from us when we want to do this.

dpsi9:
Makes sense and thanks for taking a look, happy to have explored it and I'm glad to help again whenever you decide to revist it

PR #1363, opened 2026-02-12
Add getPredicateCodec to select an encoder/decoder using a predicate

mcintyre94:
#### Problem

When we add support for v1 transactions, we will need to encode the `Transaction` data structure differently depending on the transaction version.

Currently the best way to do this would be to use `getUnionEncoder`, but this supports an arbitrary number of variants and is more complex an interface than we need here. 

#### Summary of Changes

This PR adds a higher-level helper `getPredicateEncoder(predicate, ifTrue, ifFalse)` that can be used to easily switch between two variants. Under the hood it just calls `getUnionEncoder`, but presents a simpler interface for predicate cases.

It also adds `getPredicateDecoder(predicate, decoderIfTrue, decoderIfFalse)` and `getPredicateCodec(encodePredicate, decodePredicate, codecIfTrue, codecIfFalse)`. All use the `union` codec functions under the hood. 

Note that I initially considered using an object with named fields, but our linter requires object fields to be ordered. I think the order `[ifFalse, ifTrue, predicate]` is more confusing than the object is helpful. It also doesn't match the way other kit functions are structured. I think it works as separate params, but the codec version is a bit hairy because it's two booleans and then two encoders/decoders/codecs.

### DIFF
--- a/.changeset/chubby-zoos-walk.md
+++ b/.changeset/chubby-zoos-walk.md
@@ -1,0 +1,7 @@
+---
+'@solana/codecs-data-structures': minor
+---
+
+Adds new functions `getPredicateEncoder`, `getPredicateDecoder` and `getPredicateCodec`.
+
+These can be used to write an encoder that switches between two encoders based on the value being encoded, or a decoder that switches between two decoders based on the byte array being decoded.


packages/codecs-data-structures/src/predicate.ts
```diff
@@ -0,0 +1,178 @@
...
+ * //   └── Large number encoded as u32
+ * ```
+ *
+ * @see {@link getPredicateCodec}
+ */
+export function getPredicateEncoder<T>(
+    predicate: (value: T) => boolean,
+    ifTrue: FixedSizeEncoder<T>,
+    ifFalse: FixedSizeEncoder<T>,
+): FixedSizeEncoder<T>;
```
lorisleiva: nit: We could even add a strong overload on top of that one that checks if the size is the same. Something like:

```ts
export function getPredicateEncoder<TFrom, TSize extends number>(
    predicate: (value: TFrom) => boolean,
    ifTrue: FixedSizeEncoder<TFrom, TSize>,
    ifFalse: FixedSizeEncoder<TFrom, TSize>,
): FixedSizeEncoder<TFrom, TSize>;
// ...
```

packages/codecs-data-structures/src/predicate.ts
```diff
@@ -0,0 +1,178 @@
...
+ * //   └── Small number encoded as u8
+ *
+ * encoder.encode(1000);
+ * // 0xe8030000
+ * //   └── Large number encoded as u32
+ * ```
+ *
+ * @see {@link getPredicateCodec}
+ */
+export function getPredicateEncoder<T>(
```
lorisleiva: These encoder type parameters should be `TFrom`.

packages/codecs-data-structures/src/predicate.ts
```diff
@@ -0,0 +1,178 @@
...
+ * decoder.decode(new Uint8Array([0x2a]));
+ * // 42 (decoded as u8)
+ *
+ * decoder.decode(new Uint8Array([0xe8, 0x03, 0x00, 0x00]));
+ * // 1000 (decoded as u32)
+ * ```
+ *
+ * @see {@link getPredicateCodec}
+ */
+export function getPredicateDecoder<T>(
```
lorisleiva: These encoder type parameters should be `TTo`.

packages/codecs-data-structures/src/predicate.ts
```diff
@@ -0,0 +1,178 @@
...
+ * // 0xe8030000 (encoded as u32)
+ *
+ * codec.decode(smallBytes); // 42
+ * codec.decode(largeBytes); // 1000
+ * ```
+ *
+ * @see {@link getPredicateEncoder}
+ * @see {@link getPredicateDecoder}
+ */
+export function getPredicateCodec<T>(
```
lorisleiva: These codec type parameters should use both `TFrom` and `TTo` such that `TTo extends TFrom`.


```suggestion
export function getPredicateCodec<TFrom, TTo extends TFrom>(
```

And then you can wire these `TFrom` and `TTo` parameters in their respective encoder/decoder.

mcintyre94:
Thanks @lorisleiva! I've addressed those comments. The docs will need us to publish a build first, I've pushed a branch and linked to this issue: #1365

PR #1368, opened 2026-02-13
feat(signers): add functions to create signers from UiWalletAccount

### LINKED ISSUE #1361: Add a function for UiWalletAccount -> TransactionSigner
## Motivation

This would provide a bridge between any wallet library/plugin that handles wallet-standard `UiWalletAccount` and the `TransactionSigner` interface, so any wallet can be used as a signer. 

We already have react hooks for this: https://github.com/anza-xyz/kit/blob/main/packages/react/src/useWalletAccountTransactionSigner.ts and https://github.com/anza-xyz/kit/blob/main/packages/react/src/useWalletAccountTransactionSendingSigner.ts

But we can extract the pure part of this that doesn't depend on react into a function that wouldn't depend on react

## Details

In Kit it probably makes sense to have 2 functions in the `signers` package: 

```
// requires the `solana:signTransaction` feature
createSignerFromWalletAccount(account: UiWalletAccount): TransactionModifyingSigner

// requires the `solana:signAndSendTransaction` feature
createSendingSignerFromWalletAccount(account: UiWalletAccount): TransactionSendingSigner
```

This will add a dependency on `@wallet-standard/ui`, which is already used in the react package

Note: We might want to take the opportunity to rename the `DO_NOT_USE_OR_YOU_WILL_BE_FIRED` functions in `wallet-standard/ui` at the same time

dpsi9:
#### Problem
There's no way to convert UiWalletAccount to kit's  TransactionSigner outside of react. React package does have hooks for this but core logic should be in signers package.

#### Summary of Changes
- Created two files wallet-account-signer.ts and wallet-account-sending-signer.ts in signers package and added the two functions createSignerFromWalletAccount and createSendingSignerFromWalletAccount.
- Added unit tests and type tests for both the functions
- Updated react hooks(useWalletAccountTransactionSigner.ts & useWalletAccountSendingSigner.ts to use the new functions.
- Added @wallet-standard/ui, @wallet-standard/ui-registry, @wallet-standard/errors as dependencies to signers package

Fixes #1361

### DIFF
--- a/packages/wallet-account-signer/src/__tests__/wallet-account-transaction-signer-test.ts
+++ b/packages/wallet-account-signer/src/__tests__/wallet-account-transaction-signer-test.ts
@@ -843,10 +843,8 @@ describe('createSignerFromWalletAccount', () => {
         jest.mocked(getCompiledTransactionMessageDecoder).mockReturnValue({
             decode: jest
                 .fn()
-                .mockReturnValueOnce({ lifetimeToken: 'changed-1' }) // tx1 token check
-                .mockReturnValueOnce({ lifetimeToken: 'changed-1' }) // tx1 new lifetime fetch
-                .mockReturnValueOnce({ lifetimeToken: 'changed-2' }) // tx2 token check
-                .mockReturnValueOnce({ lifetimeToken: 'changed-2' }), // tx2 new lifetime fetch
+                .mockReturnValueOnce({ lifetimeToken: 'changed-1' }) // tx1 decode lifetime
+                .mockReturnValueOnce({ lifetimeToken: 'changed-2' }), // tx2 decode lifetime
         } as unknown as ReturnType<typeof getCompiledTransactionMessageDecoder>);
 
         // Mock new lifetimes.
@@ -899,8 +897,8 @@ describe('createSignerFromWalletAccount', () => {
         // Then bytesEqual is called twice.
         expect(bytesEqual).toHaveBeenCalledTimes(2);
 
-        // And the decoder is called 4 times (2 token checks + 2 new lifetime fetches).
-        expect(getCompiledTransactionMessageDecoder().decode).toHaveBeenCalledTimes(4);
+        // And the decoder is called 2 times (once per transaction.
+        expect(getCompiledTransactionMessageDecoder().decode).toHaveBeenCalledTimes(2);
 
         // And each transaction has the new lifetime.
         expect(result[0].lifetimeConstraint).toEqual(newLifetime1);
--- a/packages/wallet-account-signer/src/wallet-account-transaction-signer.ts
+++ b/packages/wallet-account-signer/src/wallet-account-transaction-signer.ts
@@ -98,32 +98,32 @@ export function createTransactionSignerFromWalletAccount<TWalletAccount extends 
 
                         assertIsTransactionWithinSizeLimit(decodedSignedTransaction);
 
-                        // Use the input transaction at the same index if available for lifetime comparison
                         const inputTransaction = transactions[index];
                         const existingLifetime =
                             inputTransaction && 'lifetimeConstraint' in inputTransaction
                                 ? (inputTransaction as TransactionWithLifetime).lifetimeConstraint
                                 : undefined;
 
-                        if (existingLifetime) {
-                            if (
-                                inputTransaction &&
-                                bytesEqual(decodedSignedTransaction.messageBytes, inputTransaction.messageBytes)
-                            ) {
-                                // If the transaction has identical bytes, the lifetime won't have changed
-                                return Object.freeze({
-                                    ...decodedSignedTransaction,
-                                    lifetimeConstraint: existingLifetime,
-                                });
-                            }
+                        // Fast path: identical bytes means the lifetime hasn't changed
+                        if (
+                            existingLifetime &&
+                            bytesEqual(decodedSignedTransaction.messageBytes, inputTransaction.messageBytes)
+                        ) {
+                            return Object.freeze({
+                                ...decodedSignedTransaction,
+                                lifetimeConstraint: existingLifetime,
+                            });
+                        }
 
-                            // If the transaction has changed, check the lifetime constraint field
-                            const compiledTransactionMessage = getCompiledTransactionMessageDecoder().decode(
-                                decodedSignedTransaction.messageBytes,
-                            );
+                        // Decode once to inspect the lifetime token
+                        const compiledTransactionMessage = getCompiledTransactionMessageDecoder().decode(
+                            decodedSignedTransaction.messageBytes,
+                        );
+
+                        // If the token matches the existing lifetime, reuse it
+                        if (existingLifetime) {
                             const currentToken =
                                 'blockhash' in existingLifetime ? existingLifetime.blockhash : existingLifetime.nonce;
-
                             if (compiledTransactionMessage.lifetimeToken === currentToken) {
                                 return Object.freeze({
                                     ...decodedSignedTransaction,
@@ -132,10 +132,7 @@ export function createTransactionSignerFromWalletAccount<TWalletAccount extends 
                             }
                         }
 
-                        // If we get here then there is no existing lifetime, or the lifetime has changed. We need to attach a new lifetime
-                        const compiledTransactionMessage = getCompiledTransactionMessageDecoder().decode(
-                            decodedSignedTransaction.messageBytes,
-                        );
+                        // No existing lifetime or it has changed — fetch a new one
                         const lifetimeConstraint =
                             await getTransactionLifetimeConstraintFromCompiledTransactionMessage(
                                 compiledTransactionMessage,


dpsi9:
@mcintyre94, I was actually working on this from yesterday, and just submitted this PR. And regarding the DO_NOT_USE_OR_YOU_WILL_BE_FIRED function, there's already an open Issue, once that's merged and released, I'll update this code to use the new name.

packages/signers/src/wallet-account-sending-signer.ts
```diff
@@ -0,0 +1,98 @@
...
+ * const signer = createSendingSignerFromWalletAccount(walletAccount, 'solana:devnet');
+ * const [signature] = await signer.signAndSendTransactions([transaction]);
+ * ```
+ *
+ * @see {@link TransactionSendingSigner}
+ * @see {@link createSignerFromWalletAccount}
+ */
+export function createSendingSignerFromWalletAccount<TWalletAccount extends UiWalletAccount>(
+    uiWalletAccount: TWalletAccount,
+    chain: `solana:${string}`,
```
mcintyre94: Nit: I think you can use `SolanaChain` from `@solana/wallet-standard-chains` for this: https://github.com/anza-xyz/wallet-standard/blob/master/packages/core/chains/src/index.ts

packages/signers/src/wallet-account-sending-signer.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    return {
+        address: address(uiWalletAccount.address),
+
+        async signAndSendTransactions(transactions, config = {}) {
+            const { abortSignal, ...options } = config;
+            abortSignal?.throwIfAborted();
+            if (transactions.length > 1) {
+                throw new SolanaError(SOLANA_ERROR__SIGNER__WALLET_MULTISIGN_UNIMPLEMENTED);
+            }
```
mcintyre94: I know the existing hooks do this but I'm not convinced we should here

wallet-standard allows there to be multiple inputs, and wallets can decide what they want to do if you send more than one. I don't think kit should restrict this

packages/signers/src/wallet-account-sending-signer.ts
```diff
@@ -0,0 +1,98 @@
...
+                          },
+                      }
+                    : null),
+            };
+
+            // Wallet Standard returns an array of results even for a single transaction, this signer enforces exactly one.
+            const outputs = await getAbortablePromise(feature.signAndSendTransaction(input), abortSignal);
+            if (outputs.length !== 1) {
+                throw new SolanaError(SOLANA_ERROR__SIGNER__WALLET_MULTISIGN_UNIMPLEMENTED);
+            }
```
mcintyre94: Even if we keep the input restricted to 1 (we probably shouldn't IMO) we definitely don't want to throw here - the wallet might genuinely have sent as multiple transactions and returned multiple signatures. We should definitely return whatever signatures the wallet returns to us.

packages/signers/src/wallet-account-signer.ts
```diff
@@ -0,0 +1,147 @@
...
+ * const signer = createSignerFromWalletAccount(walletAccount, 'solana:devnet');
+ * const [signedTransaction] = await signer.modifyAndSignTransactions([transaction]);
+ * ```
+ *
+ * @see {@link TransactionModifyingSigner}
+ * @see {@link createSendingSignerFromWalletAccount}
+ */
+export function createSignerFromWalletAccount<TWalletAccount extends UiWalletAccount>(
+    uiWalletAccount: TWalletAccount,
+    chain: `solana:${string}`,
```
mcintyre94: Same comment re `SolanaChain`

packages/signers/src/wallet-account-sending-signer.ts
```diff
@@ -0,0 +1,98 @@
...
+    uiWalletAccount: TWalletAccount,
+    chain: `solana:${string}`,
+): TransactionSendingSigner<TWalletAccount['address']> {
+    if (!uiWalletAccount.chains.includes(chain)) {
+        throw new WalletStandardError(WALLET_STANDARD_ERROR__FEATURES__WALLET_ACCOUNT_CHAIN_UNSUPPORTED, {
+            address: uiWalletAccount.address,
+            chain,
+            featureName: SolanaSignAndSendTransaction,
+            supportedChains: [...uiWalletAccount.chains],
+            supportedFeatures: [...uiWalletAccount.features],
```
mcintyre94: I don't think we should need to spread these, does `supportedChains: uiWalletAccount.chains` not work? Might be better to just cast otherwise, but I think they should work as-is.

packages/signers/src/wallet-account-signer.ts
```diff
@@ -0,0 +1,147 @@
...
+): TransactionModifyingSigner<TWalletAccount['address']> {
+    if (!uiWalletAccount.chains.includes(chain)) {
+        throw new WalletStandardError(WALLET_STANDARD_ERROR__FEATURES__WALLET_ACCOUNT_CHAIN_UNSUPPORTED, {
+            address: uiWalletAccount.address,
+            chain,
+            featureName: SolanaSignTransaction,
+            supportedChains: [...uiWalletAccount.chains],
+            supportedFeatures: [...uiWalletAccount.features],
```
mcintyre94: Same comment re spread

packages/signers/src/wallet-account-signer.ts
```diff
@@ -0,0 +1,147 @@
...
+    const transactionCodec = getTransactionCodec();
+
+    return {
+        address: address(uiWalletAccount.address),
+        async modifyAndSignTransactions(transactions, config = {}) {
+            const { abortSignal, ...options } = config;
+            abortSignal?.throwIfAborted();
+            if (transactions.length > 1) {
+                throw new SolanaError(SOLANA_ERROR__SIGNER__WALLET_MULTISIGN_UNIMPLEMENTED);
+            }
```
mcintyre94: Same comment re supporting multiple txs

packages/signers/src/wallet-account-signer.ts
```diff
@@ -0,0 +1,147 @@
...
+            const decodedSignedTransaction = transactionCodec.decode(
+                signedTransaction,
+            ) as (typeof transactions)[number];
+
+            assertIsTransactionWithinSizeLimit(decodedSignedTransaction);
+
+            const existingLifetime =
+                'lifetimeConstraint' in transaction
+                    ? (transaction as TransactionWithLifetime).lifetimeConstraint
+                    : undefined;
```
mcintyre94: We'd need to put all this lifetime logic into a map over the returned transactions, but I can't see anything that should stop us doing that. `existingLifetime` could compare with the transaction at the same index, ie the second input transaction compared to the second output one.

That isn't perfect because the arrays could be different sizes (we would need to handle no matching index, just treat as no existing lifetime), but it's only used as an optimisation to avoid decompiling the returned transaction if it hasn't changed.

packages/signers/src/wallet-account-signer.ts
```diff
@@ -0,0 +1,147 @@
...
+                decodedSignedTransaction.messageBytes,
+            );
+            const lifetimeConstraint =
+                await getTransactionLifetimeConstraintFromCompiledTransactionMessage(compiledTransactionMessage);
+            return Object.freeze([
+                {
+                    ...decodedSignedTransaction,
+                    lifetimeConstraint,
+                },
+            ]);
```
mcintyre94: The `freeze` here should be on the object (objects after making it multiple) inside the array, not the array itself. I don't think we need to freeze the arrays.

packages/signers/src/__tests__/wallet-account-sending-signer-test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+        const result = await signer.signAndSendTransactions([tx]);
+
+        // Then the transaction is encoded and forwarded to the wallet feature.
+        expect(mockEncode).toHaveBeenCalledWith(tx);
+        expect(mockFeature.signAndSendTransaction).toHaveBeenCalled();
+
+        // And the result is frozen.
+        expect(result).toHaveLength(1);
+        expect(Object.isFrozen(result)).toBe(true);
```
mcintyre94: Ideally we should try to keep assertions per test low, just because it makes debugging easier if something fails later. In this case IMO we should have this test just do:

```ts
        // Then the transaction is encoded and forwarded to the wallet feature.
        expect(mockEncode).toHaveBeenCalledWith(tx);
        expect(mockFeature.signAndSendTransaction).toHaveBeenCalled();
```

I think in this case we don't need the array to be frozen (we use that for objects), so we can probably just drop these assertions.

But it would be useful to have another test with similar setup to this one but asserting the returned signature is correct.

packages/wallet-account-signer/src/__tests__/wallet-account-transaction-sending-signer-test.ts
```diff
@@ -0,0 +1,181 @@
...
+import { getWalletAccountFeature, UiWalletAccount } from '@wallet-standard/ui';
+import { getWalletAccountForUiWalletAccount_DO_NOT_USE_OR_YOU_WILL_BE_FIRED } from '@wallet-standard/ui-registry';
+
+import { createSendingSignerFromWalletAccount } from '../wallet-account-sending-signer';
+
+jest.mock('@wallet-standard/ui');
+jest.mock('@wallet-standard/ui-registry');
+jest.mock('@solana/transactions');
+
+describe('createSendingSignerFromWalletAccount', () => {
```
mcintyre94, line 17: We also need a test similar to "calls `signAndSendTransaction` with all options except the `abortSignal`" in packages/react/src/__tests__/useWalletAccountTransactionSendingSigner-test.ts

And one similar to "rejects when aborted"

packages/wallet-account-signer/src/__tests__/wallet-account-transaction-signer-test.ts
```diff
@@ -0,0 +1,158 @@
...
+import { getWalletAccountForUiWalletAccount_DO_NOT_USE_OR_YOU_WILL_BE_FIRED } from '@wallet-standard/ui-registry';
+
+import { createSignerFromWalletAccount } from '../wallet-account-signer';
+
+jest.mock('@wallet-standard/ui');
+jest.mock('@wallet-standard/ui-registry');
+jest.mock('@solana/transactions');
+jest.mock('@solana/transaction-messages');
+
+describe('createSignerFromWalletAccount', () => {
```
mcintyre94, line 32: Again these tests are good, but we need to cover most of the cases in packages/react/src/__tests__/useWalletAccountTransactionSigner-test.ts - in particular we need tests for how we handle the transaction lifetime here.

packages/signers/src/wallet-account-sending-signer.ts
dpsi9: I have added the @solana/wallet-standard-chains dependency with the version - ^1.1.1 and updated this to use the SolanaChain

packages/signers/src/wallet-account-sending-signer.ts
dpsi9: Yes, I verified it, we don't need to spread it, uiWalletAccount.chains already satisfies the type. I've removed the spread and passed it directly

packages/signers/src/wallet-account-sending-signer.ts
dpsi9: Yes I agree, it shouldn't be limited to one transaction,  I've mapped the transactions to input array and passed it to feature.signAndSendTransaction as it does accept multiple transactions, and also added a check to ensure that wallet returns the exact no of outputs as inputs

mcintyre94: > also added a check to ensure that wallet returns the exact no of outputs as inputs

We shouldn't do this - wallets may return a different number of transactions, the spec doesn't enforce the same number. In general we really should avoid throwing if at all possible after we've received a successful response from the wallet. At that point the wallet has signed transactions and if at all possible we should get them back to the app.

dpsi9: I've now removed the length check and it no longer throws the error if the wallet returns a different number of outputs.

dpsi9:
@mcintyre94 , I’ve addressed the review comments. Please let me know if any further changes are needed.

packages/wallet-account-signer/src/wallet-account-signer.ts
```diff
@@ -0,0 +1,150 @@
...
+                    ) as (typeof transactions)[number];
+
+                    assertIsTransactionWithinSizeLimit(decodedSignedTransaction);
+
+                    // Use the input transaction at the same index if available for lifetime comparison
+                    const inputTransaction = transactions[index];
+                    const existingLifetime =
+                        inputTransaction && 'lifetimeConstraint' in inputTransaction
+                            ? (inputTransaction as TransactionWithLifetime).lifetimeConstraint
+                            : undefined;
```
mcintyre94: Nit: I think it'd be worth extracting this lifetime code to its own function - probably everything from here to the end of the map.

mcintyre94 (commented):
Thankyou! This is looking really good, really appreciate your updates to this :) 

We've decided that we're going to put this code in a new package instead of `signers`, as it introduces a dependency and we'd like to keep the Kit umbrella package dependency-free.

I'll open a PR that creates an empty package and then we'll need to get that published. And then we'll need to move these changes into that package, which we can do in this PR.

I'm happy to take over moving this into the new package at that point though if you'd like! Either way, nothing to do until I have a new package published.

dpsi9:
Sounds good, I’m happy either way. You can take it from here, or once the new package is ready I can move the implementation over.

mcintyre94:
Sorry for the delay on this @dpsi9! We do have the new package now (wallet-account-signer) which we can move this code into.

I won't be able to review for a few days, so no rush! And happy to pick this up next week if you'd prefer too.

Thanks again for all your help with this PR so far!

dpsi9:
No worries at all, thanks for the update btw, I'll move the implementation into the new package and push updates to this PR.

.changeset/swift-ravens-yawn.md
```diff
@@ -0,0 +1,13 @@
...
+'@solana/wallet-account-signer': minor
+---
+
+Add functions to create Kit signers from [Wallet Standard](https://github.com/wallet-standard/wallet-standard) `UiWalletAccount` objects. 
+
+- `createTransactionSignerFromWalletAccount(account, chain)` returns a `TransactionModifyingSigner` that uses the wallet's `solana:signTransaction` feature.
+
+- `createTransactionSendingSignerFromWalletAccount(account, chain)` returns a `TransactionSendingSigner` that uses the wallet's `solana:signAndSendTransaction` feature.
+
+- `createMessageSignerFromWalletAccount(account)` returns a `MessageModifyingSigner` that uses the wallet's `solana:signMessage` feature.
```
lorisleiva, line 15: There's probably value in having them as separate helpers but just wanted to say that you could technically have a single helper that returns a single signer that complies with all these interfaces, i.e. `TransactionModifyingSigner & TransactionSendingSigner & MessageModifyingSigner`. Signer helpers are clever enough to figure out which signing function to use base on the context and the other signers.

packages/wallet-account-signer/src/wallet-account-transaction-signer.ts
```diff
@@ -0,0 +1,155 @@
...
+                        assertIsTransactionWithinSizeLimit(decodedSignedTransaction);
+
+                        // Use the input transaction at the same index if available for lifetime comparison
+                        const inputTransaction = transactions[index];
+                        const existingLifetime =
+                            inputTransaction && 'lifetimeConstraint' in inputTransaction
+                                ? (inputTransaction as TransactionWithLifetime).lifetimeConstraint
+                                : undefined;
+
+                        if (existingLifetime) {
```
lorisleiva: This is great. My worry is no other signer out there that uses raw bytes to sign will do as good of a job to reconstruct the signed transaction. I wonder if we should extract that logic (in a follow up PR) into the `transactions` package. Something like `reconstructEncodedTransactionFromOriginalTransaction(originalTransaction, encodedTransaction)` perhaps? Or maybe something simpler, more scoped to retrieving the lifetime, like `hasTransactionLifetimeChanged(originalLifetime, messageBytes)`.

packages/wallet-account-signer/src/wallet-account-transaction-signer.ts
```diff
@@ -0,0 +1,155 @@
...
+                            if (
+                                inputTransaction &&
```
lorisleiva: I believe this part of the conditional is redundant (and confusing) since `existingLifetime` will always be undefined if `inputTransaction` is falsy.

packages/wallet-account-signer/src/wallet-account-transaction-signer.ts
```diff
@@ -0,0 +1,155 @@
...
+                                });
+                            }
+                        }
+
+                        // If we get here then there is no existing lifetime, or the lifetime has changed. We need to attach a new lifetime
+                        const compiledTransactionMessage = getCompiledTransactionMessageDecoder().decode(
+                            decodedSignedTransaction.messageBytes,
+                        );
+                        const lifetimeConstraint =
+                            await getTransactionLifetimeConstraintFromCompiledTransactionMessage(
```
lorisleiva, line 137: This function is actually synchronous by implementation. We just force it to be asynchronous `to allow for future implementations that may fetch "lastValidBlockHeight" using an RPC` (as noted in the docblock). Just wanted to flag that to see if that's still the path we want to follow as this is literally the only reason this function is async (and we don't even require an RPC argument at the moment).

.changeset/swift-ravens-yawn.md
mcintyre94: This is a really good point. Maybe we should rename `createTransactionSignerFromWalletAccount` to `createTransactionModifyingSignerFromWalletAccount`.
 
And then have a  `createTransactionSignerFromWalletAccount ` helper that returns `TransactionSigner`. Then it returns a signer with `TransactionModifyingSigner` if `solana:signTransaction` is available, and with `TransactionSendingSigner` if `solana:signAndSendTransaction` is available. Throws if neither is available.

For the wallet plugin this is what we need for `payer: TransactionSigner`, and I think that's the right granularity to expose to most apps as well. Then the more granular functions are for more advanced use cases.

I think it probably makes sense to keep message separate because that feels like a different API surface in this context.

packages/wallet-account-signer/src/wallet-account-transaction-signer.ts
mcintyre94: This is a good idea, this logic is originally from the react hooks too (just improved to handle the multiple transactions here).

packages/wallet-account-signer/src/wallet-account-transaction-signer.ts
mcintyre94: I noticed this too, and had forgotten we decided to do that. TBH I'm not sure why we didn't just make it sync and accept that we might make it async as a breaking change in future. It's a bit more speculative than we are anywhere else, and AFAIK there's no actual plan to provide this RPC method. I think it was probably the wrong decision at the time, but also not worth a breaking change now.

.changeset/swift-ravens-yawn.md
lorisleiva: I think the `createTransactionModifyingSignerFromWalletAccount` renaming is a good change regardless. It's more consistent with the other helpers.

If we wanted to stay consistent with some other signer factories out there, we _could_ offer a `createWalletAccountSigner` that does all three. This is akin to the `createKeyPairSigner` helper that is both a `TransactionSigner` and a `MessageSigner`. Nothing wrong with having a signer object that can sign transactions and messages IMO.

mcintyre94: I like that. I think we should do `createWalletAccountSigner(account, chain): TransactionSigner | (TransactionSigner & MessageSigner)`

Require one of the transaction sign features or throw, and if message is available then also return it but don't error if it's missing. Then apps are guaranteed a `TransactionSigner`, works as payer for the plugin, and apps just need to check before using it to sign messages in case the wallet doesn't have that feature.

packages/wallet-account-signer/src/wallet-account-transaction-signer.ts
mcintyre94: Issue opened: #1466

.changeset/swift-ravens-yawn.md
mcintyre94: Added: https://github.com/anza-xyz/kit/pull/1368/changes/516a5862ad9a8d06a90a4f1d357b8e297fa1611d

packages/wallet-account-signer/src/wallet-account-transaction-signer.ts
mcintyre94: Simplified this a bit in e6905a11588e6a305299b4b42297eef93fbf7506 - also only decodes the message once now if bytes are not equal.

packages/wallet-account-signer/src/wallet-account-signer.ts
```diff
@@ -0,0 +1,118 @@
...
+ * if (isMessageSigner(signer)) {
+ *     const [signedMessage] = await signer.modifyAndSignMessages([message]);
+ * }
+ * ```
+ *
+ * @see {@link createTransactionSignerFromWalletAccount}
+ * @see {@link createTransactionSendingSignerFromWalletAccount}
+ * @see {@link createMessageSignerFromWalletAccount}
+ */
+export function createSignerFromWalletAccount<TWalletAccount extends UiWalletAccount>(
```
lorisleiva, line 68: Nice consistency with `createSignerFromKeyPair`! 👌

packages/wallet-account-signer/src/__tests__/wallet-account-transaction-signer-test.ts
```diff
@@ -843,10 +843,8 @@ describe('createSignerFromWalletAccount', () => {
         jest.mocked(getCompiledTransactionMessageDecoder).mockReturnValue({
             decode: jest
                 .fn()
-                .mockReturnValueOnce({ lifetimeToken: 'changed-1' }) // tx1 token check
-                .mockReturnValueOnce({ lifetimeToken: 'changed-1' }) // tx1 new lifetime fetch
-                .mockReturnValueOnce({ lifetimeToken: 'changed-2' }) // tx2 token check
-                .mockReturnValueOnce({ lifetimeToken: 'changed-2' }), // tx2 new lifetime fetch
+                .mockReturnValueOnce({ lifetimeToken: 'changed-1' }) // tx1 decode lifetime
+                .mockReturnValueOnce({ lifetimeToken: 'changed-2' }), // tx2 decode lifetime
```
lorisleiva, line 847: Nice, that's a good indication that it was a good change.

mcintyre94:
Awesome, merging! Thanks again @dpsi9! :)

PR #1371, opened 2026-02-13
Update the transaction encoder to encode v1 transactions

mcintyre94:
#### Problem

We need to be able to encode v1 transactions. They encode with `messageBytes` before signatures and without a prefix on the signatures array.

#### Summary of Changes

- Add a new function that inspects the message bytes to identify the transaction encoding shape. If the highest bit is 0 then it's a legacy transaction. Otherwise we mask the highest bit to get the version, which may be 0 or 1
   - For legacy and 0, we encode with the current structure, signatures first
   - For v1, we encode with message first

- Add a new function to encode with the message first. This:
  - Encodes message bytes
  - Reads the number of signatures, which is the 2nd byte of the message. We know that the first byte is a transaction version here
  - Encodes signatures using the encoder from `getSignaturesEncoderWithLength`

Note: I don't think it makes sense to add changesets to these intermediary steps, and we should just add one at the end to document v1 transactions. I think it'll be confusing to have changesets about v1 transactions before we fully support them.


mcintyre94:
* **#1374** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1374?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1373** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1373?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1371** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1371?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1371?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1372** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1372?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/transactions/src/codecs/transaction-codec.ts
```diff
@@ -1,41 +1,113 @@
...
+
+const VERSION_FLAG_MASK = 0x80;
+
+function getEncodeShapeForMessageBytes(messageBytes: ReadonlyUint8Array): EncodeShape {
+    const firstByte = messageBytes[0];
+    if ((firstByte & 0x80) === 0) {
+        // Legacy transaction: encode with signatures first
+        return 'signaturesFirst';
+    }
+    const version = firstByte & ~VERSION_FLAG_MASK;
```
lorisleiva: nit: This is looking a bit cryptic, even for bit operations. Maybe we could be a bit more explicit about the bits of the first byte we are interested in. For instance:

```ts
const LEGACY_VERSION_FLAG_MASK = 0b10000000;
const VERSION_FLAG_MASK = 0b01111111;

function getEncodeShapeForMessageBytes(messageBytes: ReadonlyUint8Array): EncodeShape {
    const firstByte = messageBytes[0];
    if ((firstByte & LEGACY_VERSION_FLAG_MASK) === 0) {
        // Legacy transaction: encode with signatures first
        return 'signaturesFirst';
    }
    const version = firstByte & VERSION_FLAG_MASK;
    // ...
}
```

packages/transactions/src/codecs/transaction-codec.ts
```diff
@@ -68,7 +140,7 @@ export function getTransactionDecoder(): VariableSizeDecoder<Transaction> {
  * @see {@link getTransactionDecoder}
  * @see {@link getTransactionEncoder}
  */
-export function getTransactionCodec(): VariableSizeCodec<Transaction> {
+export function getTransactionCodec(): Codec<Transaction> {
```
lorisleiva: Just checking if this is correct. I can see both `getTransactionEncoderWithSignaturesFirst` and `getTransactionEncoderWithMessageFirst` as `VariableSizeEncoders`.

packages/transactions/src/codecs/transaction-codec.ts
```diff
@@ -1,41 +1,113 @@
...
+            messageBytes,
+        });
+    }
+    return messageBytes[1]; // second byte
+}
+
+function getTransactionEncoderWithMessageFirst(): VariableSizeEncoder<Transaction> {
+    const bytesEncoder = getBytesEncoder();
+
+    return createEncoder({
```
lorisleiva, line 91: Interestingly, I'm pretty sure you could achieve the same using the `offsetEncoder` function.

```ts
const signatureSizeEncoder = offsetEncoder(getU8Encoder, {
    preOffset: () => 1, // Jump to absolute offset 1.
    postOffset: ({ preOffset }) => preOffset, // Return to original position.
});

const signatureEncoder = transformEncoder(
    getArrayEncoder(fixEncoderSize(getBytesEncoder(), 64), { size: signatureSizeEncoder }),
    getSignaturesToEncode,
)

return getStructEncoder([
    ['messageBytes', getBytesEncoder()],
    ['signatures', signatureEncoder],
]);
```

Probably a more confusing solution so I'm happy to keep the current one but just wanted to give a shout-out to this underuse codec primitive haha.

packages/transactions/src/codecs/transaction-codec.ts
```diff
@@ -1,41 +1,113 @@
...
+    SolanaError,
+} from '@solana/errors';
 import { SignatureBytes } from '@solana/keys';
 import { getTransactionVersionDecoder } from '@solana/transaction-messages';
 
 import { SignaturesMap, Transaction, TransactionMessageBytes } from '../transaction';
-import { getSignaturesEncoderWithSizePrefix } from './signatures-encoder';
+import { getSignaturesEncoderWithLength, getSignaturesEncoderWithSizePrefix } from './signatures-encoder';
+
+type EncodeShape = 'messageFirst' | 'signaturesFirst';
```
lorisleiva: nit: I'm guessing this will be used by the decoder as well so probably worth calling it something like `EnvelopeShape` so it can be used by both?

packages/transactions/src/codecs/transaction-codec.ts
mcintyre94: Oops nope this should be `VariableSize` too after the change to predicate.

mcintyre94:
### Merge activity

* **Feb 16, 3:31 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1371).
* **Feb 16, 3:33 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1371) rebased this pull request as part of a merge.
* **Feb 16, 3:41 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1371).

PR #1373, opened 2026-02-16
Add v1 to TransactionVersion, but not to transaction message functions

mcintyre94:
#### Summary of Changes

This commit adds v1 to the `TransactionVersion` type, and to the relevant codec and its tests. It is not added to the `TransactionMessage` type, and is explicitly disallowed by the `createTransactionMessage` type.

This allows us to use the `TransactionVersion` decoder in the `Transaction` decoder, while not yet having support for the v1 functionality in the rest of transaction messages.

Note that this also makes v1 valid in `rpc-types`, which means it's a valid transaction version to use as `maxSupportedTransactionVersion`.

PR #1372, opened 2026-02-13
Add an overload to predicate, return Variable if both inputs are

mcintyre94:
#### Problem

The `getPredicateEncoder` (etc) functions take a predicate and two encoder, and choose between them.

Currently if both inputs are `FixedSize` then we return `FixedSize`, otherwise we return `Encoder`

This loses some specificity if both inputs are `VariableSize`. In this case we know we are choosing a `VariableSize` but we just return `Encoder`

#### Summary of Changes

Add an overload for each function so that if both inputs are `VariableSize` then so is the returned encoder

This is useful for the transaction encoder where we will be switching between two variable size encoders. With this new overload, we will be able to keep the resulting transaction encoder typed as `VariableSize`. 

Note: I haven't added a changeset because we haven't published the new functions yet and the existing changeset is general enough not to need this specified.

PR #1374, opened 2026-02-16
Add support for decoding v1 Transaction bytes

mcintyre94:
#### Summary of Changes

This PR updates `getTransactionDecoder` to support decoding from v1 transaction bytes to a `Transaction` object.

As with encoding v1 transactions, we need to account for the different envelope shape for v1 transactions - they encode the message first, while legacy/v0 encode the signatures first.

v1 Transaction format SIMD: https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md 

We identify this case by comparing the first byte with the `0b10000000` mask. I've renamed this `SIGNATURE_COUNT_FLAG_MASK`, if `firstByte & SIGNATURE_COUNT_FLAG_MASK === 0`, then the first byte represents a signature count. This is the case for both legacy and v0 transactions, because they encode signatures first, with a size prefix.

If this is not the case, then we have a v1 transaction. We check the version byte and throw if it is not v1.

In the case that we are decoding with signatures first, we use the existing struct decoder (signatures, then message bytes). This gives us a `PartiallyDecodedTransaction`, which we use the existing `decodePartiallyDecodedTransaction` function (renamed to `decodePartiallyDecodedLegacyOrV0Transaction`) to decode to `Transaction`. This is existing logic - it transforms the signatures to a `Map<Address, Signature | null>` - we call this `SignaturesMap`. This requires decoding the first bytes of the message bytes to find the number of signatures and the first static addresses (which are the signers).

In the case that we are decoding with the message first, we have a new decoder to get to `PartiallyDecodedTransaction`. In this case, we read the number of signatures `n` from the second byte (first is tx version), and we treat the last `64n` bytes as the signatures which we read as an array of addresses with no prefix. We read the rest of the transaction as `messageBytes`. This is an optimisation - there is enough information in the decoded `messageBytes` to read precisely to where the signatures begin. But we would need to decode basically the whole message to do so. I think this is equivalent to the existing behaviour for signatures-first though, because currently we read signatures and then use `getBytesDecoder` with no size to read the rest of the input as `messageBytes`.

We also have new logic to decode a partially decoded v1 transaction. The idea is the same - we want to find the first signer addresses and build the `SignaturesMap`. In this case we know that the first byte is transaction version and the second is number of signatures, we don't need to handle the conditional version. We also know the offset where the addresses begin, as this is the first non-fixed size element in the new message structure. So we simply read the known number of signer addresses from this known offset.

I've also improved the unit tests for transaction codecs. Previously they all used a v0 shaped transaction. I've copied all the v0 tests into legacy tests (which just remove the leading 128 byte). I've also added all new v1 tests, with transactions containing message bytes matching the (relevant leading parts of) v1 transaction spec, and encoded with the message before the signatures.


mcintyre94:
* **#1374** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1374?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1374?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1373** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1373?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1371** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1371?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1372** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1372?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/transactions/src/codecs/transaction-codec.ts
```diff
@@ -34,12 +40,15 @@ import { getSignaturesEncoderWithLength, getSignaturesEncoderWithSizePrefix } fr
 
 type EnvelopeShape = 'messageFirst' | 'signaturesFirst';
 
-const LEGACY_VERSION_FLAG_MASK = 0b10000000;
+const SIGNATURE_COUNT_FLAG_MASK = 0b10000000;
```
mcintyre94, line 43: I renamed this because:

- when we encode messageBytes, it flags if the first byte is a signature count, and therefore we know it's a legacy tx
- when we decode a transaction, it flags if the first byte is a signature count, and therefore we know it's encoded with signatures first (but not whether it's legacy or v0).

So it works for both cases, but isn't always a legacy version.

packages/errors/src/messages.ts
```diff
@@ -760,4 +764,12 @@ export const SolanaErrorMessages: Readonly<{
...
     [SOLANA_ERROR__TRANSACTION__MALFORMED_MESSAGE_BYTES]: 'Transaction message bytes are malformed: $messageBytes',
+    [SOLANA_ERROR__TRANSACTION__CANNOT_ENCODE_WITH_EMPTY_MESSAGE_BYTES]:
+        'Transaction message bytes are empty, so the transaction cannot be encoded',
+    [SOLANA_ERROR__TRANSACTION__CANNOT_DECODE_EMPTY_TRANSACTION_BYTES]:
+        'Transaction bytes are empty, so no transaction can be decoded',
+    [SOLANA_ERROR__TRANSACTION__VERSION_ZERO_MUST_BE_ENCODED_WITH_SIGNATURES_FIRST]:
+        'Transaction version 0 must be encoded with signatures first. This transaction was encoded with first byte $firstByte, which is expected to be a signature count for v0 transactions.',
+    [SOLANA_ERROR__TRANSACTION__SIGNATURE_COUNT_TOO_HIGH_FOR_TRANSACTION_BYTES]:
+        'The provided transaction bytes encode that there should be $numExpectedSignatures signatures, but the bytes are not long enough to contain a transaction message with this many signatures. The provided bytes are $transactionBytesLength bytes long.',
```
lorisleiva:
```suggestion
        'The provided transaction bytes expect that there should be $numExpectedSignatures signatures, but the bytes are not long enough to contain a transaction message with this many signatures. The provided bytes are $transactionBytesLength bytes long.',
```

packages/transactions/src/codecs/transaction-codec.ts
lorisleiva: Yeah that's a much better name. 👌

packages/transactions/src/codecs/transaction-codec.ts
```diff
@@ -34,12 +40,15 @@ import { getSignaturesEncoderWithLength, getSignaturesEncoderWithSizePrefix } fr
...
 const VERSION_FLAG_MASK = 0b01111111;
 
 function getEncodeShapeForMessageBytes(messageBytes: ReadonlyUint8Array): EnvelopeShape {
```
lorisleiva: I might be wrong (and I should have mentioned that on a previous review) but I'm just thinking, are we not duplicating some logic here with the version codecs? Could this not be simplified by something like:

```ts
const version = getTransactionVersionDecoder().decode(messageBytes);
return version === 1 ? 'messageFirst' : 'signatureFirst';
```

nit: Should we be more explicit and rename this to `getEnvelopeShapeFromMessageBytes`?

packages/transactions/src/codecs/transaction-codec.ts
```diff
@@ -57,6 +66,32 @@ function getEncodeShapeForMessageBytes(messageBytes: ReadonlyUint8Array): Envelo
     });
 }
 
+function getDecodeShapeForBytes(transactionBytes: ReadonlyUint8Array): EnvelopeShape {
```
lorisleiva: In contrast, here this new logic is required because we don't know where the first message byte is and we need to identify the first transaction byte. So we couldn't just defer to the transaction version codecs.

nit: Should we be more explicit and rename this to `getEnvelopeShapeFromTransactionBytes`?

packages/transactions/src/codecs/transaction-codec.ts
```diff
@@ -126,15 +161,71 @@ function getTransactionEncoderWithMessageFirst(): VariableSizeEncoder<Transactio
...
+                0,
+            );
+
+            return [
+                {
+                    messageBytes: messageBytes as unknown as TransactionMessageBytes,
+                    signatures,
+                },
+                offset + transactionBytes.length,
+            ];
```
lorisleiva:
```suggestion
            const [signatures, finalOffset] = getArrayDecoder(fixDecoderSize(getBytesDecoder(), 64), { size: signatureCount }).read(
                signaturesBytes,
                messageBytesLength,
            );

            return [
                {
                    messageBytes: messageBytes as unknown as TransactionMessageBytes,
                    signatures,
                },
                finalOffset,
            ];
```

packages/transactions/src/codecs/transaction-codec.ts
```diff
@@ -126,15 +161,71 @@ function getTransactionEncoderWithMessageFirst(): VariableSizeEncoder<Transactio
...
     );
 }
 
+function getPartiallyDecodedTransactionDecoderWithMessageFirst(): VariableSizeDecoder<PartiallyDecodedTransaction> {
+    return createDecoder({
+        read(bytes, offset) {
+            const transactionBytes = bytes.subarray(offset);
+
+            // 1. Message comes first, so read signature count from message bytes
+            const signatureCount = getSignatureCountForVersionedOrThrow(transactionBytes);
```
lorisleiva: nit: It would be nice if `getSignatureCountForVersionedOrThrow` could support an offset argument so we didn't have to use `subarray` which create a new Uint8Array when there's no need for it.

packages/transactions/src/codecs/transaction-codec.ts
mcintyre94: Actually yep fair point, that's much cleaner.

mcintyre94:
### Merge activity

* **Feb 17, 12:22 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1374).
* **Feb 17, 12:23 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1374) rebased this pull request as part of a merge.
* **Feb 17, 12:34 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1374).

PR #1375, opened 2026-02-17
Add the v1 TransactionMessageconfig and functions to encode it to a mask and value array

mcintyre94:
#### Summary of Changes

The Transaction v1 format introduces a `TransactionConfig`, initially to replace the current compute budget program instructions. See from the SIMD: https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md#transactionconfigmask-1

It encodes (in order, all optional):
- total priority fee in Lamports (u64)
- compute unit limit (u32)
- loaded account data size limit (u32)
- heap size (u32)

This PR adds this type, and functions to encode it for transaction message encoding.

In the [Transaction v1 encoding spec](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md#transaction-v1-specification) the config is encoded in 2 separate places.

First, it's encoded as a 32 bit mask which identifies which values are set. The lowest 2 bits are both set if priority fee is set. The 3rd bit is set if compute unit limit is set, the 4th bit for loaded account data size, and the 5th bit for heap size. The rest of the bits are unset and reserved for future config values.

Second, the actual values are encoded into an array of u32 values. Only the values actually set are stored in this array, so it can be empty if none are set. Its length (in u32) is equivalent to the number of set bits in the mask. For the priority fee we use 8 bytes (2 u32 values), for all other current fields we use 4 bytes.

For reference, the `solana-sdk` code for encoding the mask: https://github.com/anza-xyz/solana-sdk/blob/1cf4c970e5dedbbf5cb6c2081d8f40e73f053c74/message/src/versions/v1/config.rs#L87

And for encoding the config values: https://github.com/anza-xyz/solana-sdk/blob/1cf4c970e5dedbbf5cb6c2081d8f40e73f053c74/message/src/versions/v1/message.rs#L606

### DIFF
--- a/packages/transaction-messages/src/codecs/config.ts
+++ b/packages/transaction-messages/src/codecs/config.ts
@@ -1,0 +1,62 @@
+import { createEncoder, FixedSizeEncoder, transformEncoder, VariableSizeEncoder } from '@solana/codecs-core';
+import { getU32Encoder, getU64Encoder } from '@solana/codecs-numbers';
+
+import { TransactionConfig as TransactionConfig } from '../transaction-config';
+
+const PRIORITY_FEE_LAMPORTS_BIT_MASK = 0b11;
+const COMPUTE_UNIT_LIMIT_BIT_MASK = 0b100;
+const LOADED_ACCOUNTS_DATA_SIZE_LIMIT_BIT_MASK = 0b1000;
+const HEAP_SIZE_BIT_MASK = 0b10000;
+
+/**
+ * Encode a {@link TransactionMessageConfig} into a 4 byte mask, where the lowest bits indicate which fields are set
+ * @returns An Encoder for {@link TransactionMessageConfig}
+ */
+export function getTransactionConfigMaskEncoder(): FixedSizeEncoder<TransactionConfig, 4> {
+    return transformEncoder(getU32Encoder(), config => {
+        let mask = 0;
+        // Set the lowest 2 bits for priority fee lamports
+        if (config.priorityFeeLamports !== undefined) mask |= PRIORITY_FEE_LAMPORTS_BIT_MASK;
+        // Set the 3rd lowest bit for compute unit limit
+        if (config.computeUnitLimit !== undefined) mask |= COMPUTE_UNIT_LIMIT_BIT_MASK;
+        // Set the 4th lowest bit for loaded accounts data size limit
+        if (config.loadedAccountsDataSizeLimit !== undefined) mask |= LOADED_ACCOUNTS_DATA_SIZE_LIMIT_BIT_MASK;
+        // Set the 5th lowest bit for heap size
+        if (config.heapSize !== undefined) mask |= HEAP_SIZE_BIT_MASK;
+        return mask;
+    });
+}
+
+/**
+ * Encode a {@link TransactionMessageConfig} into a variable length byte array, where the fields set are encoded based on their data size.
+ * @returns An Encoder for {@link TransactionMessageConfig}
+ */
+export function getTransactionConfigValuesEncoder(): VariableSizeEncoder<TransactionConfig> {
+    return createEncoder<TransactionConfig>({
+        getSizeFromValue(config) {
+            let size = 0;
+            // Lamports is 8 bytes, the rest are 4 bytes each
+            if (config.priorityFeeLamports !== undefined) size += 8;
+            if (config.computeUnitLimit !== undefined) size += 4;
+            if (config.loadedAccountsDataSizeLimit !== undefined) size += 4;
+            if (config.heapSize !== undefined) size += 4;
+            return size;
+        },
+        write(config, bytes, offset) {
+            let nextOffset = offset;
+            if (config.priorityFeeLamports !== undefined) {
+                nextOffset = getU64Encoder().write(config.priorityFeeLamports, bytes, nextOffset);
+            }
+            if (config.computeUnitLimit !== undefined) {
+                nextOffset = getU32Encoder().write(BigInt(config.computeUnitLimit), bytes, nextOffset);
+            }
+            if (config.loadedAccountsDataSizeLimit !== undefined) {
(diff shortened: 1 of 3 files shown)


mcintyre94:
* **#1398** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1398?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1394** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1394?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1397** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1397?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1396** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1396?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1386** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1386?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1385** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1385?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1384** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1384?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1376** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1376?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1375** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1375?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1375?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/transaction-messages/src/codecs/config.ts
```diff
@@ -0,0 +1,54 @@
...
+import { getBitArrayEncoder } from '@solana/codecs-data-structures';
+import { getU32Encoder, getU64Encoder } from '@solana/codecs-numbers';
+
+import { TransactionMessageConfig } from '../config';
+
+export function getConfigMaskEncoder(): FixedSizeEncoder<TransactionMessageConfig, 4> {
+    return createEncoder<TransactionMessageConfig, 4>({
+        fixedSize: 4,
+        write(config, bytes, offset) {
+            return fixEncoderSize(getBitArrayEncoder(1, { backward: true }), 4).write(
```
lorisleiva: As discussed offline, this could use a `transformEncoder` instead.

packages/transaction-messages/src/codecs/config.ts
```diff
@@ -0,0 +1,54 @@
+import { createEncoder, FixedSizeEncoder, fixEncoderSize, VariableSizeEncoder } from '@solana/codecs-core';
+import { getBitArrayEncoder } from '@solana/codecs-data-structures';
+import { getU32Encoder, getU64Encoder } from '@solana/codecs-numbers';
+
+import { TransactionMessageConfig } from '../config';
+
+export function getConfigMaskEncoder(): FixedSizeEncoder<TransactionMessageConfig, 4> {
```
lorisleiva: nit

```suggestion
export function getTransactionConfigMaskEncoder(): FixedSizeEncoder<TransactionMessageConfig, 4> {
```

packages/transaction-messages/src/codecs/config.ts
```diff
@@ -0,0 +1,54 @@
...
+                    config.heapSize !== undefined,
+                ],
+                bytes,
+                offset,
+            );
+        },
+    });
+}
+
+export function getConfigValuesEncoder(): VariableSizeEncoder<TransactionMessageConfig> {
```
lorisleiva: nit

```suggestion
export function getTransactionConfigValuesEncoder(): VariableSizeEncoder<TransactionMessageConfig> {
```

packages/transaction-messages/src/codecs/config.ts
mcintyre94: Agree with this - discussed offline, let's rename this `TransactionConfig` consistently.

packages/transaction-messages/src/codecs/config.ts
mcintyre94: I went back and forth a bit on this but I think this is better without using the bit array encoder, because I don't think this would correctly encode 9 bits (if we hypothetically added more config fields) as a u32 the way we'd expect. I'm kinda abusing `fixEncoder` + backward bit array to make a little-endian u32 encoder when we already have a proper implementation of that. So I think it might be more confusing in future even if it's clearer now.

I think it's better to write this as a transformed u32 encoder:

```ts
export function getTransactionConfigMaskEncoder(): FixedSizeEncoder<TransactionConfig, 4> {
    return transformEncoder(getU32Encoder(), config => {
        let mask = 0;
        // set lowest 2 bits
        if (config.priorityFeeLamports !== undefined) mask |= 0b11;
        // set 3rd lowest bit
        if (config.computeUnitLimit !== undefined) mask |= 0b100;
        // set 4th lowest bit
        if (config.loadedAccountsDataSizeLimit !== undefined) mask |= 0b1000;
        // set 5th lowest bit
        if (config.heapSize !== undefined) mask |= 0b10000;
        return mask;
    })
}
```

It also better mirrors [the Rust version](https://github.com/febo/solana-sdk/blob/ce6ae519cce00219c45c6263e5971656bc93148f/message/src/versions/v1/config.rs#L87) which will again make future changes easier to verify.

mcintyre94:
### Merge activity

* **Feb 20, 5:40 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1375).
* **Feb 20, 5:40 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1375).

PR #1391, opened 2026-02-20
Handle negative offsets when considering whether or not to slice/clone an array

### LINKED ISSUE #1119: Handle negative offsets when considering whether or not to slice/clone an array
## Overview

For efficiency, we often check to see if someone is trying to operate over a subset of an array before slicing/cloning it:

```
if (offset === 0 && length === bytes.byteLength) {
    // Operate over the original.
} else {
    // Slice the original and operate over the copy.
}
```

This is insufficient to catch the case where `offset === -bytes.byteLength` and `length === bytes.byteLength`. That negative offset is equivalent to zero.

## Steps to reproduce

These should produce identical outcomes:

```
const bytes = new Uint8Array([1,2,3]);
const thingA = doThing(bytes, 0 /* offset */, 3 /* length */);  // No slice
const thingB = doThing(bytes, -3 /* offset */, 3 /* length */);  // No slice
// thingA === thingB should be true
```

Consider code like this:

https://github.com/anza-xyz/kit/blob/238082eb602eb5f3b93a4679e9a5da5f93516117/packages/codecs-core/src/add-codec-sentinel.ts#L93

That code should copy neither when offset is zero, nor when offset is ‘negative the entire length of the buffer.’

holps-7:
**Summary of Changes**

prevent unnecessary array slicing for negative buffer length offsets

Fixes #1119

### DIFF
--- a/packages/codecs-core/src/__tests__/add-codec-sentinel-test.ts
+++ b/packages/codecs-core/src/__tests__/add-codec-sentinel-test.ts
@@ -77,11 +77,19 @@ describe('addCodecSentinel', () => {
         expect(codec.getSizeFromValue('helloworld')).toBe(12);
     });
 
-    it('does not slice the byte array when offset equals negative byteLength', () => {
+    it('is correct when offset equals negative byteLength', () => {
         const mockCodec = getMockCodec();
         mockCodec.read.mockReturnValue(['helloworld', 10]);
         const codec = addCodecSentinel(mockCodec, b('ff'));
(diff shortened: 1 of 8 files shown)


holps-7:
Hey @steveluscher can you review this PR?

holps-7:
hey @steveluscher @lorisleiva requesting a review for this PR.....

holps-7:
Thanks @mcintyre94 for the review and merging the PR, I totally missed the cases where negative offset is >= the length.

PR #1401, opened 2026-02-23
Add a codec for legacy messages, extracted from the current message codec

mcintyre94:
#### Summary of Changes

As part of adding v1 transactions, I'm adding separate codecs for v0 and legacy, instead of the combined one we currently have.

This PR adds a new codec in codecs/legacy/message. It includes just the legacy fields - the header, accounts, lifetime token, instructions.

Tests are based on the existing ones for `getCompiledMessageVersioned{}`, but expanded the coverage a bit.

Not used yet, but will replace the legacy message parts of `getCompiledMessageVersioned{}`.

PR #1407, opened 2026-02-24
Add a legacy specific compileTransactionMessage

mcintyre94:
#### Problem

Currently when we encode a transaction, we go from TransactionMessage -> CompiledTransactionMessage -> Bytes. Most of the logic is in the compile step, and the encoders are very simple. `CompiledTransactionMessage` mirrors the wire format.

In my approach to v1 transactions so far I've been treating the compile step as a thin layer, and adding the logic to the encoders. I think this was a mistake, and makes both the field encoders and the overall message v1 encoder overly complex.

Instead we should have a `CompiledTransactionMessageV1` type that matches the wire shape of v1 transactions, with plain JS logic to map from `TransactionMessage` to it.

This means that similarly to what I did with codecs, we should refactor compile to be by version so it's easier/cleaner to add v1. 

#### Summary of Changes

This PR extracts the logic for compiling legacy transactions. It's all copied from the current compile/ directory, but some changes are made.

- `accounts` has all lookup table code stripped from it. The existing version handles lookup tables and not. We could choose not to do this and leave it as-is, but since I think lookup tables are probably going to be v0 only and v1 will not use them, I think it's nicer to have this version that doesn't include them. This does duplicate complex logic though, so happy to revert this.
- `header` is just moved from compile/ with no changes. This will be used to compile all 3 versions so far.
- `instructions` is copied from compile/, but uses the simplified `OrderedAccounts` from the new `legacy/accounts`. This does mean we will have the same file duplicated in v0 (just with imports from its own `accounts`). This is a downside of copying `accounts` into the version directories. It is a small file though. 
- `lifetime-token` is moved from compile/ with no changes.
- `message` is a slightly simplified version of the existing `compileTransactionMessage`, with the version fixed and no lookup table fields. The idea is that the root `compileTransactionMessage` will just call out to this when the transaction message is legacy.

Note that a new `message-types` is used to export types currently in `compile/message` that are useful for writing these versioned `compileTransactionMessage` functions, but that we don't want to export from the package. 

This leaves for v0 (next PR):
- The original `accounts`, which handles lookup accounts. This will move unchanged into `v0/`
- `address-table-lookups` will move unchanged into `v0/` 
- `instructions` will need to be copied into `v0/` and use the `OrderedAccounts` from `v0/accounts`.
- `static-accounts`, which gets from `OrderedAccounts` just the static addresses. For legacy this is just a trivial map (all are static), so I didn't copy this in. It will move to v0/ because there it handles lookup addresses. 

Tests are all copied/moved from the current compile/ tests, and modified where necessary to strip lookup tables.

### DIFF
--- a/packages/transaction-messages/src/codecs/legacy/header.ts
+++ b/packages/transaction-messages/src/codecs/legacy/header.ts
@@ -2,7 +2,7 @@ import { FixedSizeCodec, FixedSizeDecoder, FixedSizeEncoder } from '@solana/code
 import { getStructCodec, getStructDecoder, getStructEncoder } from '@solana/codecs-data-structures';
 import { getU8Codec, getU8Decoder, getU8Encoder } from '@solana/codecs-numbers';
 
-import { getCompiledMessageHeader } from '../../compile/header';
+import { getCompiledMessageHeader } from '../../compile/legacy/header';
 
 type MessageHeader = ReturnType<typeof getCompiledMessageHeader>;
 
--- a/packages/transaction-messages/src/codecs/legacy/lifetime-token.ts
+++ b/packages/transaction-messages/src/codecs/legacy/lifetime-token.ts
@@ -8,7 +8,7 @@ import {
 import { getNullableEncoder } from '@solana/codecs-data-structures';
 import { getBase58Decoder, getBase58Encoder } from '@solana/codecs-strings';
 
-import { getCompiledLifetimeToken } from '../../compile/lifetime-token';
+import { getCompiledLifetimeToken } from '../../compile/legacy/lifetime-token';
 
 type LifetimeToken = ReturnType<typeof getCompiledLifetimeToken>;
 
--- a/packages/transaction-messages/src/compile/__tests__/static-accounts-test.ts
+++ b/packages/transaction-messages/src/compile/__tests__/static-accounts-test.ts
@@ -2,7 +2,7 @@ import { Address } from '@solana/addresses';
 import { AccountRole } from '@solana/instructions';
 
 import { OrderedAccounts } from '../../compile/accounts';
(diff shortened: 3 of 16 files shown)


mcintyre94:
* **#1409** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1409?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1408** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1408?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1407** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1407?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1407?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1406** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1406?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1401** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1401?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

mcintyre94:
Thanks, definitely good to get a second opinion here! If a future transaction version adds lookup tables back then I'll be sad and it'll probably make sense to re-think this, but I think that's probably unlikely! 🤞

mcintyre94:
### Merge activity

* **Feb 26, 11:53 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1407).
* **Feb 26, 11:53 AM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1407).

PR #1415, opened 2026-02-26
Add empty @solana/wallet-account-signer package

mcintyre94:
#### Problem

Our functions to convert `UiWalletAccount` to `Signer` introduce dependencies on `@solana/wallet-standard-` and `@wallet-standard/` packages. The current PR (https://github.com/anza-xyz/kit/pull/1368) adds these to the signers package, which I suggested in the issue. We've decided to move these into a separate package, to avoid adding these dependencies to `@solana/kit`. Therefore this package is not exported from the `@solana/kit` umbrella package.

#### Summary of Changes

This PR adds an empty new package `@solana/wallet-account-signer` which we can move these new functions to.


mcintyre94:
Will merge this after we have trusted publisher set up for the new package, changed to draft for now

PR #1420, opened 2026-03-02
Enable pattern match encoder/codec to use type narrowing

mcintyre94:
#### Problem

I want to use the pattern matching codec for the `CompiledTransactionMessage` codec

We have (simplifying) a legacy codec that takes `LegacyCompiledTransactionMessage` as input, and a v0 one that takes `v0CompiledTransactionMessage` as input

Currently we can't write:

```ts
getPatternMatchEncoder<CompiledTransactionMessage>([
  [(m) => m.version === 'legacy', getLegacyCompiledTransactionMessageEncoder()],
  [(m) => m.version === 0, getV0CompiledTransactionMessageEncoder()]
])
```

Because all the nested encoders are typed as `Encoder<TFrom>` Our version predicate narrows the type, but the encoder type is unaware of that.

#### Summary of Changes

This PR allows narrowing the type in `getPatternMatchEncoder`. Now you can pass a pattern like this:

```ts
[(value: TFrom) => value is TNarrowed, Encoder<TNarrowed>]
```

In other words, if you use a predicate function that narrows the type of the input value, you can use an encoder for that narrowed type.

The previous case with a predicate just returning a boolean remains supported, and still requires an encoder for `TFrom`. 

This means that (in the next PR), we can use the pattern match encoder for `CompiledTransactionMessage`.

PR #1419, opened 2026-03-02
Add support for compiling v1 transaction messages

mcintyre94:
#### Summary of Changes

This PR adds the `V1CompiledTransactionMessage` type and a function to compile from `TransactionMessage` to it.

- Remove `instructions` from `BaseCompiledTransactionMessage`, and move it to legacy and v0
- `configMask`: compile the bit mask for provided config values.
- `configValues`: compile to a list of tagged u32/u64 values, ready to encode with the respective codecs
- Compile instructions to `InstructionHeader` and `InstructionPayload`
- Includes `numInstructions` and `numStaticAccounts`

Note that most of the tests are copied from the original implementation of this logic in encoders.

These encoders can now be much simpler, with most of the logic moved to this compile stage.

PR #1421, opened 2026-03-02
Refactor CompiledMessageTransaction codec to use pattern match encoder/decoder

mcintyre94:
#### Summary of Changes

Now that we have type narrowing in the pattern matching codecs, we can use it here. This lets us strip out all the rest of the code from this codec, which is now in the individual versioned ones.

Note that the decoder is a tiny bit less efficient as it now reads the transaction version twice for the v0 case, once here and then once again in the v0 message decoder. We ignore the offset returned by the transaction version decoder here. We could remove the version from the v0 decoder and pass only the bytes after that from here, and that would work for future versions as well. But I think it's nice to have the message fully contained in the versioned message codecs, and worth this tiny inefficiency.

### DIFF
--- a/packages/transaction-messages/src/codecs/message.ts
+++ b/packages/transaction-messages/src/codecs/message.ts
@@ -1,83 +1,19 @@
-import { getAddressDecoder, getAddressEncoder } from '@solana/addresses';
 import {
     combineCodec,
-    createEncoder,
-    Decoder,
-    transformDecoder,
-    transformEncoder,
+    createDecoder,
     VariableSizeCodec,
     VariableSizeDecoder,
     VariableSizeEncoder,
 } from '@solana/codecs-core';
-import { getArrayDecoder, getArrayEncoder, getStructDecoder, getStructEncoder } from '@solana/codecs-data-structures';
-import { getShortU16Decoder, getShortU16Encoder } from '@solana/codecs-numbers';
+import { getPatternMatchDecoder, getPatternMatchEncoder } from '@solana/codecs-data-structures';
 
 import { CompiledTransactionMessage, CompiledTransactionMessageWithLifetime } from '../compile/message';
-import { getCompiledAddressTableLookups } from '../compile/v0/address-table-lookups';
-import { getMessageHeaderDecoder, getMessageHeaderEncoder } from './legacy/header';
-import { getInstructionDecoder, getInstructionEncoder } from './legacy/instruction';
-import { getLifetimeTokenDecoder, getLifetimeTokenEncoder } from './legacy/lifetime-token';
-import { getTransactionVersionDecoder, getTransactionVersionEncoder } from './transaction-version';
-import { getAddressTableLookupDecoder, getAddressTableLookupEncoder } from './v0/address-table-lookup';
-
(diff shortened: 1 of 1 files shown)


mcintyre94:
* **#1421** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1421?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1419](https://github.com/anza-xyz/kit/pull/1419) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1419?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1422](https://github.com/anza-xyz/kit/pull/1422) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1422?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1421?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1420** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1420?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/transaction-messages/src/codecs/message.ts
```diff
@@ -117,20 +43,26 @@ export function getCompiledTransactionMessageEncoder(): VariableSizeEncoder<
 export function getCompiledTransactionMessageDecoder(): VariableSizeDecoder<
     CompiledTransactionMessage & CompiledTransactionMessageWithLifetime
 > {
-    return transformDecoder(
-        getStructDecoder(getPreludeStructDecoderTuple()) as VariableSizeDecoder<
-            CompiledTransactionMessage &
-                CompiledTransactionMessageWithLifetime & {
-                    addressTableLookups?: ReturnType<typeof getCompiledAddressTableLookups>;
-                }
```
lorisleiva, line 125: Is the fact that we're no longer using `& { addressTableLookups? }` in the return type gonna come back to bite us somewhere else in the codebase?

mcintyre94: I don't think so - I think this cast was probably just to enable the transform function to use `({ addressTableLookups, ...restOfMessage })`. The function itself has an unchanged return type:

```
VariableSizeDecoder<
    CompiledTransactionMessage & CompiledTransactionMessageWithLifetime
>
```

And that type has never guaranteed address table lookups. So I think this cast in `getStructDecoder` was always just implementation detail, and honestly pretty hacky.

mcintyre94:
### Merge activity

* **Mar 3, 12:35 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1421).
* **Mar 3, 12:37 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1421) rebased this pull request as part of a merge.
* **Mar 3, 12:44 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1421).

PR #1424, opened 2026-03-03
Add support for compiling to v1 in compileTransactionMessage

mcintyre94:
#### Summary of Changes

This PR adds a new function to compile to a `V1CompiledTransactionMessage`, and adds support for v1 in `compileTransactionMessage`.

- New code to compile transaction config, instruction payloads + headers
- Add v1 to `TransactionMessage`, with `TransactionConfig` field. 
  - Note that it is still not supported by `createTransactionMessage`. 
  - Rename `LegacyInstruction` to `InstructionWithoutLookupTables` in our definitions of `TransactionMessage` types, since we re-use it for v1.

PR #1422, opened 2026-03-02
Add the V1CompiledTransactionMessage type

mcintyre94:
#### Summary of Changes

This PR adds the `V1CompiledTransactionMessage` type, and adds it to the `CompiledTransactionMessage` union.

This requires removing `instructions` from the `BaseCompiledTransactionMessage` type, as it is not used for V1. This has some consequences for code that assumed that field was present.

In transactions the `getTransactionLifetimeConstraintFromCompiledTransactionMessage` gets a new branch to handle v1 transactions by parsing the first instruction header and instruction payload. 

This PR does not yet add compile/decompile functions for the new type, it just deals with the consequences of adding the new type.

PR #1431, opened 2026-03-04
Add a codec for v1 CompiledTransactionMessage

mcintyre94:
#### Summary of Changes

This PR adds the codec for v1 transaciton messages, and uses it in transaction-message codecs/message.ts

- Config values are encoded as either u32 or u64, according to their `kind`
- Config values are decoded sequentially using the mask to know which values are present
- InstructionHeader is trivial from its compiled form
- InstructionPayload is simple to encode, its two fields are just sequentially encoded
- InstructionPayload is decoded using the header to read the size of its fields 

- v1 message encoder is just a struct with the `V1CompiledTransactionMessage` format
- v1 message decoder is manual, because the sizes of later fields depend on things like `numInstructions` and `numStaticAccounts`. 

New tests are added for the v1 codec

I've also reworked the `codecs/message.ts` tests to test one example of each version, these are just copied from the version-specific tests. Previously we didn't have version-specific tests so this was a mixture of different tests and versions.

I decided not to use mocks because it becomes quite messy mocking multiple encoders with different read/write functions, and I think having a case for each version here provides the same test coverage anyway.

### DIFF
--- a/packages/transaction-messages/src/codecs/message.ts
+++ b/packages/transaction-messages/src/codecs/message.ts
@@ -1,19 +1,23 @@
 import {
     combineCodec,
     createDecoder,
+    transformEncoder,
     VariableSizeCodec,
     VariableSizeDecoder,
     VariableSizeEncoder,
 } from '@solana/codecs-core';
 import { getPatternMatchDecoder, getPatternMatchEncoder } from '@solana/codecs-data-structures';
+import { SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_NOT_SUPPORTED, SolanaError } from '@solana/errors';
 
 import { CompiledTransactionMessage, CompiledTransactionMessageWithLifetime } from '../compile/message';
+import { MAX_SUPPORTED_TRANSACTION_VERSION } from '../transaction-message';
 import {
     getMessageDecoder as getLegacyMessageDecoder,
     getMessageEncoder as getLegacyMessageEncoder,
 } from './legacy/message';
 import { getTransactionVersionDecoder } from './transaction-version';
 import { getMessageDecoder as getV0MessageDecoder, getMessageEncoder as getV0MessageEncoder } from './v0/message';
+import { getMessageDecoder as getV1MessageDecoder, getMessageEncoder as getV1MessageEncoder } from './v1/message';
 
 /**
  * Returns an encoder that you can use to encode a {@link CompiledTransactionMessage} to a byte
@@ -25,12 +29,24 @@ import { getMessageDecoder as getV0MessageDecoder, getMessageEncoder as getV0Mes
 export function getCompiledTransactionMessageEncoder(): VariableSizeEncoder<
     CompiledTransactionMessage | (CompiledTransactionMessage & CompiledTransactionMessageWithLifetime)
 > {
-    return getPatternMatchEncoder<
-        CompiledTransactionMessage | (CompiledTransactionMessage & CompiledTransactionMessageWithLifetime)
-    >([
-        [m => m.version === 'legacy', getLegacyMessageEncoder()],
-        [m => m.version === 0, getV0MessageEncoder()],
-    ]);
+    return transformEncoder(
+        getPatternMatchEncoder<
+            CompiledTransactionMessage | (CompiledTransactionMessage & CompiledTransactionMessageWithLifetime)
+        >([
+            [m => m.version === 'legacy', getLegacyMessageEncoder()],
+            [m => m.version === 0, getV0MessageEncoder()],
+            [m => m.version === 1, getV1MessageEncoder()],
+        ]),
+        value => {
+            // check version is valid before encoding, so we don't get the generic pattern match error
+            if (value.version !== 'legacy' && value.version > MAX_SUPPORTED_TRANSACTION_VERSION) {
+                throw new SolanaError(SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_NOT_SUPPORTED, {
+                    unsupportedVersion: value.version,
+                });
+            }
+            return value;
+        },
+    );
 }
 
 /**
@@ -43,23 +59,16 @@ export function getCompiledTransactionMessageEncoder(): VariableSizeEncoder<
 export function getCompiledTransactionMessageDecoder(): VariableSizeDecoder<
     CompiledTransactionMessage & CompiledTransactionMessageWithLifetime
 > {
+    type ReturnType = VariableSizeDecoder<CompiledTransactionMessage & CompiledTransactionMessageWithLifetime>;
+
     return createDecoder({
         read(bytes, offset) {
             const [version] = getTransactionVersionDecoder().read(bytes, offset);
 
             return getPatternMatchDecoder([
-                [
-                    () => version === 'legacy',
-                    getLegacyMessageDecoder() as VariableSizeDecoder<
-                        CompiledTransactionMessage & CompiledTransactionMessageWithLifetime
-                    >,
-                ],
-                [
-                    () => version === 0,
-                    getV0MessageDecoder() as VariableSizeDecoder<
-                        CompiledTransactionMessage & CompiledTransactionMessageWithLifetime
-                    >,
-                ],
+                [() => version === 'legacy', getLegacyMessageDecoder() as ReturnType],
+                [() => version === 0, getV0MessageDecoder() as ReturnType],
+                [() => version === 1, getV1MessageDecoder() as ReturnType],
             ]).read(bytes, offset);
         },
     });
--- a/packages/transaction-messages/src/codecs/v1/config.ts
+++ b/packages/transaction-messages/src/codecs/v1/config.ts
@@ -1,15 +1,35 @@
+import { transformDecoder, VariableSizeDecoder, VariableSizeEncoder } from '@solana/codecs-core';
 import {
-    createDecoder,
-    createEncoder,
-    FixedSizeEncoder,
-    transformEncoder,
-    VariableSizeDecoder,
-    VariableSizeEncoder,
-} from '@solana/codecs-core';
+    getArrayEncoder,
+    getPatternMatchEncoder,
+    getStructEncoder,
+    getTupleDecoder,
+    getUnitDecoder,
+} from '@solana/codecs-data-structures';
 import { getU32Decoder, getU32Encoder, getU64Decoder, getU64Encoder } from '@solana/codecs-numbers';
 import { SOLANA_ERROR__TRANSACTION__INVALID_CONFIG_MASK_PRIORITY_FEE_BITS, SolanaError } from '@solana/errors';
 
-import { TransactionConfig } from '../../transaction-config';
+import { CompiledTransactionConfigValue } from '../../compile/v1/config';
+
+/* TODO issue #1143 - we have a type error on `getPatternMatchEncoder` where it incorrectly
+ * types the return as FixedSizeEncoder when used with differently sized FixedSizEencoder
+ * inputs. For now we cast to VariableSizeEncoder, which is what the underlying union codec
+ * actually returns.
+ */
+function getCompiledTransactionConfigValueEncoder(): VariableSizeEncoder<CompiledTransactionConfigValue> {
+    return getPatternMatchEncoder<CompiledTransactionConfigValue>([
+        [value => value.kind === 'u32', getStructEncoder([['value', getU32Encoder()]])],
+        [value => value.kind === 'u64', getStructEncoder([['value', getU64Encoder()]])],
+    ]) as unknown as VariableSizeEncoder<CompiledTransactionConfigValue>;
+}
+
+/**
+ * Encode a {@link TransactionMessageConfig} into a variable length byte array, where the fields set are encoded based on their data size.
+ * @returns An Encoder for {@link TransactionMessageConfig}
+ */
+export function getCompiledTransactionConfigValuesEncoder(): VariableSizeEncoder<CompiledTransactionConfigValue[]> {
+    return getArrayEncoder(getCompiledTransactionConfigValueEncoder(), { size: 'remainder' });
+}
 
 const PRIORITY_FEE_LAMPORTS_BIT_MASK = 0b11;
 const COMPUTE_UNIT_LIMIT_BIT_MASK = 0b100;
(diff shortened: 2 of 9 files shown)


mcintyre94:
* **#1449** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1449?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1431** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1431?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1431?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1424** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1424?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1422** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1422?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1421** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1421?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1419](https://github.com/anza-xyz/kit/pull/1419) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1419?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1420** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1420?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/transaction-messages/src/codecs/v1/config.ts
```diff
@@ -1,80 +1,47 @@
...
-    return createEncoder<TransactionConfig>({
-        getSizeFromValue(config) {
-            let size = 0;
-            // Lamports is 8 bytes, the rest are 4 bytes each
-            if (config.priorityFeeLamports !== undefined) size += 8;
-            if (config.computeUnitLimit !== undefined) size += 4;
-            if (config.loadedAccountsDataSizeLimit !== undefined) size += 4;
-            if (config.heapSize !== undefined) size += 4;
-            return size;
+export function getConfigValuesEncoder(): VariableSizeEncoder<CompiledTransactionConfigValue[]> {
```
lorisleiva: I think we discussed this before but just wanna double check: these are not exposed to the package consumer right? Just ensure a generic name like `ConfigValues` doesn't get used for something so specific.

Perhaps we should still consider using more specific names regardless like `getCompiledTransactionConfigValueEncoder`.

packages/transaction-messages/src/codecs/v1/config.ts
```diff
@@ -1,80 +1,47 @@
...
+    return createEncoder({
```
lorisleiva: Could this not use your pattern match codec?

packages/transaction-messages/src/codecs/v1/instruction.ts
```diff
@@ -84,32 +39,55 @@ export function getInstructionPayloadEncoder(): VariableSizeEncoder<CompiledInst
...
+        write(instructionPayload, bytes, offset) {
+            return getStructEncoder([
+                [
+                    'instructionAccountIndices',
+                    getArrayEncoder(getU8Encoder(), { size: instructionPayload.instructionAccountIndices.length }),
+                ],
+                ['instructionData', fixEncoderSize(getBytesEncoder(), instructionPayload.instructionData.byteLength)],
+            ]).write(instructionPayload, bytes, offset);
+        },
+    });
```
lorisleiva: Isn't that just a wrapped struct encoder?

packages/transaction-messages/src/codecs/v1/message.ts
```diff
@@ -0,0 +1,131 @@
...
+    V1CompiledTransactionMessage | (CompiledTransactionMessageWithLifetime & V1CompiledTransactionMessage)
+> {
+    return createEncoder({
+        getSizeFromValue(message) {
+            return getMessageEncoderForMessage(message).getSizeFromValue(message);
+        },
+        write(message, bytes, offset) {
+            return getMessageEncoderForMessage(message).write(message, bytes, offset);
+        },
+    });
```
lorisleiva: Isn't that just `getMessageEncoderForMessage(message)`? If so, should we inline `getMessageEncoderForMessage` here directly?

packages/transaction-messages/src/codecs/v1/message.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+export function getMessageDecoder(): VariableSizeDecoder<
+    CompiledTransactionMessageWithLifetime & V1CompiledTransactionMessage
+> {
+    return createDecoder({
```
lorisleiva, line 52: Can we not use a struct decoder here with a transformer if necessary?

packages/transaction-messages/src/codecs/message.ts
```diff
@@ -25,12 +33,41 @@ import { getMessageDecoder as getV0MessageDecoder, getMessageEncoder as getV0Mes
...
 > {
-    return getPatternMatchEncoder<
+    // check version is valid
+    const encoder = getPatternMatchEncoder<
         CompiledTransactionMessage | (CompiledTransactionMessage & CompiledTransactionMessageWithLifetime)
     >([
         [m => m.version === 'legacy', getLegacyMessageEncoder()],
         [m => m.version === 0, getV0MessageEncoder()],
+        [m => m.version === 1, getV1MessageEncoder()],
     ]);
```
lorisleiva: A better design would be to wrap this in a `transformEncoder` to check the input before passing it to the inner encoder.

packages/transaction-messages/src/codecs/v1/config.ts
mcintyre94: Only `codecs/message.ts` and `codecs/transaction-version.ts` are exported from the package, the versioned directories are just imported by `message.ts` as codecs it calls out to.

That also means there's less downside to more verbose names though, since they're just used for these codecs, so I think it makes sense to use more specific names.

packages/transaction-messages/src/codecs/v1/config.ts
mcintyre94: Yep - we can write a small encoder for each config value type (u32 and u64) that just encodes their `value` field, and then pattern match on `kind`. The reason I didn't do this was because I didn't think I could use `{size: 'remainder'}` on the array, and I thought I'd need to pass the array size to `getArrayEncoder`. Being able to use `remainder` makes this much simpler.

packages/transaction-messages/src/codecs/v1/instruction.ts
mcintyre94: Yep - similarly this becomes much simpler because we can use `{size: 'remainder'}` on these arrays. Then it's just a struct encoder.

packages/transaction-messages/src/codecs/v1/message.ts
mcintyre94: Same issue with `{size: 'remainder'}` here, I was using `getMessageEncoderForMessage` to get the array sizes in while still using a struct encoder there. Using `remainder` instead for those arrays means we don't need the wrapper.

packages/transaction-messages/src/codecs/v1/message.ts
mcintyre94: I don't think we can use `transformDecoder` here, or in general to chain decoders which is basically what we'd need to do here. We can use a `getStructDecoder` for everything up to and including static accounts. We can decode `staticAccounts` as a u8-prefixed array, because that's how it's structured in the wire format.

But after that, I don't think `transformDecoder` gives us what we'd need:

https://github.com/anza-xyz/kit/blob/f47d5cf30512bbae3233f0ddccae45462af7f309/packages/codecs-core/src/transform-codec.ts#L112-L116

Our `map` function receives the original offset, not the one after the struct decoded. We could get around this by only using the first decoder up to `numStaticAccounts`, and then everything there is fixed size and we can hardcode the offset to continue from.

But then the second problem is that the offset `transformDecoder` returns is the offset after the first decoder, and we don't have an ability to return an updated one from our next decoder. I don't think this function can currently work to chain decoders that both need to update the offset in this way.

What we definitely can do is use a `getStructDecoder` for the first part of the message, and I think that's worth doing. But I think we'll currently need a custom decoder to wrap that and handle the rest.

I'll open an issue in case I'm wrong/to add something like `chainDecoders` which I think could enable this use case.

lorisleiva: You can just use `getTupleDecoder` for this right?

You'd end up with something like:

```ts
return transformDecoder(
  getTupleDecoder([decoderA, decoderB]),
  ([valueA, valueB]) => /* Some logic that combines the two together... */,
)
```

mcintyre94: I don't think so, because you need values from `DecoderA` to write `DecoderB`. I think we need something more like 

```ts
someFn(
  Decoder<A>,
  A => Decoder<B>,
  (A, B) => C
): Decoder<C>
```

lorisleiva: Oh yeah good point. I think that's okay like that then.

One improvement we could make in the future is pass the previous state to each field inside a struct codec if it is provided as a function. For instance:

```ts
return getStructDecoder([
  [fieldA, getFieldADecoder()],
  [fieldB, getFieldBDecoder()],
  [fieldC, ({ fieldA, fieldB }) => getFieldCDecoder(fieldA, fieldB)],
  [fieldD, getFieldDDecoder()],
]);
```

We could do the same for the tuple codec.

But that's a very low-priority nice to have I think.

mcintyre94: I had the same thought when I was working on this, opened an issue with a link to a Claude prototype: https://github.com/anza-xyz/kit/issues/1448 

It would be really nice to be able to simplify this decoder, and would read a lot clearer I think. But agree it's low priority/not worth blocking on for now.

mcintyre94:
Thanks @lorisleiva!

I've updated this to use `{size: 'remainder'}` in the encoders, which simplifies them and means we don't need to include their size in the input when creating those encoders.

- config values are now encoded individually with a pattern match, and as an array using `remainder`
- instruction payload is now ended with `remainder` size on the array of accounts, and without fixing the encoder size on the data bytes (behaves the same as remainder)
- v1 message now uses `remainder` on its arrays, getting rid of the wrapper function that was being used to set its array sizes
- top-level message now uses `transformEncoder` to validate the version before the pattern match 

Decoders are still more complex, but we have a path toward simplifying the v1 message one in future (#1448)

packages/transaction-messages/src/codecs/v1/config.ts
```diff
@@ -1,80 +1,65 @@
...
-const LOADED_ACCOUNTS_DATA_SIZE_LIMIT_BIT_MASK = 0b1000;
-const HEAP_SIZE_BIT_MASK = 0b10000;
-
-/**
- * Encode a {@link TransactionMessageConfig} into a 4 byte mask, where the lowest bits indicate which fields are set
- * @returns An Encoder for {@link TransactionMessageConfig}
+/* TODO issue #1143 - these should be FixedSizeEncoder
+ * but we have a type error on `getPatternMatchEncoder` where it incorrectly types the return as FixedSizeEncoder
+ * when used with differently sized FixedSizEencoder inputs.
+ * For now we use VariableSizeEncoder here to work around this type error
```
lorisleiva: Could we not cast the encoders as `VariableSizeEncoders` instead of creating new ones if the #1143 issue is types only? Since `getPatternMatchEncoder` delegates to `getUnionEncoder` which uses `getEncodedSize` to figure out the sizes (works with fixed and variable) then I think that should work right?

packages/transaction-messages/src/codecs/v1/config.ts
```diff
@@ -1,80 +1,65 @@
...
+        },
     });
 }
 
+function getCompiledTransactionConfigValueEncoder(): VariableSizeEncoder<CompiledTransactionConfigValue> {
+    return getPatternMatchEncoder<CompiledTransactionConfigValue>([
+        [value => value.kind === 'u32', getU32ConfigValueEncoder()],
+        [value => value.kind === 'u64', getU64ConfigValueEncoder()],
+    ]);
+}
```
lorisleiva, line 24: Clean!

packages/transaction-messages/src/codecs/v1/config.ts
```diff
@@ -87,31 +72,36 @@ export function getConfigValuesDecoder(mask: number): VariableSizeDecoder<Transa
     const hasLoadedAccountsDataSizeLimit = (mask & LOADED_ACCOUNTS_DATA_SIZE_LIMIT_BIT_MASK) !== 0;
     const hasHeapSize = (mask & HEAP_SIZE_BIT_MASK) !== 0;
 
+    const u32Decoder = getU32Decoder();
+    const u64Decoder = getU64Decoder();
+
     return createDecoder({
```
lorisleiva: nit: I _think_ here you could do something like:

```ts
const unitDecoder = getUnitDecoder();

return transformDecoder(
  getTupleDecoder([
    [hasPriorityFee ? u64Decoder : unitDecoder],
    [hasComputeUnitLimit ? u32Decoder : unitDecoder],
    [hasLoadedAccountsDataSizeLimit ? u32Decoder : unitDecoder],
    [hasHeapSize ? u32Decoder : unitDecoder],
  ]),
  arr => array.filter(Boolean),
)
```

packages/transaction-messages/src/codecs/message.ts
lorisleiva: super nit: You could extract the casted type in a local type variable.

```ts
type ReturnType = VariableSizeDecoder<CompiledTransactionMessage & CompiledTransactionMessageWithLifetime>;

// ...

[ () => version === 'legacy', getLegacyMessageDecoder() as ReturnType ],
```

packages/transaction-messages/src/codecs/v1/config.ts
mcintyre94: We could make them `FixedSize` and cast them in the pattern match, is that what you mean? I think we do need to create them as new encoders either way though just because they need to explicitly encode just the `value` attribute, I'm not aware of a helper for that but might have missed something.

lorisleiva: Ooh sorry I missed the `.value` which is why I was confused.

Yeah in this case we could do:

```ts
return getPatternMatchEncoder<CompiledTransactionConfigValue>([
  [value => value.kind === 'u32', transformEncoder(getU32Encoder(), c => c.value)],
  [value => value.kind === 'u64', transformEncoder(getU64Encoder(), c => c.value)],
]);
```

mcintyre94: Oh I'm dumb, we can use `getStructDecoder` and just get the value field:

```ts
return getPatternMatchEncoder<CompiledTransactionConfigValue>([
    [value => value.kind === 'u32', getStructEncoder([['value', getU32Encoder()]])],
    [value => value.kind === 'u64', getStructEncoder([['value', getU64Encoder()]])],
])
```

Then just cast that output for now

lorisleiva: That works too! ☺️

packages/transaction-messages/src/codecs/v1/config.ts
mcintyre94: Nice! I needed a little more of a transform to turn it to the `ConfigValue` wrapper:

```ts
    return transformDecoder(
        getTupleDecoder([
            hasPriorityFee ? u64Decoder : unitDecoder,
            hasComputeUnitLimit ? u32Decoder : unitDecoder,
            hasLoadedAccountsDataSizeLimit ? u32Decoder : unitDecoder,
            hasHeapSize ? u32Decoder : unitDecoder,
        ]),
        arr =>
            arr.filter(Boolean).map(value => {
                if (typeof value === 'bigint') {
                    return { kind: 'u64', value } as CompiledTransactionConfigValue;
                } else {
                    return { kind: 'u32', value } as CompiledTransactionConfigValue;
                }
            }),
    );
```

I generally wouldn't want to rely on using number/bigint to distinguish between codecs like that, but since the spec groups them by 4 bytes, I think u32 and u64 are likely to stay the only ones used here. Easy enough to change back to something custom later if we need to.

lorisleiva: An alternative would be to change `u32Decoder` and `u64Decoder` to do that for you. Something like:

```ts
const u32Decoder = transformDecoder(getU32Decoder(), value => ({ kind: 'u32', value }));
const u64Decoder = transformDecoder(getU64Decoder(), value => ({ kind: 'u64', value }));
```

That way you don't need to rely on `typeof`.

mcintyre94: Way better, TY!

mcintyre94:
### Merge activity

* **Mar 11, 1:29 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1431).
* **Mar 11, 1:29 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1431).

PR #1454, opened 2026-03-11
test(rpc-api): add typetests for RPC methods with no arguments

### LINKED ISSUE #249: Add typetests to cover RPC methods with no arguments
Following #236, write typetests that assert that calling these methods with no arguments yields the expected response and generates no errors.

Sigmabrogz:
Fixes #249. This PR adds TypeScript type tests asserting that all RPC methods that can be called with no arguments yield the expected response without generating type errors, satisfying the requirement by using existing type-checking infrastructure in the test suite.

### DIFF
--- a/.changeset/loud-ligers-refuse.md
+++ b/.changeset/loud-ligers-refuse.md
@@ -1,0 +1,5 @@
+---
+"@solana/rpc-api": patch
+---
+
+test(rpc-api): add typetests for RPC methods with no arguments


Sigmabrogz:
Added the changeset as requested.

Sigmabrogz:
Thank you for the review. I understand now! I am unfortunately at capacity to do the full refactor to individual typetests checking the exact response shapes, so feel free to close this PR if it's not the right direction. Thanks again for your time!

PR #1475, opened 2026-03-15
feat(transactions): add reconstructEncodedTransactionFromOriginalTransaction to transactions package

### LINKED ISSUE #1466: Add a function to help `TransactionModifyingSigner` update `lifetimeConstraint`
Our API for `TransactionModifyingSigner` requires a lifetime to be returned:

```ts
    modifyAndSignTransactions(
        transactions: readonly (Transaction | (Transaction & TransactionWithLifetime))[],
        config?: TransactionModifyingSignerConfig,
    ): Promise<readonly (Transaction & TransactionWithinSizeLimit & TransactionWithLifetime)[]>;
```

As such a signer can modify the transaction, this lifetime may be different from the input one. Identifying the lifetime constraint requires decompiling the transaction and inspecting its first instruction (whether it's advance nonce determines if the token is a blockhash or nonce), so should be avoided if possible.

We have code to help with this in the react `useWalletAccountTransactionSigner` hook: https://github.com/anza-xyz/kit/blob/874a02f2aa5074c3309a66725522ead9e2fec166/packages/react/src/useWalletAccountTransactionSigner.ts#L97

And this is also used in the PR to add non-react WalletAccount -> Signer functions: https://github.com/anza-xyz/kit/pull/1368 

This code is useful for anybody writing a similar `TransactionModifyingSigner`. We could extract it to a helper in the `transactions` package to make it easier for others to write similar signers.

dpsi9:
#### Problem
The logic to determine the correct lifetime (blockhash vs durable nonce) previously existed only inside the React useWalletAccountTransactionSigner hook and was not reusable.

#### Summary of Changes
- add `reconstructEncodedTransactionFromOriginalTransaction` to the transactions package
- export the helper via the transactions package index
- refactor `useWalletAccountTransactionSigner` to use the helper
- add unit tests and typetests for the reconstructEncodedTransactionFromOriginalTransaction function
- update useWalletAccountTransactionSigner tests accordingly

Fixes #1466

### DIFF
--- a/packages/react/src/__tests__/useWalletAccountTransactionSigner-test.ts
+++ b/packages/react/src/__tests__/useWalletAccountTransactionSigner-test.ts
@@ -1,30 +1,23 @@
 import { Address } from '@solana/addresses';
-import type { VariableSizeCodec, VariableSizeDecoder } from '@solana/codecs-core';
+import type { VariableSizeCodec } from '@solana/codecs-core';
 import {
     SOLANA_ERROR__SIGNER__WALLET_MULTISIGN_UNIMPLEMENTED,
     SOLANA_ERROR__TRANSACTION__NONCE_ACCOUNT_CANNOT_BE_IN_LOOKUP_TABLE,
     SolanaError,
 } from '@solana/errors';
 import { SignatureBytes } from '@solana/keys';
-import { Blockhash } from '@solana/rpc-types';
 import {
-    CompiledTransactionMessage,
-    CompiledTransactionMessageWithLifetime,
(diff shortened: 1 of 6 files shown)


lorisleiva:
Hi there, just having a quick first look at this PR and whilst the code extraction is trivial, the tests don't seem at all consistent with the rest of the code base. For instance, if you look at the typetests, you'll see patterns that we just don't use anywhere in the repo (e.g. `X extends Y ? true : false` and asserting on that instead of just relying on `satisfies`). I believe the regular tests can also be refactored to be easier to read and more consistent with the rest of the codebase.

PR #1480, opened 2026-03-17
Add withCleanup function to @solana/plugin-core

mcintyre94:
Adds a `withCleanup` function that extends any client with a synchronous `Symbol.dispose` method, enabling the TC39 `using` declaration pattern. Note that only synchronous cleanup is supported for now — async teardown (`Symbol.asyncDispose`) is not yet handled. If the client already has a `Symbol.dispose`, the new cleanup is chained so both run when the client is disposed. Includes unit tests and a type test.

```ts
function myPlugin() {
    return <T extends object>(client: T) => {
        const socket = new WebSocket('wss://api.example.com');
        return withCleanup(
            extendClient(client, { socket }),
            () => socket.close(),
        );
    };
}

// `socket.close()` is called automatically when `client` goes out of scope:
using client = createClient(myPlugin());
```

### DIFF
--- a/packages/plugin-core/src/__tests__/client-test.ts
+++ b/packages/plugin-core/src/__tests__/client-test.ts
@@ -1,6 +1,6 @@
 import '@solana/test-matchers/toBeFrozenObject';
 
-import { createEmptyClient, extendClient } from '../client';
+import { createEmptyClient, extendClient, withCleanup } from '../client';
 
 describe('createEmptyClient', () => {
     it('creates an empty object with a use function', () => {
@@ -259,5 +259,61 @@ describe('extendClient', () => {
 
     it('returns a frozen object', () => {
         expect(extendClient({ fruit: 'apple' as const }, { vegetable: 'carrot' as const })).toBeFrozenObject();
+    });
+});
+
+describe('withCleanup', () => {
+    it('calls the cleanup function when disposed', () => {
+        const cleanup = jest.fn();
+        const result = withCleanup({}, cleanup);
+        result[Symbol.dispose]();
+        expect(cleanup).toHaveBeenCalledTimes(1);
+    });
+
+    it('disposes using the `using` syntax', () => {
+        const cleanup = jest.fn();
+        {
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            using _ = withCleanup({}, cleanup);
+        }
+        expect(cleanup).toHaveBeenCalledTimes(1);
+    });
+
+    it('does not call the cleanup function before dispose is invoked', () => {
+        const cleanup = jest.fn();
+        withCleanup({}, cleanup);
+        expect(cleanup).not.toHaveBeenCalled();
+    });
+
+    it('chains parent Symbol.dispose when client already has one', () => {
+        const callOrder: string[] = [];
+        const parentDispose = jest.fn(() => callOrder.push('parent'));
+        const cleanup = jest.fn(() => callOrder.push('cleanup'));
+        const client = { [Symbol.dispose]: parentDispose };
+        const result = withCleanup(client, cleanup);
+        result[Symbol.dispose]();
+        expect(cleanup).toHaveBeenCalledTimes(1);
+        expect(parentDispose).toHaveBeenCalledTimes(1);
+        expect(callOrder).toStrictEqual(['cleanup', 'parent']);
+    });
+
+    it('does not throw when client has no Symbol.dispose', () => {
+        const cleanup = jest.fn();
+        const result = withCleanup({}, cleanup);
+        expect(() => result[Symbol.dispose]()).not.toThrow();
+    });
+
+    it('preserves existing client properties', () => {
+        const result = withCleanup({ fruit: 'apple' as const }, () => {});
+        expect(result.fruit).toBe('apple');
+    });
+
+    it('symbol.dispose is a function on the result', () => {
+        const result = withCleanup({}, () => {});
+        expect(typeof result[Symbol.dispose]).toBe('function');
+    });
+
+    it('returns a frozen object', () => {
+        expect(withCleanup({ fruit: 'apple' as const }, () => {})).toBeFrozenObject();
     });
 });
--- a/packages/plugin-core/src/__tests__/client-test.ts
+++ b/packages/plugin-core/src/__tests__/client-test.ts
@@ -316,4 +316,35 @@ describe('withCleanup', () => {
     it('returns a frozen object', () => {
         expect(withCleanup({ fruit: 'apple' as const }, () => {})).toBeFrozenObject();
     });
+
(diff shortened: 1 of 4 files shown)


packages/plugin-core/src/__tests__/client-test.ts
```diff
@@ -261,3 +261,50 @@ describe('extendClient', () => {
...
     });
 });
+
+describe('withCleanup', () => {
+    it('calls the cleanup function when disposed', () => {
+        const cleanup = jest.fn();
+        const result = withCleanup({}, cleanup);
+        result[Symbol.dispose]();
+        expect(cleanup).toHaveBeenCalledTimes(1);
+    });
```
lorisleiva, line 271: nit: Not sure if it's possible in Jest environments but would it be possible to have a variant of this test that doesn't call `Symbol.dispose` directly? Something like:

```ts
const cleanup = jest.fn();
{
    const client = withCleanup({}, cleanup);
} // client variable is disposed here.
expect(cleanup).toHaveBeenCalledTimes(1);
```

packages/plugin-core/src/__typetests__/client-typetest.ts
```diff
@@ -233,3 +240,18 @@ const EMPTY_ASYNC_CLIENT = null as unknown as AsyncClient<object>;
         extendClient({}, 'hello');
     }
 }
+
+// [Describe] withCleanup
+{
+    // It returns a Disposable client
```
lorisleiva, line 246: nit: Should we add a test with an `AsyncClient` input for completeness?

packages/plugin-core/src/client.ts
```diff
@@ -242,3 +242,48 @@ export function extendClient<TClient extends object, TAdditions extends object>(
...
+ *         const socket = new WebSocket('wss://api.example.com');
+ *         return withCleanup(
+ *             extendClient(client, { socket }),
+ *             () => socket.close(),
+ *         );
+ *     };
+ * }
+ *
+ * // Later, when the client is no longer needed:
+ * using client = createClient(myPlugin());
```
lorisleiva: I think this example is a little broken:

```suggestion
 * const client = createEmptyClient().use(myPlugin());
 * // Later, when the client is no longer needed:
```

packages/plugin-core/src/client.ts
```diff
@@ -242,3 +242,48 @@ export function extendClient<TClient extends object, TAdditions extends object>(
...
+ *
+ * @see {@link extendClient}
+ */
+export function withCleanup<TClient extends object>(client: TClient, cleanup: () => void): Disposable & TClient {
+    const parentDispose: (() => void) | undefined =
+        Symbol.dispose in client ? (client as Disposable)[Symbol.dispose] : undefined;
+    const additions: Disposable = {
+        [Symbol.dispose]() {
+            cleanup();
+            parentDispose?.call(client);
```
lorisleiva: Interesting, the child cleanup is called before its predecessors. Is that by design to create a sort of First In Last Cleaned approach? It probably makes sense actually because if plugin B depends on plugin A, then you want to make sure plugin B is cleaned first so there's no invalid behaviour where B is still alive and A has been cleaned up.

packages/plugin-core/src/__tests__/client-test.ts
```diff
@@ -261,3 +261,50 @@ describe('extendClient', () => {
...
+    it('chains parent Symbol.dispose when client already has one', () => {
+        const callOrder: string[] = [];
+        const parentDispose = jest.fn(() => callOrder.push('parent'));
+        const cleanup = jest.fn(() => callOrder.push('cleanup'));
+        const client = { [Symbol.dispose]: parentDispose };
+        const result = withCleanup(client, cleanup);
+        result[Symbol.dispose]();
+        expect(cleanup).toHaveBeenCalledTimes(1);
+        expect(parentDispose).toHaveBeenCalledTimes(1);
+        expect(callOrder).toStrictEqual(['cleanup', 'parent']);
```
lorisleiva, line 297: Nice you test the clean up order explicitly here.

lorisleiva (approved):
Very nice! Just a couple of nits. Is the idea that we only support sync dispose for now and if we need async support we can add `withAsyncCleanup` in the future?

mcintyre94:
* **#1482** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1482?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1480** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1480?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/plugin-core/src/client.ts
mcintyre94: Yep that's the idea - if you remember the very first draft design had a stack being maintained, this is that same idea but without having to maintain that chain as separate state.

packages/plugin-core/src/__tests__/client-test.ts
mcintyre94: Nice, this test works:

```ts
    it('disposes using the `using` syntax', () => {
        const cleanup = jest.fn();
        {
            // eslint-disable-next-line @typescript-eslint/no-unused-vars
            using _ = withCleanup({}, cleanup);
        }
        expect(cleanup).toHaveBeenCalledTimes(1);
    });
```

packages/plugin-core/src/__typetests__/client-typetest.ts
mcintyre94: Good idea, added:

```ts
    // It accepts an AsyncClient
    {
        const client = null as unknown as AsyncClient<object>;
        withCleanup(client, () => {}) satisfies AsyncClient<object> & Disposable;
    }
```

mcintyre94:
We discussed offline, and decided to keep broadly this design.

I've refactored this PR to use a `DisposableStack`, which is [part of the same spec](https://github.com/tc39/proposal-explicit-resource-management?tab=readme-ov-file#the-disposablestack-and-asyncdisposablestack-container-objects). This replaces our code to manually call a parent dispose function to create a LIFO stack. The primary difference is that this provides robust error handling, if one dispose function fails then others will still be attempted.

This does add some state to the client object, which is maintained in a non-enumerable `__PRIVATE__DISPOSABLE_STACK` field. The first time we use `withCleanup` we add this stack to the client, and subsequent calls add their cleanup function to it. We also add a `[Symbol.dispose]` method which calls `stack[Symbol.dispose]()`, ie calls the cleanup functions added to the stack. If there is an existing `Symbol.dispose` on the client before the first call to `withCleanup` then this is preserved by being added to the stack first.

lorisleiva (approved):
Nice! The disposable stack API is great!

PR #1482, opened 2026-03-17
Add `withAsyncCleanup` function to `@solana/plugin-core`

mcintyre94:
This PR follows up the previous one by adding `withAsyncCleanup` to enable async cleanup functions and the `await using` syntax. 

For async cleanup, we first await the given cleanup function. We then call the parent `AsyncDispose` if there is one, falling back to a parent `Dispose` if there is one. This means that on async cleanup we call the parent cleanup function, even if it is not async.

This PR also uses types to try to force consumers to use async disposal if any plugin in the chain uses `withAsyncCleanup`:

- If the input client to `withAsyncCleanup` is `Disposable`, this is stripped and only `AsyncDisposable` is returned
- If the input client to `withCleanup` is `AsyncDisposable`, then it is returned with its type unchanged, ie it is `AsyncDisposable` and not `Disposable`. This means that if you chain a plugin with sync disposal after one with async disposal, the returned client is `AsyncDisposable` only. 

Note: We could decide not to ship this until we/someone else have an async use case, I think our wallet plugin will only need sync. But async is part of the same spec, and not having it would limit some future use cases.

### DIFF
--- a/.changeset/floppy-memes-try.md
+++ b/.changeset/floppy-memes-try.md
@@ -1,0 +1,5 @@
+---
+'@solana/plugin-core': minor
+---
+
+Add `withAsyncCleanup` function to `@solana/plugin-core`. Plugin authors can use it to register asynchronous teardown logic on a client, making it `AsyncDisposable`. If the client already implements `Symbol.asyncDispose`, it is awaited after the new cleanup; if only `Symbol.dispose` is present, it is called synchronously as a fallback.
--- a/.changeset/loose-feet-like.md
+++ b/.changeset/loose-feet-like.md
@@ -1,0 +1,5 @@
+---
+'@solana/plugin-core': minor
+---
+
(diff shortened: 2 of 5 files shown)


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1482?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1482** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1482?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1482?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1480** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1480?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

mcintyre94:
Closing for now - we'll figure out how to add this if we need to later

PR #1496, opened 2026-03-26
Enforce transaction compilation constraints in compileTransactionMessage

mcintyre94:
## Summary

- Adds four compile-time guards to `compileTransactionMessage` (legacy, v0, and v1) that enforce Solana protocol limits before a transaction reaches the network. Each violation throws a typed `SolanaError` with structured context (actual count, max allowed, and for per-instruction violations, the instruction index).
- Adds four new error codes to `@solana/errors`: `SOLANA_ERROR__TRANSACTION__TOO_MANY_ACCOUNT_ADDRESSES` (max 64), `SOLANA_ERROR__TRANSACTION__TOO_MANY_SIGNER_ADDRESSES` (max 12), `SOLANA_ERROR__TRANSACTION__TOO_MANY_INSTRUCTIONS` (max 64), and `SOLANA_ERROR__TRANSACTION__TOO_MANY_ACCOUNTS_IN_INSTRUCTION` (max 255).
- Tests cover each constraint at the boundary (N passes, N+1 throws) for all three transaction versions.

These limits are specified in the [SIMD-0385 Transaction v1 spec](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md#transaction-constraints) and apply to legacy and v0 transactions as well. They're relevant for all transaction versions, but will be easier to hit without hitting the size limits after we allow 4kb v1 transactions. 

A follow-up PR will update the transaction planner to handle these new errors in the same way that it handles transactions that exceed the size limit.

Part of #1223

PR #1497, opened 2026-03-26
Handle transaction constraint errors in the transaction planner

### LINKED ISSUE #1223: Enforce more constraints on transactions
## Motivation

Currently transactions are limited to 1232 bytes, which makes it difficult to exceed a number of constraints on transactions. These are enforced by the validator but not currently by Kit. They become trivially easy to exceed given 4096 byte transactions.

See constraints in [SIMD for V1 transactions](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md#transaction-constraints)

value | max specified by this SIMD | prior max | max implied by format
-- | -- | -- | --
transaction size | 4096 | 1232 | 4096
signatures per transaction | 12 | 12 | 42
num accounts | 64 | 64 | 96
num instructions | 64 | 64 | 255
accounts/instruction | 255 | 255 | 255

Transaction size is covered separately by #1221 and should not be changed in this issue, which is a prerequisite for that

But signature count, account count, instruction count, are all able to be exceeded with 4096 byte transactions, but explicitly not relaxed for v1 transactions. This means that these constraints apply to all transaction versions.

`accounts/instruction` is unchanged, and probably currently implicitly covered by its encoder, which likely throws if we try to encode a longer array length. But it probably makes sense to give it the same error handling as the others.

Kit should not allow transactions that violate these constraints to be compiled. 

## Details

- This is a prerequisite for #1221, as it becomes very easy to accidentally violate these constraints with a 4096 byte transaction. But these constraints also apply to legacy/v0 transactions, so this work can be done independently
- We need a helpful error for each constraint
- At a minimum, we probably want to throw in `compileTransactionMessage` if the input transaction message violates any constraints
- We need to think carefully about making sure the transaction planner API enforces these constraints as it is packing transactions

Supersedes #1102, which only considers the instruction count

mcintyre94:
#### Problem

The new errors thrown in `compileTransactionMessage` in the previous PR are not handled by the transaction planner. This means that if we can't add an instruction to the transaction message being packed because it violates one of these constraints, we will throw.

#### Summary of Changes

We now catch each of the new errors in addition to `SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN` in `selectAndMutateCandidate`. The result of this is that we will attempt to pack the instruction into the next message (and then a new one), in the same way that we do when it is too large in bytes.

Closes #1223

### DIFF
--- a/.changeset/yummy-paths-jump.md
+++ b/.changeset/yummy-paths-jump.md
@@ -1,0 +1,5 @@
+---
+"@solana/instruction-plans": patch
+---
+
+The transaction planner now handles the four new transaction compilation constraint errors (`TOO_MANY_ACCOUNT_ADDRESSES`, `TOO_MANY_SIGNER_ADDRESSES`, `TOO_MANY_INSTRUCTIONS`, `TOO_MANY_ACCOUNTS_IN_INSTRUCTION`) gracefully. When adding an instruction to an existing candidate transaction would violate a constraint, the planner splits it into a new transaction — the same behaviour it already had for transactions that exceed the byte size limit. If even a fresh transaction cannot accommodate the instruction, the constraint error propagates to the caller.
--- a/packages/instruction-plans/src/__tests__/transaction-planner-test.ts
+++ b/packages/instruction-plans/src/__tests__/transaction-planner-test.ts
@@ -1,7 +1,15 @@
 import '@solana/test-matchers/toBeFrozenObject';
 
-import { SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN, SolanaError } from '@solana/errors';
-import { Instruction } from '@solana/instructions';
+import { Address, getAddressDecoder } from '@solana/addresses';
+import {
+    SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN,
+    SOLANA_ERROR__TRANSACTION__TOO_MANY_ACCOUNT_ADDRESSES,
+    SOLANA_ERROR__TRANSACTION__TOO_MANY_ACCOUNTS_IN_INSTRUCTION,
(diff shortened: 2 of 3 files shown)


mcintyre94:
* **#1499** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1499?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1497** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1497?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1497?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1496** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1496?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lorisleiva (approved):
Thanks! Instead of mocking, how would you feel about creating four real scenarios where we add too many things and expect them to produce a new transaction even though there would have been enough space in a transaction message?

For the `propagates %s from compileTransactionMessage when even a fresh message cannot accommodate the plan` error, we could throw this in the `createTransactionMessage` callback and achieve the same result.

mcintyre94:
Agree, updated that + removed mocking. Each recoverable limit is now testing a real example, and the propagated errors throw in `createTransactionMessage`.

mcintyre94:
### Merge activity

* **Mar 31, 11:01 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1497).
* **Mar 31, 11:02 AM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1497) rebased this pull request as part of a merge.
* **Mar 31, 11:10 AM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1497).

PR #1499, opened 2026-03-26
Add version-aware transaction size limits

### LINKED ISSUE #1221: SIMD 296 - Larger Transaction Size
## Motivation

See SIMD: https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0296-larger-transactions.md

Kit needs to allow transactions to be up to 4096 bytes when the version is `v1`
Still enforce the 1232 byte limit for legacy/v0 transactions

## Example use case

V1 transactions are able to be up to 4096 bytes, and do not support lookup tables, meaning that a larger transaction size is necessary.

## Details

Prerequisites:

- Enforce transaction constraints: #1223 
- V1 transactions, this only applies to them. #1222

mcintyre94:
#### Problem

Currently we hardcode the transaction size limit to 1232 bytes. For v1 transactions this should be 4096 bytes.

#### Summary of Changes

Replace using constant transaction size limits with a function to get the size limit for a given `Transaction` or `TransactionMessage`. For `TransactionMessage` this just reads the `version` field. 

For `Transaction` we use the fact that v1 transactions encode the version byte first, and the v1 value `0x81` is not a valid number of signatures for legacy/v0. Unlike the codec, we just need to know if it's v1 or not here to determine the answer, so we don't look at anything else. This will need to be extended for future versions, but assuming they put the version byte first it'll be trivial. 

Fixes: #1221

### DIFF
--- a/packages/transactions/src/transaction-size.ts
+++ b/packages/transactions/src/transaction-size.ts
@@ -4,15 +4,20 @@ import type { TransactionMessage, TransactionMessageWithinSizeLimit } from '@sol
 
 import { getTransactionEncoder } from './codecs';
 import { Transaction } from './transaction';
+import { LEGACY_TRANSACTION_SIZE_LIMIT, V1_TRANSACTION_SIZE_LIMIT } from './transaction-size-limits';
 
 /**
  * The maximum size of a transaction packet in bytes.
+ *
+ * @deprecated Transaction size is no longer constant as v1 transactions have a larger size limit. Use `getTransactionSizeLimit` instead to get the size limit for a specific transaction based on its version.
  */
 export const TRANSACTION_PACKET_SIZE = 1280;
 
 /**
  * The size of the transaction packet header in bytes.
  * This includes the IPv6 header and the fragment header.
+ *
+ * @deprecated Transaction size is no longer constant as v1 transactions have a larger size limit. Use `getTransactionSizeLimit` instead to get the size limit for a specific transaction based on its version.
  */
 export const TRANSACTION_PACKET_HEADER =
     40 /* 40 bytes is the size of the IPv6 header. */ + 8; /* 8 bytes is the size of the fragment header. */
@@ -22,6 +27,8 @@ export const TRANSACTION_PACKET_HEADER =
  *
  * Note that this excludes the transaction packet header.
  * In other words, this is how much content we can fit in a transaction packet.
+ *
+ * @deprecated Transaction size is no longer constant as v1 transactions have a larger size limit. Use `getTransactionSizeLimit` instead to get the size limit for a specific transaction based on its version.
  */
 export const TRANSACTION_SIZE_LIMIT = TRANSACTION_PACKET_SIZE - TRANSACTION_PACKET_HEADER;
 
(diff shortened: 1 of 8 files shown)


mcintyre94:
* **#1499** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1499?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1499?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1497** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1497?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1496** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1496?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/transactions/src/transaction-size.ts
```diff
@@ -25,6 +25,11 @@ export const TRANSACTION_PACKET_HEADER =
  */
 export const TRANSACTION_SIZE_LIMIT = TRANSACTION_PACKET_SIZE - TRANSACTION_PACKET_HEADER;
 
+/**
+ * The maximum size of a version 1 transaction in bytes.
+ */
+export const V1_TRANSACTION_SIZE_LIMIT = 4096;
```
lorisleiva: I wonder if we should deprecate `TRANSACTION_SIZE_LIMIT` instead and not expose these constants. That way we ensure people go through the more accurate `getTransactionMessageSizeLimit` helper instead. Wdyt?

packages/transactions/src/transaction-size.ts
```diff
@@ -89,11 +103,19 @@ export function isTransactionWithinSizeLimit<TTransaction extends Transaction>(
 export function assertIsTransactionWithinSizeLimit<TTransaction extends Transaction>(
     transaction: TTransaction,
 ): asserts transaction is TransactionWithinSizeLimit & TTransaction {
+    if (transaction.messageBytes.length === 0) {
+        // If there are no message bytes, then the transaction is empty and thus within the size limit.
+        return;
+    }
+
+    const firstByte = transaction.messageBytes[0];
+    const sizeLimit = (firstByte & VERSION_FLAG_MASK) === 1 ? V1_TRANSACTION_SIZE_LIMIT : TRANSACTION_SIZE_LIMIT;
```
lorisleiva: I'm not a big fan that we're having another location for computing the size limit for a transaction. I get that this is because the transaction message itself is compiled and encoded so we can't just defer to `getTransactionMessageSizeLimit` but maybe we should introduce a second `getTransactionSizeLimit` helper to make it clear that these two functions are the source of truth for that logic?

lorisleiva (commented):
Nice! I made a couple of comments.

packages/transactions/src/transaction-size.ts
mcintyre94: That makes sense. I actually renamed `TRANSACTION_SIZE_LIMIT` at one point and then realised it was exported so I reverted that. Makes sense to deprecate that rather than exporting the new one.

mcintyre94: `TRANSACTION_SIZE_LIMIT` is now deprecated, and `V1_TRANSACTION_SIZE_LIMIT` no longer exported 

I've also deprecated the `TRANSACTION_PACKET_SIZE` and `TRANSACTION_PACKET_HEADER` because I think they're only relevant for calculating the legacy tx size.

I've moved both legacy and v1 sizes to a new `transaction-size-limits` file which just exports these consts, for use in transaction-size, transaction-message-size, and tests. This is intentionally not exported from the package and is just used internally.

Consumers should use `getTransactionMessageSizeLimit` and `getTransactionSizeLimit` instead.

packages/transactions/src/transaction-size.ts
mcintyre94: Agree with this. I've added (and exported) `getTransactionSizeLimit` mirroring `getTransactionMessageSizeLimit` but using the version byte. This is used by both `isTransactionWithinSizeLimit` and `assertIsTransactionWithinSizeLimit`.

lorisleiva (approved):
Thanks for the changes!

PR #1504, opened 2026-03-30
Rewrite Installation page with plugin-first tabbed quick start

lorisleiva:
The current Installation page shows a manual setup with `createSolanaRpc`, `createSolanaRpcSubscriptions`, and `sendAndConfirmTransactionFactory` — code that most developers shouldn't need to write now that Kit Plugins exist. The page needs to lead with the plugin-based client approach.

This PR rewrites the Installation page with a card-based selector that lets users pick their client type (RPC Client, Local RPC Client, LiteSVM Client, or No Client) and shows tailored install + setup steps for each. The first selected card is the default and the selection persists across page loads.

A new `CardTabs` / `CardTab` component renders the selector as a 2x2 responsive grid of cards with icons, titles, and descriptions — built on top of Fumadocs' internal Radix Tabs primitives with no additional dependencies.

Each client tab shows: package installation (using the existing `package-install` syntax for npm/yarn/pnpm tabs), program client installation, and a twoslash-annotated code block demonstrating client creation with `.use(systemProgram()).use(tokenProgram())` chaining and commented examples of what the client can do.

New devDependencies added to `docs/package.json` for twoslash type resolution: `@solana/kit-client-rpc`, `@solana/kit-client-litesvm`, `@solana-program/system`, `@solana-program/token`.

CleanShot 2026-03-27 at 18 27 18@2x

PR #1502, opened 2026-03-27
Restructure documentation for plugin-first information architecture

lorisleiva:
#### Problem

The current solanakit.com documentation teaches Kit from the low-level primitives up: manually creating RPC connections, hand-building Client types, constructing transaction messages with `pipe()`, etc. Now that Kit Plugins and ready-made clients exist (`@solana/kit-client-rpc`, `@solana/kit-client-litesvm`), the documentation should lead with the simpler plugin-based approach and position the lower-level Kit APIs as advanced content.

#### Summary of Changes

This PR restructures the documentation's information architecture without deleting any existing content. Every new page is a stub with a detailed "Planned Content" callout describing what will be written. Every existing page that will be updated or removed has a banner explaining the plan. The site builds and is fully navigable for review.

```
Documentation
├── Installation                          📝
├── Getting started                       🆕
├── Upgrade Guide                         📝
├── ❌ Compatible Clients                 🗑️
├── Guides
│   ├── Overview                          🆕
│   ├── Setting up signers                🆕
│   ├── Sending transactions              🆕
│   ├── Sending multiple transactions     🆕
│   ├── Fetching accounts                 🆕
│   ├── Using program plugins             🆕
│   └── Testing & local development       🆕
├── Clients & Plugins
│   ├── Overview                          🆕
│   ├── Available clients                 🆕
│   ├── Available plugins                 🆕
│   ├── Creating custom plugins           🆕
│   ├── Generating program plugins        🆕
│   └── Creating custom clients           🆕
├── Advanced Guides (renamed from Core Concepts)
│   ├── Overview                          📝
│   ├── Kit without a client              🆕
│   ├── Key pairs
│   ├── Signers                           📝
│   ├── Transactions                      📝
│   ├── Codecs
│   ├── RPC requests                      📝
│   ├── RPC subscriptions                 📝
│   ├── Errors                            📝
│   ├── Offchain messages
│   └── Instruction plans                 📝
└── Getting Started (Old)
    ├── ❌ Generate a signer              🗑️
    ├── ❌ Create instructions            🗑️
    ├── ❌ Build a transaction            🗑️
    ├── ❌ Send a transaction             🗑️
    └── ❌ Fetch an account               🗑️

Recipes (new top-level section)
├── Overview                              🆕
├── Transferring SOL                      🆕
├── Creating a token                      🆕
└── Airdropping tokens                    🆕

🆕 = new stub page
📝 = existing page with planned updates
🗑️ = existing page scheduled for removal
[...]

PR #1505, opened 2026-03-30
Write Getting Started one-pager tutorial

lorisleiva:
This PR adds the Getting Started page as a single-page, hands-on tutorial that walks through building a small Solana app on devnet. It replaces the previous multi-page Getting Started stub with a compact one-pager covering client creation, sending transactions, and fetching accounts.

The tutorial creates a client connected to devnet with the `rpcAirdrop`, `systemProgram`, and `tokenProgram` plugins, then demonstrates sending a single instruction via `.sendTransaction()` on the instruction itself, sending multiple instructions atomically via `client.sendTransaction([...])` to create and initialize a token mint, and fetching the resulting mint account using the token program plugin's typed fetch helper.

All code blocks use twoslash for type annotations, with hidden setup via `// ---cut-before---` for incremental snippets and a complete runnable script at the end. Kit plugin devDependencies updated from 0.8 to 0.9 which resolves a type inference issue with `getMinimumBalance` after chaining program plugins.


lorisleiva:
* **#1587** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1587?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1586** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1586?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1585** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1585?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1584** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1584?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1583** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1583?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1506** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1506?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1505** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1505?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1505?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1504** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1504?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1502** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1502?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

docs/content/docs/getting-started.mdx
```diff
@@ -3,25 +3,203 @@ title: Getting started
...
+<Callout>
+    Airdrops are only available on test networks like devnet and localhost. If the airdrop fails due
+    to rate limiting, wait a moment and try again, or request devnet SOL from the [Solana
+    Faucet](https://faucet.solana.com). For production signer options, see the [Setting Up
+    Signers](/docs/guides/setting-up-signers) guide.
 </Callout>
+
+## Send a transaction
+
+Let's create a token mint on-chain. First, we need to allocate an account for it using the System program.
```
mikemaccana: No dash - see https://solana.com/docs/references/terminology#onchain-program

docs/content/docs/getting-started.mdx
```diff
@@ -3,25 +3,203 @@ title: Getting started
...
+        newAccount: mint,
+        lamports: rentLamports,
+        space: getMintSize(),
+        programAddress: TOKEN_PROGRAM_ADDRESS,
+    }),
+    client.token.instructions.initializeMint({
+        mint: mint.address,
+        decimals: 9,
+        mintAuthority: client.payer.address,
+    }),
```
amilz: fine w/ this, but imo i've been trying to highlight your helpers like `createMint`. you could either 
1. replace this w/ it OR
2. have an additional callout below just highlighting the 3 token helpers that abract multiple ix in a row (can link to a guide on plans to learn more)

docs/content/docs/getting-started.mdx
```diff
@@ -3,25 +3,203 @@ title: Getting started
...
+const payer = await generateKeyPairSigner();
+const client = createClient({ url: 'https://api.devnet.solana.com', payer })
+    .use(rpcAirdrop())
+    .use(systemProgram())
+    .use(tokenProgram());
+// ---cut-before---
+const mint = await generateKeyPairSigner();
+const rentLamports = await client.getMinimumBalance(getMintSize());
+
+await client.system.instructions
```
amilz: could be helpful showing how to log the txn id.

docs/content/docs/getting-started.mdx
```diff
@@ -3,25 +3,203 @@ title: Getting started
...
+@solana/kit @solana/kit-client-rpc @solana/kit-plugin-rpc
+```
 
-**Set up a signer** — Generate a keypair signer with `generateKeyPairSigner` and fund it by requesting an airdrop on devnet.
+Then install program clients for the Solana programs you want to interact with. For this tutorial, we'll use the System program to create accounts and the Token program to create a token.
 
-**Send a transaction** — Build a transfer instruction using `@solana-program/system`, send it via `client.sendTransaction`, and confirm the result.
+```package-install
+@solana-program/system @solana-program/token
+```
```
mikemaccana, line 20: How often do people use Solana and *not* use the `system` or `token` programs? I don't think I've ever done that. What about you @lorisleiva and @amilz ? What situation is there where people don't use system or token and how common is it? 

If the answer is 'rarely': these should be dependencies.

docs/content/docs/getting-started.mdx
```diff
@@ -3,25 +3,203 @@ title: Getting started
...
+    .createAccount({
+        newAccount: mint,
+        lamports: rentLamports,
+        space: getMintSize(),
+        programAddress: TOKEN_PROGRAM_ADDRESS,
+    })
+    .sendTransaction();
+```
+
+When an instruction is created through a program client like `systemProgram()`, you can call `.sendTransaction()` directly on it. The client handles building the transaction, signing, sending, and confirming — all in one call. Notice that we don't need to pass a `payer` here since the client fills it from `client.payer` automatically.
```
mikemaccana: Hrm. `createAccount()` intutively feels like it's going to return an account. But I bet you don't want `createCreateAccountInstruction()`.  

Maybe the best thing to do here is to be explicit?

```suggestion

> [!NOTE]  
> While `client.system.instructions.createAccount()` might seem like it's going to create an account, it's actually going to return an instruction for the `system` program to create an account. 

When you can call `.sendTransaction()` on the created instruction, the client handles building the transaction with that instruction, signing, sending, and confirming - all in one call. Notice that we don't need to pass a `payer` here since the client fills it from `client.payer` automatically.

```

Alternatively, you could do `await client.sendTransaction([...])` like you do elsewhere and introduce `.sendTransaction()` as a shortcut later - remember, many people reading this guide will be new to Solana so it's important they understand that one transaction can have many instructions. Sending an instruction as a transaction could be confusing.

docs/content/docs/getting-started.mdx
mikemaccana: +1. I'd also print the Explorer address, so people can see what they did on devnet, after each step in this guide.

docs/content/docs/getting-started.mdx
```diff
@@ -3,25 +3,203 @@ title: Getting started
...
+
+Let's create a token mint on-chain. First, we need to allocate an account for it using the System program.
+
+```ts twoslash
+import { generateKeyPairSigner, lamports } from '@solana/kit';
+import { createClient } from '@solana/kit-client-rpc';
+import { rpcAirdrop } from '@solana/kit-plugin-rpc';
+import { systemProgram } from '@solana-program/system';
+import { getMintSize, TOKEN_PROGRAM_ADDRESS, tokenProgram } from '@solana-program/token';
+const payer = await generateKeyPairSigner();
```
mikemaccana: Idea: let them grind a prefix. `generateKeyPairSigner(prefix: 'mm', file: 'welcome-to-solana-kit.json')` or whatever. That way they can easily follow the account they made in Explorer later.

docs/content/docs/getting-started.mdx
mikemaccana: If you want to show off multiple instructions (which I feel is the intention here), a token transfer and a memo are a great combination - you can easily see both operations in Solana Explorer.

docs/content/docs/getting-started.mdx
```diff
@@ -3,25 +3,203 @@ title: Getting started
...
+        decimals: 9,
+        mintAuthority: client.payer.address,
+    }),
+]);
+
+// Fetch the mint account.
+const mintAccount = await client.token.accounts.mint.fetch(mint.address);
+console.log('Decimals:', mintAccount.data.decimals);
+console.log('Mint authority:', mintAccount.data.mintAuthority);
+console.log('Supply:', mintAccount.data.supply);
```
mikemaccana:
```suggestion
console.log('Supply:', mintAccount.data.supply);

console.log(`✅ Success! You've connected to Solana, funded a wallet, created a token mint, and read its data.`)
```

mikemaccana: It's just more fun if people get a big green tick when they win! 😃

docs/content/docs/getting-started.mdx
mikemaccana: Plus airdrop limits can be a hassle on devnet, and a keypair lets them persist the SOL they have.

docs/content/docs/getting-started.mdx
lorisleiva: Yeah the intention was to show how to handle multiple instruction / instruction plans in one transaction. maybe transfer SOL and memo is best since transfer token requires quite a lot of set up for a small getting started guide (we can add that to the recipes).

But yeah ideally I'd like to show-case:
- `client.<program>.instruction.<instruction>.sendTransaction()`
- `client.<program>.instruction.<instructionPlan>.sendTransaction()`
- `client.sendTransaction([<multiple instructions/instructionPlans>])`.

So far, we're missing the middle one so I just need to find a nice way to document this.

docs/content/docs/getting-started.mdx
lorisleiva: We'll have something like `.use(essentialPrograms())` soon which will fix this.

docs/content/docs/getting-started.mdx
lorisleiva: I think it's a good point that we should probably introduce `client.sendTransaction([...])` first so they get used to the transaction/instruction model and only after we can say something like btw, here's a shortcut for you.

docs/content/docs/getting-started.mdx
lorisleiva: I like the idea but as discussed offline, these should be separate composable primitives.

PR #1521, opened 2026-04-06
fix: error code had one extra zero

ChALkeR:
The reserved range is `3230000-3230999`
Neighboring codes were `3230000` and `3230002`
`32300001` was a mistype that should have been `3230001`


mcintyre94:
Thankyou for this! You're correct but unfortunately we can't update error codes like this. We need developers to be able to decode this error using any future version of the kit `@solana/errors` CLI, and if we change or remove errors codes then this will break. 

Luckily this shouldn't cause any real issues. We would get a compile error if we tried to re-use this error code in error context or messages. And the grouping of errors codes is just a code organisation thing, rather than being used for decoding the error.

I'll close this because we unfortunately can't do anything about the incorrect error code.

PR #1524, opened 2026-04-07
Add reactive() for pending subscriptions, that returns a reactive store for the subscription

mcintyre94:
#### Summary of Changes

This PR adds an alternative API for RPC subscriptions:

```typescript
const store = rpcSubscriptions.slotNotifications().reactive({ abortSignal: abortController.signal })
```

This returns an object with fields `getState()`, `getError()` and `subscribe()`

- `getState()` returns the latest published value from the subscription, or `undefined` until a value is published
- `getError()` returns the first error from the subscription, or `undefined` if there has not been any error
- `subscribe(listener)`registers a function to be called if either data or an error occurs

Together they enable reactive frameworks to consume subscriptions:, eg react with an error boundary

```typescript
const slot = useSyncExternalStore(store.subscribe, () => {
  if (store.getError()) throw store.getError();
  return store.getState();
});
```

Or without error boundaries:

```typescript
const slot = useSyncExternalStore(store.subscribe, store.getState)
const error = useSyncExternalStore(store.subscribe, store.getError)
```

And svelte:

```typescript
let slot = $state(store.getState());
let error = $state(store.getError());
$effect(() => store.subscribe(() => {
    slot = store.getState();
    error = store.getError();
}));
```

Vanilla JS:

```typescript
store.subscribe(() => {
    if (store.getError()) console.error(store.getError());
    else el.textContent = store.getState().slot;
});
```

It's also compatible with libraries like SWR subscriptions:

```typescript
useSWRSubscription('slot', (key, { next }) => {
    const abortController = new AbortController();
    rpcSubscriptions
        .slotNotifications()
        .reactive({ abortSignal: abortController.signal })
        .then(store => {
            store.subscribe(() => {
                const error = store.getError();
                if (error) next(error);
                else next(null, store.getState());
            });
        });
    return () => abortController.abort();
});
```

Note that we intentionally separate error handling, and `getState()` never throws. This gives consumers maximum flexibility and allows access to the stale latest value while recovering from an error.

This change is purely additive to the existing async generator API for subscriptions, and architecturally lives alongside it as an alternative interface to the `DataPublisher` abstraction. It's intended to make subscriptions easier to use in reactive UIs.

### DIFF
--- a/packages/subscribable/src/reactive-store.ts
+++ b/packages/subscribable/src/reactive-store.ts
@@ -1,0 +1,146 @@
+import { AbortController } from '@solana/event-target-impl';
+
+import { DataPublisher } from './data-publisher';
+
+type Config = Readonly<{
+    /**
+     * Triggering this abort signal will cause the store to stop updating and will disconnect it from
+     * the underlying data publisher.
+     */
+    abortSignal: AbortSignal;
+    /**
+     * Messages from this channel of `dataPublisher` will be used to update the store's state.
+     */
+    dataChannelName: string;
+    // FIXME: It would be nice to be able to constrain the type of `dataPublisher` to one that
+    //        definitely supports the `dataChannelName` and `errorChannelName` channels, and
+    //        furthermore publishes `TData` on the `dataChannelName` channel. This is more difficult
+    //        than it should be: https://tsplay.dev/NlZelW
+    dataPublisher: DataPublisher;
+    /**
+     * Messages from this channel of `dataPublisher` will cause subscribers to be notified without
+     * updating the state, so that they can respond to the error condition.
+     */
+    errorChannelName: string;
+}>;
+
+/**
+ * A reactive store that holds the latest value published to a data channel and allows external
+ * systems to subscribe to changes. Compatible with `useSyncExternalStore`, Svelte stores, Solid's
+ * `from()`, and other reactive primitives that expect a `{ subscribe, getState }` contract.
+ *
+ * @example
+ * ```ts
+ * // React — throw error from snapshot function to surface via Error Boundary
+ * const state = useSyncExternalStore(store.subscribe, () => {
+ *     if (store.getError()) throw store.getError();
+ *     return store.getState();
+ * });
+ *
+ * // Vue — check error reactively in a composable
+ * const data = shallowRef(store.getState());
+ * const error = shallowRef(store.getError());
+ * store.subscribe(() => {
+ *     data.value = store.getState();
+ *     error.value = store.getError();
+ * });
+ * ```
+ *
+ * @see {@link createReactiveStoreFromDataPublisher}
+ */
+export type ReactiveStore<T> = {
+    /**
+     * Returns the error published to the error channel, or `undefined` if no error has occurred.
+     * Once set, the error is preserved — subsequent errors do not overwrite it.
+     */
+    getError(): unknown;
+    /**
+     * Returns the most recent value published to the data channel, or `undefined` if no
+     * notification has arrived yet. On error, continues to return the last known value.
+     */
+    getState(): T | undefined;
+    /**
+     * Registers a callback to be called whenever the state changes or an error is received.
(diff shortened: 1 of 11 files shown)


mcintyre94:
* **#1524** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1524?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1525](https://github.com/anza-xyz/kit/pull/1525) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1525?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1528](https://github.com/anza-xyz/kit/pull/1528) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1528?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1524?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/subscribable/src/reactive-store.ts
```diff
@@ -0,0 +1,146 @@
...
+        data => {
+            currentState = data as TData;
+            subscribers.forEach(cb => cb());
+        },
+        options,
+    );
+    dataPublisher.on(
+        errorChannelName,
+        err => {
+            if (currentError !== undefined) return;
```
lorisleiva, line 123: That's the only part of this design I'm not entirely sure about as it introduces a different behaviour than the underlying data emitter. That being said, I believe this is how the async generator behaves, that is once the first error is thrown, we leave the for-loop and therefore stop subscribing. Although here we technically allow the user to ignore these errors by only relying on `getState` if they want to. Therefore I wonder if we should have `getError` return the latest error thrown instead of the first one?

I don't feel too strongly about that, I just want to make sure we don't end up in a situation where the user keeps reading data from the publisher despite some earlier errors and later on the user being confused about `getError()` not reflecting what they would expect because we only kept the first error.

lorisleiva (approved):
This is amazing! And so much cleaner than what I thought it was going to be. Just left a comment on the error handling side but happy to merge as-is.

packages/subscribable/src/reactive-store.ts
mcintyre94: So the actual design decision here is to match the async generator and abort on the first error, in this block we do `abortController.abort(err)`. That's the signal that's passed to `dataPublisher.on` (as options), which passes it to `addEventListener`, so we get unsubscribed when it aborts there. So the consumer can't stay subscribed and keep reading `getState`, the subscriptions are ended on this first error. 

The check there is just defensive against a race condition/edge case where another error gets published during that tear down. I suspect it's actually redundant because unsubscribing event listeners should probably be synchronous, but probably worth keeping just in case.

I think we could design this to be recoverable and not tear everything down on the first error, but it gets really tricky to reason about the different combinations of fresh/stale data/errors. I think something like a resubscribe on error wrapper would be better if we wanted to support that use case.

So the current design when we get an error is that we persist that error, unsubscribe our listeners, and then notify our subscribers. This is the last time we'll notify them because our listeners have gone. If they call `getState()` they'll get the last data (if any ever published) and when they call `getError()` they'll get the error.

lorisleiva: Oh gotcha that makes a lot more sense. It didn't click with me that, after the first error, we would abort the whole thing and therefore unsubscribe from the transport anyway. Then I think you design is perfect.

mcintyre94:
### Merge activity

* **Apr 8, 1:12 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1524).
* **Apr 8, 1:12 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1524).

PR #1525, opened 2026-04-07
Add a subscription for current slot to example react app

mcintyre94:
#### Summary of Changes

This PR updates the react example app to include a component that subscribes to the `slotSubscriptions` RPC subscription method, and displays the current slot. Simple example to demonstrate the new reactive API for subscriptions.

### DIFF
--- a/examples/react-app/package.json
+++ b/examples/react-app/package.json
@@ -34,8 +34,8 @@
         "@vitejs/plugin-react-swc": "^4.2.3",
         "eslint": "^9.39.2",
         "eslint-plugin-react-refresh": "^0.5.2",
-        "jest": "^30.0.0-alpha.6",
         "globals": "^17.3.0",
+        "jest": "^30.0.0-alpha.6",
         "vite": "^7.3.1"
     }
 }
--- a/examples/react-app/src/components/SlotIndicator.tsx
+++ b/examples/react-app/src/components/SlotIndicator.tsx
@@ -1,0 +1,61 @@
+import { Link, Text } from '@radix-ui/themes';
+import { useContext, useEffect, useState, useSyncExternalStore } from 'react';
+
+import { ChainContext } from '../context/ChainContext';
+import { RpcContext } from '../context/RpcContext';
+
+type SlotNotification = Readonly<{
+    parent: bigint;
+    root: bigint;
+    slot: bigint;
+}>;
+
+type ReactiveStore<T> = {
+    getError(): unknown;
+    getState(): T | undefined;
+    subscribe(callback: () => void): () => void;
+};
+
+function createNoopStore(error?: unknown): ReactiveStore<SlotNotification> {
+    return {
+        getError: () => error,
+        getState: () => undefined,
+        subscribe: () => () => { },
+    };
+}
+
+const slotFormatter = new Intl.NumberFormat();
+
+export function SlotIndicator() {
+    const { rpcSubscriptions } = useContext(RpcContext);
+    const { chain, solanaExplorerClusterName } = useContext(ChainContext);
+    const [store, setStore] = useState<ReactiveStore<SlotNotification>>(createNoopStore);
+
+    useEffect(() => {
+        const abortController = new AbortController();
+        rpcSubscriptions
+            .slotNotifications()
+            .reactive({ abortSignal: abortController.signal })
+            .then(setStore)
+            .catch(e => setStore(createNoopStore(e)));
+        return () => abortController.abort();
+    }, [rpcSubscriptions, chain]);
+
+    const slot = useSyncExternalStore(store.subscribe, () => {
+        if (store.getError()) throw store.getError();
+        return store.getState();
+    });
+
+    if (!slot) {
+        return <Text>{'\u2013'}</Text>;
+    }
+
+    return (
+        <Link
+            href={`https://explorer.solana.com/block/${slot.slot}?cluster=${solanaExplorerClusterName}`}
+            target="_blank"
+        >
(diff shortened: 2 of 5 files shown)


mcintyre94:
* **#1525** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1525?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1525?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1524** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1524?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1528](https://github.com/anza-xyz/kit/pull/1528) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1528?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

examples/react-app/package.json
```diff
@@ -19,6 +19,7 @@
         "@solana-program/system": "^0.12.0",
         "@solana/kit": "workspace:*",
         "@solana/react": "workspace:*",
+        "@solana/subscribable": "workspace:*",
```
mcintyre94: Just to flag - this is because the `ReactiveStore` type is in `subscribable` and we don't re-export that from Kit  
  
I think this is probably ok, this dependency would be in a react helper library rather than apps for the most part, and doesn't necessarily feel core. But we could re-export `subscribable` (just this type from it?) if we want to, I don't have a strong view either way

examples/react-app/src/components/SlotIndicator.tsx
```diff
@@ -0,0 +1,40 @@
...
+            .slotNotifications()
+            .reactive({ abortSignal: abortController.signal })
+            .then(setStore);
+        return () => abortController.abort();
+    }, [rpcSubscriptions, chain]);
+
+    const slot = useSyncExternalStore(store.subscribe, () => {
+        if (store.getError()) throw store.getError();
+        return store.getState();
+    });
```
mcintyre94: In reality I'd expect everything here (including `NOOP_STORE` to be in a helper library that just exposes `slot` as `useSlotSubscription` or something like that

examples/react-app/src/components/SlotIndicator.tsx
```diff
@@ -0,0 +1,56 @@
...
+        rpcSubscriptions
+            .slotNotifications()
+            .reactive({ abortSignal: abortController.signal })
+            .then(setStore)
+            .catch(e => setStore(createNoopStore(e)));
+        return () => abortController.abort();
+    }, [rpcSubscriptions, chain]);
+
+    const slot = useSyncExternalStore(store.subscribe, () => {
+        if (store.getError()) throw store.getError();
```
lorisleiva, line 45: So if the publisher throws an error, this whole component throws right? Is the idea that users can wrap this component in some of generic retry component if needed?

examples/react-app/package.json
lorisleiva: I don't mind too much. Since this is just a type though could we need get it via some `typeof` references?

examples/react-app/src/components/SlotIndicator.tsx
lorisleiva: Yeah we could even have a generic react hook like `useRpcNotification(rpcSubscriptions.slotNotifications())` which all the specific hook (e.g. `useSlotSubscription`) could use internally to make them one-liners.

lorisleiva (approved):
Nice! Thanks for adding a concrete example.

examples/react-app/src/components/SlotIndicator.tsx
mcintyre94: In this example yep the component throws. You could wrap it in a retry component, in this example we just wrap it in a React error boundary:

```
<ErrorBoundary fallback={<Text>&ndash;</Text>} key={chain}>
  <SlotIndicator />
</ErrorBoundary>
```

Without something like error boundaries you'd probably want a wrapper that separately tracks `getState` and `getError` but that'd be framework-specific.

examples/react-app/src/components/SlotIndicator.tsx
mcintyre94: Agreed - and all of that would be genuinely react-specific logic, so it feels like we have the right boundary here.

examples/react-app/package.json
mcintyre94: I've just inlined the type for now, since we have it with `createNoopStore` anyway. We can figure out if it should be exported from Kit later, easy enough to do if we want to.

examples/react-app/src/components/SlotIndicator.tsx
lorisleiva: Actually it might be best to accept a `ReactiveStore` from that hook (e.g. `useReactiveStore`) so that we can also plug your higher level `createReactiveStoreFromRpcAndRpcSubscriptions` helper into it.

mcintyre94:
### Merge activity

* **Apr 8, 1:12 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1525).
* **Apr 8, 1:14 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1525) rebased this pull request as part of a merge.
* **Apr 8, 1:24 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1525).

PR #1527, opened 2026-04-07
Make typescript a devDependency

djrenren:
#### Problem

Listing typescript as a peer dependency prevents consumers from using their own version of typescript (e.g. typescript 6). This change marks typescript as a `devDependency` so that consumers are unconstrained.

Fixes #1511

### DIFF
--- a/.changeset/clean-ants-end.md
+++ b/.changeset/clean-ants-end.md
@@ -1,0 +1,50 @@
+---
+'@solana/accounts': patch
+'@solana/addresses': patch
+'@solana/assertions': patch
+'@solana/codecs': patch
+'@solana/codecs-core': patch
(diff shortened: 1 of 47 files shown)


lorisleiva:
The optional peer dependency is set to ensure consumer apps won't break as we use certain modern TypeScript features by requiring a minimum version. We simply need to bump the peer dependency version to also include version 6.0.0 and above.

mcintyre94:
Agree with @lorisleiva . This is my mistake in #1187 - I changed it from `>=5.9.3` to `^5.0.0` but it should have been `>=5.0.0`

mcintyre94:
Updated to `>=5` in #1529 - thanks for flagging this!

PR #1528, opened 2026-04-08
Add a function `createReactiveStoreFromRpcAndSubscription`

mcintyre94:
#### Problem

In my earlier PR I added an API to get a reactive store (subscribe, getState, getError) for RPC subscriptions. But another case, which is more complex for apps, is where you want to get the current state and then keep it updated with a subscription. An example of this was the balance functionality in our react app, which makes a call to `getBalance`and then subscribes to account updates, and keeps track of the slot to always accept only the most recent data. Just having a reactive subscription is insufficient for this use case.

#### Summary of Changes

This PR adds a new function `createReactiveStoreFromRpcAndSubscription`. It's added to `pakcages/kit` since it combines RPC and subscriptions.

It takes as input:

- An RPC request that returns a `SolanaRpcResponse`, ie provides a slot in context
- A map function from the response shape of that request, to `TItem`
- An RPC subscription request that returns notifications with `SolanaRpcResponse`, providing a slot in context
- A map function from the notification shape of that request, to `TItem`

It then returns a `ReactiveStore` as defined in the previous PR which provides the latest data (by slot number) from either data source, and notifies subscribers only when newer data is available (or an error occurs). It handles the unified abort signal and error handling. 

This generalises the approach used in `balance.ts`, while providing an output shape compatible with any reactive framework.

The PR also updates the `balance.ts` in the react-app to demonstrate this, binding the store to an SWR subscription.

This would enable frameworks targeting react, SWR, or any other reactive library etc to provide bindings like `useBalance()` with minimal framework logic.

Outside of balance, it would be useful for any case where we have an RPC and a subscription providing the same data that can change. Anything on accounts (getAccountInfo + account notifications), and potentially also transaction signatures (`getSignatureStatuses`\+ signature notifications). 

I also realised afterward that we could provide an async generator API for this, which only yields values when the slot is newer. Same logic, but presented as an async generator API. I'll add that later as a follow up PR.

### DIFF
--- a/packages/kit/src/create-reactive-store-from-rpc-and-subscription.ts
+++ b/packages/kit/src/create-reactive-store-from-rpc-and-subscription.ts
@@ -114,6 +114,7 @@ export function createReactiveStoreFromRpcAndSubscription<TRpcValue, TSubscripti
     rpcRequest
         .send({ abortSignal: signal })
         .then(({ context: { slot }, value }) => {
+            if (signal.aborted) return;
             if (slot < lastUpdateSlot) return;
             lastUpdateSlot = slot;
             currentState = rpcValueMapper(value);
@@ -128,6 +129,7 @@ export function createReactiveStoreFromRpcAndSubscription<TRpcValue, TSubscripti
                 context: { slot },
                 value,
             } of notifications) {
+                if (signal.aborted) return;
                 if (slot < lastUpdateSlot) continue;
                 lastUpdateSlot = slot;
                 currentState = subscriptionValueMapper(value);
--- a/packages/kit/src/create-reactive-store-from-rpc-and-subscription.ts
+++ b/packages/kit/src/create-reactive-store-from-rpc-and-subscription.ts
@@ -1,0 +1,153 @@
+import type { PendingRpcRequest } from '@solana/rpc';
+import type { PendingRpcSubscriptionsRequest } from '@solana/rpc-subscriptions';
+import type { SolanaRpcResponse } from '@solana/rpc-types';
+import type { ReactiveStore } from '@solana/subscribable';
+
+type CreateReactiveStoreFromRpcAndSubscriptionConfig<TRpcValue, TSubscriptionValue, TItem> = Readonly<{
+    /**
+     * Triggering this abort signal will cancel the pending RPC request and subscription, and
+     * disconnect the store from further updates.
+     */
+    abortSignal: AbortSignal;
+    /**
+     * A pending RPC request whose response will be used to set the store's initial state.
+     * The response must be a {@link SolanaRpcResponse} so that its slot can be compared with
+     * subscription notifications.
+     */
+    rpcRequest: PendingRpcRequest<SolanaRpcResponse<TRpcValue>>;
+    /**
+     * Maps the value from the RPC response to the item type stored in the reactive store.
+     */
+    rpcValueMapper: (value: TRpcValue) => TItem;
+    /**
+     * A pending RPC subscription request whose notifications will be used to keep the store
+     * up to date. Each notification must be a {@link SolanaRpcResponse} so that its slot can be
+     * compared with the initial RPC response and other notifications.
+     */
+    subscriptionRequest: PendingRpcSubscriptionsRequest<SolanaRpcResponse<TSubscriptionValue>>;
+    /**
+     * Maps the value from a subscription notification to the item type stored in the reactive store.
+     */
+    subscriptionValueMapper: (value: TSubscriptionValue) => TItem;
+}>;
+
+/**
+ * Creates a {@link ReactiveStore} that combines an initial RPC fetch with an ongoing subscription
+ * to keep its state up to date.
+ *
+ * The store uses slot-based comparison to ensure that only the most recent value is kept,
+ * regardless of whether it came from the initial RPC response or a subscription notification.
+ * This prevents stale data from overwriting newer data when the RPC response and subscription
+ * notifications arrive out of order.
+ *
+ * Things to note:
+ *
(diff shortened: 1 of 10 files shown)


mcintyre94:
* **#1528** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1528?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1528?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1524** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1524?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1525](https://github.com/anza-xyz/kit/pull/1525) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1525?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

packages/kit/src/create-reactive-store-from-rpc-and-subscription.ts
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * A pending RPC subscription request whose notifications will be used to keep the store
+     * up to date. Each notification must be a {@link SolanaRpcResponse} so that its slot can be
+     * compared with the initial RPC response and other notifications.
+     */
+    subscriptionRequest: PendingRpcSubscriptionsRequest<SolanaRpcResponse<TSubscriptionValue>>;
+    /**
+     * Maps the value from a subscription notification to the item type stored in the reactive store.
+     */
+    subscriptionValueMapper: (value: TSubscriptionValue) => TItem;
```
lorisleiva: nit: these should be prefixed with `rpc` to stay consistent with the `rpcSubscriptions` naming convention.


```suggestion
    rpcSubscriptionRequest: PendingRpcSubscriptionsRequest<SolanaRpcResponse<TSubscriptionValue>>;
    /**
     * Maps the value from a subscription notification to the item type stored in the reactive store.
     */
    rpcSubscriptionValueMapper: (value: TSubscriptionValue) => TItem;
```

packages/kit/src/create-reactive-store-from-rpc-and-subscription.ts
```diff
@@ -0,0 +1,155 @@
...
+ * const unsubscribe = balanceStore.subscribe(() => {
+ *     const error = balanceStore.getError();
+ *     if (error) console.error('Error:', error);
+ *     else console.log('Balance:', balanceStore.getState());
+ * });
+ * ```
+ *
+ * @see {@link ReactiveStore}
+ */
+export function createReactiveStoreFromRpcAndSubscription<TRpcValue, TSubscriptionValue, TItem>({
```
lorisleiva: I wonder if the name is clear enough to convey that it's an RPC Subscription with an initial data fetch. Probably something like `createReactiveStoreWithInitialValue` would be more explicit?

Additionally, a big portion of that value added by this function is tracking slots to ensure we only keep newer data regardless of the message ordering. So maybe even something like `createReactiveStoreWithInitialValueAndSlotTracking`?

packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts
```diff
@@ -0,0 +1,155 @@
...
+            for await (const {
+                context: { slot },
+                value,
+            } of notifications) {
+                if (signal.aborted) return;
+                if (slot < lastUpdateSlot) continue;
+                lastUpdateSlot = slot;
+                currentState = subscriptionValueMapper(value);
+                notifySubscribers();
+            }
```
lorisleiva, line 141: Interesting, why not using the `reactive` API here since we're essentially wrapping a reactive API into another?

lorisleiva (approved):
Nice! I love that you can wrap a reactive store in another reactive store whilst providing an RPC pending request for the initial fetch. 👌

packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts
mcintyre94: I think the async generator API is a bit nicer in this case, and is generally likely to be better for any kind of processing of the notifications like this. I think the equivalent with a reactive store would be:

```ts
subscriptionRequest
    .reactive({ abortSignal: signal })
    .then(subscriptionStore => {
        subscriptionStore.subscribe(() => {
            const error = subscriptionStore.getError();
            if (error) { handleError(error); return; }
            const notification = subscriptionStore.getState();
            if (!notification || signal.aborted) return;
            const { context: { slot }, value } = notification;
            if (slot < lastUpdateSlot) return;
            lastUpdateSlot = slot;
            currentState = subscriptionValueMapper(value);
            notifySubscribers();
        });
    })
    .catch(handleError);
```

You add indirection by having to subscribe, I think the for loop syntax is nicer and clearer here. And the error handling that makes the store so flexible for UI is less helpful here than the generator that throws.

packages/kit/src/create-reactive-store-from-rpc-and-subscription.ts
mcintyre94: That's fair, it probably makes sense to be verbose and use `createReactiveStoreWithInitialValueAndSlotTracking` here.

packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts
lorisleiva: Gotcha! Yeah that does look a bit less clear than the for-loop API.

PR #1529, opened 2026-04-08
Change typescript peer dependency to >5

mcintyre94:
#### Problem

A while ago we changed our minimum typescript version to 5, but also changed it from `>=5.x.y` to `^5`. See #1187 

This means it doesn't match typescript 6.x or future versions

#### Summary of Changes

Change the version to `>=5.0.0`. This matches any version of 5.x or any future version of typescript


lorisleiva (approved):
Thanks!

PR #1531, opened 2026-04-08
Add `extractable` option to `generateKeyPair` and `generateKeyPairSigner`

lorisleiva:
This PR adds an optional `extractable` boolean argument to `generateKeyPair` in `@solana/keys` and threads it through `generateKeyPairSigner` in `@solana/signers`. The argument defaults to `false`, preserving the existing secure-by-default behavior that prevents the bytes of the private key from being visible to JS. When set to `true`, the generated private key can be exported via `crypto.subtle.exportKey()`, which is useful for scenarios like persisting a generated key pair. This brings `generateKeyPair` and `generateKeyPairSigner` in line with the sibling helpers `createKeyPairFromBytes`, `createKeyPairFromPrivateKeyBytes`, `createKeyPairSignerFromBytes`, and `createKeyPairSignerFromPrivateKeyBytes`, which already accept an `extractable` parameter.

### DIFF
--- a/.changeset/clean-breads-march.md
+++ b/.changeset/clean-breads-march.md
@@ -1,0 +1,6 @@
+---
+'@solana/signers': minor
+'@solana/keys': minor
+---
+
+Add an optional `extractable` argument to `generateKeyPair` and `generateKeyPairSigner`. It defaults to `false`, preserving the existing secure-by-default behavior, but can be set to `true` when you need to export the generated private key bytes via `crypto.subtle.exportKey()`.
--- a/packages/keys/src/__tests__/key-pair-test.ts
+++ b/packages/keys/src/__tests__/key-pair-test.ts
@@ -39,6 +39,16 @@ describe('key-pair', () => {
             const { privateKey } = await generateKeyPair();
             expect(privateKey).toHaveProperty('extractable', false);
         });
+        it('generates a non-extractable private key when `extractable` is explicitly `false`', async () => {
+            expect.assertions(1);
+            const { privateKey } = await generateKeyPair(false);
+            expect(privateKey).toHaveProperty('extractable', false);
+        });
+        it('generates an extractable private key when `extractable` is `true`', async () => {
+            expect.assertions(1);
+            const { privateKey } = await generateKeyPair(true);
+            expect(privateKey).toHaveProperty('extractable', true);
+        });
         it('generates a private key usable for signing operations', async () => {
             expect.assertions(1);
             const { privateKey } = await generateKeyPair();
--- a/packages/keys/src/key-pair.ts
+++ b/packages/keys/src/key-pair.ts
@@ -15,6 +15,10 @@ import { signBytes, verifySignature } from './signatures';
  * Generates an Ed25519 public/private key pair for use with other methods in this package that
  * accept [`CryptoKey`](https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey) objects.
  *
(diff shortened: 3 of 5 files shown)


lorisleiva:
* **#1531** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1531?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1531?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

trevor-cortex (commented):
## Summary

Adds an optional `extractable: boolean = false` parameter to `generateKeyPair` (`@solana/keys`) and threads it through `generateKeyPairSigner` (`@solana/signers`). When `true`, the generated Ed25519 private key can be exported via `crypto.subtle.exportKey()`. The default remains `false`, preserving the existing secure-by-default behavior.

This aligns the signatures of `generateKeyPair` / `generateKeyPairSigner` with their sibling helpers (`createKeyPairFromBytes`, `createKeyPairFromPrivateKeyBytes`, `createKeyPairSignerFromBytes`, `createKeyPairSignerFromPrivateKeyBytes`), which already expose the same flag with the same default.

## What to watch out for

- **Backwards compatible**: the new parameter is optional and defaults to `false`, so existing call sites keep their current (non-extractable) behavior. The changeset is correctly marked as a `minor` bump for both packages.
- **Secure-by-default preserved**: the old inline comment "Prevents the bytes of the private key from being visible to JS" is now captured in the TSDoc for the `extractable` parameter, so the security rationale isn't lost.
- **Consistency**: parameter name, default, and TSDoc wording match the sibling helpers in `packages/keys/src/key-pair.ts`.

## Test coverage

- `packages/keys/src/__tests__/key-pair-test.ts` adds two cases: explicit `false` and explicit `true`, asserting the resulting `privateKey.extractable` flag.
- `packages/signers/src/__tests__/keypair-signer-test.ts` updates the existing default-call test to additionally assert `generateKeyPair` was called with `false`, and adds a new test verifying that `generateKeyPairSigner(true)` forwards `true` to `generateKeyPair`.

- Worth a quick sanity check that the docs site / API reference picks up the new `@param extractable` TSDoc block for both functions.
- No runtime changes beyond the new parameter pass-through — the `crypto.subtle.generateKey` call now receives `extractable` directly instead of a hardcoded `false`, which is the only behavioral change.

Looks good to me. 🚀

beeman:
Super happy to finally see this as I think it means I can update my [answer here](https://solana.stackexchange.com/a/18633/98) 🙏

lorisleiva:
@beeman Yeah the other factory helpers already have that `extractable` option anyway so it's more consistent that way and we of course keep the non-extractable behaviour by default. I'm about to add grind and write helpers as well.

beeman:
@lorisleiva amazing! Way better that these are part of the official package so that people don't have to write ad-hoc implementations of this. 🙏

mcintyre94 (approved):
I'm happy with this, I think it makes sense to have as an option, with false as the default

I think the argument not to do this would be that with the `FromBytes` helpers, you already have the bytes in insecure JS when you call it, so extracting doesn't add any additional risk. Whereas when you generate, assuming a sane browser implementation, those bytes are never available insecurely. 

That said if you have malicious JS in your sandbox (browser extension etc) then it can hijack the crypto APIs anyway so I don't think anything we can do actually holds up to that threat.

PR #1534, opened 2026-04-08
Add `grindKeyPair` and `grindKeyPairSigner` for mining vanity addresses

lorisleiva:
This PR adds `grindKeyPair`, `grindKeyPairs`, `grindKeyPairSigner`, and `grindKeyPairSigners` for mining vanity Solana addresses whose base58-encoded public key matches a `RegExp` or a custom predicate. Key pairs are generated in parallel batches of `concurrency` (defaulting to `32`), with each batch also exporting and base58-decoding the public key bytes concurrently so the matching phase is fully synchronous. Cancellation is handled via `getAbortablePromise` so that firing the `abortSignal` rejects the grind immediately without waiting for in-flight key generations to settle. Regex matchers have their literal characters validated against the base58 alphabet up front — after stripping escapes, character classes, quantifiers, and groups — to catch common typos like `/^sol0/` before an infinite grind starts, with the new `SOLANA_ERROR__KEYS__INVALID_BASE58_IN_GRIND_REGEX` error code. The signer variants are thin wrappers around the key-pair variants that pipe results through `createSignerFromKeyPair`.


lorisleiva:
* **#1537** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1537?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1534** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1534?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

trevor-cortex (commented):
## Summary

This PR adds `grindKeyPair`/`grindKeyPairs` in `@solana/keys` and `grindKeyPairSigner`/`grindKeyPairSigners` in `@solana/signers` for mining vanity Solana addresses. The implementation generates key pairs in parallel batches of `concurrency` (default 32), tests each candidate against a `RegExp` or predicate, and supports `abortSignal`-based cancellation via `getAbortablePromise` so the grind rejects immediately without waiting for in-flight generations. Regex matchers are statically validated against the base58 alphabet up front (after stripping escapes, character classes, quantifiers, and groups) to catch typos like `/^sol0/`, surfaced via a new `SOLANA_ERROR__KEYS__INVALID_BASE58_IN_GRIND_REGEX` error code. The signer variants are thin wrappers around the key-pair variants that pipe results through `createSignerFromKeyPair`.

## Things to watch

- **Regex validation soundness.** The strip-then-check approach in `assertGrindRegexIsValid` is pragmatic but lossy. I've flagged a few specific cases inline (notably nested groups and `/\0/`-style escapes whose *semantics* can never match base58). No hard correctness bugs that I can construct, but the comment "cannot be analyzed reliably without a full regex parser" is accurate — reviewers should decide if they're comfortable with the current trade-offs.
- **Redundant address derivation in the signer path.** `grindKeyPairs` exports + base58-decodes every candidate's public key during matching, and then `createSignerFromKeyPair` calls `getAddressFromPublicKey` *again* for each winning key pair. Minor; mentioned inline.
- **Test flakiness.** The `grindKeyPair` tests that use the real `generateKeyPair` with `/^1/` are probabilistic (~1/58 per key). With `concurrency` 32 this should virtually always finish within Jest's default timeout, but it's worth noting.
- **`concurrency` overshoot.** A request for `amount: 1` with default `concurrency: 32` always runs a full batch, so the `"generates key pairs in batches of concurrency"` test is correct — but callers targeting rare patterns with very small `amount` pay for a full batch even after a match is found mid-batch. Documented behaviour, just flagging it.

- Please double-check the regex-stripping logic against your own pathological inputs — especially nested groups, escaped brackets/parens (`/\[0\]/`), and things like `/\d/` (currently skipped via the escape rule, which is the right call but worth confirming).
- Confirm `crypto.subtle.exportKey('raw', publicKey)` works for Ed25519 on all supported runtimes. Public keys are always extractable regardless of the `extractable` flag, so this should be fine, but the code path is new.
- The new `@solana/promises` workspace dep in `packages/keys/package.json` should be sanity-checked against the overall dependency graph policy for `@solana/keys`.
- Docs look thorough and the examples match the API surface. Changeset is correct (minor bump on three packages).

trevor-cortex:
Re: `STRIP_UNVALIDATED_REGEX_PARTS` — a few edge cases to consider for `packages/keys/src/grind-keypair.ts`:

1. **Nested groups aren't fully stripped.** `[^)]*` stops at the first `)`, so for `/((a)0)/` the strip consumes `((a)` and leaves `0)` → `0` is then rejected. The rejection happens to be *correct* here (outer group literally requires `a0`), but the reasoning inside the validator is wrong: the `0` is not actually at the top level of the regex. For a less contrived example, consider something like `/(a(b)c)/` — the strip consumes `(a(b)` leaving `c)`, which happens to be fine, but the behaviour is fragile. A hand-rolled depth-counting scan over the source would be more robust than a single regex with `[^)]*`, and wouldn't cost much.

2. **Escaped literal brackets leak their contents.** `/\[foo\]/` has source `\[foo\]`. The `\\.` alternative matches `\[` and `\]` first, so the strip leaves `foo`. That's fine because `foo` is base58, but `/\[foo0\]/` would surface a validation error whose `character: '0'` message refers to a character that is *inside* what the user sees as a literal `[foo0]`. The rejection is still semantically correct (the regex can never match an address containing `[foo0]`), but the error message may be confusing. Worth a comment, or adjusting the error wording.

3. **`/\0/` is accepted.** The test `{ regex: /\0/ }` is in the accepted list — `\0` is the NUL character, which of course never appears in a base58 address, so this regex is guaranteed to never match. The escape strip makes this pass validation, which is a false negative. Not a blocker, and detecting this generally is hard, but you might want a separate check for `\0`, `\n`, `\t`, `\x00`, `\u0000`, etc. — or just document that the validator only catches literal-character typos, not semantically-impossible regexes.

4. **`BASE58_ALPHABET_REGEX` uses `*` instead of `+`.** `/^[1-9A-HJ-NP-Za-km-z]*$/` accepts the empty string. In practice `isBase58Character` is only ever called with a single code point from `for (const character of stripped)`, so this is harmless — but `+` would be marginally more defensive.

None of these are blockers; the current implementation correctly rejects the common typos it's designed to catch. Just flagging for visibility.

EDIT (Loris): Hmm I'm pretty sure the regex is greedy by default and therefore it will eat the whole group and everything inside of it.

trevor-cortex:
Minor perf note on `packages/signers/src/grind-keypair-signer.ts`:

`grindKeyPairs` already does `crypto.subtle.exportKey('raw', keyPair.publicKey)` + base58 decode for every candidate key pair during the grind loop. Then `createSignerFromKeyPair` (called via `Promise.all(keyPairs.map(createSignerFromKeyPair))`) calls `getAddressFromPublicKey(keyPair.publicKey)`, which performs the same export + decode again for each *winning* key pair.

For `amount: 1` this is negligible (one extra export), but for large `amount` it's N redundant exports. If you want to eliminate it, `grindKeyPairs` could optionally return `{ keyPair, address }` tuples (or expose an internal variant that does), and the signer wrapper could construct the signer without re-deriving the address. Happy to leave this for a follow-up — just flagging since the grind path is explicitly optimized for throughput.

trevor-cortex:
Re: `packages/keys/src/__tests__/grind-keypair-test.ts` — the two tests that use the real `generateKeyPair` (`'returns a key pair whose address matches the provided regex'` and `'returns a key pair whose address satisfies the provided predicate'`) both match `/^1/`, which has ~1/58 probability per candidate. With `concurrency: 32` that's >42% chance per batch, so median time to find is ~1 batch, but the tail is long — there's a real chance of 10+ batches occasionally. At default Jest timeouts this should be fine, but if CI is tight you might prefer `/^[1-9]/` (prob ~9/58 ≈ 15.5%) for robustness, or mock `generateKeyPair` and feed it deterministic keys.

The other tests correctly mock `generateKeyPair` with `() => true` matchers, which is the right pattern for the rest of the suite.

packages/keys/src/grind-keypair.ts
```diff
@@ -0,0 +1,301 @@
...
+     */
+    matches: GrindKeyPairMatches;
+}>;
+
+// Matches regex parts that should be skipped during base58 validation:
+//   - `\<any>` escape sequences
+//   - `[...]` character classes
+//   - `{...}` quantifiers
+//   - `(...)` groups
+const STRIP_UNVALIDATED_REGEX_PARTS = /\\.|\[[^\]]*\]|\{[^}]*\}|\([^)]*\)/g;
```
mcintyre94, line 71: Yo dawg, I heard you like regex 😅

mcintyre94 (approved):
Looks good! Not sure if it'd just be worth a comment in the tests to keep matchers very simple to avoid flakiness there. Makes sense that `WebCrypto` can't be seeded and we shouldn't make the tests do something else just to make them deterministic.

packages/keys/src/grind-keypair.ts
lorisleiva: Imma regex ya regex! 😅

packages/keys/src/__tests__/grind-keypair-test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+// Some tests in this file exercise the real `generateKeyPair` implementation,
+// which uses WebCrypto under the hood and cannot be seeded. As a result, the
+// addresses produced are genuinely random. To keep these tests fast and
+// deterministic (i.e. non-flaky), any test that does NOT mock
+// `generateKeyPair` must use an extremely permissive matcher — typically a
+// single-character regex prefix like `/^1/` or a trivial `() => true`
+// predicate — so that the grind loop converges within one batch. Do not
+// tighten these matchers: if you need to assert more specific behaviour,
+// mock `generateKeyPair` the same way the forwarding tests below do.
```
lorisleiva, line 29: @mcintyre94 Addressed your comment here.

lorisleiva:
### Merge activity

* **Apr 9, 9:41 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1534).
* **Apr 9, 9:41 PM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1534).

PR #1535, opened 2026-04-08
Refactor `createReactiveStoreWithInitialValueAndSlotTracking` to return slot as well as value

mcintyre94:
#### Summary of Changes

Just rethinking this API a little before we ship

The stored value is now `SolanaRpcResponse<TItem>` instead of just `TItem`, ie the app also has access to the slot that the value comes from. This means that if you use the identity function as the mapper for the subscription, code consuming this store would be identical to if it was just calling `reactive()` on the subscription. 

This enables use cases like displaying the freshness of data, acting conditionally on which slot data is from, etc. 

Note that since this hasn't shipped I've just updated the existing changeset.

### DIFF
--- a/.changeset/giant-falcons-invite.md
+++ b/.changeset/giant-falcons-invite.md
@@ -2,4 +2,4 @@
 '@solana/kit': minor
 ---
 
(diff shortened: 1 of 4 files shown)


lorisleiva (approved):
Good shout!

PR #1536, opened 2026-04-09
Add createAsyncGeneratorWithInitialValueAndSlotTracking

mcintyre94:
#### Summary of Changes

This PR adds `createAsyncGeneratorWithInitialValueAndSlotTracking`. Same idea as the reactive store version, but presented as an async generator API.

It allows you to take any RPC request and subscription with a `SolanaRpcResponse` shape and that can map to the same value, and iterate over them while silently dropping any information from an older slot than has already been received.

While the store only needs to provide the most recent data, this requires us to queue all responses received (from either source) in order, where the slot is not older than the previous last seen slot.

Note that avoiding race conditions is mostly implicit on the JS single threaded runtime, and the fact that Promise `then` blocks are synchronous.

### DIFF
--- a/packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
+++ b/packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
@@ -104,7 +104,7 @@ function createMockSubscriptionRequest(): {
 }
 
 function rpcResponse(slot: number, value: TestValue): SolanaRpcResponse<TestValue> {
-    return { context: { slot: BigInt(slot) }, value } as SolanaRpcResponse<TestValue>;
+    return { context: { slot: BigInt(slot) }, value };
 }
 
 /** Build the expected `{ context: { slot }, value }` shape for a yielded value. */
--- a/packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
+++ b/packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
@@ -2,7 +2,7 @@ import type { PendingRpcRequest } from '@solana/rpc';
 import type { PendingRpcSubscriptionsRequest } from '@solana/rpc-subscriptions';
 import type { SolanaRpcResponse } from '@solana/rpc-types';
 
-import { createReactiveStoreWithInitialValueAndSlotTracking } from '../create-reactive-store-with-initial-value-and-slot-tracking';
+import { createAsyncGeneratorWithInitialValueAndSlotTracking } from '../create-async-generator-with-initial-value-and-slot-tracking';
 
 /** Flush all pending microtasks by waiting for a macrotask boundary. */
 const flushMicrotasks = () => new Promise(resolve => setTimeout(resolve, 0));
@@ -104,10 +104,25 @@ function createMockSubscriptionRequest(): {
 }
 
 function rpcResponse(slot: number, value: TestValue): SolanaRpcResponse<TestValue> {
+    return { context: { slot: BigInt(slot) }, value } as SolanaRpcResponse<TestValue>;
+}
+
+/** Build the expected `{ context: { slot }, value }` shape for a yielded value. */
+function expectedResponse(slot: number, value: number) {
     return { context: { slot: BigInt(slot) }, value };
 }
 
-describe('createReactiveStoreWithInitialValueAndSlotTracking', () => {
+/** Collect all values yielded by the generator until it's done (or collect `n` values). */
+async function collectValues<T>(gen: AsyncGenerator<T>, n?: number): Promise<T[]> {
+    const values: T[] = [];
+    for await (const value of gen) {
+        values.push(value);
+        if (n !== undefined && values.length >= n) break;
+    }
+    return values;
+}
+
+describe('createAsyncGeneratorWithInitialValueAndSlotTracking', () => {
     let abortController: AbortController;
 
     beforeEach(() => {
@@ -118,24 +133,12 @@ describe('createReactiveStoreWithInitialValueAndSlotTracking', () => {
         abortController.abort();
     });
 
-    describe('getState()', () => {
-        it('returns `undefined` before any data arrives', () => {
-            const { mockRequest: rpcRequest } = createMockRpcRequest();
-            const { mockRequest: rpcSubscriptionRequest } = createMockSubscriptionRequest();
-            const store = createReactiveStoreWithInitialValueAndSlotTracking({
-                abortSignal: abortController.signal,
-                rpcRequest,
-                rpcSubscriptionRequest,
-                rpcSubscriptionValueMapper: v => v.count,
-                rpcValueMapper: v => v.count,
-            });
-            expect(store.getState()).toBeUndefined();
-        });
-        it('updates with the RPC response value', async () => {
+    describe('yielded values', () => {
+        it('yields the RPC response value', async () => {
             expect.assertions(1);
             const { mockRequest: rpcRequest, resolve } = createMockRpcRequest();
             const { mockRequest: rpcSubscriptionRequest } = createMockSubscriptionRequest();
-            const store = createReactiveStoreWithInitialValueAndSlotTracking({
+            const gen = createAsyncGeneratorWithInitialValueAndSlotTracking({
                 abortSignal: abortController.signal,
                 rpcRequest,
                 rpcSubscriptionRequest,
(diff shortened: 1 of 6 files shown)


trevor-cortex (commented):
## Summary

Adds `createAsyncGeneratorWithInitialValueAndSlotTracking`, an async generator API that merges an initial RPC response with an ongoing subscription into a single stream, using slot numbers to silently drop out-of-order (stale) values. It's the async-iterator analogue of the existing `createReactiveStoreWithInitialValueAndSlotTracking`.

The implementation looks sound overall: single-threaded JS runtime + synchronous promise `.then` continuations mean the slot-comparison-and-update sequence in each source handler is race-free, which is the right intuition noted in the PR description. Test coverage is thorough — yield ordering, slot dropping from both directions, buffering, error propagation from each source, abort-signal plumbing, early abort, early break cleanup. Nice.

## Things to watch out for

1. **Behavioral divergence from the reactive store when the subscription ends before the RPC resolves.** The subscription's `.then` handler unconditionally sets `done = true` once the iterator completes, which terminates the generator and discards any RPC response still in flight (the later `enqueue` call is gated by `if (done || signal.aborted) return`). The reactive store doesn't do this — the RPC can still populate the store after the subscription ends. This isn't necessarily wrong (and the JSDoc does say "completes when the subscription ends"), but it's an easy edge case to miss, and there's no test for "subscription completes while RPC is still pending". Worth adding one to pin the intended semantics.

2. **Queue drain on abort.** When the abort signal fires, `onAbort` sets `done = true` but leaves any already-buffered items in `queue`. If a consumer calls `.next()` after abort, they'll still receive buffered values before `done` is observed. The `yields values received before abort, then completes` test happens to exercise a similar path but collects the value *before* calling `abort()`, so the actual drain-after-abort behavior isn't asserted. Either is defensible, but please make the intent explicit with a test.

3. **Minor code smells** (see inline comments): a redundant `pendingError` check, a `waitingReject = null` missing from one branch (benign — resolve-after-settle is a no-op), an unused `reactive` mock field, and an inconsistent `as SolanaRpcResponse<TestValue>` cast in the new test file vs the one being removed in the existing file in the same PR.

- The slot-ordering/race reasoning hinges on the fact that `check lastUpdateSlot` → `update lastUpdateSlot` → `enqueue` runs without any intervening `await` in each source handler. Worth double-checking that reasoning if anyone refactors this to introduce an `await` between those steps.
- The subscription-completion semantics (point 1 above) is the main design question I'd want a second opinion on. Everything else is polish.

packages/kit/src/create-async-generator-with-initial-value-and-slot-tracking.ts
```diff
@@ -0,0 +1,199 @@
...
+            for await (const {
+                context: { slot },
+                value,
+            } of notifications) {
+                if (signal.aborted) return;
+                if (slot < lastUpdateSlot) continue;
+                lastUpdateSlot = slot;
+                enqueue({ context: { slot }, value: rpcSubscriptionValueMapper(value) });
+            }
+            // Subscription completed normally.
```
trevor-cortex, line 181: When the subscription completes, `done = true` is set unconditionally, which terminates the generator even if the RPC request hasn't resolved yet. If that race happens, the pending RPC response is silently dropped when it eventually arrives (blocked by the `if (done || signal.aborted) return` guard in `enqueue`).

This differs from `createReactiveStoreWithInitialValueAndSlotTracking`, where the store keeps accepting the RPC response even after the subscription ends. If this divergence is intentional, consider calling it out explicitly in the JSDoc and adding a test for "subscription ends before RPC resolves" to lock in the behavior. If not, you probably want to only flip `done` once *both* sources have settled.

trevor-cortex (commented):
Inline notes follow.

packages/kit/src/create-async-generator-with-initial-value-and-slot-tracking.ts
```diff
@@ -0,0 +1,199 @@
...
+            done = true;
+            if (waitingResolve) {
+                const resolve = waitingResolve;
+                waitingResolve = null;
+                resolve({ done: true, value: undefined });
```
trevor-cortex: Missing `waitingReject = null;` here, unlike the other branches that clear both. It's benign today because resolving an already-resolved promise is a no-op, so a later `handleError` that tries to `reject(...)` this same promise will just be ignored — but it's inconsistent with the pattern used in `enqueue`, `handleError`, and `onAbort`, and makes the invariant "`waitingResolve` and `waitingReject` are always set/cleared together" a bit harder to reason about.

packages/kit/src/create-async-generator-with-initial-value-and-slot-tracking.ts
```diff
@@ -0,0 +1,199 @@
...
+        while (true) {
+            if (pendingError) throw pendingError;
+            if (queue.length > 0) {
+                yield queue.shift()!;
+            } else if (done) {
+                if (pendingError) throw pendingError;
+                return;
+            } else {
+                // if no value queued or error, wait for the next value or error
+                const result: IteratorResult<SolanaRpcResponse<TItem>> = await new Promise((resolve, reject) => {
```
trevor-cortex, line 196: This inner `if (pendingError) throw pendingError;` is unreachable — the top of the loop already threw if `pendingError` was set, so by the time we get here it's guaranteed to be falsy. Can be dropped.

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
```diff
@@ -0,0 +1,512 @@
...
+    };
+
+    return {
+        complete,
+        error,
+        mockRequest: {
+            reactive: jest.fn().mockRejectedValue(new Error('not implemented')),
+            subscribe: jest.fn().mockResolvedValue(asyncIterable),
+        },
+        pushNotification,
```
trevor-cortex, line 94: Inconsistent with the cleanup you're doing in `create-reactive-store-with-initial-value-and-slot-tracking-test.ts` in this same PR, where you're *removing* the `as SolanaRpcResponse<TestValue>` cast. Since both test files use the same `TestValue` shape and the object literal is structurally compatible with `SolanaRpcResponse<TestValue>`, you should be able to drop the cast here too.

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
```diff
@@ -0,0 +1,512 @@
...
+        if (waitingResolve) {
+            const resolve = waitingResolve;
+            waitingResolve = null;
+            resolve({ done: true, value: undefined });
+        }
+    };
+
+    return {
+        complete,
+        error,
```
trevor-cortex, line 89: The `reactive` field on the mock is never exercised — the production code only calls `.subscribe()`. Looks like a leftover from copying the reactive-store test harness; safe to remove to keep the mock focused on what's actually used.

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
```diff
@@ -0,0 +1,512 @@
...
+            await flushMicrotasks();
+            abortController.abort();
+            // RPC resolves after abort.
+            resolve(rpcResponse(100, { count: 42 }));
+            const values = await valuesPromise;
+            expect(values).toEqual([]);
+        });
+        it('aborts the signal passed to the RPC request', async () => {
+            expect.assertions(2);
+            const { mockRequest: rpcRequest } = createMockRpcRequest();
```
trevor-cortex, line 537: Consider adding a test for the queue-drain-after-abort behavior: push a couple of notifications (so they buffer), then fire the abort signal, then call `.next()` repeatedly and assert what you expect — either "buffered values are still yielded, then `{done: true}`" or "everything is dropped, first `.next()` returns `{done: true}`". Right now the implementation does the former (since `onAbort` sets `done` but leaves the queue alone), but no test actually pins that down.

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
```diff
@@ -0,0 +1,512 @@
...
+        it('throws the error even after yielding values', async () => {
+            expect.assertions(2);
+            const { mockRequest: rpcRequest, resolve } = createMockRpcRequest();
+            const { mockRequest: rpcSubscriptionRequest, error } = createMockSubscriptionRequest();
+            const gen = createAsyncGeneratorWithInitialValueAndSlotTracking({
+                abortSignal: abortController.signal,
+                rpcRequest,
+                rpcSubscriptionRequest,
+                rpcSubscriptionValueMapper: v => v.count,
+                rpcValueMapper: v => v.count,
```
trevor-cortex, line 333: Missing coverage for the "subscription completes before RPC resolves" case. With the current implementation, the generator returns immediately and the RPC response is discarded when it eventually arrives — which is a sharp edge worth a dedicated test to make the intended semantics explicit. See the main review comment on the source file.

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
mcintyre94: Kept this in - it's part of the pending subscription type and I'd prefer to implement a no-op field than mess with types to get around it

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Re-review

The main design concern from the previous review — premature termination when the subscription ends before the RPC resolves — has been fixed cleanly with the `rpcDone`/`subscriptionDone` flags and the `markSourcesDone()` helper. The `waitingReject = null` inconsistency is also fixed. 👍

However, **two tests will hang under the new semantics** because they were updated to not resolve the RPC but still expect the generator to complete after `complete()`. Details inline. These should fail / time out in CI — worth running the suite locally to confirm.

### Things to watch out for
- The two hanging tests (inline) are the main blocker.
- Please add a test for the specific fix: **subscription completes before RPC resolves, then RPC resolves, and the yielded values include the RPC response**. This directly exercises the behavior change and would have caught the hanging-test issue too.
- Minor: the `reactive` field on the mock is still unused, and the `as SolanaRpcResponse<TestValue>` cast in the test helper is still present even though the reactive store test removes its equivalent in this same PR.

### For subsequent reviewers
- Focus on the termination logic in the two `.then(() => { ... markSourcesDone(); })` branches and the interaction with `handleError` / abort.
- Verify that the test suite actually passes — the two tests I flagged look like they hang rather than fail loudly.

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
```diff
@@ -0,0 +1,595 @@
...
+            });
+            const valuesPromise = collectValues(gen);
+            await flushMicrotasks();
+            pushNotification(rpcResponse(100, { count: 1 }));
+            await flushMicrotasks();
```
trevor-cortex: **This test will hang under the new termination semantics.**

You removed the `resolve` / `rpcRequest` destructure here, so the RPC request is never resolved. With the new logic in the source file, the generator only terminates when **both** `rpcDone` **and** `subscriptionDone` are true. Calling `complete()` alone sets `subscriptionDone = true`, but since `rpcDone` stays false, `markSourcesDone()` never runs — the outer `while (true)` loop is parked on `waitingResolve` forever, and `collectValues(gen)` never returns.

`afterEach` won't rescue it either, since it only runs after the `it` callback's promise settles.

Fix: before `complete()`, also resolve the RPC at an old slot so it's dropped but still marks `rpcDone`:

```ts
resolve(rpcResponse(0, { count: 0 }));
await flushMicrotasks();
complete();
```

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
```diff
@@ -0,0 +1,595 @@
...
+            const firstNext = gen.next();
+            await flushMicrotasks();
+            // First notification resolves the waiting promise; the generator yields it.
+            pushNotification(rpcResponse(100, { count: 1 }));
+            await expect(firstNext).resolves.toEqual({ done: false, value: expectedResponse(100, 1) });
+            // Generator is now suspended at the yield. Push more values — these
+            // buffer into the internal queue because the consumer hasn't called next().
+            pushNotification(rpcResponse(200, { count: 2 }));
+            pushNotification(rpcResponse(300, { count: 3 }));
+            await flushMicrotasks();
```
trevor-cortex: **Same hang as the previous test.** The RPC is never resolved, so after `complete()` the generator stays parked waiting for the RPC path to finish, and the final `gen.next()` expecting `{ done: true, value: undefined }` never resolves.

Fix: resolve the RPC at an old slot before `complete()`:

```ts
resolve(rpcResponse(0, { count: 0 }));
await flushMicrotasks();
complete();
await expect(gen.next()).resolves.toEqual({ done: true, value: undefined });
```

(Also requires adding `resolve` back to the `createMockRpcRequest()` destructure at the top of the test.)

packages/kit/src/create-async-generator-with-initial-value-and-slot-tracking.ts
```diff
@@ -0,0 +1,210 @@
...
+        .then(async notifications => {
+            for await (const {
+                context: { slot },
+                value,
+            } of notifications) {
+                if (signal.aborted) return;
+                if (slot < lastUpdateSlot) continue;
+                lastUpdateSlot = slot;
+                enqueue({ context: { slot }, value: rpcSubscriptionValueMapper(value) });
+            }
```
trevor-cortex, line 180: Nice fix on the termination semantics — this correctly waits for both sources now and matches the reactive store behavior. Consider adding a dedicated test that exercises the specific case: subscription completes (with or without yielding), then RPC resolves later, and the RPC value is still yielded. That directly covers the behavior change and makes the intent explicit.

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
```diff
@@ -0,0 +1,595 @@
...
+    const complete = () => {
+        done = true;
+        if (waitingResolve) {
+            const resolve = waitingResolve;
+            waitingResolve = null;
+            resolve({ done: true, value: undefined });
+        }
+    };
+
+    return {
```
trevor-cortex, line 87: Still worth removing — this generator never calls `.reactive()`, so the mock field is dead weight and slightly misleading for anyone reading the test.

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
```diff
@@ -0,0 +1,595 @@
...
+        complete,
+        error,
+        mockRequest: {
+            reactive: jest.fn().mockRejectedValue(new Error('not implemented')),
+            subscribe: jest.fn().mockResolvedValue(asyncIterable),
+        },
+        pushNotification,
+    };
```
trevor-cortex, line 95: Still inconsistent with the cast removal in `create-reactive-store-with-initial-value-and-slot-tracking-test.ts` in this same PR. The shape matches `SolanaRpcResponse<TestValue>` exactly, so the cast isn't needed and just adds noise.

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
lorisleiva: @trevor-cortex Callum already made a comment about this on your previous review. Are you able to see it?

trevor-cortex: No, unfortunately I can't — each time I'm triggered I only see the immediate reply context for the thread I'm responding to, not the full history of prior replies on earlier reviews. If Callum's comment addressed a point I'm re-raising, could you either quote it here or point me to which thread it's on? Happy to take it into account.

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
mcintyre94: False positive, the test does what's suggested:

```ts
            // Resolve RPC at old slot (dropped) so the generator can complete.
            resolve(rpcResponse(0, { count: 0 }));
            await flushMicrotasks();
            complete();
            const values = await valuesPromise;
            expect(values).toEqual([expectedResponse(100, 1), expectedResponse(300, 3), expectedResponse(400, 4)]);
```

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
mcintyre94: False positive, the test does this:

```ts
            // Consume — values should drain from the queue.
            await expect(gen.next()).resolves.toEqual({ done: false, value: expectedResponse(200, 2) });
            await expect(gen.next()).resolves.toEqual({ done: false, value: expectedResponse(300, 3) });
            // Resolve RPC at old slot (dropped) and complete subscription.
            resolve(rpcResponse(0, { count: 0 }));
            await flushMicrotasks();
            complete();
            await expect(gen.next()).resolves.toEqual({ done: true, value: undefined });
```

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
mcintyre94: Repeat of https://github.com/anza-xyz/kit/pull/1536#discussion_r3056884385 I think? Which was resolved. No cast to `SolanaRpcResponse` in the current test

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
```diff
@@ -0,0 +1,595 @@
...
+    mockRequest: PendingRpcRequest<SolanaRpcResponse<TestValue>>;
+    reject(error: unknown): void;
+    resolve(response: SolanaRpcResponse<TestValue>): void;
+} {
+    let resolve!: (response: SolanaRpcResponse<TestValue>) => void;
+    let reject!: (error: unknown) => void;
+    const promise = new Promise<SolanaRpcResponse<TestValue>>((res, rej) => {
+        resolve = res;
+        reject = rej;
+    });
```
lorisleiva: nit: you can use `Promise.withResolvers` here. 

```ts
let { promise, resolve, reject } = Promise<SolanaRpcResponse<TestValue>>.withResolvers();
```

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
```diff
@@ -0,0 +1,595 @@
+import type { PendingRpcRequest } from '@solana/rpc';
+import type { PendingRpcSubscriptionsRequest } from '@solana/rpc-subscriptions';
+import type { SolanaRpcResponse } from '@solana/rpc-types';
+
+import { createAsyncGeneratorWithInitialValueAndSlotTracking } from '../create-async-generator-with-initial-value-and-slot-tracking';
+
+/** Flush all pending microtasks by waiting for a macrotask boundary. */
+const flushMicrotasks = () => new Promise(resolve => setTimeout(resolve, 0));
```
lorisleiva: I think you can do the same with Jest using `await jest.runAllTimersAsync();` directly in the test. This requires you call `jest.useFakeTimers();` before the first describe block.

lorisleiva (approved):
Jesus my brain hurts haha. Looks good though. That being said, I have more faith in Trevor than me on that one. 😅

packages/kit/src/__tests__/create-async-generator-with-initial-value-and-slot-tracking-test.ts
mcintyre94: Nice I forgot about that, thanks! I had to add `lib: "ES2024"` to make this work, so just making sure that works in CI 🤞

PR #1537, opened 2026-04-09
Add `writeKeyPair` and `writeKeyPairSigner` helpers

lorisleiva:
This PR adds `writeKeyPair` to `@solana/keys` and `writeKeyPairSigner` to `@solana/signers`, giving callers a one-line way to save an extractable key pair as a 64-byte JSON array that's byte-for-byte compatible with `solana-keygen new`. Missing parent directories are created automatically, and the resulting file is written with mode `0600` so only the owner can read or write it. Both helpers refuse to overwrite an existing file by default and require the caller to explicitly set `unsafelyOverwriteExistingKeyPair: true` to replace one — the name is deliberately alarming because overwriting a key pair file permanently destroys the previous key and any assets it controls. Calling either helper in a browser or React Native environment throws `SOLANA_ERROR__KEYS__WRITE_KEY_PAIR_UNSUPPORTED_ENVIRONMENT` up front.

To keep browser and React Native bundles free of any `node:fs/promises` or `node:path` references, filesystem access is routed through a new private `@solana/fs-impl` package that follows the same pattern as `@solana/crypto-impl` and friends. Its Node entry re-exports `mkdir`, `writeFile`, and `dirname`; its browser entry provides structurally-identical stubs that throw a new generic `SOLANA_ERROR__FS__UNSUPPORTED_ENVIRONMENT` error as defense-in-depth beneath the helper-specific guard. The package is added to `noExternal` in `getBaseConfig.ts` so it gets inlined at build time.


lorisleiva:
* **#1537** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1537?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1537?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1534** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1534?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/fs-impl/package.json
```diff
@@ -0,0 +1,60 @@
+{
+    "name": "@solana/fs-impl",
+    "version": "0.0.0",
+    "private": true,
+    "exports": {
+        "edge-light": {
+            "types": "./dist/types/index.node.d.ts",
+            "import": "./dist/index.node.mjs",
+            "require": "./dist/index.node.cjs"
```
trevor-cortex: The `edge-light` and `workerd` conditions point to the Node entry (`index.node.mjs` / `index.node.cjs`), but those runtimes don't have `node:fs/promises` or `node:path`. Unlike `@solana/crypto-impl` (where edge runtimes *do* have `crypto.subtle`), filesystem access is genuinely unsupported here.

These should point to the browser stubs so callers get the structured `SOLANA_ERROR__FS__UNSUPPORTED_ENVIRONMENT` error instead of a module-not-found crash:

```json
"edge-light": {
    "types": "./dist/types/index.node.d.ts",
    "import": "./dist/index.browser.mjs",
    "require": "./dist/index.browser.cjs"
},
"workerd": {
    "types": "./dist/types/index.node.d.ts",
    "import": "./dist/index.browser.mjs",
    "require": "./dist/index.browser.cjs"
},
```

packages/keys/src/write-keypair.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    const [privateKeyPkcs8, publicKeyRaw] = await Promise.all([
+        crypto.subtle.exportKey('pkcs8', keyPair.privateKey),
+        crypto.subtle.exportKey('raw', keyPair.publicKey),
+    ]);
+    // The PKCS#8 header added by `createPrivateKeyFromBytes` (see `./private-key.ts`) is
+    // a fixed 16-byte prefix for Ed25519 keys, so slicing it off yields the raw 32-byte
+    // seed that `solana-keygen` writes as the first half of its output.
+    const privateKeyBytes = new Uint8Array(privateKeyPkcs8).slice(16);
+    const publicKeyBytes = new Uint8Array(publicKeyRaw);
```
trevor-cortex, line 108: Nit (non-blocking): `privateKeyBytes` will always be 32 bytes given the fixed PKCS#8 header, but there's no assertion guarding that invariant. If a future runtime ever emitted a different PKCS#8 envelope, this would silently write a malformed file. An assertion like `if (privateKeyBytes.length !== 32) throw ...` would make the failure mode obvious. Totally fine to leave as-is though — the header is part of the spec and isn't going to change.

trevor-cortex (commented):
## Summary

This PR adds `writeKeyPair` (`@solana/keys`) and `writeKeyPairSigner` (`@solana/signers`) — one-line helpers to persist an extractable Ed25519 key pair to disk as a 64-byte JSON array, byte-for-byte compatible with `solana-keygen new`. A new private `@solana/fs-impl` package isolates Node filesystem access (following the `crypto-impl` / `event-target-impl` pattern), with browser stubs that throw a structured error. The implementation is clean, well-documented, and thoroughly tested.

## Key observations

### `@solana/fs-impl` export conditions: `edge-light` / `workerd` should use browser stubs

The `edge-light` and `workerd` export conditions in `packages/fs-impl/package.json` currently point to the **Node** entry (`index.node.mjs` / `index.node.cjs`). Edge runtimes like Vercel Edge Functions and Cloudflare Workers don't have `node:fs/promises` or `node:path`, so these should resolve to the **browser** stubs instead — the same way `@solana/crypto-impl` maps `edge-light` / `workerd` to the browser entry for types but still points to Node for runtime (since those runtimes *do* have `crypto.subtle`). For `fs-impl`, however, there's no filesystem in those runtimes, so both the runtime *and* types should use the browser entry.

This is the only issue I'd flag as needing a fix — the rest of the PR looks great.

### Everything else looks solid

- **PKCS#8 slice-at-16 approach** — correctly documented and matches the existing `createPrivateKeyFromBytes` implementation that prepends the same fixed 16-byte header.
- **`wx` flag** for default no-overwrite is the right primitive — lets the OS enforce atomicity.
- **`unsafelyOverwriteExistingKeyPair`** naming is appropriately scary.
- **Error codes** are properly registered with context, messages, and the union type.
- **Tests** cover the important cases: round-trip, file mode, recursive dir creation, non-extractable keys, overwrite protection, and browser environments.
- **`writeKeyPairSigner`** is a clean thin wrapper that delegates to `writeKeyPair`.
- **Docs** (`@see` cross-references between `createKeyPairFromBytes` ↔ `writeKeyPair` ↔ `writeKeyPairSigner`) are helpful.

- Verify that the `@solana/fs-impl` package is wired into the monorepo's CI (turbo pipeline, lint, typecheck, treeshakability).
- The PKCS#8 header length (16 bytes) is hard-coded — this is correct for Ed25519 but worth noting in case the codebase ever adds other curve types.
- `@solana/fs-impl` is a `devDependency` of `@solana/keys` (since it gets inlined via `noExternal`). Confirm that the published `@solana/keys` bundle doesn't carry a runtime dependency reference to `@solana/fs-impl`.

packages/fs-impl/package.json
lorisleiva: Addressed

packages/keys/src/write-keypair.ts
lorisleiva: Addressed

trevor-cortex (commented):
## Re-review

All feedback from the previous pass has been addressed. LGTM — submitting as COMMENT since I can't approve without explicit repo access.

### Changes since last review

1. **`@solana/fs-impl` export conditions fixed** ✅ — `edge-light` and `workerd` now resolve to the browser stubs (`index.browser.mjs` / `.cjs`) for both `import` and `require`, matching the `browser` condition. Edge runtimes will no longer pull in `node:fs/promises` / `node:path`.

2. **New PKCS#8 length guard** in `writeKeyPair` — after slicing the 16-byte PKCS#8 header, the code now asserts `privateKeyBytes.byteLength === 32` and throws `SOLANA_ERROR__KEYS__INVALID_PRIVATE_KEY_BYTE_LENGTH` otherwise. Good forward-compat defense if WebCrypto ever changes the Ed25519 envelope shape. The inline comment explaining the 16-byte prefix is helpful context for future readers.

3. **README documentation added** for both `writeKeyPair` (in `@solana/keys`) and `writeKeyPairSigner` (in `@solana/signers`). Both snippets correctly show `generateKeyPair(true)` / `generateKeyPairSigner(true)` to highlight the extractable requirement, and both call out the overwrite-destroys-funds footgun.

### Minor observations (non-blocking)

- The new PKCS#8 length branch is effectively unreachable today (Ed25519 PKCS#8 envelopes are fixed at 48 bytes → 32 after the slice), so it's correctly omitted from tests — just flagging it for coverage-report readers who might wonder.
- `WriteKeyPairConfig` is defined in `@solana/keys` and re-exported from `@solana/signers` via `export type { WriteKeyPairConfig }`. Clean — avoids type drift between the two helpers.

- Still worth verifying `@solana/fs-impl` is wired into CI (turbo pipeline, lint, typecheck, treeshakability scripts are all declared in its `package.json`, but confirm the root turbo config picks it up).
- The `writeKeyPair` round-trip test asserts the written bytes are compatible with `createKeyPairFromBytes`, which is the key correctness guarantee — good coverage.

packages/errors/src/messages.ts
```diff
@@ -553,6 +557,8 @@ export const SolanaErrorMessages: Readonly<{
         'The provided private key does not match the provided public key.',
     [SOLANA_ERROR__KEYS__SIGNATURE_STRING_LENGTH_OUT_OF_RANGE]:
         'Expected base58-encoded signature string of length in the range [64, 88]. Actual length: $actualLength.',
+    [SOLANA_ERROR__KEYS__WRITE_KEY_PAIR_UNSUPPORTED_ENVIRONMENT]:
+        'Writing a key pair to disk is only supported in Node environments.',
```
mcintyre94: Nit: I slightly prefer the "not supported in this evironment" wording used for `SOLANA_ERROR__FS__UNSUPPORTED_ENVIRONMENT`. We probably do this elsewhere, but specifying Node feels a bit more specific than we need to be - other server runtimes with a persistent disk would be fine too.

mcintyre94 (approved):
Nice, LGTM!

lorisleiva:
### Merge activity

* **Apr 9, 9:41 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1537).
* **Apr 9, 9:42 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1537) rebased this pull request as part of a merge.
* **Apr 9, 9:51 PM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1537).

PR #1541, opened 2026-04-12
fix(codecs): Correct return types for predicate and pattern-match codec overloads

### LINKED ISSUE #1443: Incorrect type overloads for predicate/pattern match codecs
Both these codecs are higher-level wrappers over the union codec. The union codec has runtime behaviour that means if all inputs are not FixedSize _with the same size_, the returned codec is VariableSize. Its types also reflect this.

These codecs have the same behaviour, because they use the union codec. But their type overloads incorrectly return a FixedSize encoder in this case, with the size a union of the sizes of the inputs.

Their types need to be changed to match the way the union codec is typed, so that they return a VariableSized codec unless all inputs are FixedSize with the same size.

agentmila:
## Problem

Fixes #1443

Both `getPredicateEncoder/Decoder/Codec` and `getPatternMatchEncoder/Decoder/Codec` wrap the union codec internally. The union codec has runtime behavior where it returns a `VariableSize` codec when all inputs are `FixedSize` but have **different sizes** (see `getUnionFixedSize` in `union.ts`).

However, overload #2 in both predicate and pattern-match incorrectly promised `FixedSize` returns when both inputs are `FixedSize` without constraining them to the same size.

## Example of the bug

```ts
const encoder = getPredicateEncoder(
  (n: number) => n < 256,
  getU8Encoder(),   // FixedSize<1>
  getU32Encoder(),  // FixedSize<4>
);
// TypeScript says: FixedSizeEncoder<number>
// Runtime says: VariableSizeEncoder
```

## Changes

Changed overload #2 return types to match runtime behavior. 4 files, 12 lines changed.

- `predicate.ts` — Encoder/Decoder/Codec overload #2: FixedSize → Encoder/Decoder/Codec
- `pattern-match.ts` — Encoder/Decoder/Codec overload #2: FixedSize → Encoder/Decoder/Codec  
- Typetests updated accordingly

Overload #1 (same TSize) still correctly returns FixedSize with preserved size.

### DIFF
--- a/packages/codecs-data-structures/src/__typetests__/pattern-match-typetest.ts
+++ b/packages/codecs-data-structures/src/__typetests__/pattern-match-typetest.ts
@@ -31,7 +31,7 @@ const stringTypePredicate = null as unknown as (value: number | string) => value
         getPatternMatchEncoder([
             [numberValuePredicate, {} as FixedSizeEncoder<number>],
             [numberValuePredicate, {} as FixedSizeEncoder<number>],
-        ]) satisfies FixedSizeEncoder<number>;
+        ]) satisfies Encoder<number>;
 
         // It maintains the size if all encoders are FixedSizeEncoder with the same size
         getPatternMatchEncoder([
@@ -115,7 +115,7 @@ const stringTypePredicate = null as unknown as (value: number | string) => value
     getPatternMatchDecoder([
         [bytesPredicate, {} as FixedSizeDecoder<number>],
         [bytesPredicate, {} as FixedSizeDecoder<number>],
-    ]) satisfies FixedSizeDecoder<number>;
+    ]) satisfies Decoder<number>;
 
     // It maintains the size if all decoders are FixedSizeDecoder with the same size
     getPatternMatchDecoder([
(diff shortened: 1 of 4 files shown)


lorisleiva:
That's not quite what we had in mind. If you look at the union codec, you'll see the runtime returns a fixed-size codec if and only if all of the variant are fixed-size _and_ they are all the same size. However, the types are a little more relaxed than this, i.e. they return a fixed-size if all variants are fixed-size too without checking the actual size. This is because there are cases where TypeScript cannot know the exact size of a fixed-size codec before runtime.

Ideally the union codec type signature would:
- Return a `FixedSizeCodec` of size `S` if all variant codecs are `FixedSizeCodecs` of size `S` as well.
- Return a `Codec` (i.e. we don't know) if all variant codecs are `FixedSizeCodecs` but we cannot guarantee they all have the same size.
- Return a `VariableSizeCodec` for all other cases.

Once the union codec signature is fixed, the other codecs that use it under hood will need adjusting accordingly.

I will close this PR since the required solution is quite different (and we're getting lots of PRs to review these days so we need to be more direct with our reviewing process) but feel free to open a new one.

PR #1542, opened 2026-04-15
Add NO_SIGNER_CONNECTED and SIGNER_NOT_AVAILABLE wallet errors

mcintyre94:
#### Summary of Changes

These are new errors that will be thrown when accessing `client.identity` or `client.payer` while those are controlled by a wallet plugin and there is not a signer for a connected wallet available.

There are 2 cases:

- `SOLANA_ERROR__WALLET__NO_SIGNER_CONNECTED` will be used when no wallet is connected. It includes a `status`, which in the wallet plugin will be one of its statuses, such as `disconnected` or `pending`. In the error definition it's just a string though.
- `SOLANA_ERROR__WALLET__SIGNER_NOT_AVAILABLE` will be used when a wallet is connected, but a signer was not created for it because of missing features. In this case as `client.identity` and `client.payer` are both signer fields, we cannot set them and will throw this error. 

Note that the first is differentiated from `SOLANA_ERROR__WALLET__NOT_CONNECTED` because we're trying to access a field controlled by the wallet, not perform an operation on the wallet. Its message `Cannot $operation: no wallet connected` doesn't work, and its context does not include `status`.

Also note that these don't specify the field that is being accessed, I think the stacktrace will provide that and it makes these error codes more reusable and less tied to precisely how plugins use them.

### DIFF
--- a/.changeset/calm-swans-stay.md
+++ b/.changeset/calm-swans-stay.md
@@ -1,0 +1,5 @@
+---
+'@solana/errors': minor
+---
+
+Add `SOLANA_ERROR__WALLET__NO_SIGNER_CONNECTED` and `SOLANA_ERROR__WALLET__SIGNER_NOT_AVAILABLE` error codes for wallet-signer availability checks.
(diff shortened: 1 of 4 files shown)


lorisleiva (approved):
Thanks!

PR #1544, opened 2026-04-15
Use latest stable validator version in CI

mcintyre94:
#### Problem

CI is broken: https://github.com/anza-xyz/kit/actions/runs/24452992515/job/71487609515?pr=1542

This is because the [anza releases page](https://api.github.com/repos/anza-xyz/agave/releases) that we use to find the validator version no longer has any 2.3.x versions on the first page of results

#### Summary of Changes

This PR changes the approach so that we use the latest stable (not prerelease, no alpha/beta/rc in tag name) version. Currently this is 3.1.13, as all 4.x releases are betas. 

This might turn out to be a bad approach, but I think it's probably better for our CI to be running against the validator the network is currently using. See https://www.validators.app/ - currently around 73% of the network are on 3.1.13 

We also update the types (at least those covered by tests), to match what the 3.1.13 validator is returning. This is purely additive, which justifies a minor changeset. Our tests do validate precise values (to catch fields missing from our types) and so needed to be updated. Adding new fields to types is not a breaking change in Kit though.

There is one exception: The RPC now returns `replacementBlockhash: null` when `replaceRecentBlockhash` is false in simulateTransaction. Previously it excluded the field. This would be a breaking change to Kit types, so we don't include it in this PR to unblock CI. #1546 

Going forward we can decide at each breaking version whether to pin to eg 3.1.x, or immediately update to keep up with the current stable version. But I think staying permanently pinned to an old version (2.3.x) is the worst of all options. Consumers are going to be receiving responses from a network on 3.1.x currently. 

CI run with `--force` to ensure all tests pass: https://github.com/anza-xyz/kit/actions/runs/24504824860/job/71620554038


mcintyre94:
@trevor-cortex

scripts/get-latest-validator-release-version.sh
```diff
@@ -2,7 +2,7 @@
 (
     set -e
     version=$(node -e \
-      'fetch("https://api.github.com/repos/anza-xyz/agave/releases").then(res => res.json().then(rs => rs.filter(r => !r.prerelease && r.tag_name.startsWith("v2.3."))).then(x => console.log(x[0].tag_name)));'
+      'fetch("https://api.github.com/repos/anza-xyz/agave/releases").then(res => res.json()).then(rs => { const r = rs.find(r => !r.prerelease && !/alpha|beta|rc/.test(r.tag_name)); if (r) console.log(r.tag_name); });'
```
trevor-cortex, line 5: Nit: the PR description mentions filtering out `test` versions too, but the regex only covers `alpha|beta|rc`. Worth adding `test` to the pattern?

```suggestion
      'fetch("https://api.github.com/repos/anza-xyz/agave/releases").then(res => res.json()).then(rs => { const r = rs.find(r => !r.prerelease && !/alpha|beta|rc|test/.test(r.tag_name)); if (r) console.log(r.tag_name); });'
```

trevor-cortex (commented):
## Summary

CI was broken because the GitHub releases API pagination no longer returns any 2.3.x versions on the first page. This PR fixes it by switching from filtering for `v2.3.*` to finding the latest stable (non-prerelease, no alpha/beta/rc in tag name) release instead.

## Notes

This looks good to me! A couple of small things to be aware of:

- The regex `/alpha|beta|rc/` doesn't filter out `test` versions, which was mentioned in the PR description. Not sure if that's intentional or if it just doesn't exist in practice. Might be worth adding it for safety.
- This still only looks at the first page of results (30 items by default from the GitHub API). If all 30 happen to be prereleases at some point, the script will fall through to the cache-busting path. Unlikely in practice, but worth noting.
- The `find` approach (first match) works well here since GitHub returns releases in reverse chronological order.

scripts/get-latest-validator-release-version.sh
mcintyre94: oops misread the regex while writing the PR description, but we don't use `-test` in tag names so not needed. [Last 100 releases](https://api.github.com/repos/anza-xyz/agave/releases?per_page=100) have only used alpha/beta.

scripts/start-shared-test-validator.sh
```diff
@@ -24,7 +24,7 @@ mkdir -p $LOCK_DIR
   flock -s 200 || exit 1
   (
     if flock -nx 200; then
-      $TEST_VALIDATOR --ledger $TEST_VALIDATOR_LEDGER --reset --quiet --account-dir $FIXTURE_ACCOUNTS_DIR --rpc-pubsub-enable-vote-subscription --rpc-pubsub-enable-block-subscription >/dev/null &
+      $TEST_VALIDATOR --ledger $TEST_VALIDATOR_LEDGER --reset --quiet --account-dir $FIXTURE_ACCOUNTS_DIR --rpc-pubsub-enable-vote-subscription >/dev/null &
```
mcintyre94, line 27: Seems like this flag has been removed, caused an error on startup

beeman: I ran into the exact same issue yesterday (wallet-ui repo is a fork of this base repo)

https://github.com/wallet-ui/wallet-ui/pull/477

mcintyre94: Thanks @beeman! I'm hoping this PR pinning to the latest stable release is going to be the right longer term fix here

scripts/get-latest-validator-release-version.sh
trevor-cortex, line 5: Previous review noted this: the regex filters `alpha|beta|rc` but not `test` (which was mentioned in the original PR description). Looks like the description was updated to drop `test` — just confirming that's intentional and not an oversight.

packages/rpc-api/src/simulateTransaction.ts
```diff
@@ -96,14 +98,34 @@ type WithInnerInstructionsConfig = Readonly<{
...
+    /** Token balances for each token account before the transaction was processed */
+    preTokenBalances: readonly TokenBalance[] | null;
+    /**
+     * The blockhash and last valid block height used for the simulation. Only contains a
+     * replacement when `replaceRecentBlockhash` is `true`.
+     */
+    replacementBlockhash: TransactionBlockhashLifetime | null;
     /** The most-recent return data generated by an instruction in the transaction */
     returnData: Readonly<{
         /** The return data itself, as base-64 encoded binary data */
```
trevor-cortex, line 126: Nit: `replacementBlockhash` is now in the base response type as `T | null`, but `SimulateTransactionApiResponseWithReplacementBlockhash` still exists and narrows it to non-null. This works correctly via intersection (`&`), but it does mean `replacementBlockhash` appears in two places. Might be worth a comment on the base field noting that it gets narrowed by the intersection type for `replaceRecentBlockhash: true` overloads.

packages/rpc-api/src/__tests__/simulate-transaction-test.ts
```diff
@@ -248,9 +255,12 @@ describe('simulateTransaction', () => {
             )
             .send();
 
-        await expect(resultPromise).rejects.toThrow(
-            new SolanaError(SOLANA_ERROR__JSON_RPC__SERVER_ERROR_TRANSACTION_SIGNATURE_VERIFICATION_FAILURE),
-        );
+        await expect(resultPromise).resolves.toStrictEqual({
```
trevor-cortex, line 258: Good catch on this behavioral change. The validator now returns `SignatureFailure` as a result error rather than throwing a JSON-RPC error. Worth noting in a changelog or migration guide if this repo publishes one — consumers catching `SERVER_ERROR_TRANSACTION_SIGNATURE_VERIFICATION_FAILURE` will need to update their error handling.

trevor-cortex (commented):
## Summary

This PR fixes broken CI by switching from the pinned `v2.3.*` validator version to the latest stable release (currently 3.1.13). It also updates all affected types and tests to match the 3.1.13 validator's responses, including:

- New fields on `SimulateTransactionApiResponseBase`: `fee`, `loadedAddresses`, `preBalances`, `postBalances`, `preTokenBalances`, `postTokenBalances`, and `replacementBlockhash` (now always present, nullable).
- New fields on `JsonParsedVoteAccount`: `blockRevenueCollector`, `blockRevenueCommissionBps`, `blsPubkeyCompressed`, `inflationRewardsCollector`, `inflationRewardsCommissionBps`, `pendingDelegatorRewards`, and `latency` on votes.
- New fields on `RpcSimulateTransactionResult` (errors package) to stay in sync with the RPC API types.
- Test assertions relaxed to use `expect.objectContaining()` / `expect.arrayContaining()` to accommodate new fields from the validator without brittle exact matching.
- Behavioral change: invalid signature with `sigVerify: true` now returns a result with `err: 'SignatureFailure'` instead of throwing `SERVER_ERROR_TRANSACTION_SIGNATURE_VERIFICATION_FAILURE`.
- Removed `--rpc-pubsub-enable-block-subscription` flag from the test validator start script (presumably dropped in 3.x).

## Key things to watch

- The `replacementBlockhash` type change is a **breaking change for consumers**: it was previously conditional on `replaceRecentBlockhash` config, but is now always present as `T | null` in the base response. The typetest update reflects this. Downstream users who relied on `replacementBlockhash` not existing in the response type when `replaceRecentBlockhash` is false will need to handle the `null` case.
- The `SimulateTransactionApiResponseWithReplacementBlockhash` type still narrows `replacementBlockhash` to non-null `TransactionBlockhashLifetime` for the `replaceRecentBlockhash: true` overloads — this is correct since `&` will intersect to the non-null version.
- The signature verification behavior change (returning `SignatureFailure` in the result rather than throwing) seems like a validator-level change. Worth documenting if this affects any public error handling guidance.

- Verify that all new `JsonParsedVoteAccount` fields match the actual Agave 3.1.13 vote account schema. The fields look reasonable (BLS key support, revenue collectors, delegator rewards) but worth cross-referencing with Agave source.
- The `postTokenBalances` and `preTokenBalances` are typed as `unknown[]` in `RpcSimulateTransactionResult` (errors package) but as `readonly TokenBalance[]` in `simulateTransaction.ts` — this is fine since the errors package is looser by design, but worth noting the asymmetry.
- The `loadedAccountsDataSize` field is now optional (`?:`) in `SimulateTransactionApiResponseBase` — check if 3.1.13 still returns it or if it's been removed.

LGTM 👍

mcintyre94:
@trevor-cortex Is "The replacementBlockhash type change" the **only** breaking change?

trevor-cortex:
No, there are several breaking changes in this PR beyond `replacementBlockhash`. Here's the full list:

**1. New fields added to `SimulateTransactionApiResponseBase` (and `RpcSimulateTransactionResult`)**
- `fee: Lamports | null`
- `loadedAddresses: { readonly: Address[], writable: Address[] } | null`
- `postBalances: readonly Lamports[] | null`
- `postTokenBalances: readonly TokenBalance[] | null`
- `preBalances: readonly Lamports[] | null`
- `preTokenBalances: readonly TokenBalance[] | null`

These are added to the base type, so any code doing exhaustive checks on the response shape will see new properties.

**2. `replacementBlockhash` moved to the base type**
Previously it was only present when `replaceRecentBlockhash: true` (via the `SimulateTransactionApiResponseWithReplacementBlockhash` intersection type). Now it's `TransactionBlockhashLifetime | null` in the base response, always present. The narrowing type still exists and intersects to `TransactionBlockhashLifetime` (non-null) when `replaceRecentBlockhash: true`. This is a type-level breaking change — code that previously relied on `replacementBlockhash` _not existing_ on the response when `replaceRecentBlockhash` is false will now see it as `| null`.

**3. Signature verification behavior change**
With the 2.3.x validator, sending a transaction with an invalid signature + `sigVerify: true` threw `SOLANA_ERROR__JSON_RPC__SERVER_ERROR_TRANSACTION_SIGNATURE_VERIFICATION_FAILURE`. With 3.1.x, it now returns a successful response with `err: 'SignatureFailure'`. This is a runtime behavior change that could break error handling code expecting the old throw pattern.

**4. New fields on `JsonParsedVoteAccount`**
- `blockRevenueCollector: Address`
- `blockRevenueCommissionBps: bigint`
- `blsPubkeyCompressed: string | null`
- `inflationRewardsCollector: Address`
- `inflationRewardsCommissionBps: bigint`
- `pendingDelegatorRewards: StringifiedBigInt`
- `latency: bigint` added to vote entries

**5. Rent sysvar values changed**
- `exemptionThreshold`: 2 → 1
- `lamportsPerByteYear`: '3480' → '6960'

These are runtime data changes rather than type changes, but they'll break any snapshot/exact-match tests.

**6. `--rpc-pubsub-enable-block-subscription` flag removed**
The test validator startup script drops this flag — presumably it was removed or made default in 3.1.x.

So TL;DR: the `replacementBlockhash` type change is one of several. The signature verification behavior change (#3) is probably the most impactful one for downstream consumers.

mcintyre94:
I disagree that these are major breaking changes. Importantly, this PR is not upgrading the validator version for consumers, that's already happened and is independent of Kit. We pick up the breaking changes from the validator intentionally in our tests to inform our types.

1) adding fields to types to match validator is not breaking IMO
2) agreed, reverted. #1546 for this breaking change later
3) outside of kit's API. This is a breaking change for consumers caused by 3.x, not Kit
4) same as 1
5) same as 3
6) not our change

lorisleiva (approved):
Amazing! Tysm for keeping up with these changes. 🫶

PR #1549, opened 2026-04-21
Add `isAbortError` to `@solana/promises`

mcintyre94:
#### Summary of Changes

This PR adds a new function `isAbortError` to `@solana/promises`

It provides an easy way to filter for an abort error, in the same way we do with `isSolanaError`. Pairs with `getAbortablePromise`, but useful for anything throwing abort errors (like fetch) too.

### DIFF
--- a/packages/promises/src/__tests__/abortable-test.ts
+++ b/packages/promises/src/__tests__/abortable-test.ts
@@ -1,4 +1,4 @@
-import { getAbortablePromise } from '../abortable';
+import { getAbortablePromise, isAbortError } from '../abortable';
 
 describe('getAbortablePromise()', () => {
     let promise: Promise<unknown>;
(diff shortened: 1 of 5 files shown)


mcintyre94:
* **#1549** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1549?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1549?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/promises/src/__tests__/abortable-test.ts
```diff
@@ -74,3 +74,39 @@ describe('getAbortablePromise()', () => {
...
+    it('returns `false` for a regular `Error`', () => {
+        expect(isAbortError(new Error('nope'))).toBe(false);
+    });
+    it('returns `false` for a `TypeError`', () => {
+        expect(isAbortError(new TypeError('nope'))).toBe(false);
+    });
+    it('returns `false` for a non-`Error` object whose `name` is `AbortError`', () => {
+        expect(isAbortError({ name: 'AbortError' })).toBe(false);
+    });
+    it('returns `false` for `undefined`', () => {
```
trevor-cortex, line 106: Optional: since the docblock specifically calls out `DOMException` as the real-world aborted-fetch rejection shape, consider adding a case that uses the real thing:

```ts
it('returns `true` for a `DOMException` whose `name` is `AbortError`', () => {
    expect(isAbortError(new DOMException('The operation was aborted.', 'AbortError'))).toBe(true);
});
```

This exercises the `instanceof Error` branch against the actual platform class rather than a synthesized `Error` with a patched `name`, and guards against any future environment where `DOMException` stops extending `Error`. Non-blocking.

trevor-cortex (commented):
Nice, tidy addition. Adds `isAbortError(err)` to `@solana/promises` as a narrow predicate — `err instanceof Error && err.name === 'AbortError'` — so callers don't have to `instanceof`-check each platform's error class. Pairs well with `getAbortablePromise` and standard fetch abort behavior.

**What looks good**
- Export is automatic via the existing `export * from './abortable'` in `src/index.ts`, no index change needed.
- `minor` changeset is correct for a new public API.
- JSDoc follows the repo's docblock skill (summary, `@example`, `@see {@link getAbortablePromise}`).
- README section mirrors the style of the adjacent `getAbortablePromise` entry.
- Tests cover the happy path, a subclass, the real abort flow through `getAbortablePromise`, and the main negative cases (regular `Error`, `TypeError`, duck-typed object, `undefined`, `null`). `expect.assertions(1)` is correctly used only in the async test, per CLAUDE.md.

**For subsequent reviewers**
- The docblock specifically mentions `DOMException` as the real-world aborted-fetch rejection, but the tests simulate it with `Object.assign(new Error(), { name: 'AbortError' })` rather than a real `new DOMException('...', 'AbortError')`. `DOMException` extends `Error` in all environments Kit targets (Node ≥17 current/lts, browsers, RN) so the predicate works, but a direct `DOMException` test would tie the behavior more tightly to the documented use case. Non-blocking.
- Worth sanity-checking that `@solana/promises` genuinely isn't intended to be re-exported from the `@solana/kit` facade — it currently isn't (neither `getAbortablePromise` nor the new `isAbortError` will be reachable from `@solana/kit`). That's pre-existing and out of scope for this PR, just flagging.

packages/promises/src/__tests__/abortable-test.ts
mcintyre94: Good idea - had to add this as a browser only test because of a limitation of Jest's node env.

lorisleiva (approved):
Thanks! I double-checked and it also works when passing nothing to `AbortController.abort` ([See playground](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAhgIxAJygYXFZIA2OCmyMAvDGPgO4wCCSqGYWuByA3AFCIrqbZ6EA6LqgAUASnbtQkWAAdsAWwCWEfCRgi4EAJ5hgGsSQB8MAN7sY8Oj0Z8WAiEoDmYODgFQAFtgoBJAGa03PgAJuIWMMj4UACuyGBkMXgcAL5i4h6e+GAi0hC4+AI4IE5iAsBwUMCeIoTYRMQmeQVFJbXI9QKuCvhiYkA)). nit: Maybe we could add a test case like this.

mcintyre94:
Nice, thanks! Added to the browser tests because it has the same quirk in Jest's Node environment.

PR #1550, opened 2026-04-21
Add `ReactiveActionStore` to `@solana/subscribable`

mcintyre94:
#### Summary of Changes

This PR adds a new `createReactiveActionStore` to `@solana/subscribable`, which turns any async function into a state machine with subscribe/getState functionality, similar to the `ReactiveStreamStore` we built for subscriptions.

It's designed to be the framework-agnostic reactive-UI-friendly core for all async functions a reactive UI needs. Eg in react this would be the core of hooks like `useConnectWallet` and `useSendTransaction`.

React example:

```typescript
import { createReactiveActionStore } from '@solana/subscribable';
import { useSyncExternalStore } from 'react';

const searchStore = createReactiveActionStore(async (signal, query: string) => {
    const response = await fetch(`/api/search?q=${query}`, { signal });
    return response.json() as Promise<SearchResult[]>;
});

function SearchBox() {
    const state = useSyncExternalStore(searchStore.subscribe, searchStore.getState);
    return (
        <>
            <input onChange={e => searchStore.dispatch(e.target.value)} />
            {state.status === 'running' && <Spinner />}
            {state.status === 'error' && <ErrorBanner error={state.error} />}
            {state.data && <Results items={state.data} />}
        </>
    );
}
```

Functionality:

- Use `dispatch` to send the promise, fire-and-forget for UI handlers. The state is surfaced through the store, not the result
- Use `dispatchAsync` to await the promise and handle any errors. Used for imperative handlers, eg await and then navigate.  
- Reactive state: idle, running, success, error
- Allows calls to supersede earlier ones, by aborting the earlier signal and dropping their result. Note that this is different behaviour to eg tanstack query, and is designed for how Kit apps should behave. Clicking twice doesn't send a transaction twice.
- stale-while-revalidate, the previous data is available if present in the running/error states
- compatible with `useSyncExternalStore` and all similar reactive UI primitives

This is intended to enable a generic `useAction` hook in React, which other async functionality like `useSendTransaction` would build on. Other UI libraries would build a similar primitive. Example (simpified):

```typescript
export function useAction<TArgs extends unknown[], TResult>(
    fn: (signal: AbortSignal, ...args: TArgs) => Promise<TResult>,
): UseActionResult<TArgs, TResult> {
    const [store] = useState(() =>
        createActionStore<TArgs, TResult>((signal, ...args) => fnRef.current(signal, ...args)),
    );

    const snapshot = useSyncExternalStore(store.subscribe, store.getState, store.getState);

    return useMemo(
        () => ({
            ...snapshot, // data, error, status 
            reset: store.reset,
            send: store.dispatch,
        }),
        [snapshot, store],
    )
}
```

Why not just use tanstack query etc?

This duplicates some of the functionality of libraries like tanstack query and SWR. The reasons are:

- this is a framework-agnostic primitive that works with/without any UI library/caching layer
- mentioned above, but this adds automatically aborting previous requests on double click which is the right behaviour for things like sending transactions. That's not default in tanstack etc and would require fiddly plumbing. Better to do it here and make that wrapper trivial (and optional)


mcintyre94:
* **#1550** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1550?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1550?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/subscribable/README.md
```diff
@@ -73,6 +91,41 @@ target.dispatchEvent(new CustomEvent('candyVended', { detail: { flavor: 'raspber
...
+Things to note:
+
+- Starts at `{ status: 'idle' }`. `getState()` always returns a defined snapshot.
+- `dispatch` is a stable reference — safe to pass into memoized callbacks without re-renders.
+- Calling `dispatch` while one is in flight aborts the previous call; its promise rejects with an `AbortError` and its outcome is dropped.
+- `dispatch` rethrows on failure in addition to surfacing the error on state, so callers can `try/catch` at the call site.
+- `data` survives across transitions: a fresh `running` or `error` snapshot carries the last successful result so call sites can keep rendering stale content while a retry is in flight. Only `reset()` clears it.
+- `reset()` aborts the in-flight dispatch and restores the idle snapshot, clearing both `data` and `error`.
+- Pair with [`isAbortError`](../promises#isabortererr) from `@solana/promises` to ignore abort rejections at `dispatch` call sites.
+- Subscribers are notified only when the snapshot's `status`, `data`, or `error` actually changes, so redundant transitions (`dispatch` while already `running` with the same `data`, `reset` while already `idle`) are silent.
```
trevor-cortex, line 140: Nit: this anchor is broken. `isAbortError(err)` in `packages/promises/README.md` slugifies to `#isaborterrorerr` (GitHub keeps the `err` parameter in the slug), not `#isabortererr`. Either fix the anchor or link to the heading without the anchor fragment.

trevor-cortex (commented):
## Summary

Adds `createActionStore` to `@solana/subscribable`: a framework-agnostic state machine (`idle` / `running` / `success` / `error`) that wraps an async function and exposes `{ dispatch, getState, subscribe, reset }`. Each `dispatch` aborts the previous in-flight call via a fresh `AbortController` piped through `getAbortablePromise`, preserves the last successful `data` across subsequent `running` / `error` states (stale-while-revalidate), and bridges cleanly into `useSyncExternalStore`-style primitives. Ships with a `minor` changeset, a new `@solana/promises` dependency, a `lib: ES2024.Promise` tsconfig bump for `Promise.withResolvers`, thorough README docs, and 388 lines of tests.

## What I checked

- **Supersede semantics**: `dispatch` #2 synchronously aborts #1's controller and sets `running` before awaiting. On #1's late settlement, the `signal.aborted` guard correctly skips the stale `setState`. `previousData` is captured from `state.data` before the `running` transition, so stale-while-revalidate preserves the genuinely-last-successful value. ✓
- **Reset during running**: `reset()` aborts, sets idle, and the still-pending `dispatch` awaits then sees `signal.aborted` → skips setState. The "superseded call resolves after reset" test covers this. ✓
- **Dependency choices**: `@solana/event-target-impl` stays a devDep (inlined at build time per CLAUDE.md). `@solana/promises` is a legitimate runtime dep — `getAbortablePromise` is load-bearing for the AbortError rejection contract when `fn` doesn't respect the signal. ✓
- **Changeset**: present, correctly marks `@solana/subscribable` as `minor` for a new public API. ✓
- **`tsconfig` bump to `ES2024.Promise`**: required for `Promise.withResolvers` in the tests. No runtime impact since tsup handles compilation. ✓

## Things to watch / notes for reviewers

1. **Unhandled-rejection ergonomics.** `dispatch` always rethrows, including AbortError on supersede/reset. That's a deliberate API choice (callers can `try/catch`), but the intended fire-and-forget usage pattern (the `searchStore.dispatch(e.target.value)` example in the PR description, and anywhere `dispatch` is wired to `onChange`/`onClick`) will produce unhandled AbortError rejections the moment a second dispatch supersedes a first. The tests all `.catch(() => {})` these, but the README React example doesn't. Worth an explicit call-out in the README's "Things to note" list, or consider swallowing AbortErrors on the returned promise (keeping them on `state.error` is still useful, though you'd want to distinguish abort from user error there too). Flagged inline.

2. **Broken README anchor.** `[isAbortError](../promises#isabortererr)` won't resolve — the actual heading `isAbortError(err)` slugifies to `#isaborterrorerr`. Flagged inline.

3. **Listener re-entrancy.** `listeners.forEach(l => l())` with a `Set` handles unsubscribe-during-notify correctly, but a listener that synchronously calls `dispatch` / `reset` / `subscribe` will recurse. In practice reactive-UI consumers never do this (and the README's target audience is `useSyncExternalStore` et al.), so I'm not asking for a snapshot-the-listeners change — just noting it.

4. **`currentController` bookkeeping.** After a dispatch settles successfully, `currentController` still points at the completed controller, so a later `reset()` will `.abort()` an already-aborted-or-completed signal (harmless, just slightly untidy). Not worth changing.

Overall this is a clean, well-tested addition with good API ergonomics and thoughtful docs. My two inline comments are both small.

packages/subscribable/src/reactive-action-store.ts
```diff
@@ -0,0 +1,124 @@
...
+            throw error;
+        }
+    };
+
+    return {
+        dispatch,
+        getState: () => state,
+        reset: () => {
+            currentController?.abort();
+            currentController = undefined;
```
trevor-cortex, line 135: Ergonomic footgun worth documenting (or handling) more loudly: because `dispatch` always rethrows — including the AbortError on supersede/reset — any fire-and-forget call site will produce an unhandled rejection the moment a second dispatch arrives. The intended reactive-UI usage (e.g. `onChange={e => store.dispatch(e.target.value)}`, and the `searchStore` example in the PR description) hits this by default. Your tests all `.catch(() => {})` these, but real consumers won't know to.

Options:
- Keep current behavior and add an explicit "you must `.catch(isAbortError)` or `void` with a handler" bullet to the README's "Things to note" (the `isAbortError` link is the right move, but make the obligation clearer than "Pair with").
- Swallow AbortErrors on the returned promise (resolve with `undefined`? return a `Promise<TResult | undefined>`? reject with a sentinel?) so only genuine failures propagate — at the cost of making the type slightly less honest.
- Document a recommended `useAction`-style wrapper that `.catch`es AbortError internally.

Not blocking — the current contract is defensible — but the gap between the docstring "so callers can try/catch at the call site" and the "dispatch straight from an event handler" ergonomic target is worth closing.

mcintyre94:
I've updated the API so there are now two functions to send the promise. I expect we would expose both in eg `useAction` hooks. 

- `dispatch` is a sync fire-and-forget that is suitable for UI event handlers. It swallows all errors, and surfaces the result only through the reactive store state
- `dispatchAsync` is an async function that resolves to the result, and passes through all errors, including an abort error. This is useful for imperative event handlers, that need to eg await and then navigate.

Motivating example:

```tsx
const { sendAsync, isRunning, error } = useSendTransaction();

async function onClick() {
    try {
        const result = await sendAsync(ix);
        navigate(`/tx/${result.signature}`);  // needs the awaited value
    } catch (err) {
        if (isAbortError(err)) return;
        toast.error(String(err));
    }
}

// needs store state
return <button onClick={onClick} disabled={isRunning}>{error ? 'Retry' : 'Send'}</button>;
```

This matches Tanstack mutations: https://tanstack.com/query/v5/docs/framework/react/guides/mutations#promises 

This means that UI using `dispatch` behaves as expected: no errors to catch, no aborts to handle, everything is exposed through the store. An abort due to being superseded will never surface.

Advanced use cases that need to `await` in an event handler are also available using `dispatchAsync`, and can use `isAbortError` to avoid exposing abort errors, while still catching them. 

Also stacked the branch on `is-abort-error` which should fix the readme link. They're tangential, but it's useful to document `isAbortError` as the way to filter those errors here.

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Re-review

Thanks for the fast turnaround. The `dispatch` / `dispatchAsync` split is a clean solution to the unhandled-rejection ergonomics — fire-and-forget never rejects (failures surface on state), and imperative callers get a proper promise. API shape, docs, and test coverage all look good.

### Still open

- **Broken anchor in README** — `[isAbortError](../promises#isabortererr)` still resolves to a non-existent anchor. GitHub's slug for the `isAbortError(err)` heading is `#isaborterrorerr` (it includes the `err` param). I flagged this last round; I don't think it was picked up in the latest push.

### New minor nit

- **`dispatchAsync` can resolve with a superseded call's value in a narrow race.** If `fn` resolves *and then* the signal aborts before the `await` continuation runs, `getAbortablePromise` has already won with the value, the `!signal.aborted` guard correctly skips `setState`, but the function still `return`s `result` to the awaiting caller. That contradicts the doc claim that `dispatchAsync` "rejects … with an `AbortError` when the call is superseded". The easy fix is to check `signal.aborted` after the await and throw an `AbortError` in that branch before returning. Not worth blocking on — the window is tiny and most callers won't notice — but the docs currently overpromise. Flagged inline.

- The only substantive behaviour change since my last pass is the `dispatch` / `dispatchAsync` split; the core state machine, supersede semantics, and reset behaviour are unchanged and were verified last round.
- Tests added coverage for both dispatch variants; I spot-checked the new cases and they look right.

packages/subscribable/README.md
```diff
@@ -73,6 +91,42 @@ target.dispatchEvent(new CustomEvent('candyVended', { detail: { flavor: 'raspber
...
+    - `dispatchAsync(...)` — returns a promise that resolves to the wrapped function's result. Rejects on failure and with an `AbortError` when superseded or `reset()`. Use from imperative code that needs the resolved value; pair with [`isAbortError`](../promises#isabortererr) from `@solana/promises` to filter abort rejections.
+- Calling either dispatch while one is in flight aborts the previous call; its outcome is dropped from state regardless of which variant started it.
+- `data` survives across transitions: a fresh `running` or `error` snapshot carries the last successful result so call sites can keep rendering stale content while a retry is in flight. Only `reset()` clears it.
+- `reset()` aborts the in-flight dispatch and restores the idle snapshot, clearing both `data` and `error`.
+- Subscribers are notified only when the snapshot's `status`, `data`, or `error` actually changes, so redundant transitions (`dispatch` while already `running` with the same `data`, `reset` while already `idle`) are silent.
```
trevor-cortex, line 140: Still broken from my last round — `#isabortererr` doesn't resolve. The heading `isAbortError(err)` slugifies to `#isaborterrorerr` on GitHub (it keeps the `err` param). Easy one-char fix: `../promises#isaborterrorerr`.

packages/subscribable/src/action-store.ts
```diff
@@ -0,0 +1,143 @@
...
+        setState({ data: previousData, error: undefined, status: 'running' });
+        try {
+            const result = await getAbortablePromise(fn(signal, ...args), signal);
+            if (!signal.aborted) {
+                setState({ data: result, error: undefined, status: 'success' });
+            }
+            return result;
+        } catch (error) {
+            if (!signal.aborted) {
+                setState({ data: previousData, error, status: 'error' });
```
trevor-cortex: Narrow race: if `fn` resolves and *then* the signal aborts before this continuation runs, `getAbortablePromise` has already settled with `result`, the `!signal.aborted` guard correctly skips `setState`, but we still `return result` to the caller — contradicting the doc promise that `dispatchAsync` rejects with `AbortError` when superseded.

Consider:

```ts
const result = await getAbortablePromise(fn(signal, ...args), signal);
if (signal.aborted) {
    // Superseded between fn resolving and us observing it.
    throw new DOMException('Aborted', 'AbortError');
}
setState({ data: result, error: undefined, status: 'success' });
return result;
```

(Or whatever abort-error construction matches the rest of the codebase — `@solana/promises` likely has a helper.)

Minor; the window is tiny, but the alternative is softening the doc language.

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Re-review

Both open items from the last round are resolved:

- **README anchor** — now `#isaborterrorerr`. ✓
- **`dispatchAsync` supersede race** — `signal.aborted` is now checked after both the success `await` and in the catch block, throwing `signal.reason` in either case. The catch-branch guard is a nice addition I hadn't explicitly asked for: it prevents a real `fn` rejection from masking the AbortError when the call is superseded between `fn` rejecting and the continuation running. Two new tests (`resolves ... before the continuation runs`, `rejects ... before the continuation runs`) cover both windows. ✓

The `CLAUDE.md` addition codifying `Promise.withResolvers` over hand-rolled deferreds is a nice bonus.

Nothing else regressed from my perspective — happy with this now.

CLAUDE.md
```diff
@@ -70,6 +70,7 @@ All errors use the `SolanaError` class from `@solana/errors`. Key rules:
 - **Dev-only code**: Guard with `__DEV__` (e.g. verbose error messages, debug assertions).
 - **Formatting**: ESLint via `@solana/eslint-config-solana`, Prettier via `@solana/prettier-config-solana`. Run `pnpm style:fix` to auto-fix.
 - **All publishable packages share a fixed version** (currently in lockstep).
+- **Deferred promises**: Use `Promise.withResolvers<T>()` instead of hand-rolling a `new Promise((resolve, reject) => ...)` with captured externals. Do not reintroduce a `deferred()` helper — `Promise.withResolvers` already returns `{ promise, resolve, reject }`.
```
lorisleiva, line 73: Nice!

lorisleiva (commented):
This is super nice!

I like how `ReactiveStore` and `ActionStore` follow similar reactive patterns but for different purposes. One is for streaming values whereas the other allows you to dispatch new values yourself.

That being said, I think we could probably improve the way these two concepts sit next to each other. I feel like they are a lot closer than this API makes out. For instance, we should consider breaking the `ActionState` type into multiple getters like we do with the `ReactiveStore`.

|           | `ReactiveStore` | `ActionStore` (now) | `ActionStore` (proposed) |
|-----------|-----------------|---------------------|--------------------------|
| `TResult` | `getState()`    | `getState().data`   | `getState()`             |
| error     | `getError()`    | `getState().error`  | `getError()`             |
| status    | N/A [1]         | `getState().status` | `getStatus()`            |

Alternatively (if we’re worried about `getStatus()` and `getState()` not being fetched automatically) we could do the opposite and make `ReactiveStore` return a `ReactiveState`.

Either way, once we unify their APIs and place `ReactiveStore` next to `ActionStore`, it becomes clearer that the main difference between the two is that `ReactiveStore` is readonly whereas `ActionStore` is a dispatchable version of `ReactiveStore`. Therefore, we could make that association more obvious by renaming `ActionStore` to `DispatchableReactiveStore`.

What do you think?

[1]: Side note, we could even envisage adding a `getStatus()` function to `ReactiveStores` for API symmetry even though the the `running` status wouldn’t be applicable here.

mcintyre94:
@lorisleiva I like the idea of the two APIs being closer together, but I don't think doing it that way would work as well as it appears.

For reactive-store the error case is fatal, rather than being part of the lifecycle. The intended use is (currently) something like:

```ts
// ErrorBoundary somewhere above
const data = useSyncExternalStore(store.subscribe, () => {
  if (store.getError()) throw store.getError();
  return store.getState();
});
```

Here `getState()` is returning the latest data, and if there's an error somewhere then that's the end of the stream and we throw it as an error. It's designed to mirror the async generator in that way.

Whereas with action-store the intended use case is that you dispatch a function, and it could throw, but that's handled by the store. An error because the user rejected the transaction is recoverable by dispatching again, and the store handles a retried response superseding the previous one.

I also think it's important that we have `getState()` return all the data, so that `useSyncExternalStore(store.subscribe, store.getState)` gives you a cohesive type. Otherwise you'd need a separate `useSyncExternalStore` for each field that you want to track, and you wouldn't have the narrowing provided by `ActionState` - if status === 'success' then data is defined and error is undefined etc. I also don't think the react (etc) lifecycle would work as well if each field was a separate `useSyncExternalStore` because we only have one `subscribe`, not one per field.

We could expose `getData` etc functions that just do `getStore().data`, but they'd be strictly weaker typed. And given the expectation is that consumers are using `store.subscribe` (through `useSyncExternalStore` or otherwise), I don't think they'd be useful in most cases.

I think that the action-store design is generally a bit better than reactive-store, because of the more cohesive `getState` type  (with status) and the retry-ability.

We likely can't make `createReactiveStoreFromDataPublisher` retry-able as-is, because it's taking a DataPublisher as input. We can't restart that if there's an error and it's dead, unless we change it to receive a factory function that creates a `DataPublisher`, then it could recreate it. Maybe that's worth doing - we could just deprecate the current version that doesn't use a factory to avoid the breaking change. Then we could make `ReactiveStore` include a `retry` or `restart` function. Noop if running, recreate if there was an error.

This would also work for `createReactiveStoreWithInitialValueAndSlotTracking` as-is, since that takes pending RPC/subscription requests as input and could restart them. This would make it more similar to action-store in that way - it still tears down the subscription on error but knows how to rebuild it if you want to. This would make the store returned by that function more similar to this action-store.

There is also an argument to make `getState()` return something more like the discriminated union state type used here, which would be a breaking change. But at that point the error case becomes more like action-store: a recoverable part of the lifecycle that you might just want to flag in the UI. Maybe we add that as `getUnifiedState()` or something inevitably annoying, and then in a breaking change we can make `getState()` change to that.

This would essentially lead us to a similar outcome of the stores having a more similar shape, but from the opposite direction - bringing the more flexible API here into reactive-store. The state machines are much more similar, and the `getState` snapshot state `{data, error, status}` is the same (status values probably different though).

TLDR: I think there's a stronger case for unifying by bringing the unified state type and retry functionality into reactive-store

lorisleiva:
@mcintyre94 Yeah that's what I meant with my alternative solution:

> Alternatively (if we’re worried about getStatus() and getState() not being fetched automatically) we could do the opposite and make ReactiveStore return a ReactiveState.

Which I now realise contain a typo and `automatically` should say `atomically`! 😅 

I think the reverse makes a lot of sense so you get these "snapshots" in one call to ensure the data is coherent together. As long as the APIs get closer together I think that's good. I think a temporary `getUnifiedState` on `ReactiveStore` makes sense to facilitate the v7 migration.

What do you think about the name `DispatchableReactiveStore` instead of `ActionStore`? Do you think it describes what an action store is better? Or do you prefer coining a new concept in case the APIs diverge further from each other in the future?

mcintyre94:
@lorisleiva Gotcha, I'll work on a PR to make those changes to reactive-store independently.

I think it makes sense to have them as different concepts, because they still remain quite different. I don't feel too strongly either way though because in any case this is a low level helper. But IMO the relationship can't really be described as "reactive-store, but you can dispatch to it":

- Lifecycle is quite different, action-store is idle until dispatched, while reactive-store is controlled by its underlying subscription. Arguably `Dispatchable` captures this, but it's more than just being able to dispatch to it
- The fact it's dispatch-able changes the signature quite a bit: action-store takes params that are passed to the wrapped function, while reactive-store has everything baked into its subscription. `retry()` doesn't take arguments (when added to reactive-store), while each call to `dispatch` does. You can pass different params each time in action-store.
- For action-store it's one shot, so after success it's done until you dispatch again. Whereas with reactive-store it keeps streaming new data. `DispatchableReactiveStore` suggests behaviour more similar to `ReactiveStore` than we actually have IMO  
- Because of that we  have the async/sync dispatch distinction here but not in reactive-store - you can't await the end of a stream while you can await a `dispatchAsync` here 

(why doesn't github have threads...)

lorisleiva:
@mcintyre94 That's fair. Looking back I think the main thing bothering me here in the `ReactiveStore`/`ActionStore` naming is that it implies `ActionStore` isn't reactive when both are meant to be used as reactive primitives.

Maybe naming them something like `ReactiveStreamStore`/`ReactiveActionStore` (or better name) makes more sense?

mcintyre94:
@lorisleiva Yep I think that makes sense - the mistake was probably using `ReactiveStore` for a quite specific implementation of the concept. I think it's probably not worth a breaking change, but maybe we should re-export `ReactiveStore` as `ReactiveStreamStore` for now? And then rename this one `ReactiveActionStore`.

lorisleiva:
@mcintyre94 Sounds perfect! 👌

mcintyre94:
Draft for now - I'll move this to stack after the changes to `ReactiveStore`, because the rename here will conflict otherwise, and it'll improve the documentation here if we land those changes first. 

Edit: Rebased + renamed ActionStore to ReactiveActionStore throughout

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Re-review (rename round)

The `Reactive`-prefix rename is mechanical and clean — `ActionStore` → `ReactiveActionStore`, `createActionStore` → `createReactiveActionStore`, file/test renamed accordingly, README and barrel updated. Implementation is byte-for-byte the same as last round modulo identifier names; the supersede / reset / dispatchAsync race fixes from prior rounds are preserved. No new code-level concerns.

## On Loris's API-unification proposal

For subsequent reviewers — my two cents on Loris's comment, since it's a design call rather than a correctness one:

**The two APIs are already heading toward the same shape, just under different names.** `ReactiveStreamStore` already exposes `getUnifiedState()` returning a single discriminated `ReactiveState<T>` snapshot, and the per-field `getState()` / `getError()` getters on it are explicitly `@deprecated` in favour of that unified getter (see the README and JSDoc on `reactive-stream-store.ts`). So the modern, recommended `ReactiveStreamStore` API is _already_ a single-snapshot getter, exactly like what `ReactiveActionStore` ships in this PR.

Given that, I'd lean against the first option in Loris's table (splitting `ReactiveActionState` back into `getStatus()` / `getError()` / etc.) — that would walk away from the direction Stream is already headed and re-introduce the snapshot-identity / `useSyncExternalStore` footgun that `getUnifiedState()` was added to fix.

The friction Loris is pointing at feels more like **naming inconsistency** than shape inconsistency:

| Concept | Stream (today) | Action (this PR) |
|---|---|---|
| Discriminated snapshot getter | `getUnifiedState()` | `getState()` |
| Snapshot type | `ReactiveState<T>` | `ReactiveActionState<TResult>` |

IMO the cleanest unification is to align _names_, not shapes. Two reasonable directions:

1. **Action adopts Stream's vocabulary** — rename to `getUnifiedState()` here too, and call the snapshot type `ReactiveActionState<…>` (already does). Pro: zero churn on the Stream side, which already has deprecation cycles in flight. Con: `getUnifiedState` is a slightly awkward name when there's nothing to unify _against_ on the action side.
2. **Stream adopts Action's vocabulary** — promote `getState()` to be the unified getter on Stream (the deprecated value-only `getState()` is in the way, but it's already deprecated, so a future major can reclaim the name). Pro: `getState()` reads more naturally and matches `useSyncExternalStore` docs. Con: adds churn to the Stream deprecation plan.

I'd weakly prefer (2) long-term but (1) is the lower-risk move for this PR if you want to land it without dragging Stream into scope.

On the **`DispatchableReactiveStore` rename** — I'd push back gently. "ActionStore" is widely-recognised vocabulary (Redux/Flux/Zustand lineage) and `ReactiveActionStore` already telegraphs the relationship via the shared `Reactive*Store` prefix. `DispatchableReactiveStore` is more architecturally precise but loses the discoverability of "action" as a concept. The shared prefix arguably already does the symmetry work the rename is trying to do.

Not blocking either way — flagging this for discussion before the next push so Loris and Callum can land on a direction without a third rename round.

- Behaviour and tests are unchanged from the previous approval-shaped pass; no need to re-verify the supersede / reset / `dispatchAsync` race semantics.
- The only outstanding question is the API-shape one above; once that's settled the code itself is in good shape.

lorisleiva (approved):
Thanks for the changes!

mcintyre94:
### Merge activity

* **May 7, 2:56 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1550).
* **May 7, 3:22 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1550) rebased this pull request as part of a merge.
* **May 7, 3:32 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1550).

PR #1551, opened 2026-04-21
Add ClientWithSubscribeToPayer/Identity interfaces

mcintyre94:
This PR adds new types `ClientWithSubscribeToPayer` and `ClientWithSubscribeToIdentity` in `plugin-interfaces`. These allow a plugin (such as the wallet plugin) to notify subscribers when they modify these fields, without those consumers needing to know which plugin has made them reactive.

It's already the case that `client.payer` can be made dynamic by the wallet plugin, by using a get() function. This retains compatibility with the `payer: TransactionSigner` type. But in order to make this useful for a reactive UI that may for example render based on the current value of `client.payer`, we also need to be able to subscribe to changes.

Reactive consumers can then use `client.subscribeToPayer(() => {})` to receive these notifications.

Alternatives considered:

- Add these types to the wallet plugin only. This would work, but this would need to be duplicated by any other plugin that wants to make these fields reactive. In practice we'd be binding consumers (like a react library) to the wallet plugin's types. I think it makes more sense to define the types as plugin-interfaces and allow consumers to be agnostic to how the fields are made dynamic
- Make payer and identity actually act as a reactive store (with subscribe and getValue functions), this would be a breaking change and annoying for most uses. These types provides the same functionality only to consumers that need to react to changes to these values. This is likely only reactive UI - anything using client.payer already gets the correct updated value

Extracted from https://github.com/anza-xyz/kit-plugins/pull/200, which originally added these types to kit-plugin-signer.

### DIFF
--- a/packages/plugin-interfaces/README.md
+++ b/packages/plugin-interfaces/README.md
@@ -72,6 +72,32 @@ function nftPlugin() {
 }
 ```
 
+### `ClientWithSubscribeToPayer` / `ClientWithSubscribeToIdentity`
+
+Some plugins set `client.payer` or `client.identity` reactively — the connected wallet may change, an account may be swapped, or a signer may be cleared on disconnect. Plugins that participate in this pattern advertise it by installing a sibling `subscribeTo<Capability>` function on the client:
+
+| Type                            | Sibling function      | Advertises that…                       |
+| ------------------------------- | --------------------- | -------------------------------------- |
+| `ClientWithSubscribeToPayer`    | `subscribeToPayer`    | `client.payer` may change over time    |
+| `ClientWithSubscribeToIdentity` | `subscribeToIdentity` | `client.identity` may change over time |
+
+Reactive consumers (framework hooks, stores, effects) can then observe changes without having to know which plugin installed the capability — they duck-type on the subscribe function:
+
+```ts
+import { ClientWithPayer, ClientWithSubscribeToPayer } from '@solana/plugin-interfaces';
+
+function observePayer() {
+    return <T extends ClientWithPayer & ClientWithSubscribeToPayer>(client: T) => {
+        client.subscribeToPayer(() => {
+            console.log('payer is now', client.payer);
+        });
+        return client;
+    };
+}
+```
+
+Plugins that leave the signer fixed for the lifetime of the client do **not** need to install these hooks — there is nothing to subscribe to. The convention is meant for plugins that reassign `client.payer` / `client.identity` as the user connects, switches accounts, or disconnects.
+
 ### `ClientWithAirdrop`
 
 Represents a client that can request SOL airdrops (typically on devnet/testnet). The airdrop succeeds when the promise resolves. Some implementations (e.g., LiteSVM) update balances directly without a transaction, so no signature is returned in those cases.
--- a/.changeset/itchy-geese-crash.md
+++ b/.changeset/itchy-geese-crash.md
@@ -1,0 +1,6 @@
+---
+'@solana/plugin-interfaces': minor
+'@solana/kit': minor
+---
+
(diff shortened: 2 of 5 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds two new type-only exports to `@solana/plugin-interfaces`: `ClientWithSubscribeToPayer` and `ClientWithSubscribeToIdentity`, each advertising a sibling `subscribeTo<Capability>(listener): () => void` function on the client. The goal is to let reactive consumers (UI hooks, stores) observe changes to `client.payer` / `client.identity` without naming the specific plugin that made them reactive.

The PR is type-only — no runtime code, no behavior change for existing consumers. Comes with a changeset (`minor` on both `plugin-interfaces` and `kit`), README entry with a clear table, and compile-time type tests covering both interfaces and their composability.

## Design notes

The convention is sound and fits the package's existing style. A few things worth calling out for subsequent reviewers / implementers:

- **Listener takes no arguments** and consumers must re-read `client.payer` / `client.identity` after a notification. This is deliberate and matches the snapshot-based model used by React's `useSyncExternalStore` — the docblock explains it and notes that over-notification is acceptable. Good choice; a value-passing listener would make it too easy to leak stale references.
- **Idempotent unsubscribe** is a documented contract (`"must be idempotent"`) but not enforced by the type system. Worth verifying this actually holds in the wallet-plugin implementation ([kit-plugins#200](https://github.com/anza-xyz/kit-plugins/pull/200)) — calling unsubscribe twice should not throw or double-remove.
- **Duck-typing via intersection** (`ClientWithPayer & ClientWithSubscribeToPayer`) is the intended usage pattern shown in the README and `@example`. Plugins that don't make the capability reactive simply don't install the hook, and downstream libraries that want to stay agnostic can fall back to a no-op subscribe as the docblock suggests.
- **`readonly` modifier**: the new types use `readonly subscribeToPayer: ...` while existing sibling interfaces (`ClientWithPayer`, `ClientWithIdentity`) use plain `{ payer: TransactionSigner }`. Not worth changing — `readonly` is semantically correct here (consumers shouldn't reassign the subscribe fn), just noting the minor stylistic divergence.

## Things to watch out for

- The type name `SubscribeToCapability` reads ambiguously — it's the *signature* of a subscribe function, not "subscribe to a capability." Minor bikeshed, happy to leave as-is given the docblock clarifies it, but a name like `SubscribeToFn` would be slightly more self-describing. Not blocking.
- Since this lands in `plugin-interfaces`, any future plugin that wants to participate in the convention needs to use exactly these names (`subscribeToPayer`, `subscribeToIdentity`) — the whole point is duck-typing. Worth keeping that in mind if a third pattern emerges later (e.g. `subscribeToRpc`); the convention should probably be documented somewhere more permanent than just this module's top-of-file comment once there are 3+ instances.

Looks good to me — no blocking concerns.

mcintyre94:
Renamed `SubscribeToCapability` type to `SubscribeToFn`.

packages/plugin-interfaces/README.md
```diff
@@ -72,6 +72,29 @@ function nftPlugin() {
...
+Reactive consumers (framework hooks, stores, effects) can then observe changes without having to know which plugin installed the capability — they duck-type on the subscribe function:
+
+```ts
+import { ClientWithPayer, ClientWithSubscribeToPayer } from '@solana/plugin-interfaces';
+
+function observePayer(client: ClientWithPayer & ClientWithSubscribeToPayer) {
+    return client.subscribeToPayer(() => {
+        console.log('payer is now', client.payer);
+    });
+}
```
lorisleiva, line 96: nit: The convention on this README is to use plugin factories as examples. For consistency this should look something like:

```ts
import { ClientWithPayer, ClientWithSubscribeToPayer } from '@solana/plugin-interfaces';

function observePayer() {
    return <T extends ClientWithPayer & ClientWithSubscribeToPayer>(client: T) => {
        client.subscribeToPayer(() => {
            console.log('payer is now', client.payer);
        });
        return client;
    };
}
```

lorisleiva (approved):
Thanks!

PR #1552, opened 2026-04-21
Add `retry` and `getUnifiedState` to `ReactiveStore`

mcintyre94:
#### Problem

`ReactiveStore` exposed `getState()` and `getError()` as separate getters with no unified snapshot, which made `useSyncExternalStore` integration awkward. Errors were also terminal — once a stream failed, the only recovery was to rebuild the whole store, losing subscribers and last-known state in the process.

#### Summary of Changes

This PR brings `ReactiveStore` closer to the proposed `ActionStore`: https://github.com/anza-xyz/kit/pull/1550. See discussion: https://github.com/anza-xyz/kit/pull/1550#pullrequestreview-4147500351 

- `ReactiveStore<T>` gains `retry(): void` and `getUnifiedState(): ReactiveState<T>`. `ReactiveState<T>` is a discriminated union over `loading | loaded | error | retrying` with stable snapshot identity across reads. The existing `getState()` and `getError()` getters are preserved but marked `@deprecated`. A future breaking change can make `getState()` the new `getUnifiedState`
- New `createReactiveStoreFromDataPublisherFactory` accepts a `() => Promise<DataPublisher>` factory and supports real retry, wiring a fresh inner abort signal for each connection attempt.
- The existing `createReactiveStoreFromDataPublisher` is now `@deprecated`. Its `retry()` throws a new `SolanaError` with code `SOLANA_ERROR__SUBSCRIBABLE__RETRY_NOT_SUPPORTED` since a ready-made `DataPublisher` can't be restarted.
- `createReactiveStoreWithInitialValueAndSlotTracking` (in `@solana/kit`) now implements `retry()` by re-invoking the pending RPC request and subscription on a new inner abort signal, preserving `lastUpdateSlot` and the last known value.
- Updated tests for both factories and the slot-tracking store; updated the subscribable README.


mcintyre94:
@trevor-cortex

packages/subscribable/src/reactive-store.ts
```diff
@@ -76,66 +135,188 @@ export type ReactiveStore<T> = {
...
+            return currentState.data;
+        },
+        getUnifiedState(): ReactiveState<TData> {
             return currentState;
         },
+        retry(): void {
+            if (currentState.status !== 'error') return;
+            currentState = { data: currentState.data, error: undefined, status: 'retrying' };
+            notify();
+            connect();
```
trevor-cortex, line 322: **Bug:** If the caller's `abortSignal` has fired before `retry()` is called, this transitions `currentState` to `'retrying'`, calls `notify()`, and then `connect()` short-circuits on `if (outerController.signal.aborted) return;` — leaving the store permanently stuck in `'retrying'` with no further transitions possible. Any subscriber that reads `getUnifiedState()` after this will see `status: 'retrying'` forever.

This also contradicts the docblock on `FactoryConfig.abortSignal`: _"Subsequent calls to `retry()` are no-ops once this signal has fired."_

Suggested fix — short-circuit before mutating state:

```ts
retry(): void {
    if (outerController.signal.aborted) return;
    if (currentState.status !== 'error') return;
    currentState = { data: currentState.data, error: undefined, status: 'retrying' };
    notify();
    connect();
},
```

The existing `retry() after abort is a no-op` test only asserts `mockRequest` wasn't called; it doesn't assert the store's state, so it passes despite the bug. Worth extending that test to also assert `store.getUnifiedState().status !== 'retrying'` (e.g. still `'error'`).

packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts
```diff
@@ -95,66 +104,81 @@ export function createReactiveStoreWithInitialValueAndSlotTracking<TRpcValue, TS
...
+            return currentState.data;
+        },
+        getUnifiedState(): ReactiveState<SolanaRpcResponse<TItem>> {
             return currentState;
         },
+        retry(): void {
+            if (currentState.status !== 'error') return;
+            currentState = { data: currentState.data, error: undefined, status: 'retrying' };
+            notify();
+            connect();
```
trevor-cortex, line 184: Same `retry()`-after-abort issue as in `reactive-store.ts`: this sets `status: 'retrying'` and notifies before `connect()` gets a chance to short-circuit on `outerController.signal.aborted`, leaving the store stuck.

Suggested fix:

```ts
retry(): void {
    if (outerController.signal.aborted) return;
    if (currentState.status !== 'error') return;
    currentState = { data: currentState.data, error: undefined, status: 'retrying' };
    notify();
    connect();
},
```

The `is a no-op after the caller has aborted` test here has the same gap — it only checks call counts, not the resulting `status`.

packages/subscribable/src/reactive-store.ts
```diff
@@ -76,66 +135,188 @@ export type ReactiveStore<T> = {
...
+        const forwardAbort = () => innerController.abort(outerController.signal.reason);
+        outerController.signal.addEventListener('abort', forwardAbort, { signal: innerController.signal });
+        createDataPublisher().then(
+            publisher => {
+                if (innerController.signal.aborted) return;
+                publisher.on(
+                    dataChannelName,
+                    data => {
+                        currentState = { data: data as TData, error: undefined, status: 'loaded' };
+                        notify();
```
trevor-cortex, line 281: Nice — the `{ signal: innerController.signal }` scoping on this `addEventListener('abort', ...)` call ensures `forwardAbort` is removed when the inner signal aborts, so repeated `retry()` calls don't leak one listener per retry onto the outer signal. Worth a one-line comment here to call that out explicitly, since it's a load-bearing subtlety that's easy to "simplify away" in a future refactor:

```ts
// Scope this forwarder to the inner signal so it's removed on reconnection
// and we don't accumulate listeners on the outer signal across retries.
outerController.signal.addEventListener('abort', forwardAbort, { signal: innerController.signal });
```

packages/subscribable/README.md
```diff
@@ -105,7 +117,9 @@ Things to note:
 
 ### `createReactiveStoreFromDataPublisher({ abortSignal, dataChannelName, dataPublisher, errorChannelName })`
```
trevor-cortex, line 119: Minor, but worth making the crash behaviour more prominent here: readers might skim the note and assume "deprecated but still works everywhere", then wire `onRetry={store.retry}` to a button (matching the React example in the `ReactiveStore<T>` section above) and get an uncaught `SolanaError` at click-time.

Consider hoisting the "calling `retry()` throws" clause to its own sentence (or bold fragment) so it reads more like a warning than a footnote, or mention it in the React example for the type itself. Not a blocker.

packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts
```diff
@@ -95,66 +104,81 @@ export function createReactiveStoreWithInitialValueAndSlotTracking<TRpcValue, TS
...
+        function handleValue(value: SolanaRpcResponse<TItem>) {
+            currentState = { data: value, error: undefined, status: 'loaded' };
+            notify();
+        }
+
+        rpcRequest
+            .send({ abortSignal: innerSignal })
+            .then(({ context: { slot }, value }) => {
+                if (innerSignal.aborted) return;
+                if (slot < lastUpdateSlot) return;
```
trevor-cortex, line 145: Worth a brief comment here noting that `lastUpdateSlot` is intentionally *not* reset on retry — so a reconnect can't cause the store to regress to an older slot. It's mentioned in the PR description and the top-of-file docblock, but a comment at the site of the check would help a future reader chasing a "why is my retry stuck in `retrying`" bug understand it's by design, and that if the retried RPC happens to return a slot below `lastUpdateSlot`, the store will stay in `retrying` until the subscription catches up.

trevor-cortex (commented):
Nice, focused evolution of `ReactiveStore` toward the `ActionStore` shape. The unified `{ data, error, status }` snapshot with stable identity is exactly what `useSyncExternalStore` needs, and preserving the existing getters (deprecated) keeps this non-breaking. Factoring the factory variant out as a separate function rather than overloading the existing one is the right call — the semantics of `retry()` differ enough to justify the split.

One correctness bug worth fixing before merge, plus a few smaller notes below.

### Things to watch out for

- **`retry()` after the outer abort leaves the store stuck in `retrying`** (see inline). The guard is in `connect()` but `retry()` has already transitioned state by the time it's hit. Both the factory variant and the slot-tracking variant have this.
- **Asymmetric `retry()` behaviour across factories.** The deprecated `createReactiveStoreFromDataPublisher` *throws* from `retry()`; the factory variant is a silent no-op when not in `error`. Defensible (the throw surfaces the deprecation loudly), but it means a React component that wires `retry` to a button can crash if it happens to be rendered against a non-retryable store. Worth flagging more prominently in the README snippet for the deprecated function.
- **`lastUpdateSlot` is preserved across retry in the slot-tracking store** (intentional and documented). Be aware this means a retry whose RPC + subscription both initially return slots `< lastUpdateSlot` leaves the store stuck in `retrying` until the subscription catches up. In practice that window is tiny, but it's worth a brief comment at the site of the slot check.

- The inner/outer abort-controller pattern in the factory variant is the trickiest part to eyeball. Specifically worth re-checking: (a) stale listeners from a prior connection are aborted before `retry()` wires up the new one, and (b) the `outerController.signal.addEventListener('abort', forwardAbort, { signal: innerController.signal })` self-unregistration pattern (so we don't leak one listener per retry onto the outer signal). Both looked correct to me, but they're load-bearing.
- Test coverage for the new API is thorough across all three sites. The one gap I'd suggest is that the "retry after abort is a no-op" tests only assert call counts, not the resulting `status` — which is why they pass despite the state-machine bug flagged below.
- Changeset bump is `minor` across `@solana/errors`, `@solana/subscribable`, `@solana/kit` — matches the CLAUDE.md convention for new APIs. The existing getters are only `@deprecated` (not removed), so this is additive.
- Error code `SOLANA_ERROR__SUBSCRIBABLE__RETRY_NOT_SUPPORTED = 8195000` opens a new reserved range `[8195000-8195999]` for the `SUBSCRIBABLE` domain. Looks consistent with the existing registry layout.

mcintyre94:
* **#1552** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1552?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1552?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/subscribable/README.md
mcintyre94: I think this is clear enough in practice
Screenshot 2026-04-21 at 18 04 41

lorisleiva (approved):
Reviewed this the wrong way around but I get what you're doing haha. Nice changes!

mcintyre94:
### Merge activity

* **May 7, 2:56 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1552).
* **May 7, 2:57 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1552).

PR #1553, opened 2026-04-21
Add sync reactiveStore() to pending RPC subscription

mcintyre94:
#### Summary of Changes

The previous PR updated `ReactiveStore` , and added a new `createReactiveStoreFromDataPublisherFactory` function that takes a `createDataPublisher(): Promise<DataPublisher>`as input, instead of just `DataPublisher`. This enables retry functionality, where the publisher is re-created and re-subscribed to, without losing other store state.

But it also enables `rpcSubscriptions.someNotification(...).reactive()` to be made synchronous: previously it needed to create the `DataPublisher` asynchronously, now it can pass that promise to `createReactiveStoreFromDataPublisherFactory`. That function already handles the lifecycle of creating the `DataPublisher` from the factory. 

This is a significantly better API for reactive consumers, as they have a synchronous store available immediately that will begin publishing data. They no longer need to handle the async store creation. The store's `loading` state covers both the `DataPublisher` being created and waiting for the first publish, and handles errors creating the `DataPublisher`. In react specifically, there's no need to awkwardly wrap the promise in a `useEffect` and handle extra state. 

React example:

```tsx
function SlotDisplay({ rpcSubscriptions }: Props) {
    const { store, controller } = useMemo(() => {
        const controller = new AbortController();
        const store = rpcSubscriptions
            .slotNotifications()
            .reactiveStore({ abortSignal: controller.signal });
        return { store, controller };
    }, [rpcSubscriptions]);

    useEffect(() => () => controller.abort(), [controller]);

    const state = useSyncExternalStore(store.subscribe, store.getUnifiedState);

    if (state.status === 'loading' || state.status === 'retrying') return <Spinner />;
    if (state.status === 'error') return <button onClick={store.retry}>Retry</button>;
    return <div>Slot: {state.data.slot}</div>;
}
```

Making `reactive()` synchronous would be a breaking change, so this PR adds `.reactiveStore()` as a sync method on pending RPC subscriptions. `reactive()` is deprecated.

TBH I think `reactiveStore()` might be the better name, but we could switch back to `reactive()` as a breaking change later if we want to.


mcintyre94:
* **#1553** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1553?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1553?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
### Summary

Adds a synchronous `reactiveStore()` method to `PendingRpcSubscriptionsRequest` alongside the existing (now `@deprecated`) async `reactive()`. The new method delegates to `createReactiveStoreFromDataPublisherFactory` from the prior PR, so consumers get a store they can use immediately — `status: 'loading'` covers publisher creation *and* waiting for the first notification, and `retry()` re-invokes the transport. This is a clear UX win for `useSyncExternalStore` and similar reactive primitives, where having to await a promise before you have a store is awkward.

The changeset correctly marks this as a `minor` bump on `@solana/rpc-subscriptions-spec` and `@solana/kit`. Since the repo is pre-1.0 and `reactive()` is only soft-deprecated, this is appropriately non-breaking.

### Key things to watch for

- **`reactive()` is only JSDoc-deprecated**, still fully functional. Downstream mocks in `kit` and `transaction-confirmation` add a no-op `reactiveStore: jest.fn()` purely to satisfy the interface — confirms nothing internal has migrated yet, which is fine for this PR.
- **Naming**: as the author calls out, `reactiveStore()` is more descriptive than `reactive()`. Worth locking in the name before 1.0 — if the intent is eventually to rename back to `reactive()` as a breaking change, committing to one or the other now would avoid carrying both long-term.
- **Test relocation**: the old `rpc-subscriptions-reactive-test.ts` was deleted and its suite moved verbatim into `rpc-subscriptions-test.ts` alongside the new `reactiveStore()` suite. Pure reorganization, no behavioural change.

- The `reactiveStore()` implementation in `rpc-subscriptions.ts` is a thin pass-through: it forwards the caller's `abortSignal` both to the factory (as the outer signal) and into the transport on each connect. Mirrors how `reactive()`/`subscribe()` already handle abort propagation.
- The new test suite covers loading, notifications, error surfacing, and retry. Two gaps worth considering as follow-ups (not blocking): (1) a test that aborting the caller's signal actually stops the store from receiving further notifications, and (2) a test that `getUnifiedState()` returns a stable object reference between unrelated calls — that stability is explicitly documented as important for `useSyncExternalStore`.
- `flushMicrotasks()` in the new test ticks twice — enough for the `.then()` inside `connect()` plus the factory resolution. A one-line comment explaining the two ticks would help future readers.

Overall shape LGTM.

packages/rpc-subscriptions-spec/src/__tests__/rpc-subscriptions-test.ts
```diff
@@ -31,3 +33,168 @@ describe('createSubscriptionRpc', () => {
...
+        expect(store.getError()).toBe(error);
+    });
+});
+
+describe('PendingRpcSubscriptionsRequest.reactiveStore()', () => {
+    let mockTransport: jest.MockedFunction<RpcSubscriptionsTransport>;
+    let mockOn: jest.Mock;
+    let mockDataPublisher: DataPublisher;
+    let rpcSubscriptions: RpcSubscriptions<TestRpcSubscriptionNotifications>;
+    function publish(type: string, payload: unknown) {
```
trevor-cortex, line 110: Minor: a quick comment explaining why two microtask ticks are needed (one for the `createDataPublisher()` promise inside `connect()`, one for the `.then` handler that wires up the `on(...)` listeners) would save the next reader from having to reverse-engineer it. Non-blocking.

packages/rpc-subscriptions-spec/src/__tests__/rpc-subscriptions-test.ts
```diff
@@ -31,3 +33,168 @@ describe('createSubscriptionRpc', () => {
...
+        const store = rpcSubscriptions
+            .thingNotifications()
+            .reactiveStore({ abortSignal: new AbortController().signal });
+        await flushMicrotasks();
+        publish('error', new Error('stream died'));
+        store.retry();
+        await flushMicrotasks();
+        expect(mockTransport).toHaveBeenCalledTimes(2);
+    });
+});
```
trevor-cortex, line 223: Nice coverage of the retry path. Two worthwhile follow-ups, not blocking:

1. A test that triggering the caller's `abortSignal` stops further notifications from reaching the store — the factory's outer/inner controller wiring is the subtle bit, worth pinning down.
2. A test that `getUnifiedState()` returns the same object reference across back-to-back calls when state hasn't changed. The `ReactiveStore` JSDoc explicitly calls this out as important for `useSyncExternalStore`, so it would be good to guard it against regressions.

packages/rpc-subscriptions-spec/src/rpc-subscriptions-request.ts
```diff
@@ -28,8 +28,30 @@ export type PendingRpcSubscriptionsRequest<TNotification> = {
...
+     * ```ts
+     * const store = rpc.accountNotifications(address).reactiveStore({ abortSignal });
+     * // React — the unified snapshot has stable identity per update.
+     * const state = useSyncExternalStore(store.subscribe, store.getUnifiedState);
+     * if (state.status === 'error') return <ErrorMessage error={state.error} onRetry={store.retry} />;
+     * if (state.status === 'loading') return <Spinner />;
+     * return <View data={state.data} />;
+     * ```
+     */
+    reactiveStore(options: RpcSubscribeOptions): ReactiveStore<TNotification>;
```
trevor-cortex, line 54: Nit on naming, echoing the PR description: `reactiveStore()` is genuinely clearer than `reactive()` (tells you what you get, not just that it's "reactive"). If there's appetite for making this *the* name long-term, it might be worth committing now rather than leaving both around — carrying a deprecated `reactive()` alongside `reactiveStore()` in perpetuity is a bit of surface area. Happy either way; the shape of this PR is right regardless.

mcintyre94: Kit is not pre v1.0, deprecating is the right thing

lorisleiva (approved):
Nice! I like that.

mcintyre94:
### Merge activity

* **May 7, 2:56 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1553).
* **May 7, 2:59 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1553) rebased this pull request as part of a merge.
* **May 7, 3:09 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1553).

PR #1554, opened 2026-04-22
Rename ReactiveStore to ReactiveStreamStore

mcintyre94:
#### See discussion: https://github.com/anza-xyz/kit/pull/1550#issuecomment-4290107357

We decided to rename `ReactiveStore` to `ReactiveStreamStore`, to make space for an orthogonal `ReactiveActionStore`.

PR #1555, opened 2026-04-22
Add reactiveStore to PendingRpcRequest

mcintyre94:
#### Summary of Changes

This PR adds `.reactiveStore()`to `PendingRpcRequest`, similar to how we use it for `PendingRpcSubscriptionRequest`. This sends the request and synchronously returns a `ReactiveActionStore.`

As with our other reactive store features, this is intended to provide a better API for reactive consumers:

```typescript
const store = rpc.getAccountInfo(address).reactiveStore();
const state = useSyncExternalStore(store.subscribe, store.getState);
if (state.status === 'error') return <ErrorMessage error={state.error} onRetry={store.dispatch} />;
if (state.status === 'running' && !state.data) return <Spinner />;
return <View data={state.data!} />;
```

Note that this has auto-dispatch semantics, meaning calling it has the same sort of behaviour as calling `send()` or the similar `reactiveStore()` on subscriptions.

Consumers that want to control sending precisely can use `createActionStore((signal) => rpc.getAccountInfo(address).send({abortSignal: signal})` - this is just a thin wrapper that adds auto-dispatch and therefore skips the `idle` state.

### DIFF
--- a/packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
+++ b/packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
@@ -419,6 +419,9 @@ describe('account loader', () => {
                 // First we should see `getMultipleAccounts` used for the first two layers
                 rpc.getMultipleAccounts
                     .mockImplementationOnce(() => ({
+                        reactiveStore: jest.fn().mockImplementation(() => {
+                            throw new Error('not implemented');
+                        }),
                         send: () =>
                             Promise.resolve(
                                 getMultipleAccountsMockResponse([
@@ -442,6 +445,9 @@ describe('account loader', () => {
                             ),
                     }))
                     .mockImplementationOnce(() => ({
+                        reactiveStore: jest.fn().mockImplementation(() => {
+                            throw new Error('not implemented');
+                        }),
                         send: () =>
                             Promise.resolve(
                                 getMultipleAccountsMockResponse([
@@ -464,6 +470,9 @@ describe('account loader', () => {
                 // Then we should see `getAccountInfo` used for the single
                 // account in the last layer
                 rpc.getAccountInfo.mockReturnValue({
+                    reactiveStore: jest.fn().mockImplementation(() => {
+                        throw new Error('not implemented');
+                    }),
                     send: jest.fn().mockResolvedValueOnce({
                         context: {
                             slot: 0,
--- a/packages/rpc-spec/src/__tests__/rpc-test.ts
+++ b/packages/rpc-spec/src/__tests__/rpc-test.ts
@@ -80,6 +80,114 @@ describe('JSON-RPC 2.0', () => {
             expect(rpc).not.toHaveProperty('then');
         });
     });
+    describe('when calling reactiveStore() on a pending request', () => {
+        let execute: jest.Mock;
+        let rpc: Rpc<TestRpcMethods>;
+        beforeEach(() => {
+            jest.useFakeTimers();
+            execute = jest.fn(
+                () =>
+                    new Promise(() => {
+                        /* never resolve */
+                    }),
+            );
+            rpc = createRpc({
+                api: new Proxy({} as RpcApi<TestRpcMethods>, {
+                    get() {
+                        return (..._params: unknown[]): RpcPlan<unknown> => ({ execute });
+                    },
+                }),
+                transport: makeHttpRequest,
+            });
+        });
+        afterEach(() => {
+            jest.useRealTimers();
+        });
+        it('fires the request on creation with a non-aborted signal', () => {
+            rpc.someMethod(123).reactiveStore();
+            expect(execute).toHaveBeenCalledTimes(1);
+            const { signal } = execute.mock.calls[0][0];
+            expect(signal).toBeInstanceOf(AbortSignal);
+            expect(signal.aborted).toBe(false);
+        });
+        it('forwards the transport to the plan on creation', () => {
+            rpc.someMethod(123).reactiveStore();
+            expect(execute).toHaveBeenCalledWith(expect.objectContaining({ transport: makeHttpRequest }));
+        });
+        it('returns a store synchronously in the `running` status', () => {
+            const store = rpc.someMethod(123).reactiveStore();
+            expect(store.getState()).toStrictEqual({
+                data: undefined,
+                error: undefined,
+                status: 'running',
+            });
+        });
+        it('transitions to `success` with resolved data once the plan resolves', async () => {
+            expect.assertions(1);
+            const { promise, resolve } = Promise.withResolvers<number>();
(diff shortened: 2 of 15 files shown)


mcintyre94:
* **#1555** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1555?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1555?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/rpc-spec/src/__tests__/rpc-test.ts
```diff
@@ -80,6 +80,111 @@ describe('JSON-RPC 2.0', () => {
             expect(rpc).not.toHaveProperty('then');
         });
     });
+    describe('when calling reactiveStore() on a pending request', () => {
+        let execute: jest.Mock;
+        let rpc: Rpc<TestRpcMethods>;
+        beforeEach(() => {
+            jest.useFakeTimers();
+            execute = jest.fn(
```
trevor-cortex, line 88: Minor: `jest.useFakeTimers()` is installed in this `beforeEach` but never torn down. The next `describe` block (`when calling a method having a concrete implementation`) inherits fake timers as a result. Its one test is synchronous so it still passes today, but this is a landmine for future additions. Consider an `afterEach(() => { jest.useRealTimers(); })` in this block.

packages/rpc-spec/src/rpc.ts
```diff
@@ -96,6 +118,11 @@ function createPendingRpcRequest<TRpcMethods, TRpcTransport extends RpcTransport
     plan: RpcPlan<TResponse>,
 ): PendingRpcRequest<TResponse> {
     return {
+        reactiveStore(): ReactiveActionStore<[], TResponse> {
+            const store = createReactiveActionStore<[], TResponse>(signal => plan.execute({ signal, transport }));
```
trevor-cortex, line 122: Question (non-blocking): would it be worth accepting an optional `{ abortSignal }` here, mirroring `send(options)`? Use cases: binding the store's lifetime to a parent controller (e.g. Svelte `onDestroy`, Vue `onScopeDispose`, or a composed cancellation tree) without having to manually wire `reset()`. The escape-hatch pattern in the PR description (`createActionStore(signal => ...send({ abortSignal: signal }))`) works but loses the one-liner ergonomics. Happy to defer — this is purely additive and can be layered on later.

packages/rpc-graphql/src/loaders/__tests__/account-loader-test.ts
```diff
@@ -419,6 +419,7 @@ describe('account loader', () => {
                 // First we should see `getMultipleAccounts` used for the first two layers
                 rpc.getMultipleAccounts
                     .mockImplementationOnce(() => ({
+                        reactiveStore: jest.fn(),
```
trevor-cortex: Nit: the mocks in `kit` and `accounts` use `jest.fn().mockImplementation(() => { throw new Error('not implemented'); })` so that any accidental call surfaces loudly, while here (and in `transaction-confirmation`) `reactiveStore: jest.fn()` silently returns `undefined`. Harmless since these tests don't invoke it, but aligning on one style would be nice for consistency and for future debugging.

trevor-cortex (commented):
Nice, clean addition. The `reactiveStore()` method mirrors the existing `PendingRpcSubscriptionRequest.reactiveStore()` pattern, and the implementation is a tiny, easy-to-reason-about wrapper: `createReactiveActionStore` + immediate auto-dispatch. Docblock + README + changeset + tests are all in place. LGTM modulo the comments below.

**Summary**
- New `reactiveStore(): ReactiveActionStore<[], TResponse>` method on `PendingRpcRequest`, auto-fires on construction and returns a store with the standard `{ subscribe, getState, dispatch, dispatchAsync, reset }` contract.
- `@solana/subscribable` added as a runtime dep of `@solana/rpc-spec`.
- `ES2024.Promise` added to the `rpc-spec` tsconfig `lib` (needed for `Promise.withResolvers()` in the new tests).
- Mock `PendingRpcRequest` shapes updated across several packages (`accounts`, `kit`, `rpc-graphql`, `transaction-confirmation`) to satisfy the expanded type.
- New CLAUDE.md guidance about preferring `jest.useFakeTimers()` + `runAllTimersAsync()` over hand-rolled microtask flushers, which the new tests themselves follow.

**Things to watch / notes for subsequent reviewers**
- **No external `abortSignal` option on `reactiveStore()`** — unlike `send({ abortSignal })`. Consumers tie cancellation to `store.reset()` or let `dispatch()` supersede in-flight calls. The PR description mentions the escape hatch (`createActionStore((signal) => rpc.getAccountInfo(address).send({ abortSignal: signal }))`), which is fine, but worth confirming this is the intended ergonomic — a follow-up that accepts an external signal wouldn't be a breaking change, so deferring is reasonable.
- **Auto-dispatch semantics**: the `ReactiveActionStore` returned here can re-enter `idle` via `reset()` but not re-enter `running` without an explicit `dispatch()`. Documented in the changeset/README; flagging because it's the main behavioural subtlety.
- **`@solana/subscribable` in `@solana/rpc-spec`'s runtime deps** pulls the `event-target-impl` / `promises` chain into the spec package. Consistent with the subscription-side precedent, but it does slightly expand the spec package's surface — worth a sanity-check.
- Test hygiene: the new `describe` installs fake timers in `beforeEach` with no matching `afterEach` to restore real timers — see inline comment.
- Bundle-size impact on `@solana/kit` via the new transitive dep — BundleMon will tell us, but worth eyeballing.

packages/rpc-spec/src/rpc.ts
mcintyre94: Considered this but I don't think it's needed. There's nothing that can leak here - the http request is abortable and doesn't leave anything behind. In contrast to subscriptions where we have a subscription to clean up.

Having both signals requires fiddly wiring in subscriptions, which is worth it there but would just be complexity for no value here IMO.

lorisleiva (approved):
Sorry for the late review, looks good to me! 👌

mcintyre94:
### Merge activity

* **May 7, 2:56 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1555).
* **May 7, 3:34 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1555) rebased this pull request as part of a merge.
* **May 7, 3:43 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1555).

PR #1563, opened 2026-04-24
Add fixed-point arithmetic

lorisleiva:
This PR is part of the stack implementing the fixed-point number types proposed in #1545. It adds the six arithmetic operations for both kinds.

- `add*` / `subtract*` require both operands to share the full shape (signedness, total bits, scale).
- `multiply*` / `divide*` accept a same-signedness fixed-point of any shape or a bare `bigint` as the second operand. The result always takes the first operand's shape. Both default to `'strict'` rounding and accept an optional `RoundingMode`.
- `negate*` only accepts signed values (enforced at the type level and at runtime).
- `absolute*` accepts both signednesses.

```ts
const usd = decimalFixedPoint('unsigned', 64, 2);
const rate = decimalFixedPoint('unsigned', 64, 4);
addDecimalFixedPoint(usd('1.50'), usd('2.25'));      // 3.75
multiplyDecimalFixedPoint(usd('100'), rate('0.0025')); // 0.25
divideDecimalFixedPoint(usd('10.50'), 3n, 'round');   // 3.50
```

All results are range-checked and throw `ARITHMETIC_OVERFLOW` with the operation name in the error context when they don't fit the target shape. Division by zero throws `DIVISION_BY_ZERO`. Shape mismatches throw `SHAPE_MISMATCH`. Inexact operations under `'strict'` throw `STRICT_MODE_PRECISION_LOSS`.

Bigint second operands are especially handy for scalar-by-integer operations like doubling or halving:

```ts
const audioSample = binaryFixedPoint('signed', 16, 15);
multiplyBinaryFixedPoint(audioSample('0.25'), 2n);  // 0.5
divideBinaryFixedPoint(audioSample('0.5'), 2n);     // 0.25
```

Shape-matching is enforced both at compile time (via TypeScript's built-in `NoInfer` utility type, which forces the second operand's shape parameters to match the first operand's exactly) and at runtime (as a defense in case the type safety is bypassed). All returned values are `Object.freeze`-d. The `@solana/fixed-points` README will be written in one pass at the end of this stack.

PR #1562, opened 2026-04-24
Bump TypeScript minimum version to 5.4

lorisleiva:
Bumps the declared TypeScript peer dependency floor from `>=5.0.0` to `>=5.4.0` across all 46 publishable packages. TypeScript 5.4 was released in March 2024 and introduces the `NoInfer` utility type, which Kit starts relying on in `@solana/fixed-points`. Consumers still on TypeScript 5.0–5.3 will see an optional peer dependency warning at install time.

PR #1561, opened 2026-04-24
Add fixed-point factories and guards

lorisleiva:
This PR is part of the stack implementing the fixed-point number types proposed in #1545. It adds the first runtime code on top of the type and error scaffolding from previous PRs.

Adds three curried factory families per kind (outer call validates the shape once, inner call constructs values):

- `decimalFixedPoint` / `binaryFixedPoint` parse decimal strings.
- `rawDecimalFixedPoint` / `rawBinaryFixedPoint` accept a pre-scaled raw bigint.
- `ratioDecimalFixedPoint` / `ratioBinaryFixedPoint` construct from a `numerator / denominator`.

```ts
const usdc = decimalFixedPoint('unsigned', 64, 6);
usdc('42.5');                                        // raw === 42500000n
rawDecimalFixedPoint('unsigned', 64, 6)(42500000n);  // same value
ratioDecimalFixedPoint('unsigned', 64, 6)(1n, 4n);   // 0.25
```

Factories default to `'strict'` rounding and throw when inputs can't be represented exactly. Pass `'floor'`, `'ceil'`, `'trunc'`, or `'round'` to opt into lossy construction. Invalid shape parameters, out-of-range raw values, zero denominators, and malformed strings throw their respective `FIXED_POINTS` error codes. All constructed values are `Object.freeze`-d.

```ts
const q1_15 = binaryFixedPoint('signed', 16, 15);
q1_15('0.5');           // raw === 16384n (exact)
q1_15('0.1');           // throws STRICT_MODE_PRECISION_LOSS
q1_15('0.1', 'round');  // raw === 3277n
```

Also adds `isBinaryFixedPoint` / `assertIsBinaryFixedPoint` and decimal equivalents. Each shape parameter is independently optional; pass `undefined` to leave a field unconstrained. `assertIs*` is the source of truth and `is*` wraps it in `try/catch`.

```ts
if (isBinaryFixedPoint(value, 'signed', 16, 15)) {
    value satisfies BinaryFixedPoint<'signed', 16, 15>;
}
assertIsBinaryFixedPoint(value, 'signed'); // narrows to BinaryFixedPoint<'signed', number, number>
```

The `@solana/fixed-points` README will be written in one pass at the end of this stack, once the full public API is in place.

PR #1560, opened 2026-04-24
Add fixed-point types

lorisleiva:
This PR is part of the stack implementing the fixed-point number types proposed in #1545. It lands the public type surface that every follow-up PR will build on.

Adds the four core types:

- `Signedness` — `'signed' | 'unsigned'`.
- `RoundingMode` — `'floor' | 'ceil' | 'trunc' | 'round' | 'strict'`. The `'round'` mode breaks ties away from zero (e.g. `0.5 → 1`, `-0.5 → -1`, `-1.5 → -2`), which is symmetric around zero and differs from `Math.round`. The `'strict'` mode rejects inputs that would require rounding and throws instead.
- `BinaryFixedPoint<TSignedness, TTotalBits, TFractionalBits>` — a fixed-point value whose scale is a power of 2.
- `DecimalFixedPoint<TSignedness, TTotalBits, TDecimals>` — a fixed-point value whose scale is a power of 10.

PR #1564, opened 2026-04-24
Add fixed-point comparisons

lorisleiva:
This PR adds comparison helpers to `@solana/fixed-points`: `cmp`, `eq`, `lt`, `lte`, `gt`, and `gte` variants for both `BinaryFixedPoint` and `DecimalFixedPoint`. The `cmp*` function returns `-1 | 0 | 1` and does the shape assertion; the other five delegate to it and return a boolean.

Comparisons constrain only `kind` and scale (`fractionalBits` for binary, `decimals` for decimal). `signedness` and `totalBits` are allowed to differ because they are storage concerns that do not affect the mathematical value being compared — `u8(5)` and `i32(5)` are the same number. Mismatches on the constrained dimensions throw `SOLANA_ERROR__FIXED_POINTS__SHAPE_MISMATCH`. Compile-time enforcement of the scale dimension uses `NoInfer` on the second operand, matching the pattern already used by `add`/`subtract`/`multiply`/`divide`.

### DIFF
--- a/packages/fixed-points/src/__tests__/binary-comparisons-test.ts
+++ b/packages/fixed-points/src/__tests__/binary-comparisons-test.ts
@@ -1,0 +1,152 @@
+import { SOLANA_ERROR__FIXED_POINTS__SHAPE_MISMATCH, SolanaError } from '@solana/errors';
+
+import {
+    binaryFixedPoint,
+    cmpBinaryFixedPoint,
+    eqBinaryFixedPoint,
+    gtBinaryFixedPoint,
+    gteBinaryFixedPoint,
+    ltBinaryFixedPoint,
+    lteBinaryFixedPoint,
+    rawBinaryFixedPoint,
+} from '../binary';
+
+describe('cmpBinaryFixedPoint', () => {
+    it('returns -1 when the first operand is smaller', () => {
+        const q1_15 = binaryFixedPoint('signed', 16, 15);
+        expect(cmpBinaryFixedPoint(q1_15('0.25'), q1_15('0.5'))).toBe(-1);
+    });
+
+    it('returns 0 when the two operands are equal', () => {
+        const q1_15 = binaryFixedPoint('signed', 16, 15);
+        expect(cmpBinaryFixedPoint(q1_15('0.5'), q1_15('0.5'))).toBe(0);
+    });
+
+    it('returns 1 when the first operand is greater', () => {
+        const q1_15 = binaryFixedPoint('signed', 16, 15);
+        expect(cmpBinaryFixedPoint(q1_15('0.75'), q1_15('0.5'))).toBe(1);
+    });
+
+    it('compares negative and positive signed values correctly', () => {
+        const q1_15 = binaryFixedPoint('signed', 16, 15);
+        expect(cmpBinaryFixedPoint(q1_15('-0.5'), q1_15('0.5'))).toBe(-1);
+        expect(cmpBinaryFixedPoint(q1_15('-0.5'), q1_15('-0.5'))).toBe(0);
+        expect(cmpBinaryFixedPoint(q1_15('-0.25'), q1_15('-0.5'))).toBe(1);
+    });
+
+    it('treats zero as equal to zero', () => {
+        const q1_15 = binaryFixedPoint('signed', 16, 15);
+        expect(cmpBinaryFixedPoint(q1_15('0'), q1_15('0'))).toBe(0);
+    });
+
+    it('allows operands with different signedness', () => {
+        const signed = rawBinaryFixedPoint('signed', 16, 15);
+        const unsigned = rawBinaryFixedPoint('unsigned', 16, 15);
+        expect(cmpBinaryFixedPoint(signed(-1n), unsigned(1n))).toBe(-1);
+        expect(cmpBinaryFixedPoint(signed(1n), unsigned(1n))).toBe(0);
+        expect(cmpBinaryFixedPoint(signed(2n), unsigned(1n))).toBe(1);
+    });
+
+    it('allows operands with different totalBits', () => {
+        const small = rawBinaryFixedPoint('unsigned', 8, 4);
+        const large = rawBinaryFixedPoint('unsigned', 32, 4);
+        expect(cmpBinaryFixedPoint(small(16n), large(16n))).toBe(0);
+        expect(cmpBinaryFixedPoint(small(16n), large(32n))).toBe(-1);
+    });
+
+    it('allows operands with different signedness and totalBits combined', () => {
+        const signed8 = rawBinaryFixedPoint('signed', 8, 0);
+        const unsigned32 = rawBinaryFixedPoint('unsigned', 32, 0);
+        expect(cmpBinaryFixedPoint(signed8(5n), unsigned32(5n))).toBe(0);
+    });
+
+    it('throws SHAPE_MISMATCH when fractionalBits differ', () => {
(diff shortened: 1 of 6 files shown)


lorisleiva:
* **#1595** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1578** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1578?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1570** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1568** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1568?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1567** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1567?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1566** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1566?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1565** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1565?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1564** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1564?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1564?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1563** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1563?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1562** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1562?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1561** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1561?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1560** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1560?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1559** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1559?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1558** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

mcintyre94:
Sorry I didn't get to review these properly when we were together!

Reviewing this and the arithmetic PR, and the `usdc = decimalFixedPoint('unsigned', 64, 6)` example, I'm wondering if this API might benefit from adding a symbol to the fixed points.

As a motivating example:

```ts
const usdc = decimalFixedPoint('unsigned', 64, 6)
const sol = decimalFixedPoint('unsigned', 64, 9)
```

You can add/compare instances of `usdc`, but not one of `usdc` and one of `sol`.

```ts
const eurc = decimalFixedPoint('unsigned', 64, 6)
```

But you can both add and compare instances of `usdc` and `eurc`, because they're the same shape.

Mathematically this all makes sense, but when we're documenting this as the way to define your `usdc` and `sol` shaped values, I think it might be confusing.

A suggestion would be to add `TSymbol extends string | undefined = undefined` as a type param on `binaryFixedPoint` and `decimalFixedPoint`, and an optional `symbol?: TSymbol` param. Any operator that takes two fixed points would throw (I guess a new symbol mismatch error) if the symbols don't match, in addition to being a compile error. This would be opt in functionality, undefined matches undefined, so if you never use symbols then everything works as in the current implementation.

But now you could write:

```ts
const usdc = decimalFixedPoint('unsigned', 64, 6, 'usdc')
const sol = decimalFixedPoint('unsigned', 64, 9, 'sol')
const eurc = decimalFixedPoint('unsigned', 64, 6, 'eurc')
```

And mixing instances of any two different symbols in add/compare would be an error.

I think this could prevent a class of developer errors which could be difficult to debug, and would reduce confusion around when instances with the same shape should be considered compatible.

WDYT?

lorisleiva:
I did think about symbol support for this API but decided against it for a few reasons.

There are valid use cases for performing operations on fixed-points with different shapes. In fact, if you look at the multiply/divide functions, you'll see that the shapes don't even need to match. That's how you can multiply say a USDC amount with a fee rate in basis points. Similarly, the comparison operations don't request that the whole shape match. It is totally valid IMO to compare two amounts in different units.

The fixed-point types only concern themselves with storage information. We need the signedness, the total bits and the scale (decimals or fractional bits) in order to get the full picture on how to store and validate the amount. Symbols are not part of the minimum viable solution to describe fixed points and therefore I think they are best composed with opaque types instead of added inside the core types.

It also makes formatting awkward because now we have the expectation that the symbol should also be part of the formatted strings when the string is supposed to provide a lossless representation of the value.

That being said, I do see the value in having TypeScript validate against two units that should not match (addition for example). Maybe we should adjust the function signatures so that they reject these cases when you add your own opaque types?

mcintyre94:
I figured you probably would have thought about it! Multiply is definitely a good counter-example to not wanting to mix symbols. And agree it's not necessary as the minimum viable solution.

Agree that a higher-level helper makes sense for this too. I suspect we'd be better off leaving fixed points as-is and creating a higher-level type that combines a fixed point and a symbol. And then some subset of the API with `addChecked` type naming maybe.

Probably worth seeing how fixed points get used and if there's demand for that later though.

packages/fixed-points/src/__tests__/binary-comparisons-test.ts
```diff
@@ -0,0 +1,152 @@
...
+        const q1_15 = binaryFixedPoint('signed', 16, 15);
+        expect(cmpBinaryFixedPoint(q1_15('-0.5'), q1_15('0.5'))).toBe(-1);
+        expect(cmpBinaryFixedPoint(q1_15('-0.5'), q1_15('-0.5'))).toBe(0);
+        expect(cmpBinaryFixedPoint(q1_15('-0.25'), q1_15('-0.5'))).toBe(1);
+    });
+
+    it('treats zero as equal to zero', () => {
+        const q1_15 = binaryFixedPoint('signed', 16, 15);
+        expect(cmpBinaryFixedPoint(q1_15('0'), q1_15('0'))).toBe(0);
+    });
```
mcintyre94, line 40: Worth adding a test for `0` and `-0`?

mcintyre94 (approved):

lorisleiva:
### Merge activity

* **May 5, 8:27 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1564).
* **May 5, 8:48 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1564) couldn't merge this pull request because a downstack PR #1560 failed to merge.
* **May 5, 9:18 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1564).
* **May 5, 9:43 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1564) rebased this pull request as part of a merge.
* **May 5, 9:53 PM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1564).

PR #1566, opened 2026-04-24
Add fixed-point rescaling

lorisleiva:
This PR adds `rescaleBinaryFixedPoint` and `rescaleDecimalFixedPoint` to `@solana/fixed-points`. These helpers change a fixed-point's `totalBits` and scale (`fractionalBits` for binary, `decimals` for decimal) in a single call while preserving signedness, matching the API shape established in the proposal's bridge example.

When the requested shape matches the input shape exactly, the same reference is returned with no new allocation. Scale-up is always exact; scale-down consults the optional `RoundingMode` argument, which defaults to `'strict'` and throws `SOLANA_ERROR__FIXED_POINTS__STRICT_MODE_PRECISION_LOSS` on inexact results. Rescaled values are range-checked against the new `totalBits` and throw `SOLANA_ERROR__FIXED_POINTS__ARITHMETIC_OVERFLOW` (with `operation: 'rescale'`) when they do not fit. Invalid targets reuse the existing `INVALID_TOTAL_BITS`, `INVALID_FRACTIONAL_BITS`, `INVALID_DECIMALS`, and `FRACTIONAL_BITS_EXCEED_TOTAL_BITS` codes. No new error codes were introduced.

PR #1565, opened 2026-04-24
Add fixed-point signedness conversions

lorisleiva:
This PR adds four helpers to `@solana/fixed-points` for converting between signed and unsigned variants at the same `totalBits` and scale: `toSignedBinaryFixedPoint`, `toUnsignedBinaryFixedPoint`, `toSignedDecimalFixedPoint`, and `toUnsignedDecimalFixedPoint`.

Each helper accepts input of either signedness. If the input already has the target signedness, the same reference is returned — no new object is allocated. Otherwise the raw value is checked against the target range and a new frozen value is returned with the updated `signedness` field. Values that do not fit the target range (for example a negative number being converted to unsigned, or an unsigned value above `2 ** (totalBits - 1) - 1` being converted to signed) throw the existing `SOLANA_ERROR__FIXED_POINTS__VALUE_OUT_OF_RANGE` error. No new error codes were introduced.

PR #1559, opened 2026-04-24
Add `FIXED_POINTS` error codes

lorisleiva:
This PR is part of the stack implementing the fixed-point number types proposed in #1545. It lands every error code, context shape, and message that the new package will throw, so later PRs in the stack can just reference them.

Reserves the range `[8090000-8090999]` for the `FIXED_POINTS` domain and allocates 12 codes covering invalid shape parameters (total bits, fractional bits, decimals, and the binary-only "fractional bits exceed total bits" check), out-of-range raw values, malformed raw values, malformed strings, zero-denominator ratios, division by zero, arithmetic overflow (shared by add/subtract/multiply/divide/negate/absolute/rescale/signedness conversion), strict shape mismatches in operations and codecs, and strict-mode precision loss when factories and operations cannot produce an exact result without rounding.

Here is an example message for each new code:

- `SOLANA_ERROR__FIXED_POINTS__INVALID_TOTAL_BITS`:
  ```
  Invalid `totalBits`. Expected a positive integer, got foo.
  ```
- `SOLANA_ERROR__FIXED_POINTS__INVALID_FRACTIONAL_BITS`:
  ```
  Invalid `fractionalBits`. Expected a non-negative integer, got -1.
  ```
- `SOLANA_ERROR__FIXED_POINTS__INVALID_DECIMALS`:
  ```
  Invalid `decimals`. Expected a non-negative integer, got 1.5.
  ```
- `SOLANA_ERROR__FIXED_POINTS__FRACTIONAL_BITS_EXCEED_TOTAL_BITS`:
  ```
  `fractionalBits` (64) must not exceed `totalBits` (16).
  ```
- `SOLANA_ERROR__FIXED_POINTS__VALUE_OUT_OF_RANGE`:
  ```
  Fixed-point value of kind `decimalFixedPoint` is out of range for unsigned 8-bit storage. Expected a raw bigint in [0, 255], got 999.
  ```
- `SOLANA_ERROR__FIXED_POINTS__INVALID_STRING`:
  ```
  Invalid string `abc` for fixed-point value of kind `binaryFixedPoint`.
  ```
- `SOLANA_ERROR__FIXED_POINTS__INVALID_ZERO_DENOMINATOR_RATIO`:
  ```
  Invalid ratio 3/0 for fixed-point value of kind `decimalFixedPoint`. Denominator must be non-zero.
  ```
- `SOLANA_ERROR__FIXED_POINTS__ARITHMETIC_OVERFLOW`:
  ```
[...]

PR #1558, opened 2026-04-24
Scaffold empty `@solana/fixed-points` package

lorisleiva:
This PR is the first in a stack implementing the fixed-point number types proposed in #1545. It lands an empty `@solana/fixed-points` package so the scaffolding can be reviewed in isolation before any real API arrives.

The package is wired into `@solana/codecs` via a re-export, matching the `@solana/options` precedent. `@solana/kit` picks it up transitively with no changes. A changeset covering the whole new package will be added at the end of this stack.

PR #1567, opened 2026-04-24
Add fixed-point string and number conversions

lorisleiva:
This PR adds `binaryFixedPointToString`, `decimalFixedPointToString`, `binaryFixedPointToNumber`, and `decimalFixedPointToNumber` to `@solana/fixed-points`. The string variants accept an optional `FixedPointToStringOptions` bag (`decimals`, `padTrailingZeros`, `rounding`) and throw `SOLANA_ERROR__FIXED_POINTS__STRICT_MODE_PRECISION_LOSS` (with `operation: 'toString'`) when capping requires a lossy rescale under the default `'strict'` mode.

The number variants are lossy escape hatches bounded by IEEE 754's ~53-bit mantissa. `binaryFixedPointToNumber` splits the raw value into integer and fractional parts before coercing, preserving exactness whenever the result magnitude fits `Number.MAX_SAFE_INTEGER` even if the raw bigint does not.

### DIFF
--- a/packages/fixed-points/src/formatting.ts
+++ b/packages/fixed-points/src/formatting.ts
@@ -1,0 +1,81 @@
+import { roundDivision, type RoundingMode } from './rounding';
+
+/**
+ * Options accepted by `binaryFixedPointToString` and
+ * `decimalFixedPointToString` to control the emitted representation.
+ *
+ * - `decimals`: caps the number of fractional digits in the output. When
+ *   this is lower than the value's native precision the raw value is
+ *   rescaled using `rounding` (defaults to `'strict'`, which throws
+ *   `SOLANA_ERROR__FIXED_POINTS__STRICT_MODE_PRECISION_LOSS` on inexact
+ *   results). When higher, the extra precision is zero-padded only if
+ *   `padTrailingZeros` is also set.
+ * - `padTrailingZeros`: emits exactly as many fractional digits as
+ *   requested by `decimals`. When `decimals` is omitted, pads to the
+ *   value's native scale (`decimals` for decimal values,
+ *   `fractionalBits` for binary values — the length of the exact
+ *   base-10 expansion). Defaults to `false`, which trims trailing zeros
+ *   (and drops the decimal point altogether for whole numbers).
+ * - `rounding`: only consulted when `decimals` forces a scale-down.
+ *   Defaults to `'strict'`.
+ */
+export type FixedPointToStringOptions = {
+    decimals?: number;
+    padTrailingZeros?: boolean;
+    rounding?: RoundingMode;
+};
+
+/**
+ * Rescales `raw` from `currentDecimals` decimal digits to `options.decimals`
+ * decimal digits (when set), respecting `options.rounding`. Returns the
+ * raw value to format and the number of fractional digits implied by it.
+ *
+ * @internal
+ */
+export function applyDecimalsOption(
+    kind: 'binaryFixedPoint' | 'decimalFixedPoint',
+    raw: bigint,
+    currentDecimals: number,
+    options: FixedPointToStringOptions | undefined,
+): { decimals: number; raw: bigint } {
+    const targetDecimals = options?.decimals;
+    if (targetDecimals === undefined || targetDecimals === currentDecimals) {
+        return { decimals: currentDecimals, raw };
+    }
+    if (targetDecimals > currentDecimals) {
+        return {
+            decimals: targetDecimals,
+            raw: raw * 10n ** BigInt(targetDecimals - currentDecimals),
+        };
+    }
+    const divisor = 10n ** BigInt(currentDecimals - targetDecimals);
+    const rescaled = roundDivision(kind, 'toString', raw, divisor, options?.rounding ?? 'strict');
+    return { decimals: targetDecimals, raw: rescaled };
+}
+
+/**
+ * Formats a scaled bigint `(raw, decimals)` as a canonical decimal
+ * string. When `padTrailingZeros` is `true`, the output emits exactly
+ * `decimals` fractional digits; otherwise trailing zeros are trimmed and
+ * the decimal point is dropped if the fractional part becomes empty.
+ *
+ * @internal
+ */
+export function formatScaledBigint(raw: bigint, decimals: number, padTrailingZeros: boolean): string {
+    if (decimals === 0) {
+        return raw.toString();
+    }
+    const isNegative = raw < 0n;
+    const absDigits = (isNegative ? -raw : raw).toString();
+    const padded = absDigits.padStart(decimals + 1, '0');
+    const integerPart = padded.slice(0, -decimals);
+    let fractionalPart = padded.slice(-decimals);
+    if (!padTrailingZeros) {
+        fractionalPart = fractionalPart.replace(/0+$/, '');
+    }
+    const sign = isNegative ? '-' : '';
+    if (fractionalPart.length === 0) {
+        return `${sign}${integerPart}`;
+    }
+    return `${sign}${integerPart}.${fractionalPart}`;
+}
--- a/packages/fixed-points/src/__tests__/binary-formatting-test.ts
+++ b/packages/fixed-points/src/__tests__/binary-formatting-test.ts
@@ -1,0 +1,134 @@
+import { SOLANA_ERROR__FIXED_POINTS__STRICT_MODE_PRECISION_LOSS, SolanaError } from '@solana/errors';
+
+import {
+    binaryFixedPoint,
+    binaryFixedPointToNumber,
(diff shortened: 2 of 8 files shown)


lorisleiva:
* **#1595** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1578** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1578?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1570** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1568** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1568?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1567** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1567?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1567?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1566** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1566?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1565** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1565?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1564** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1564?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1563** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1563?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1562** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1562?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1561** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1561?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1560** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1560?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1559** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1559?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1558** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/fixed-points/src/formatting.ts
```diff
@@ -0,0 +1,81 @@
...
+    let fractionalPart = padded.slice(-decimals);
+    if (!padTrailingZeros) {
+        fractionalPart = fractionalPart.replace(/0+$/, '');
+    }
+    const sign = isNegative ? '-' : '';
+    if (fractionalPart.length === 0) {
+        return `${sign}${integerPart}`;
+    }
+    return `${sign}${integerPart}.${fractionalPart}`;
+}
```
mcintyre94, line 81: I think we could use `Intl.NumberFormat` here

```ts
return new Intl.NumberFormat('en-US', {
  minimumFractionDigits: padTrailingZeros ? decimals : 0, 
  maximumFractionDigits: decimals, 
  useGrouping: false
}).format(`${raw}E-${decimals}`)
```

should be equivalent to this I think.

Edit: Confirmed this passes all tests 

I also wonder if we should expose a function to get from a fixed point to `{raw, decimals}` (ie everything before the call to this function) so that apps can add their own `Intl.NumberFormat` call with custom locale and formatting. Could be added later though.

mcintyre94 (approved):

packages/fixed-points/src/formatting.ts
lorisleiva: That's not a bad idea but we would need a cast because `NumberFormat` technically only works on `number` types.

```ts
interface NumberFormat {
    format(value: number): string;
    resolvedOptions(): ResolvedNumberFormatOptions;
}
```

I went down a small rabbit hole on this. I was initially worried that providing a string as argument would internally transform it to a `number` and therefore lose precision when reaching the boundaries of JavaScript numbers. Turns out since ES2023, the spec actually does say string arguments are parsed as a `StringNumericLiteral` and stored as an exact "mathematical value", so precision is preserved in fully-compliant environments. That said, I still think an explicit homemade solution might be worth keeping mainly due to inconsistencies between environments. For instance, (according to Claude) Hermes / React Native added string-arg support much later and therefore we may run into some issues when running in older environments. I also think this custom solution should be faster (not that it matters a lot) since it doesn't need to account for all the formatting logic that `NumberFormat` provides.

This is why I thought of offering lossy `toNumber` helpers so apps can explicitly rely on `NumberFormat` when they want locale-aware output and have accepted the precision tradeoff. And as you suggested, they can even use strings now if they are in the correct environments (although they will need a type-cast because TypeScript hasn't caught up with the new API yet).


Regarding the function that outputs `{ raw, decimals }`, am I right to say this would be for binary fixed-point only (since decimal fixed-points are already described as such) and it would wrap the following conversion?

```ts
// Convert the base-2 representation to an exact base-10 representation:
// raw / 2 ** F === (raw * 5 ** F) / 10 ** F, which terminates cleanly.
// The transformed raw carries exactly F decimal digits of precision.
const base10Decimals = value.fractionalBits;
const base10Raw = base10Decimals === 0 ? value.raw : value.raw * 5n ** BigInt(base10Decimals);
```

If so, I'm happy to add a `binaryFixedPointToBase10` function (name TBD) that returns `{ raw, decimals }`. I initially wanted to suggest a `decimalFixedPointToBinaryFixedPoint`/`binaryFixedPointToDecimalFixedPoint` duo but turns out this is a bit more complicated than that because you will likely need to adjust the `totalBits`/`fractionalBits` to keep the range of values intact (or throw arithmetic overflows).

Let me know what you think and I can push a new PR on top of that stack.

mcintyre94: Yea it's weirdly missing from Typescript still, but I learnt about it from Steve's issue about adding math way back: https://github.com/anza-xyz/kit/issues/29. I've used it a bunch since and find it really nice.

I definitely think `Intl.NumberFormat` with string scientific notation is the most flexible API for display rather than going through number there, but having `toNumber` conversions is still worth it for other use cases where you need a number IMO.

Agree the custom solution should be faster, and if `Intl.NumberFormat` doesn't have the support yet then I'm good to keep it as-is.

On the function for `{raw, decimals}`, the only other thing the base10 one would give us is the rounding - but `NumberFormat` has a lot of rounding modes and I think it's a superset of ours. So probably not worth adding anything there. I think just adding binary -> base10 like you said is the way to go there. Then anyone who wants to use `Intl.NumberFormat` with either decimal or binary fixed points can easily do so, with whatever locale and formatting suit them.

lorisleiva: That's a really good shout, I had forgotten about this part of the original issue. How about on top of the `toNumber` and `toString` helpers, I provide the following helpers in a subsequent PR in the stack:

- `formatDecimalFixedPoint(formatter: Intl.NumberFormat, value: DecimalFixedPoint): string`
- `formatBinaryFixedPoint(formatter: Intl.NumberFormat, value: BinaryFixedPoint): string`
- `binaryFixedPointToBase10(value: BinaryFixedPoint): { raw: bigint, decimals: number }` (used internally by the binary fixed-point helpers)

That way, people can mostly rely on the `format` helpers when they need to display values but the `toString` helpers stay lossless and environment-safe. Wdyt?

mcintyre94: That sounds perfect!

lorisleiva:
### Merge activity

* **May 5, 8:27 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1567).
* **May 5, 8:48 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1567) couldn't merge this pull request because a downstack PR #1560 failed to merge.
* **May 5, 9:18 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1567).
* **May 5, 10:18 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1567) rebased this pull request as part of a merge.
* **May 5, 10:27 PM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1567).

PR #1569, opened 2026-04-24
Fix `extendClient` to allow plugins to override existing keys

lorisleiva:
Fixes a `TypeError: Cannot redefine property` thrown when two plugins set the same key on a client. For example:

```ts
createClient()
    .use(c => extendClient(c, { payer: signerA }))
    .use(c => extendClient(c, { payer: signerB })); // throws TypeError: Cannot redefine property: payer
```

The root cause: `createClient` and `extendClient` freeze their output, which marks every own property as `configurable: false`. When a later `extendClient` call copied those descriptors onto its intermediate object, the subsequent `Object.defineProperties(result, additions)` could not redefine any conflicting key.

The fix normalizes the source descriptors to `configurable: true` before copying, so `additions` can always win on conflicts — matching `extendClient`'s documented behavior. The returned client is still frozen, so external immutability is unchanged.

Added two regression tests covering single-key and multi-key override scenarios.

### DIFF
--- a/.changeset/clever-gifts-find.md
+++ b/.changeset/clever-gifts-find.md
@@ -1,0 +1,5 @@
+---
+'@solana/plugin-core': patch
+---
+
+Fix `extendClient` so that a later plugin can override a key previously set by an earlier plugin. Previously, chaining two plugins that set the same key threw `TypeError: Cannot redefine property` because the frozen client's non-configurable property descriptors were copied verbatim onto the intermediate object.
--- a/packages/plugin-core/src/__tests__/client-test.ts
+++ b/packages/plugin-core/src/__tests__/client-test.ts
@@ -284,6 +284,21 @@ describe('extendClient', () => {
     it('returns a frozen object', () => {
         expect(extendClient({ fruit: 'apple' as const }, { vegetable: 'carrot' as const })).toBeFrozenObject();
     });
+
+    it('allows a later plugin to override a key set by an earlier plugin', () => {
+        const client = createClient()
+            .use(c => extendClient(c, { payer: 'A' as const }))
+            .use(c => extendClient(c, { payer: 'B' as const }));
+        expect(client.payer).toBe('B');
+    });
+
+    it('allows override when the earlier plugin set multiple keys', () => {
+        const client = createClient()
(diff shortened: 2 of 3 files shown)


lorisleiva:
* **#1569** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1569?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
Clean, minimal fix for a real bug. LGTM — leaving as a comment rather than approval since I don't have explicit approval rights on this repo.

The root cause analysis is correct: `Object.freeze` marks descriptors as `configurable: false`, and copying those verbatim onto the intermediate object blocks the subsequent `Object.defineProperties(result, additions)` from redefining any conflicting key. Normalizing to `configurable: true` before the first copy lets `additions` win on conflicts — which matches the documented contract ("When the same key exists on both, `additions` wins"). The final `Object.freeze` still guarantees external immutability, so nothing changes for consumers.

`toConfigurableDescriptors` correctly uses `Reflect.ownKeys` (handles symbol-keyed properties, matching `Object.getOwnPropertyDescriptors` semantics) and preserves all other descriptor attributes via spread. Tests cover both the single-key override and the multi-key case where unrelated keys should be preserved — good coverage for the fix. Changeset is present with the right bump level and package.

- Worth confirming there's no intentional reason the original code left descriptors non-configurable (e.g. to protect against accidental collisions). Given the existing docblock explicitly states `additions` should win on conflicts, I'm confident this is just a bug, not a deliberate guard — but a maintainer signoff on that intent is worthwhile.
- The `addUse` path uses the same `Object.getOwnPropertyDescriptors` + `Object.freeze` pattern but never has conflicting keys to redefine (the `use` key is unique to that intermediate), so it doesn't need the same treatment.
- No public API surface change; behavior change is strictly additive (previously-throwing case now works).

mcintyre94 (approved):
Nice, thankyou!

PR #1576, opened 2026-04-27
feat(errors): add create coded error class

amilz:
#### Problem
Network-related protocols (e.g., Kora or Keychain) and Client generating tools (e.g., Codama) cannot fully leverage the @solana/errors package b/c it relies on storing all errors in this package--that's not viable for scale. Codama addresses this by effectively cloning the error class. 

#### Summary of Changes
- Add `createCodedErrorClass` so downstream packages can define strongly typed coded error classes with a generated constructor, code-narrowing guard, and human-readable message helper.
- Refactor `SolanaError` to use the shared coded-error factory while preserving dev/prod messages, decode hints, `cause` handling, frozen contexts, and instruction-index suffix behavior.
- Harden coded-error guards against same-name foreign errors and align exposed context types with the runtime frozen context.

#### Test Plan
- `pnpm --filter @solana/errors test:typecheck`
- `pnpm --filter @solana/errors test:lint`
- `pnpm --filter @solana/errors test:prettier`
- `pnpm --dir packages/errors exec jest -c ../../node_modules/@solana/test-config/jest-unit.config.node.js --rootDir . --runTestsByPath src/__tests__/coded-error-test.ts src/__tests__/error-test.ts`
- `pnpm --dir packages/errors exec jest -c ../../node_modules/@solana/test-config/jest-unit.config.browser.js --rootDir . --runTestsByPath src/__tests__/coded-error-test.ts src/__tests__/error-test.ts`

#### Concept

```ts
import { createCodedErrorClass } from '@solana/errors';

export const KORA_ERROR__ACCOUNT_NOT_FOUND = -32050 as const;
export const KORA_ERROR__RATE_LIMIT_EXCEEDED = -32030 as const;

type KoraErrorCode = typeof KORA_ERROR__ACCOUNT_NOT_FOUND | typeof KORA_ERROR__RATE_LIMIT_EXCEEDED;
type KoraErrorContext = {
    [KORA_ERROR__ACCOUNT_NOT_FOUND]: { address: string };
    [KORA_ERROR__RATE_LIMIT_EXCEEDED]: undefined;
};

export const { ErrorClass: KoraError, isError: isKoraError } = createCodedErrorClass<KoraErrorCode, KoraErrorContext>({
    messages: {
        [KORA_ERROR__ACCOUNT_NOT_FOUND]: 'Account $address not found',
        [KORA_ERROR__RATE_LIMIT_EXCEEDED]: 'Rate limit exceeded',
    },
    name: 'KoraError',
});

try {
    /* ... */
} catch (e) {
    if (isKoraError(e, KORA_ERROR__ACCOUNT_NOT_FOUND)) {
        displayError(`Missing account ${e.context.address}`);
    }
}
```


packages/errors-core/src/coded-error.ts
```diff
@@ -0,0 +1,308 @@
...
+    class CodedErrorImpl extends Error {
+        readonly context: Readonly<Record<string, unknown> & { __code: TCode }>;
+        constructor(code: TCode, contextAndErrorOptions?: ErrorOptions & Record<string, unknown>) {
+            let context: Record<string, unknown> | undefined;
+            let errorOptions: ErrorOptions | undefined;
+            if (contextAndErrorOptions) {
+                Object.entries(Object.getOwnPropertyDescriptors(contextAndErrorOptions)).forEach(
+                    ([propName, descriptor]) => {
+                        // If the `ErrorOptions` type ever changes, update this code.
+                        if (propName === 'cause') {
```
raushan728, line 263: cause: undefined is stripped from context but still forwarded to ErrorOptions so error.cause exists but is undefined rather than absent. Consumers doing 'cause' in error checks might get surprised.

packages/errors-core/src/coded-error.ts
```diff
@@ -0,0 +1,308 @@
...
+                            errorOptions = { cause: descriptor.value };
+                        } else {
+                            if (context === undefined) {
+                                context = { __code: code };
```
raushan728, line 267: __code feels like an implicit reserved key but it's not documented anywhere. If a consumer accidentally includes __code in their context shape, it would get silently overwritten with no warning.

packages/errors/src/message-formatter.ts
```diff
@@ -16,18 +15,28 @@ type State = Readonly<{
...
+ * `$foo`). Use a backslash to escape a `$` that should not be treated as a variable (e.g. `\$foo`).
+ *
+ * This is the low-level formatter shared by {@link getHumanReadableErrorMessage} (which layers on
+ * the {@link SolanaErrorMessages} lookup and the instruction-error-index suffix) and by
+ * {@link createCodedErrorClass} (which layers on a consumer-provided message map).
+ *
+ * @param messageFormatString The message template containing `$variable` tokens.
+ * @param context             An object whose keys correspond to the variables in the template.
+ */
+export function formatMessageTemplate(messageFormatString: string | undefined, context: object = {}): string {
```
raushan728: formatMessageTemplate is now exported as a public API but it was previously internal. Should this be marked `@internal` in JSDoc if it's only meant for use by createCodedErrorClass and getHumanReadableErrorMessage? Exposing it publicly means any breaking change to it would be semver-major.

packages/errors-core/src/coded-error.ts
```diff
@@ -0,0 +1,308 @@
...
+ * }
+ * ```
+ *
+ * @see {@link SolanaError} — the reference implementation this factory mirrors.
+ */
+export function createCodedErrorClass<TCode extends number, TContextMap extends CodedErrorContextMap<TCode>>(
+    definition: CodedErrorDefinition<TCode>,
+): CodedErrorClassBundle<TCode, TContextMap> {
+    const { messagePostProcessor, messages, name, prodDecodeCommand, prodMessagePrefix } = definition;
+    const prefix = prodMessagePrefix ?? name;
```
raushan728, line 220: prodMessagePrefix falls back to name silently. If someone sets name: 'KoraError' expecting prod messages to also say "KoraError #-32050" that works fine but if they later add prodMessagePrefix without realizing name was the fallback, the behavior change could be confusing. Worth calling out in the JSDoc example.

raushan728:
First time reviewing a PR still a 2nd year student. Happy to be corrected if I've misread anything.

packages/errors/src/message-formatter.ts
raushan728: formatMessageTemplate is marked `@internal` but it's still exported from the package via `index.ts` (`export * from './coded-error`' pulls in everything). `@internal` in JSDoc doesn't prevent consumers from importing it would a barrel-file exclusion or a TypeScript declaration override be needed to actually enforce this?

packages/errors/src/__tests__/coded-error-test.ts
```diff
@@ -0,0 +1,345 @@
...
+        name: 'TestError',
+    });
+}
+
+describe('createCodedErrorClass', () => {
+    let originalDev: boolean | undefined;
+    beforeEach(() => {
+        originalDev = (globalThis as { __DEV__?: boolean }).__DEV__;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (globalThis as any).__DEV__ = true;
```
mcintyre94: Nit: should either do any or `{ __DEV__?: boolean }` each time

packages/errors-core/src/__tests__/coded-error-test.ts
```diff
@@ -0,0 +1,345 @@
...
+            expect(err.name).toBe('TestError');
+            expect(err.constructor.name).toBe('TestError');
+        });
+        it('freezes the context object and sets __code', () => {
+            const { ErrorClass } = makeBundle();
+            const err = new ErrorClass(CODE_WITH_CONTEXT, { count: 3, name: 'world' });
+            expect(err.context).toHaveProperty('__code', CODE_WITH_CONTEXT);
+            expect(err.context).toHaveProperty('name', 'world');
+            expect(err.context).toHaveProperty('count', 3);
+            expect(err.context).toBeFrozenObject();
```
mcintyre94, line 60: I think this should be split up. Maybe 3 tetss: `__code`, properties (name and count), frozen object

packages/errors/src/__tests__/coded-error-test.ts
```diff
@@ -0,0 +1,345 @@
...
+        });
+        it('freezes the context object and sets __code', () => {
+            const { ErrorClass } = makeBundle();
+            const err = new ErrorClass(CODE_WITH_CONTEXT, { count: 3, name: 'world' });
+            expect(err.context).toHaveProperty('__code', CODE_WITH_CONTEXT);
+            expect(err.context).toHaveProperty('name', 'world');
+            expect(err.context).toHaveProperty('count', 3);
+            expect(err.context).toBeFrozenObject();
+        });
+        it('exposes a default empty-ish context for codes without context', () => {
```
mcintyre94: Nit: Something like "context with only __code" would be clearer

packages/errors/src/__tests__/coded-error-test.ts
```diff
@@ -0,0 +1,345 @@
...
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            (globalThis as any).__DEV__ = false;
+            const messagesProxy = new Proxy(
+                {},
+                {
+                    get() {
+                        throw new Error('messages map was accessed in prod');
+                    },
+                },
+            ) as Readonly<Record<TestCode, string>>;
```
mcintyre94: Could this be done with a jest mock and a `.not.toHaveBeenCalled()` assertion?

packages/errors-core/src/__tests__/coded-error-test.ts
```diff
@@ -0,0 +1,345 @@
...
+        });
+        it('preserves accessor (getter) context properties', () => {
+            const { ErrorClass } = makeBundle();
+            let reads = 0;
+            const ctx = {
+                count: 1,
+                get name() {
+                    reads++;
+                    return 'alice';
+                },
```
mcintyre94, line 202: Again I think this could be a mock, then assert it's called

packages/errors/src/__tests__/coded-error-test.ts
```diff
@@ -0,0 +1,345 @@
...
+        });
+        it('strips `cause: undefined` from context but forwards it to ErrorOptions', () => {
+            const { ErrorClass } = makeBundle();
+            const err = new ErrorClass(CODE_WITH_CONTEXT, {
+                cause: undefined,
+                count: 1,
+                name: 'x',
+            } as unknown as TestContext[typeof CODE_WITH_CONTEXT]);
+            expect(err.context).not.toHaveProperty('cause');
+        });
```
mcintyre94: Is this test missing an assert for the "forwards it to ErrorOptions"?

packages/errors/src/__tests__/coded-error-test.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    describe('isError guard additional hardening', () => {
+        it('rejects a same-name error whose context.__code is not a number', () => {
+            const { isError } = makeBundle();
+            const foreign = new Error('bogus');
+            foreign.name = 'TestError';
+            (foreign as unknown as { context: object }).context = { __code: 'not-a-number' };
+            expect(isError(foreign)).toBe(false);
+        });
+    });
```
mcintyre94: I'd just pull this into the `isError` checks block

packages/errors-core/src/__tests__/coded-error-test.ts
```diff
@@ -0,0 +1,345 @@
...
+                prodDecodeCommand: 'npx @test/errors decode --',
+                prodMessagePrefix: 'Test error',
+            });
+            const err = new ErrorClass(CODE_WITHOUT_CONTEXT);
+            expect(err.name).toBe('TestError');
+            expect(err.message).toBe(
+                `Test error #${CODE_WITHOUT_CONTEXT}; Decode this error by running \`npx @test/errors decode -- ${CODE_WITHOUT_CONTEXT}\``,
+            );
+        });
+        it('appends an encoded context when present', () => {
```
mcintyre94, line 327: Missing test: does it also include encoded context when no `prodDecodeCommand` is configured?

packages/errors/src/__tests__/error-test.ts
```diff
@@ -1,11 +1,31 @@
...
+import { formatMessageTemplate } from '../message-formatter';
 
 jest.mock('../message-formatter');
 
 describe('SolanaError', () => {
+    let originalDev: boolean | undefined;
+    beforeEach(() => {
+        originalDev = (globalThis as { __DEV__?: boolean }).__DEV__;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (globalThis as any).__DEV__ = true;
```
mcintyre94: Same as in coded-error-test, either use `any` or not for all of these

packages/errors/src/__tests__/message-formatter-test.ts
```diff
@@ -1,221 +1,161 @@
...
-            expect(message).toBe(
-                "Solana error #123; Decode this error by running `npx @solana/errors decode -- 123 'ENCODED_CONTEXT'`",
-            );
-        });
-        it('renders no encoded context in the decoding advice when the context has no keys', () => {
-            jest.mocked(encodeContextObject).mockReturnValue('ENCODED_CONTEXT');
-            const context = {};
-            const message = getErrorMessage(123 as SolanaErrorCode, context);
-            expect(message).toBe('Solana error #123; Decode this error by running `npx @solana/errors decode -- 123`');
+describe('getHumanReadableErrorMessage', () => {
```
mcintyre94, line 12: For review purposes, is the change here just renaming to `getHumanReadableErrorMessage` and dropping the outer describe prod/dev blocks, or are there meaningful changes to these tests?

mcintyre94 (changes requested):
I think this is generally a good idea - I think our error infrastructure here is really good and it makes sense to make it easy for other projects in the ecosystem to do something similar.

I wonder though whether we should split the infrastructure here into a separate package, maybe `@solana/errors-core` based on some other package naming? Then `@solana/errors` is just Kit errors using these functions.

I think that'd make more space for a focused readme. It would also be useful to explain the rules that we have buried in `codes.ts` about not changing error codes etc if you're publishing a decode helper.

I think we should also include in that readme a minimal version of the `@solana/errors` decode helper (and an explanation of how to add that script to your package), because you really should be publishing one if you're using this infra. Encoded errors with context are quite useless without it.

PR #1600, opened 2026-05-07
ci: get dependabot use github app token

yihau:
#### Problem

looks like the default github token couldn't trigger actions in merge queue as well.

#### Summary of Changes

use github app token to merge


mcintyre94 (approved):
LGTM! I think we'll also need to add `PRIVATE_KEY` to dependabot secrets: https://github.com/anza-xyz/kit/settings/secrets/dependabot (I don't have that key to add it)

yihau:
yeah, iirc dependabot for some reason doesn't read secrets from the vault. let't just merge this for now and see how it goes!

PR #1606, opened 2026-05-07
Add framework-agnostic source types for reactive bindings

mcintyre94:
#### Summary of Changes

This PR adds some types that are useful for reactive UI consumers:

- `ReactiveStreamSource<T>` — anything with a `reactiveStore({ abortSignal })` method that returns a `ReactiveStreamStore<T>`. `PendingRpcSubscriptionsRequest<T>` satisfies this by design.
- `ReactiveActionSource<T>` — anything with a zero-argument `reactiveStore()` method that returns a `ReactiveActionStore<[], T>`. `PendingRpcRequest<T>` satisfies this by design.

These are intended to allow plugins to expose their own functionality to reactive UI acting on actions/streams, without having to fit the precise `PendingRpc[Subscriptions]Request` types. React UI hooks will use these types as input.

`CreateReactiveStoreWithInitialValueAndSlotTrackingConfig` is now exported from `@solana/kit`, was private. This type is useful for plugins to return, eg `kit-plugin-rpc`can provide a function that returns the config for a `useBalance` type subscription in this shape, which can then be used with any reactive UI, without depending on `@solana/react`.

### DIFF
--- a/packages/rpc-subscriptions-spec/src/__typetests__/reactive-stream-source-typetest.ts
+++ b/packages/rpc-subscriptions-spec/src/__typetests__/reactive-stream-source-typetest.ts
@@ -1,0 +1,11 @@
+import type { ReactiveStreamSource } from '@solana/subscribable';
+
+import { PendingRpcSubscriptionsRequest } from '../rpc-subscriptions-request';
+
+// `PendingRpcSubscriptionsRequest<T>` is structurally assignable to `ReactiveStreamSource<T>` —
+// the duck-type reactive-framework bindings consume so they don't have to name a concrete
+// producer type.
+null as unknown as PendingRpcSubscriptionsRequest<number> satisfies ReactiveStreamSource<number>;
+null as unknown as PendingRpcSubscriptionsRequest<{ foo: string }> satisfies ReactiveStreamSource<{
+    foo: string;
+}>;
--- a/packages/subscribable/src/reactive-action-store.ts
+++ b/packages/subscribable/src/reactive-action-store.ts
@@ -48,6 +48,28 @@ export type ReactiveActionStore<TArgs extends readonly unknown[], TResult> = {
     readonly subscribe: (listener: () => void) => () => void;
 };
 
+/**
+ * Duck-type for objects that build a {@link ReactiveActionStore} on demand via a zero-argument
+ * `reactiveStore()` method. Satisfied by `PendingRpcRequest<T>`. The `[]` argument tuple is
+ * intentional — the operation's arguments are already baked into the pending request, so each
+ * `dispatch()` re-fires the same call.
+ *
+ * @typeParam T - The value type resolved by the wrapped operation.
+ *
+ * @example
+ * ```ts
+ * function bind<T>(source: ReactiveActionSource<T>) {
+ *     return source.reactiveStore();
+ * }
+ * ```
+ *
+ * @see {@link ReactiveActionStore}
+ * @see {@link ReactiveStreamSource}
+ */
+export type ReactiveActionSource<T> = {
+    reactiveStore(): ReactiveActionStore<[], T>;
+};
+
 const IDLE_STATE: ReactiveActionState<never> = Object.freeze({
     data: undefined,
     error: undefined,
--- a/packages/subscribable/src/reactive-stream-store.ts
+++ b/packages/subscribable/src/reactive-stream-store.ts
@@ -132,6 +132,28 @@ export type ReactiveStreamStore<T> = {
 export type ReactiveStore<T> = ReactiveStreamStore<T>;
 
 /**
+ * Duck-type for objects that build a {@link ReactiveStreamStore} on demand via a
+ * `reactiveStore({ abortSignal })` method. Satisfied by `PendingRpcSubscriptionsRequest<T>`.
+ * Reactive-framework bindings (e.g. React's `useSubscription`) consume this duck-type so they
+ * don't have to name a concrete producer type.
+ *
+ * @typeParam T - The value type emitted by the resulting stream store.
+ *
+ * @example
+ * ```ts
+ * function bind<T>(source: ReactiveStreamSource<T>, abortSignal: AbortSignal) {
+ *     return source.reactiveStore({ abortSignal });
+ * }
+ * ```
+ *
+ * @see {@link ReactiveStreamStore}
+ * @see {@link ReactiveActionSource}
+ */
+export type ReactiveStreamSource<T> = {
+    reactiveStore(options: { abortSignal: AbortSignal }): ReactiveStreamStore<T>;
+};
+
+/**
  * Returns a {@link ReactiveStreamStore} given a data publisher.
  *
  * The store will update its state with each message published to `dataChannelName` and notify all
--- a/.changeset/open-cameras-sit.md
+++ b/.changeset/open-cameras-sit.md
@@ -1,0 +1,17 @@
+---
+'@solana/subscribable': minor
+'@solana/kit': minor
+---
+
+Add framework-agnostic source duck-types for reactive bindings.
+
+`@solana/subscribable` now exports two new types:
+
+- `ReactiveStreamSource<T>` — anything with a `reactiveStore({ abortSignal })` method that returns a `ReactiveStreamStore<T>`. `PendingRpcSubscriptionsRequest<T>` satisfies this by design.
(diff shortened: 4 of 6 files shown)


mcintyre94:
* **#1606** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1606?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1606?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/subscribable/src/reactive-stream-store.ts
```diff
@@ -131,6 +131,19 @@ export type ReactiveStreamStore<T> = {
...
+ * Duck-type for objects that build a {@link ReactiveStreamStore} on demand via a
+ * `reactiveStore({ abortSignal })` method. Satisfied by `PendingRpcSubscriptionsRequest<T>`.
+ * Reactive-framework bindings (e.g. React's `useSubscription`) consume this duck-type so they
+ * don't have to name a concrete producer type.
+ *
+ * @see {@link ReactiveStreamStore}
+ * @see {@link ReactiveActionSource}
+ */
+export type ReactiveStreamSource<T> = {
+    reactiveStore(options: { abortSignal: AbortSignal }): ReactiveStreamStore<T>;
```
trevor-cortex, line 153: Per the repo's JSDoc convention (see `CLAUDE.md` → "TypeScript Docblocks"): exported generic types should carry `@typeParam` for each type parameter, and an `@example` where helpful. Suggest adding:

```ts
/**
 * ...existing prose...
 *
 * @typeParam T - The notification value type emitted by the produced {@link ReactiveStreamStore}.
 *
 * @example
 * ```ts
 * function bindStream<T>(source: ReactiveStreamSource<T>, abortSignal: AbortSignal) {
 *     const store = source.reactiveStore({ abortSignal });
 *     return store.subscribe(() => { /* ... *\/ });
 * }
 * ```
 */
```

(The promoted `CreateReactiveStoreWithInitialValueAndSlotTrackingConfig` in this same PR is a good model — fully tagged with `@typeParam` × 3.)

packages/subscribable/src/reactive-action-store.ts
```diff
@@ -48,6 +48,19 @@ export type ReactiveActionStore<TArgs extends readonly unknown[], TResult> = {
...
+/**
+ * Duck-type for objects that build a {@link ReactiveActionStore} on demand via a zero-argument
+ * `reactiveStore()` method. Satisfied by `PendingRpcRequest<T>`. The `[]` argument tuple is
+ * intentional — the operation's arguments are already baked into the pending request, so each
+ * `dispatch()` re-fires the same call.
+ *
+ * @see {@link ReactiveActionStore}
+ * @see {@link ReactiveStreamSource}
+ */
+export type ReactiveActionSource<T> = {
```
trevor-cortex, line 69: Same JSDoc-convention point as on `ReactiveStreamSource`: please add `@typeParam T - ...` and ideally an `@example` showing how a binding consumes the source (e.g. constructing the store and subscribing). Keeps the new public types consistent with the rest of the package's docs.

packages/rpc-subscriptions-spec/src/__typetests__/reactive-stream-source-typetest.ts
```diff
@@ -0,0 +1,11 @@
+import type { ReactiveStreamSource } from '@solana/subscribable';
+
+import { PendingRpcSubscriptionsRequest } from '../rpc-subscriptions-request';
+
+// `PendingRpcSubscriptionsRequest<T>` is structurally assignable to `ReactiveStreamSource<T>` —
+// the duck-type reactive-framework bindings consume so they don't have to name a concrete
+// producer type.
+null as unknown as PendingRpcSubscriptionsRequest<number> satisfies ReactiveStreamSource<number>;
```
trevor-cortex, line 8: Minor: as a stronger signal of intent, consider also asserting the inverse direction is *not* required — i.e. that a `ReactiveStreamSource<T>` is **not** assignable back to `PendingRpcSubscriptionsRequest<T>`. Something like:

```ts
// @ts-expect-error — the duck-type does not require `subscribe()` / `reactive()`.
null as unknown as ReactiveStreamSource<number> satisfies PendingRpcSubscriptionsRequest<number>;
```

This pins down the "strictly broader" relationship and would catch a future change that accidentally tightens `ReactiveStreamSource` to require methods that only the concrete producer has. Not blocking — the existing assertion already covers the contract the bindings rely on.

trevor-cortex (commented):
## Summary

Adds two structural duck-types in `@solana/subscribable` that decouple reactive-framework bindings from the concrete `PendingRpcRequest` / `PendingRpcSubscriptionsRequest` producer types:

- `ReactiveStreamSource<T>` — anything with `reactiveStore({ abortSignal })` returning a `ReactiveStreamStore<T>`.
- `ReactiveActionSource<T>` — anything with a zero-arg `reactiveStore()` returning a `ReactiveActionStore<[], T>`.

Also promotes `CreateReactiveStoreWithInitialValueAndSlotTrackingConfig` from file-private to a public export so non-React consumers can name the config shape, and adds two compile-only typetests asserting the producer types satisfy the new duck-types.

The changes are purely additive types — no runtime impact, tree-shake-safe. Both producer types already happen to match the duck-types exactly, so the typetests should hold without any signature changes.

## Things to watch out for

- **`ReactiveStreamSource` / `ReactiveActionSource` are not re-exported from `@solana/kit`.** `packages/kit/src/index.ts` doesn't blanket re-export `@solana/subscribable`, and nothing in this PR adds these two types to the kit barrel. Given that the PR description positions them as the duck-types React hooks and plugin authors will consume — and plugin authors typically import from `@solana/kit` — this looks like a gap worth closing in this PR (or worth an explicit decision to require importing from `@solana/subscribable` directly). The `@solana/kit` minor bump in the changeset is still justified by the newly-exported `CreateReactiveStoreWithInitialValueAndSlotTrackingConfig`, so this is about API surface, not the bump.

- **Per-repo JSDoc convention.** `CLAUDE.md` requires `@typeParam` on exported generics and an `@example` when helpful. The new `ReactiveStreamSource<T>` / `ReactiveActionSource<T>` types have prose docblocks but no `@typeParam T` and no `@example`. The promoted `CreateReactiveStoreWithInitialValueAndSlotTrackingConfig` is fully documented and is a good model.

- **Inlined options shape.** `ReactiveStreamSource` inlines `{ abortSignal: AbortSignal }` rather than reusing `RpcSubscribeOptions`. That's the right call here — `@solana/subscribable` mustn't depend on `@solana/rpc-subscriptions-spec` — but it does mean the duck-type and the producer share the shape only structurally. If `RpcSubscribeOptions` ever grows a non-optional field, the duck-type silently won't match anymore. Not an issue today; worth keeping in mind.

- Verify whether the intent is to expose these new types via `@solana/kit` (likely yes, given the PR motivation). A one-line addition to `packages/kit/src/index.ts` re-exporting `@solana/subscribable` (or a targeted `export type { ... }`) would do it.
- The two new typetest files use the existing repo idiom (`null as unknown as X satisfies Y`) and are correctly placed in packages that already depend on `@solana/subscribable` — no `package.json` changes needed.
- No runtime code paths are touched, so existing tests should be sufficient; the compile-time tests are the meaningful coverage here.

packages/subscribable/src/reactive-stream-store.ts
mcintyre94: Updated these

mcintyre94:
> ReactiveStreamSource / ReactiveActionSource are not re-exported from @solana/kit. packages/kit/src/index.ts doesn't blanket re-export @solana/subscribable, and nothing in this PR adds these two types to the kit barrel. Given that the PR description positions them as the duck-types React hooks and plugin authors will consume — and plugin authors typically import from @solana/kit — this looks like a gap worth closing in this PR (or worth an explicit decision to require importing from @solana/subscribable directly). The @solana/kit minor bump in the changeset is still justified by the newly-exported CreateReactiveStoreWithInitialValueAndSlotTrackingConfig, so this is about API surface, not the bump.

This is intentional - we're exporting these from `@solana/subscribable` and not re-exporting that package from Kit.

lorisleiva (approved):
Good shout!

PR #1607, opened 2026-05-07
Add `ClientProvider`, `useClient`, and `useClientCapability`

mcintyre94:
#### Summary of Changes

This PR is the first of a stack of additions to `@solana/react` to provide full react functionality for Kit apps

It adds the `ClientProvider` and two hooks to consume it.

`ClientProvider` allows you to expose any `Client` to the tree. It supports async clients (ie those using an async plugin), by implementing a shim of `React.use`for react <19. The client is owned by the app outside of the provider lifecycle.

`useClient<TClient>` is a hook to get that client anywhere in the tree

`useClientCapability<TClient>` is the same, but with runtime checking for specific capabilities (provided by plugins) and a friendly error message to identify what's missing. This is what plugin hooks should typically use. Eg the wallet plugin would use this to provide functionality that depends on having `client.wallet` installed.


mcintyre94:
* **#1607** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1607?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1607?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/react/src/ClientProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+ *     </ClientProvider>
+ * </Suspense>
+ * ```
+ *
+ * @see {@link useClient}
+ */
+export function ClientProvider({ children, client }: ClientProviderProps): React.ReactElement {
+    const resolved = isPromiseLike(client) ? usePromise(client) : client;
+    return <ClientContext.Provider value={resolved}>{children}</ClientContext.Provider>;
+}
```
trevor-cortex, line 77: Conditional hook call (see review summary). Consider making the call site unconditional — e.g. `usePromise` returns `value` directly when `!isPromiseLike(value)` and the provider always calls `usePromise(client)`. That makes the hook order independent of the prop shape, satisfies `react-hooks/rules-of-hooks`, and removes the unwritten "don't toggle sync↔async on one provider" constraint.

packages/react/src/usePromise.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * React 18 fallback for `React.use(promise)` — suspends by throwing the promise on first render,
+ * returns the resolved value on subsequent renders, and re-throws the rejection if the promise
+ * settles with an error.
+ *
+ * The promise identity must be stable across renders so the cache keyed off it can find the
+ * settled entry on the second render. The {@link ClientProvider}'s consumer-facing contract
+ * documents this — pass a memoised or module-scope promise.
+ */
+function usePromiseShim<T>(promise: PromiseLike<T>): T {
```
trevor-cortex, line 33: Minor: `let entry` could be `const` if you reshape this as `cache.get(promise) ?? (() => { const tracked = trackedPromise(promise); const e = { promise: tracked, status: 'pending' as const }; cache.set(promise, e); return e; })()` — or just leave it; not worth contorting for. Ignore if you prefer the current shape.

packages/react/src/usePromise.ts
```diff
@@ -0,0 +1,55 @@
...
+
+function trackedPromise<T>(promise: PromiseLike<T>): PromiseLike<unknown> {
+    return promise.then(
+        value => {
+            cache.set(promise, { status: 'fulfilled', value });
+        },
+        reason => {
+            cache.set(promise, { reason, status: 'rejected' });
+        },
+    );
```
trevor-cortex: `trackedPromise` returns the *derivative* promise (`promise.then(...)`) rather than the original. That's deliberately fine for Suspense (React just needs something that settles), but worth a one-liner comment so a future reader doesn't "fix" it back to returning `promise` and accidentally cause the throw to happen with a promise whose handlers haven't run yet.

packages/react/src/useClientCapability.ts
```diff
@@ -0,0 +1,75 @@
...
+    hookName,
+    providerHint,
+}: UseClientCapabilityConfig): Client<TClient> {
+    const client = useClient<TClient>();
+    const required = typeof capability === 'string' ? [capability] : capability;
+    const missing = required.filter(name => !(name in client));
+    if (missing.length > 0) {
+        throw new SolanaError(SOLANA_ERROR__REACT__MISSING_CAPABILITY, {
+            capabilities: missing,
+            hookName,
```
trevor-cortex, line 70: Minor: `(name in client)` will return `true` for inherited properties too. In practice every Kit client is built via `Object.defineProperties({}, ...)` so it's effectively own-properties only — but if a future plugin ever returns a class instance with capabilities on the prototype, the check would still pass while a `Object.hasOwn` check wouldn't. Either is defensible; flagging in case the team has a stronger opinion on the intended contract here ("capability declared" vs "capability is an own property").

packages/react/src/__tests__/ClientProvider-test.browser.tsx
```diff
@@ -0,0 +1,96 @@
...
+            });
+            function Probe() {
+                useClient();
+                return <div data-testid="probe">ready</div>;
+            }
+            const { queryByTestId } = render(
+                <Suspense fallback={<div data-testid="fallback">loading</div>}>
+                    <ClientProvider client={clientPromise}>
+                        <Probe />
+                    </ClientProvider>
```
trevor-cortex, line 101: Consider adding a third async case for the rejected path — render a `<ClientProvider client={Promise.reject(new Error('boom'))}>` inside an error boundary and assert the boundary catches. That locks in the `usePromiseShim` rejected branch, which currently has no coverage.

packages/react/README.md
```diff
@@ -13,6 +13,87 @@
...
+    const clientPromise = useMemo(() => createClient().use(someAsyncPlugin()), []);
+    return (
+        <Suspense fallback={<Splash />}>
+            <ClientProvider client={clientPromise}>
+                <Shell />
+            </ClientProvider>
+        </Suspense>
+    );
+}
+```
```
trevor-cortex, line 56: Tiny doc nit: "the provider suspends via the nearest `<Suspense>` boundary" appears here and again three lines down. The duplication reads a little odd in the rendered docs — could trim one of them.

trevor-cortex (commented):
## Summary

First PR in the stack adding the React/Kit-client bindings: a `ClientProvider`, two consumer hooks (`useClient`, `useClientCapability`), and a `usePromise` shim that backports `React.use` to React 18. Two new errors (`SOLANA_ERROR__REACT__MISSING_PROVIDER`, `SOLANA_ERROR__REACT__MISSING_CAPABILITY`) are added in a fresh `[9000000-9000999]` range. The shape is clean: the provider is a thin value channel, the runtime capability check lives in one place, and the typetests pin the generic narrowing behaviour.

Overall this looks good. One real concern around Rules-of-Hooks in `ClientProvider`, plus a few smaller polish items below.

## Things to watch out for

### `ClientProvider` calls `usePromise` conditionally

```tsx
const resolved = isPromiseLike(client) ? usePromise(client) : client;
```

This is a conditional hook call. As long as a given `<ClientProvider>` instance is consistently fed *either* a sync client *or* a promise it works in practice — and the README does ask for a stable reference — but:

- It will trip `react-hooks/rules-of-hooks` (assuming the repo's eslint config has the rule enabled, which it typically does for a `*.tsx` package with a React peer dep).
- It quietly relies on the consumer never flipping between the two shapes for one provider instance. A toggle that swaps between a precomputed sync client and `createClient().use(asyncPlugin())` would crash with a hook-order mismatch — and the README's advice is about reference stability, not sync-vs-async stability.

Two straightforward fixes, either is fine:

1. Always go through the unwrap path and have the helper short-circuit on non-thenables, e.g. `usePromise(client)` where `usePromise` returns `value` directly when not promise-like (no hook calls at all on that branch is also OK because the call site itself is unconditional now).
2. Split into two components — `<SyncClientProvider>` and `<AsyncClientProvider>` — and let `ClientProvider` pick one based on the prop shape. Each child has a single, unconditional hook call.

I'd lean toward (1) for simplicity.

### No coverage for the rejected-promise path

`usePromiseShim` re-throws `entry.reason` on subsequent renders, but there's no test that a rejected client promise surfaces to the nearest error boundary. Worth adding alongside the existing pending/fulfilled cases — both because it's a real path users will hit (failing async plugin) and because it locks in the contract against future refactors of the cache.

### Note for subsequent reviewers

- The PR depends on `@solana/plugin-core` being part of `@solana/react`'s dependency graph. The README examples import `createClient` and `Client*` types from `@solana/kit` — if `kit` doesn't yet re-export from `plugin-core`, those snippets won't run as-is. Worth a sanity check that the umbrella exports land in the same release window as this PR, or that the docs use `@solana/plugin-core` directly until then.
- Two error codes are reserved in a brand-new `[9000000-9000999]` range. Per CLAUDE.md these are stable forever — fine here, just calling it out so the range allocation gets a second pair of eyes.
- `usePromise.ts` is intentionally not re-exported from `index.ts`, keeping it package-internal. Good.

## Smaller polish

Inline below.

packages/react/src/ClientProvider.tsx
mcintyre94: Yep! Updated to inline the `isPromiseLike` into `usePromise` and call that hook unconditionally

packages/react/src/useClientCapability.ts
mcintyre94: Updated to `Object.hasOwn` - matches how `extendClient` uses `Object.getOwnPropertyDescriptors`.

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Re-review — LGTM, all of my prior points have been addressed:

- **Rules-of-Hooks**: `ClientProvider` now calls `usePromise(client)` unconditionally, with the runtime check moved inside `usePromise`. The call site is clean and won't trip the lint rule.
- **Rejected-promise coverage**: new test using `react-error-boundary` exercises the rethrow path in `usePromiseShim`. Nice.
- **`trackedPromise` returns the chained promise**: now documented inline with a clear explanation (retry runs strictly after the cache update).
- **`Object.hasOwn`**: switched over, with `ES2022.Object` added to the tsconfig lib accordingly.
- **README**: the duplicated suspends-via-Suspense sentence is gone.

## One residual subtlety (non-blocking)

`usePromise` itself still has a conditional `React.use` call internally: on the sync branch it returns immediately and no hook is invoked, on the async branch it invokes `React.use` (or the shim). From the caller's perspective (`ClientProvider`) this is a single unconditional function call, so ESLint is satisfied — but if a given `<ClientProvider>` instance ever flipped between a sync client and a promise-shape client across renders, hook order would still mismatch.

In practice this can't really happen given the README's stable-reference contract and the natural lifecycle of a client, so I think this is fine as-is. Worth a one-line comment in `usePromise` noting the assumption ("caller must not flip between sync/async for a given instance") but I wouldn't hold the PR on it.

- The `useClient` outside-provider test (`renderHook` without a wrapper) will log React's noisy uncaught-error warning. Cosmetic — can be silenced with a `console.error` spy if it bothers anyone.
- `usePromiseShim`'s cache is a module-level `WeakMap` keyed by promise identity. Fine for the documented contract, but note that the cache persists for the module lifetime — tests that reuse promise instances across cases would need to be aware. None do today.
- Two error codes still occupy the fresh `[9000000-9000999]` range. Same flag as before — worth a second pair of eyes on the range allocation.

packages/react/src/useClientCapability.ts
```diff
@@ -0,0 +1,75 @@
...
+ * @see {@link ClientProvider}
+ */
+export function useClientCapability<TClient extends object>({
+    capability,
+    hookName,
+    providerHint,
+}: UseClientCapabilityConfig): Client<TClient> {
+    const client = useClient<TClient>();
+    const required = typeof capability === 'string' ? [capability] : capability;
+    const missing = required.filter(name => !Object.hasOwn(client, name));
```
lorisleiva, line 66: I think that check should be enough 99% of the time but just wanted to flag there might be cases where a plugin functionality goes beyond an attribute check (e.g. asserting a specific RPC method is available). In these cases though, people can write their own logic and throw `SOLANA_ERROR__REACT__MISSING_CAPABILITY` directly so I think that's fine.

packages/react/src/usePromise.ts
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+
+type CacheEntry<T> =
+    | Readonly<{ promise: PromiseLike<unknown>; status: 'pending' }>
+    | Readonly<{ reason: unknown; status: 'rejected' }>
+    | Readonly<{ status: 'fulfilled'; value: T }>;
+
+const cache = new WeakMap<PromiseLike<unknown>, CacheEntry<unknown>>();
```
lorisleiva, line 8: This module variable scares me a little because of how `node_modules` can totally allow multiple versions and/or builds of the same package to coexist at runtime. Could you put my mind at ease? 😅

packages/react/src/usePromise.ts
```diff
@@ -0,0 +1,69 @@
...
+
+function trackedPromise<T>(promise: PromiseLike<T>): PromiseLike<unknown> {
```
lorisleiva: I think here you always return `PromiseLike<undefined>` right? Since the `then` function doesn't actually return any value.

lorisleiva (approved):
Love that we can add all this to `@solana/react` directly. 👌

packages/react/src/useClientCapability.ts
mcintyre94: Yep agreed, this is only going as far as checking whether the capability is installed at all. It's mostly to enable things like the wallet plugin to flag on mount if you try to use its hooks on a client without the right plugin. We could expand it to add an assertion function that runs on the capability, which could support some extra cases. Though the proxy on RPC means we can't really runtime check it synchronously either. Agree that a custom hook is the best answer for now.

packages/react/src/usePromise.ts
mcintyre94: Valid concern! A few reasons I think it's ok:

- `usePromise` isn't exported, only used internally by `ClientProvider`. So an external caller can't stitch together calls from different builds
- It should be fine to have different caches for different builds. It's keyed on the promise, and that can only resolve once and to one value. 
- Worst case if there's a cache miss is just one microtask delay calling `.then` on a resolved promise. The cache is important so that each render doesn't suspend, but if each independent build has a cache miss once that'd be invisible

This is also how SWR stores their global state: https://github.com/vercel/swr/blob/main/src/_internal/utils/global-state.ts

To be transparent, the alternative AFAICT is to mutate the user-owned promise, which seems to be how React does it: https://github.com/facebook/react/blob/d5736f098edee62c44f27b053e6e48f5fa443803/packages/react-reconciler/src/ReactFiberThenable.js#L107 The current method feels a bit less invasive/more temporary (it disappears for React 19) but I think we could re-work it to this approach

packages/react/src/usePromise.ts
mcintyre94: Yep agreed, I made it `PromiseLike<void>` which I think best represents what's going on here

mcintyre94:
This PR has been slightly updated to use `StrictMode` for all the react tests. Added here because it's a shared helper used by all react hooks added in the stack.

packages/react/src/usePromise.ts
lorisleiva: Thank you for putting my mind at ease. ☺️

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1607).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1607) couldn't merge this PR because it had merge conflicts.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1607).
* **Jun 26, 1:39 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1607).

PR #1611, opened 2026-05-08
feat(transaction-introspection): add txn introspection package

amilz:
For consideration/discussion.

Adds `@solana/transaction-introspection`, a new package that bridges a getTransaction response and the auto-generated @solana-program/* parseXInstruction clients. Decodes responses encoded as base64, base58, or json; resolves account indices against static and ALT-loaded addresses; normalizes inner instructions from meta.innerInstructions; and exposes walkInstructions for streaming traversal of every instruction (outer and inner). Re-exported from @solana/kit.

Also hoists the inline GetTransactionApi response shapes in @solana/rpc-api into named exported types (Base64/Base58/Json/JsonParsed) so the new package can consume them directly.


packages/transaction-introspection/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+const { compiledMessage, loadedAddresses } = decodeTransactionFromRpcResponse(rpcTx);
+
+const stream = walkInstructions({ compiledMessage, loadedAddresses, meta: rpcTx.meta });
+for (const { trace, instruction } of filterInstructionsForProgram(stream, TOKEN_PROGRAM_ADDRESS)) {
+    if (identifyTokenInstruction(instruction) !== TokenInstruction.SyncNative) continue;
+    console.log(
+        `SyncNative at ${trace.kind === 'outer' ? `outer[${trace.index}]` : `inner[${trace.outerIndex}/${trace.innerIndex}]`}`,
+    );
+}
```
mcintyre94, line 46: Just commenting on the API for now

- This is a great idea, making it easy to parse `getTransaction` using our existing tooling would be useful 
- I think a stream of instructions including inner is a good API
- I'm not sure if it's necessarily worth using a stream, because (TIL recently) there's a cap of 64 instructions including inner ones per transaction
- Looking at your `ResolvedInstruction` it extends `IInstruction`. If `walkInstruction` returned `ResolvedInstruction & { trace }`, ie it also extends `IInstruction` then you wouldn't need the filter function. You should be able to just do `if(isInstructionForProgram(instruction, TOKEN_PROGRAM_ADDRESS) && identifyTokenInstruction(instruction)) !== TokenInstruction.SyncNative` inside the for loop

packages/transaction-introspection/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+`{ instruction: ResolvedInstruction, trace: InstructionTrace }` — one entry yielded by `walkInstructions`.
+
+### `MetaWithInnerInstructions`
+
+A structural type capturing the minimum shape of `getTransaction`'s `meta` field that `getInnerInstructionsFromMeta` needs. Accepting a structural type keeps callers free to pass the full RPC response without coupling to a specific overload.
+
+## Notes
+
+- V1 transaction messages are not currently supported. The helpers throw a `SolanaError` with code `SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_NOT_SUPPORTED` for them. Pin `maxSupportedTransactionVersion: 0` when calling `getTransaction`.
```
mcintyre94: Would it be difficult to support v1? We're about to have v1 on testnet and everything else in Kit should support them, so it'd be a shame to have a new API without support for them. In my testing I think the relevant part of `getTransaction` response is unchanged?

.changeset/crisp-squids-flow.md
```diff
@@ -0,0 +1,6 @@
+---
+'@solana/transaction-introspection': minor
+'@solana/kit': minor
+---
+
+Add `@solana/transaction-introspection`, a new package that bridges a `getTransaction` response and the auto-generated `@solana-program/*` `parseXInstruction` clients. Decodes the transaction (`encoding: 'base64'`, `'base58'`, or `'json'`), resolves account indices against static + ALT-loaded addresses, normalizes inner instructions from `meta.innerInstructions`, and exposes `walkInstructions` to enumerate every instruction (outer and inner) with a `trace` recording its location. Each returned instruction is a `ResolvedInstruction & { trace }` directly usable with `isInstructionForProgram` from `@solana/instructions` and with the auto-generated `identifyXInstruction` / `parseXInstruction` helpers. Supports `legacy`, `v0`, and `v1` compiled transaction messages. Re-exported from `@solana/kit`.
```
mcintyre94: Could you add a code example to this changeset?

packages/transaction-introspection/src/__tests__/decode-rpc-transaction-test.ts
```diff
@@ -0,0 +1,214 @@
...
+        // The wire-decoder path sets `lifetimeToken` from the encoded message.
+        expect(result.compiledMessage.lifetimeToken).toBe('11111111111111111111111111111111');
+    });
+
+    it('returns empty loaded addresses when meta is null', () => {
+        const b64 = buildBase64Tx();
+        const rpcTx = { meta: null, transaction: [b64, 'base64'] } as unknown as Base64GetTransactionResponse<0>;
+        const result = decodeTransactionFromRpcResponse(rpcTx);
+        expect(result.loadedAddresses).toStrictEqual({ readonly: [], writable: [] });
+    });
```
mcintyre94: I think this is testing the same path as the test below (returns empty loaded addresses for a legacy response (meta has no `loadedAddresses` key)'), because AFAICT the transaction tested is version: legacy. The only thing doing v0 here is a type cast

packages/transaction-introspection/src/__tests__/walk-instructions-test.ts
```diff
@@ -0,0 +1,189 @@
...
+                innerInstructions: [
+                    {
+                        index: 1,
+                        instructions: [{ accounts: [0], data: innerData, programIdIndex: 1, stackHeight: 2 }],
+                    },
+                ],
+            },
+        });
+        expect(out).toHaveLength(3);
+        expect(out[2].trace).toStrictEqual({ innerIndex: 0, kind: 'inner', outerIndex: 1, stackHeight: 2 });
```
mcintyre94: This ordering works, but would it be more clear to do inner instructions after their outer in order, ie like explorer displays them? outer1, inner1-1, inner1-2, outer2, outer3, inner3-1, inner3-2, inner3-3, etc.

packages/transaction-introspection/src/__tests__/walk-instructions-test.ts
```diff
@@ -0,0 +1,189 @@
...
+        expect(out[2].programAddress).toBe('program-a');
+    });
+
+    it('returned items are usable directly with isInstructionForProgram', () => {
```
mcintyre94, line 89: Nice! Great to be able to use our normal filter functions on these instructions.

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,273 @@
...
+ * @example
+ * ```ts
+ * const rpcResponse: Base64GetTransactionResponse<0> = await rpc
+ *     .getTransaction(sig, { encoding: 'base64', maxSupportedTransactionVersion: 0 })
+ *     .send();
+ * ```
+ */
+export type Base64GetTransactionResponse<
+    TMaxSupportedTransactionVersion extends TransactionVersion | void = TransactionVersion | void,
+> = GetTransactionApiResponseBase64<TMaxSupportedTransactionVersion>;
```
mcintyre94: Are these just renaming eg `GetTransactionApiResponseBase64` to `Base64GetTransactionResponse`?

Can we just use eg `GetTransactionApiResponseBase64` in this file?

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,273 @@
...
+
+    const header = {
+        numReadonlyNonSignerAccounts: message.header.numReadonlyUnsignedAccounts,
+        numReadonlySignerAccounts: message.header.numReadonlySignedAccounts,
+        numSignerAccounts: message.header.numRequiredSignatures,
+    };
+
+    // The envelope only carries `version` when `maxSupportedTransactionVersion`
+    // was set on the request; otherwise the response is necessarily legacy.
+    const version: 'legacy' | 0 = 'version' in rpcTx ? (rpcTx.version as 0) : 'legacy';
```
mcintyre94: Missing v1 support here?

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,273 @@
...
+                      'addressTableLookups' in message
+                          ? message.addressTableLookups.map(l => ({
+                                lookupTableAddress: l.accountKey,
+                                readonlyIndexes: l.readonlyIndexes,
+                                writableIndexes: l.writableIndexes,
+                            }))
+                          : [],
+                  header,
+                  instructions: compiledInstructions,
+                  lifetimeToken: message.recentBlockhash,
```
mcintyre94: According to the `getTransaction` type, this will be the nonce for a durable nonce transaction - so this is correct. Could you add a comment here so this doesn't trip us up in future?

https://github.com/anza-xyz/kit/blob/93191af2fd088cd1c56cbed65b2ba1acd2a49ff6/packages/rpc-api/src/getTransaction.ts#L74-L79

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,273 @@
...
+
+    // The envelope only carries `version` when `maxSupportedTransactionVersion`
+    // was set on the request; otherwise the response is necessarily legacy.
+    const version: 'legacy' | 0 = 'version' in rpcTx ? (rpcTx.version as 0) : 'legacy';
+
+    // Mirror the wire-decoder path, which produces a message intersected with
+    // `CompiledTransactionMessageWithLifetime`. The recent-blockhash lives in
+    // `lifetimeToken` so consumers see the same field on both encodings.
+    const compiledMessage: CompiledTransactionMessage & CompiledTransactionMessageWithLifetime =
+        version === 0
```
mcintyre94: Looks like this is missing v1 support here - falls back to the `legacy` branch below.

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,273 @@
...
+        | Base64GetTransactionResponse<TMaxSupportedTransactionVersion>
+        | JsonGetTransactionResponse<TMaxSupportedTransactionVersion>,
+): DecodedRpcTransaction {
+    const tx = rpcTx as AnyGetTransactionResponse;
+    if (isBase64Response(tx)) return decodeFromBase64(tx);
+    if (isBase58Response(tx)) return decodeFromBase58(tx);
+    if (isJsonResponse(tx)) return decodeFromJson(tx);
+    throw new SolanaError(SOLANA_ERROR__TRANSACTION__MALFORMED_MESSAGE_BYTES, {
+        messageBytes: new Uint8Array(0),
+    });
```
mcintyre94: I think it would make sense to throw a new error code here. `TRANSACTION_INTROSPECTION__CANNOT_DECODE_JSON_PARSED_TRANSACTION` or something like that

packages/transaction-introspection/src/get-all-addresses.ts
```diff
@@ -0,0 +1,60 @@
...
+ * 3. ALT-loaded readonly accounts.
+ *
+ * For a legacy transaction (no address lookup tables), `loadedAddresses` may
+ * be omitted; only the static accounts are returned.
+ *
+ * Most callers should use {@link getAccountMetasFromCompiledTransactionMessage}
+ * instead — it returns the same ordering with `AccountRole` bits attached,
+ * which is what the auto-generated `parseXInstruction` clients expect. This
+ * helper is exposed for the rare case where you only need the address strings
+ * (e.g. an indexer indexing by account-index without caring about roles).
```
mcintyre94: Based on this comment I think we should drop this as a public export, it sounds like callers could use `getAccountMetasFromCompiledTransactionMessage` and then map to just addresses.

packages/transaction-introspection/src/get-instructions.ts
```diff
@@ -0,0 +1,202 @@
...
+ *     identifyTokenInstruction(ix);
+ * }
+ * ```
+ */
+export type ResolvedInstruction<TProgramAddress extends string = string> = Instruction<
+    TProgramAddress,
+    readonly AccountMeta[]
+> &
+    InstructionWithAccounts<readonly AccountMeta[]> &
+    InstructionWithData<ReadonlyUint8Array>;
```
mcintyre94: How important is it for this code to normalize empty accounts to `InstructionWithAccounts` and empty data to `InstructionWithData`?

I think in general it's more useful that if an instruction doesn't have any data then it doesn't have the `data` field. Then `isInstructionWithData` does what you'd expect.

mcintyre94 (changes requested):
Thanks for updating this! I made a few comments, the most important is that I think there is still a v1 gap in `decodeFromJson`.

I'll also kick off an AI review to see what I missed

mcintyre94:
@trevor-cortex

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,273 @@
...
+    const compiledMessage: CompiledTransactionMessage & CompiledTransactionMessageWithLifetime =
+        version === 0
+            ? ({
+                  addressTableLookups:
+                      'addressTableLookups' in message
+                          ? message.addressTableLookups.map(l => ({
+                                lookupTableAddress: l.accountKey,
+                                readonlyIndexes: l.readonlyIndexes,
+                                writableIndexes: l.writableIndexes,
+                            }))
```
trevor-cortex: **v1 gap (echoing Callum).** This branch collapses everything to either `'legacy'` or `0`. If a `'json'` response ever carries `version: 1` (or anything else), `rpcTx.version as 0` silently lies and we emit a `V0CompiledTransactionMessage`-shaped object tagged `version: 0` from a v1 source. Downstream `normalizeCompiledInstructions` then takes the legacy/v0 branch and reads `ix.accountIndices` / `ix.data` off a structure that doesn't have those fields, so the failure mode is a corrupt `CompiledTransactionMessage` rather than a clean error.

My suggestion would be to make the version check explicit and throw `SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_NOT_SUPPORTED` for anything that isn't `'legacy'` or `0`, matching what `normalizeCompiledInstructions` already does for unknown compiled versions:

```ts
const rawVersion = 'version' in rpcTx ? rpcTx.version : 'legacy';
if (rawVersion !== 'legacy' && rawVersion !== 0) {
    throw new SolanaError(SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_NOT_SUPPORTED, {
        unsupportedVersion: rawVersion as number,
    });
}
const version: 'legacy' | 0 = rawVersion;
```

That also gives JSON-path callers the same error surface as `getInstructionsFromCompiledTransactionMessage`, which is nice for consumer ergonomics — anyone catching that code already handles "version I don't understand".

packages/transaction-introspection/src/get-instructions.ts
```diff
@@ -0,0 +1,202 @@
...
+    const accounts: AccountMeta[] = ix.accountIndices.map(i => {
+        const meta = metas[i];
+        if (!meta) {
+            throw new SolanaError(
+                SOLANA_ERROR__TRANSACTION__FAILED_TO_DECOMPILE_INSTRUCTION_PROGRAM_ADDRESS_NOT_FOUND,
+                { index: i },
+            );
+        }
+        return meta;
+    });
```
trevor-cortex: Reusing `SOLANA_ERROR__TRANSACTION__FAILED_TO_DECOMPILE_INSTRUCTION_PROGRAM_ADDRESS_NOT_FOUND` for an account-index lookup is a semantic mismatch — the error name says "program address", but here we're throwing it for any account index. Same pattern in `get-inner-instructions.ts` for both the inner program-id and the inner account indices.

Going by the conventions in `packages/errors/src/codes.ts` (`_OUT_OF_RANGE` for value-out-of-range, never reuse codes), the cleanest fix is probably to add a sibling code in the `5663xxx` transaction range — something like `SOLANA_ERROR__TRANSACTION__FAILED_TO_DECOMPILE_INSTRUCTION_ACCOUNT_INDEX_OUT_OF_RANGE` — and use it in both `resolveInstruction` and `getInnerInstructionsFromMeta` for the account-index branches. The program-index branches can keep the existing code.

If you'd rather not grow the error registry in this PR, a follow-up is also fine, but please at least leave a `TODO` here and in `get-inner-instructions.ts` so the conflation doesn't quietly become precedent.

.gitignore
```diff
@@ -27,3 +27,5 @@ test-ledger/
 
 # ignore JetBrains IDE
 .idea
+.claude/settings.local.json
+.codex/config.toml
```
trevor-cortex, line 34: These look like personal editor / agent configs (`.claude/settings.local.json`, `.codex/config.toml`) rather than something the project as a whole needs to ignore. Recommend moving them to your global `~/.config/git/ignore` (or `core.excludesFile`) and dropping them from the shared `.gitignore` so the repo isn't accumulating per-contributor tooling state.

If the project wants to standardise on agent configs being tracked-but-personal, that's a wider conversation worth having explicitly rather than slipping it in here.

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,273 @@
...
+                  instructions: compiledInstructions,
+                  lifetimeToken: message.recentBlockhash,
+                  staticAccounts,
+                  version: 0,
+              } satisfies CompiledTransactionMessageWithLifetime & V0CompiledTransactionMessage)
+            : ({
+                  header,
+                  instructions: compiledInstructions,
+                  lifetimeToken: message.recentBlockhash,
+                  staticAccounts,
```
trevor-cortex: Minor consistency nit on the JSON path. The wire-decoder branch always populates `instructions[i].data` (even as an empty `Uint8Array`), but here you only attach `accountIndices` / `data` when the source field is non-empty:

```ts
...(ix.accounts.length ? { accountIndices: [...ix.accounts] } : null),
...(ix.data != null ? { data: base58.encode(ix.data) } : null),
```

`normalizeCompiledInstructions` defaults both to empty when missing, so this is internally consistent today. But it does mean the JSON-derived `CompiledTransactionMessage` differs in shape from the wire-derived one (missing fields vs empty fields) for empty cases, which is the kind of thing that bites consumers who try to e.g. re-serialise it. Suggest always attaching the fields:

```ts
accountIndices: [...ix.accounts],
data: base58.encode(ix.data ?? ''),
programAddressIndex: ix.programIdIndex,
```

Up to you — pure tidiness, not a bug today.

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,273 @@
...
+ * — one of the input types accepted by {@link decodeTransactionFromRpcResponse}.
+ *
+ * @example
+ * ```ts
+ * const rpcResponse: Base64GetTransactionResponse<0> = await rpc
+ *     .getTransaction(sig, { encoding: 'base64', maxSupportedTransactionVersion: 0 })
+ *     .send();
+ * ```
+ */
+export type Base64GetTransactionResponse<
```
trevor-cortex: Heads-up that the default here (`TransactionVersion | void`) is wider than the underlying `GetTransactionApiResponseBase64` default in `packages/rpc-api/src/getTransaction.ts`, which is `void`. Same for the `Base58` and `Json` wrappers below.

Not necessarily wrong — the wider default makes the introspection-package types more permissive at the call site, which is reasonable for a consumer-facing helper. But it does mean a `Base64GetTransactionResponse` from `@solana/rpc-api` and one from `@solana/transaction-introspection` are not the same type when unparameterised. Worth either:

- aligning the default to `void` to match rpc-api, or
- leaving a one-line `@remarks` on each re-exported alias calling out the intentional widening so future readers don't "fix" it.

The overload signatures of `decodeTransactionFromRpcResponse` itself look fine — they constrain `TMaxSupportedTransactionVersion` independently.

trevor-cortex (commented):
## Summary

This PR adds a new `@solana/transaction-introspection` package that closes the gap between a `getTransaction` response and the auto-generated `@solana-program/*` `parseXInstruction` / `identifyXInstruction` clients. It decodes base64/base58/json responses into a `CompiledTransactionMessage` (+ wire `Transaction` for the binary encodings), resolves account indices against static + ALT-loaded addresses with proper signer/writable roles, normalises inner instructions from `meta.innerInstructions`, and exposes `walkInstructions` to enumerate every outer + inner instruction with a `trace` recording its location. Each returned item is itself a `ResolvedInstruction`, so it flows straight into `isInstructionForProgram` and the auto-generated `identify`/`parse` helpers.

It also tidies `packages/rpc-api/src/getTransaction.ts` by hoisting the four inline non-null response shapes (`Base64`, `Base58`, `Json`, `JsonParsed`) into named exported types, so the new package can consume them directly without re-deriving them. The underlying overload shapes look unchanged — purely an extraction.

Nice quality bar overall: thorough JSDoc, a real README with a quickstart and per-symbol docs, type tests for both the `decodeTransactionFromRpcResponse` overload narrowing and `TracedInstruction` interop with `isInstructionForProgram`, and unit tests covering legacy/v0/v1 + the JSON path + the ALT-loaded v0 walk. The `_exhaustiveCheck: never` in `normalizeCompiledInstructions` is a nice touch — a future `CompiledTransactionMessage` variant will fail to typecheck rather than silently misbehaving.

## Key things to address

Echoing Callum's point — the v1 gap in `decodeFromJson` is the headline issue and I think it needs to be addressed before this lands. Details inline. A few smaller things alongside it (error-code semantics, `.gitignore` additions, PR-description / changeset drift). Everything else is healthy.

- The biggest semantic question is **how strict the JSON path should be about transaction versions it does not understand**. Today, anything other than `'legacy'` is forced to `0` (`rpcTx.version as 0`), which silently mis-shapes v1+ responses. Worth deciding before this is published.
- Worth a sanity-check on whether `@solana/rpc-api` should appear in the changeset. It now exports four new named types (`GetTransactionApiResponseBase58/Base64/Json/JsonParsed`). All publishable packages are version-locked via `fixed`, so the bump will propagate regardless, but the changeset entry as-is doesn't list rpc-api as user-facing; if you'd rather the release notes call out the new exports, the changeset should be expanded.
- The `rpc-api` extraction is a structural refactor of a public type. I scanned the diff and the four hoisted aliases look line-for-line equivalent to the previous inline shapes, but worth a second pair of eyes to confirm nothing widens or narrows (especially the `TMaxSupportedTransactionVersion extends void` branches).
- Inner-instructions normalisation in `get-inner-instructions.ts` reuses `SOLANA_ERROR__TRANSACTION__FAILED_TO_DECOMPILE_INSTRUCTION_PROGRAM_ADDRESS_NOT_FOUND` for both program-index *and* account-index out-of-range cases. Same code is reused in `get-instructions.ts` for account indices. See inline.
- The PR description mentions `filterInstructionsForProgram`, but it isn't in `src/index.ts` or the README — the description is stale and the README correctly says you can use `isInstructionForProgram` directly. Worth tidying the description before merge.

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,273 @@
...
+
+    return { compiledMessage, loadedAddresses: getLoadedAddresses(rpcTx.meta) };
+}
+
+function isBase64Response(rpcTx: AnyGetTransactionResponse): rpcTx is Base64GetTransactionResponse {
+    const t = rpcTx.transaction;
+    return Array.isArray(t) && t[1] === 'base64';
+}
+
+function isBase58Response(rpcTx: AnyGetTransactionResponse): rpcTx is Base58GetTransactionResponse {
```
trevor-cortex: The shape sniff is solid for the cases tested, but I'd suggest making the rejection path more informative. Right now an unsupported JSON shape (e.g. `jsonParsed`) lands here:

```ts
throw new SolanaError(SOLANA_ERROR__TRANSACTION__MALFORMED_MESSAGE_BYTES, {
    messageBytes: new Uint8Array(0),
});
```

`MALFORMED_MESSAGE_BYTES` with an empty byte array is a bit of a lie when the input is structurally a JSON object — there are no "message bytes" at all to be malformed. From a consumer-DX standpoint, when the failure is specifically "this looks like `jsonParsed`", it would be much friendlier to either:

- detect `jsonParsed` explicitly (e.g. instructions with `programId` but no `programIdIndex`) and throw a more targeted error or a `SolanaError` with a custom message that names the encoding, or
- at minimum mention in the thrown error's context that the input was JSON-shaped but not `'json'` encoded.

The docblock already calls out that `jsonParsed` isn't supported — propagating that as a specific runtime error would close the loop nicely.

.gitignore
mcintyre94: Note: pulled these into #1737

amilz:
@mcintyre94 @trevor-cortex thx for the feedback
Gone through each comment. Main changes
- added handling for v1
- reordered output to match explorer style (outer-1, inner-1, inner-n, outer-2, inner-1, etc.) 
- leveraged the suggested kit types where suggested
- added new errors
- added/updated some tests per feedback
- updated docs per feedback
- rebased b/c some conflicts w/ error commits

packages/transaction-introspection/src/__tests__/get-instructions-test.ts
```diff
@@ -0,0 +1,208 @@
...
+    it('throws if a program address index is out of range', () => {
+        const broken = {
+            ...compiled,
+            instructions: [{ accountIndices: [0], data: new Uint8Array(), programAddressIndex: 99 }],
+        } as CompiledTransactionMessage;
+        const err = getThrownError(() => getInstructionsFromCompiledTransactionMessage(broken));
+        expect(err).toBeInstanceOf(SolanaError);
+        expect((err as SolanaError).context.__code).toBe(
+            SOLANA_ERROR__TRANSACTION__FAILED_TO_DECOMPILE_INSTRUCTION_PROGRAM_ADDRESS_NOT_FOUND,
+        );
```
mcintyre94: Not sure if I'm missing something, but I think this should be able to be written like:

```
expect(getInstructionsFromCompiledTransactionMessage(broken)).toThrow(new SolanaError(SOLANA_ERROR__TRANSACTION__FAILED_TO_DECOMPILE_INSTRUCTION_PROGRAM_ADDRESS_NOT_FOUND)))`
```

The tests in `get-inner-instructions-test.ts‎` use this and I think it'd work here too 

Not relevant here but FWIW this works with async tests too: `await expect(...).rejects.toThrow`.

(Also applies to the other tests using `getThrownError` here)

mcintyre94 (dismissed):
This looks really good to me now, thanks for making all those changes!

To flag a relevant PR, https://github.com/anza-xyz/kit/pull/1776 adds `getTransactionForAddress` to the Kit RPC. If my PR to add `version` to that API is merged (https://github.com/solana-rpc/superbank/pull/20) then AFAICT this would also be compatible with transactions from `getTransactionsForAddress` which would be super nice.

I'd suggest that we handle that as a follow up though, rather than blocking this until that's added.

Will do another AI review in case I missed anything!

packages/transaction-introspection/src/get-inner-instructions.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (!meta.innerInstructions) return [];
+    const base58 = getBase58Encoder();
+    const result: TracedInstruction[] = [];
+    for (const group of meta.innerInstructions) {
+        for (let innerIndex = 0; innerIndex < group.instructions.length; innerIndex++) {
+            const ix = group.instructions[innerIndex];
+            const programMeta = accountMetas[ix.programIdIndex];
+            if (!programMeta) {
+                throw new SolanaError(
+                    SOLANA_ERROR__TRANSACTION__FAILED_TO_DECOMPILE_INSTRUCTION_PROGRAM_ADDRESS_NOT_FOUND,
```
trevor-cortex, line 67: Nit: the inner `const meta` here shadows the outer `meta: MetaWithInnerInstructions` parameter. Not a bug — the outer one isn't referenced again inside this `.map` callback — but it reads a bit confusingly when you're scanning the file. Renaming the local to `accountMeta` (or pulling the lookup into a small helper) would make the intent slightly clearer.

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,310 @@
...
+ *     encoding: 'base64',
+ *     maxSupportedTransactionVersion: 0,
+ * }).send();
+ * if (!rpcResponse) throw new Error('not found');
+ *
+ * const { compiledMessage, loadedAddresses } = decodeTransactionFromRpcResponse(rpcResponse);
+ * const instructions = getInstructionsFromCompiledTransactionMessage(compiledMessage, loadedAddresses);
+ * ```
+ */
+export function decodeTransactionFromRpcResponse<
```
trevor-cortex, line 286: The docblock enumerates the two `TRANSACTION_INTROSPECTION__*` errors but doesn't mention that `SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_NOT_SUPPORTED` can also bubble up here (from `decodeFromJson` for an unrecognized `rpcTx.version`, and indirectly via the wire decoder for malformed binary inputs). A short `@throws` note covering it would round out the API surface for callers wrapping this in `try/catch`.

packages/transaction-introspection/src/decode-rpc-transaction.ts
```diff
@@ -0,0 +1,310 @@
...
+
+/**
+ * Detects an `encoding: 'json'` response specifically: its message carries
+ * the compiled-message `header` (signer/readonly counts). A `jsonParsed`
+ * message has no `header` — the server has already resolved the roles onto
+ * each of its `accountKeys` — so checking for it distinguishes the two
+ * encodings regardless of how many instructions the transaction has.
+ */
+function isJsonResponse(rpcTx: AnyGetTransactionResponse): rpcTx is GetTransactionApiResponseJson {
+    const message = getJsonShapedMessage(rpcTx);
```
trevor-cortex, line 228: Heuristic question (not a blocker): `isJsonResponse` vs `isJsonParsedResponse` is discriminated purely on `typeof message.header === 'object'`. This is the right call given today's RPC contract — `jsonParsed` strips the header and inlines roles on `accountKeys` — but it does mean any future server-side change that re-introduced a `header` on a parsed response would silently route into the JSON path. If you wanted belt-and-braces, you could also check that `message.instructions[0]?.programIdIndex` is a `number` (the JSON discriminator) rather than relying on `header` alone. Up to you whether it's worth the extra check.

packages/transaction-introspection/src/get-instructions.ts
```diff
@@ -0,0 +1,215 @@
...
+ * helper is also useful for resolving inner instructions.
+ */
+export function getAccountMetasFromCompiledTransactionMessage(
+    compiledMessage: CompiledTransactionMessage,
+    loadedAddresses?: LoadedAddresses | null,
+): AccountMeta[] {
+    const { header, staticAccounts } = compiledMessage;
+    const numWritableSignerAccounts = header.numSignerAccounts - header.numReadonlySignerAccounts;
+    const numWritableNonSignerAccounts =
+        staticAccounts.length - header.numSignerAccounts - header.numReadonlyNonSignerAccounts;
```
trevor-cortex, line 75: Minor robustness: if `staticAccounts.length < header.numSignerAccounts + header.numReadonlyNonSignerAccounts` (malformed input — shouldn't happen with anything the RPC returns), `numWritableNonSignerAccounts` goes negative and the `for (let n = 0; n < negativeNumber; ...)` loop is silently skipped, producing an `AccountMeta[]` shorter than `staticAccounts`. Subsequent index lookups would then throw `FAILED_TO_DECOMPILE_INSTRUCTION_*_INDEX_OUT_OF_RANGE`, which is technically correct but somewhat oblique — the real issue is the message header. Not worth fixing if you're confident only well-formed inputs reach here, just noting it for the record.

trevor-cortex (commented):
Took a fresh pass over the diff against my previous notes. All the substantive points from the last round are addressed and the package looks ready to ship.

## What's been resolved since the last review

- **v1 gap in `decodeFromJson` (the headline)** is closed. `version` is now read from the envelope and dispatched through a `switch` with a `_exhaustiveCheck: never` default that throws `SOLANA_ERROR__TRANSACTION__VERSION_NUMBER_NOT_SUPPORTED`. The legacy / v0 / v1 branches each build the correct compiled-message shape, and there's a dedicated test covering an unknown `version: 99`.
- **Error-code overloading** for account-index lookups is fixed with the new `SOLANA_ERROR__TRANSACTION__FAILED_TO_DECOMPILE_INSTRUCTION_ACCOUNT_INDEX_OUT_OF_RANGE` (5663038), plus a new `TRANSACTION_INTROSPECTION` domain at 5664xxx with `CANNOT_DECODE_JSON_PARSED_TRANSACTION` and `UNRECOGNIZED_GET_TRANSACTION_RESPONSE`. `jsonParsed` now has its own dedicated, message-clear error rather than a generic decode failure.
- **`.gitignore`** is cleaned up — only `.docs/` and `dist/` remain in the package-local file, no personal tooling state.
- **JSON instruction shape consistency** is now explicit: the JSON path's `getCompiledInstructions` omits `accountIndices` / `data` when empty, with a comment noting it matches the wire decoder.
- **Default type-parameter widening** is consistent — every overload defaults to `TransactionVersion | void`.
- **Changeset** now correctly lists `@solana/errors`, `@solana/rpc-api`, `@solana/kit`, and `@solana/transaction-introspection`, and the description matches the README (no stale `filterInstructionsForProgram`).
- **`loaded-addresses.ts`** extracted cleanly as a tiny module — nice.

The new tests are a notable jump in coverage: v0 + v1 JSON paths, unknown version, unrecognized shape, ALT-loaded v0 walking with mixed static/ALT-writable/ALT-readonly indices, and the v1 `INSTRUCTION_HEADERS_PAYLOADS_MISMATCH` case. The two type tests (`decode-rpc-transaction-typetest.ts` and `traced-instruction-typetest.ts`) lock down both the overload narrowing of `transaction` and the `isInstructionForProgram` / `isInstructionWithData` ergonomics for `TracedInstruction`.

## Remaining observations (all minor)

None block merge — these are polish-level. See inline.

- The `isJsonResponse` / `isJsonParsedResponse` discrimination relies on the presence of `message.header`. That's a reasonable heuristic given the RPC contract, but it's worth being aware of: any future change to the `jsonParsed` shape that re-introduced a `header` field on `message` would silently route into the JSON path. The conditional is currently the only thing distinguishing the two encodings structurally — there's no `encoding` discriminator on the response itself.
- Worth double-checking that `getAccountMetasFromCompiledTransactionMessage` produces the same role assignments as the kit `decompileTransactionMessage` helper for v0 messages with ALT-loaded addresses — the order (static → ALT writable → ALT readonly) matches the runtime, and the `walk-instructions-test.ts` ALT case exercises it, but if there's a canonical decoder elsewhere in the kit it would be reassuring to confirm shape parity once.
- Callum already mentioned this in their approval, but flagging here for posterity: the related PR https://github.com/anza-xyz/kit/pull/1611's discussion of follow-up compatibility with `getTransactionsForAddress` (once `version` lands there) is a reasonable follow-up rather than a blocker.

packages/transaction-introspection/src/__tests__/get-instructions-test.ts
amilz: good catch. cleaner... 92c66d6

packages/transaction-introspection/src/get-inner-instructions.ts
amilz: changed in 92c66d6

packages/transaction-introspection/src/decode-rpc-transaction.ts
amilz: added comment in 92c66d6

packages/transaction-introspection/src/get-instructions.ts
amilz: agree not worth changing

packages/transaction-introspection/src/decode-rpc-transaction.ts
amilz: don't think we need this.

amilz:
@mcintyre94 updated that test and a couple of nits from trevor

mcintyre94 (approved):
I think we're good to ship this! Thankyou! 🚀

PR #1612, opened 2026-05-08
Add `useAction` for tracked async actions

mcintyre94:
#### Summary of Changes

This PR adds the `useAction` hook in React. This bridges any async action into the `ReactiveActionStore` state machine:

```tsx
function PostMessageButton({ url, body }: { url: string; body: string }) {
    const { send, isRunning, error } = useAction(async (signal, content: string) => {
        const res = await fetch(url, { body: content, method: 'POST', signal });
        if (!res.ok) throw new Error(`HTTP ${res.status}`);
        return res.json();
    });

    return (
        <button disabled={isRunning} onClick={() => send(body)}>
            {isRunning ? 'Posting…' : error ? 'Retry' : 'Post'}
        </button>
    );
}
```

The function passed to `useAction` doesn't need to be memoised etc, we handle that lifecycle internally with a ref.

The status comes from `ReactiveActionStore` and can be idle, running, success, error. The react hook adds boolean aliases `isRunning` etc. Later requests abort previous ones, and previous data remains available for SWR.

This is available for apps for any async functionality they need, and plugin hooks such as `useSendTransaction` will also be built on top of it.


mcintyre94:
* **#1612** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1612?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1612?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds `useAction` to `@solana/react`: a generic React hook that bridges any async function into a tracked action by wrapping `createReactiveActionStore` from `@solana/subscribable`. The hook is the React-side adapter for the existing framework-agnostic state machine — store is created once, the user's `fn` is held in a layout-effect-synced ref so the latest closure is invoked on every `send(...)` (no `deps` array), state is consumed via `useSyncExternalStore`, and the store is `reset()` on unmount to abort any in-flight call. The shape (`status` + `data` + `error` + derived `is*` booleans + stable `send` / `reset`) matches the supersede-on-second-call / stale-while-revalidate semantics already encoded in the underlying store.

Good, well-scoped PR. Nice docblocks and a thorough test suite covering all four state transitions, supersede semantics, SWR data preservation, fresh-closure capture, and reference stability of `send`/`reset`. Changeset is present and well-written; the `tsconfig.json` `ESNext.Promise` → `ES2024.Promise` narrowing is appropriate (Promise.withResolvers landed in ES2024 and the tests rely on it). README placement under the foundational-hooks section before `## Hooks` (wallet hooks) reads consistently with the existing layout.

## Things to watch / discussion points

Nothing blocking — the points below are suggestions worth a glance:

1. **Store creation via `useMemo([])`** — React technically reserves the right to discard `useMemo` caches; the canonical "create once, never re-create" pattern for store-like objects is `const [store] = useState(() => createReactiveActionStore(...))` (or `useRef` with lazy init). In practice this doesn't bite today, but `useState` makes the lifetime guarantee explicit and is what `zustand`, `jotai`, etc. use for the same situation. Inline comment with more detail.

2. **`useLayoutEffect` and SSR** — `useLayoutEffect` emits the well-known no-op-on-server warning under React SSR. This package builds a Node entrypoint and `useSyncExternalStore` itself is SSR-safe, so consumers may render this in an SSR shell. The common workaround is `const useIsoLayoutEffect = typeof window !== 'undefined' ? useLayoutEffect : useEffect`. Worth confirming whether `@solana/react` considers SSR a target.

3. **Ref-staleness during the layout-effect window** — the latest-`fn` ref is assigned in `useLayoutEffect`, which means a `send(...)` invoked from a *child* component's layout effect (which runs before the parent's) could still see the previous `fn`. Vanishingly unlikely in practice, and TanStack/wagmi have the same edge — flagging only for awareness.

## For subsequent reviewers

- Verify the `useLayoutEffect` choice against whatever SSR posture `@solana/react` targets — if SSR is supported, swap for the `typeof window` guarded variant.
- Sanity-check the assumption baked into the silenced `react-hooks/refs`: it relies on `createReactiveActionStore` only invoking its callback inside `dispatch`/`dispatchAsync`, never during construction or `getState`. That matches the current implementation in `packages/subscribable/src/reactive-action-store.ts`, so the rule suppression is sound — but it's an invariant worth keeping in mind if that store ever gains eager behavior.
- The `keeps stable send / reset references` test passes trivially today because both come straight off the store, which is created with `[]` deps. If point (1) above changes the construction strategy, that test still holds.

packages/react/src/useAction.ts
```diff
@@ -0,0 +1,125 @@
...
+    const store = useMemo(
+        () =>
+            // eslint-disable-next-line react-hooks/refs
+            createReactiveActionStore<TArgs, TResult>((signal, ...args) => fnRef.current(signal, ...args)),
+        [],
+    );
+
+    // Reset on unmount so any in-flight call is aborted and state is dropped.
+    useEffect(() => () => store.reset(), [store]);
+
```
trevor-cortex, line 124: Minor: prefer `useState`-with-lazy-init over `useMemo` for store-like objects whose identity must be stable for the lifetime of the component:

```ts
const [store] = useState(() =>
    // eslint-disable-next-line react-hooks/refs
    createReactiveActionStore<TArgs, TResult>((signal, ...args) => fnRef.current(signal, ...args)),
);
```

`useMemo` is documented as best-effort cache — React reserves the right to throw the memoized value away and recompute on a later render. In practice this doesn't currently happen for `useMemo([])`, but `useState` makes the "created exactly once, never replaced" guarantee explicit and matches what most external-store libraries do (`zustand`, `jotai`, `valtio` all use `useState` lazy init or a ref).

If the store were ever replaced mid-mount, the unmount-cleanup `useEffect(() => () => store.reset(), [store])` would also fire on the discarded store, leaving the new one without its abort-on-unmount safety net until the next render commits.

Not a correctness bug today, just a more defensive primitive choice.

packages/react/src/useAction.ts
```diff
@@ -0,0 +1,125 @@
...
+    useLayoutEffect(() => {
+        fnRef.current = fn;
+    });
+
+    // `createReactiveActionStore` only reads the callback when the returned `send` is called,
+    // not during render. The `react-hooks/refs` rule doesn't know that, so we silence it here.
+    const store = useMemo(
+        () =>
+            // eslint-disable-next-line react-hooks/refs
+            createReactiveActionStore<TArgs, TResult>((signal, ...args) => fnRef.current(signal, ...args)),
```
trevor-cortex: If `@solana/react` is meant to be SSR-friendly, `useLayoutEffect` will emit the standard "useLayoutEffect does nothing on the server" warning during SSR rendering. The usual fix is a one-liner isomorphic helper:

```ts
const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffect : useEffect;
```

Functionally equivalent here — the ref only needs to be fresh by the time an event handler fires, which can't happen on the server anyway. Skip if SSR isn't a supported target for this package.

packages/react/src/__typetests__/useAction-typetest.ts
```diff
@@ -0,0 +1,43 @@
...
+
+// [DESCRIBE] useAction
+{
+    // It infers TArgs and TResult from the wrapped function
+    {
+        const result = useAction(async (_signal: AbortSignal, value: number) => `n=${value}`);
+        result satisfies ActionResult<[value: number], string>;
+        result.send(7) satisfies Promise<string>;
+        result.data satisfies string | undefined;
+        result.error satisfies unknown;
```
trevor-cortex: Optional: `result.error satisfies unknown` passes trivially because `error` is already typed `unknown` — anything `satisfies unknown`. If you want this line to actually pull weight, consider asserting the inverse with `@ts-expect-error`, e.g.:

```ts
// @ts-expect-error - error is unknown, not Error
result.error satisfies Error;
```

Non-blocking nit.

mcintyre94:
@trevor-cortex

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Re-review

_Would approve if I had repo write access — GitHub blocks `APPROVE` reviews from accounts without explicit access, so leaving this as a comment._

All three points from the prior review are addressed:

1. **Store lifetime** — `useMemo([])` → `useState(() => createReactiveActionStore(...))`. Lifetime is now an explicit React guarantee rather than a best-effort cache.
2. **SSR safety** — `useIsomorphicLayoutEffect` shim added with a clear inline comment explaining why falling back to `useEffect` on the server is functionally equivalent (no event handlers can fire during SSR, so the ref-sync timing doesn't matter there).
3. **Typetest** — the trivial `result.error satisfies unknown` was dropped; the remaining assertions (`ActionResult<...>` shape, `send(7) satisfies Promise<string>`, `data satisfies string | undefined`) are all meaningful.

The `eslint-disable react-hooks/refs` suppression on the `useState` initializer is correct: `fnRef.current` is only read inside the `dispatchAsync` body in `createReactiveActionStore`, never at construction or during `getState`, so the rule's concern doesn't apply. The accompanying comment captures this invariant well.

LGTM from my side. 👍

packages/react/src/useAction.ts
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Trigger the action. Resolves with the wrapped function's result, or rejects with the thrown
+     * error. Calling `send` again while a prior call is in flight aborts the first and rejects its
+     * promise with an `AbortError`. Stable reference.
+     *
+     * Fire-and-forget callers can ignore the returned promise and render from `status` / `data` /
+     * `error`. Awaiters that read the resolved value (e.g. to navigate on success) should filter
+     * supersede rejections with `isAbortError` from `@solana/promises`.
+     */
+    send: (...args: TArgs) => Promise<TResult>;
```
lorisleiva: Should this not be `dispatch` to match the `ReactiveActionStore` API?

```ts
export type ReactiveActionStore<TArgs extends readonly unknown[], TResult> = {
    readonly dispatch: (...args: TArgs) => void;
    readonly dispatchAsync: (...args: TArgs) => Promise<TResult>;
    readonly getState: () => ReactiveActionState<TResult>;
    readonly reset: () => void;
    readonly subscribe: (listener: () => void) => () => void;
};
```

I feel like introducing a new dispatch term here could be more confusing than helping.

lorisleiva (commented):
Looks good! Just one important comment IMO to avoid introducing two terms for the same concept.

packages/react/src/useAction.ts
mcintyre94: Changed to `dispatch`, I think this is mostly better. The one thing I'd flag is that it's actually mirroring `dispatchAsync` (ie you can await the result if you need to), but I think that's unlikely to be confusing in practice. I don't think there's value in distinguishing `dispatch` and `dispatchAsync` in the React layer, and I think using `dispatchAsync` would be confusing since it's mostly going to be fire-and-forget.

lorisleiva: Agreed! Appreciate the heads up though. 🙏

lorisleiva (approved):

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1612).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1612) couldn't merge this pull request because a downstack PR #1607 failed to merge.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1612).
* **Jun 26, 1:41 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1612) rebased this pull request as part of a merge.
* **Jun 26, 1:51 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1612).

PR #1614, opened 2026-05-11
Add `withSignal` to reactive action store for per-dispatch cancellation

mcintyre94:
#### Summary of Changes

Previously I didn't add an abort signal to the `ReactiveActionStore`, but I've realised that as this can be used with eg`PendingRpcRequest` it is useful for consumers to be able to control the abort logic - eg to pass `AbortSignal.timeout(N)`.

I think it makes sense for this signal to be per dispatch, ie not a single abort signal passed into the store on create.

This PR adds an optional API: `store.withSignal(signal).dispatch(...)` to call `dispatch` or `dispatchAsync` with an additional abort signal. It's combined with the store controller signal using `AbortSignal.any`. 

`store.dispatch(...)` and `store.dispatchAsync(...)` remain unchanged. 

The same per-request logic also applies for the `ReactiveStreamStore`, this will be addressed in a later PR. 

Note that we use `AbortSignal.any` for combining signals. This required Claude to polyfill it for jsdom because Jest, but [it's widely supported and is in Baseline 2024](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static). Much better than the signal juggling we have in the stream store (which will be refactored in the upstack PR too)

### DIFF
--- a/packages/subscribable/src/__tests__/reactive-action-store-test.ts
+++ b/packages/subscribable/src/__tests__/reactive-action-store-test.ts
@@ -174,6 +174,124 @@ describe('createReactiveActionStore', () => {
         });
     });
 
+    describe('withSignal()', () => {
+        it('forwards a non-aborted internal signal to `fn` when called via the bare `dispatch`', async () => {
+            expect.assertions(1);
+            const fn = jest.fn((_signal: AbortSignal) => Promise.resolve('ok'));
+            const store = createReactiveActionStore(fn);
+            await store.dispatchAsync();
+            expect(fn.mock.calls[0][0].aborted).toBe(false);
+        });
+
+        it('aborts the in-flight dispatch and transitions to `error` when the caller-provided signal fires', async () => {
+            expect.assertions(2);
+            const ctrl = new AbortController();
+            const reason = new Error('per-request abort');
+            const store = createReactiveActionStore(() => new Promise<string>(() => {}));
+            const dispatched = store.withSignal(ctrl.signal).dispatchAsync();
+            ctrl.abort(reason);
+            await expect(dispatched).rejects.toBe(reason);
+            expect(store.getState()).toStrictEqual({
+                data: undefined,
+                error: reason,
+                status: 'error',
+            });
+        });
+
+        it('preserves stale data when the caller-provided signal aborts after a prior success', async () => {
+            expect.assertions(1);
+            const ctrl = new AbortController();
+            const reason = new Error('abort');
+            const { promise: second } = Promise.withResolvers<string>();
+            const results = [Promise.resolve('first'), second];
+            const store = createReactiveActionStore(() => results.shift()!);
+            await store.dispatchAsync();
+            const dispatched = store.withSignal(ctrl.signal).dispatchAsync();
+            ctrl.abort(reason);
+            await dispatched.catch(() => {});
+            expect(store.getState()).toStrictEqual({
+                data: 'first',
+                error: reason,
+                status: 'error',
+            });
+        });
+
+        it('lets later dispatches recover when a prior call started with an already-aborted signal', async () => {
+            expect.assertions(2);
+            const store = createReactiveActionStore(() => Promise.resolve('ok'));
+            await store
+                .withSignal(AbortSignal.abort(new Error('first')))
+                .dispatchAsync()
+                .catch(() => {});
+            expect(store.getState().status).toBe('error');
+            await store.dispatchAsync();
+            expect(store.getState()).toStrictEqual({
+                data: 'ok',
+                error: undefined,
+                status: 'success',
+            });
+        });
+
+        it('when the caller signal is already aborted, transitions to error without invoking `fn`', async () => {
+            expect.assertions(2);
+            const reason = new Error('pre-aborted');
+            const fn = jest.fn(() => Promise.resolve('ok'));
+            const store = createReactiveActionStore(fn);
+            await store
+                .withSignal(AbortSignal.abort(reason))
+                .dispatchAsync()
+                .catch(() => {});
+            expect(fn).not.toHaveBeenCalled();
+            expect(store.getState()).toStrictEqual({
+                data: undefined,
+                error: reason,
+                status: 'error',
+            });
+        });
+
+        it('passes a combined signal to `fn` that fires when the caller-provided signal aborts', async () => {
+            expect.assertions(1);
+            const ctrl = new AbortController();
+            let captured: AbortSignal | undefined;
+            const store = createReactiveActionStore((signal: AbortSignal) => {
+                captured = signal;
+                return new Promise<string>(() => {});
+            });
+            store.withSignal(ctrl.signal).dispatch();
+            ctrl.abort(new Error('boom'));
+            await Promise.resolve();
+            expect(captured?.aborted).toBe(true);
+        });
+
+        it('lets the caller vary the signal across dispatches on the same store', async () => {
+            expect.assertions(2);
+            const fn = jest.fn((_signal: AbortSignal) => Promise.resolve('ok'));
+            const store = createReactiveActionStore(fn);
+            const ctrlA = new AbortController();
+            const ctrlB = new AbortController();
+            await store.withSignal(ctrlA.signal).dispatchAsync();
+            await store.withSignal(ctrlB.signal).dispatchAsync();
+            // Aborting one controller only fires its own dispatch's composed signal.
+            ctrlA.abort(new Error('only-A'));
+            expect(fn.mock.calls[0][0].aborted).toBe(true);
+            expect(fn.mock.calls[1][0].aborted).toBe(false);
+        });
+
+        it('lets a wrapper be reused as a "kill switch" across dispatches', async () => {
+            expect.assertions(2);
+            const killCtrl = new AbortController();
+            const fn = jest.fn((_signal: AbortSignal) => Promise.resolve('ok'));
+            const store = createReactiveActionStore(fn);
+            const killable = store.withSignal(killCtrl.signal);
+            await killable.dispatchAsync();
+            await killable.dispatchAsync();
+            killCtrl.abort(new Error('killed'));
+            // Both completed dispatches saw a signal that's now aborted.
+            expect(fn.mock.calls[0][0].aborted).toBe(true);
+            expect(fn.mock.calls[1][0].aborted).toBe(true);
+        });
+    });
+
     describe('reset()', () => {
         it('returns the store to idle from a success state', async () => {
             expect.assertions(1);
--- a/packages/subscribable/src/reactive-action-store.ts
+++ b/packages/subscribable/src/reactive-action-store.ts
@@ -31,6 +31,7 @@ export type ReactiveActionStore<TArgs extends readonly unknown[], TResult> = {
      * no state update. Use from UI event handlers; there's no promise to handle or `.catch`.
      *
      * @see {@link ReactiveActionStore.dispatchAsync} when you need the resolved value or propagated errors.
+     * @see {@link ReactiveActionStore.withSignal} to attach a caller-provided `AbortSignal` to a dispatch.
      */
     readonly dispatch: (...args: TArgs) => void;
     /**
@@ -46,20 +47,49 @@ export type ReactiveActionStore<TArgs extends readonly unknown[], TResult> = {
     readonly reset: () => void;
     /** Registers a listener called on every state change. Returns an unsubscribe function. */
     readonly subscribe: (listener: () => void) => () => void;
+    /**
+     * Returns a thin wrapper exposing `dispatch` / `dispatchAsync` that compose `signal` with the
+     * store's internal per-dispatch controller via `AbortSignal.any` — aborting either cancels
+     * the in-flight call. Aborting the caller-provided signal surfaces the abort reason on state
+     * as `{ status: 'error' }`; the internal controller path (supersession by a newer dispatch or
+     * `reset()`) is silent by design so the newer dispatch owns state. Use this to attach a
+     * caller-provided cancellation source (per-attempt timeout, shared kill switch, parent-context
+     * signal) without touching the bare `dispatch` / `dispatchAsync` API.
+     *
+     * - Per-attempt timeout: `store.withSignal(AbortSignal.timeout(5_000)).dispatch(args)` — fresh
+     *   clock per call.
+     * - Permanent kill switch: hold one `AbortController`, bind the wrapper once
+     *   (`const killable = store.withSignal(killCtrl.signal)`), and use `killable.dispatch(...)`
+     *   everywhere; aborting the controller cancels in-flight and short-circuits future calls.
+     *
+     * The wrapper exposes only `dispatch` / `dispatchAsync` — `getState` / `subscribe` / `reset`
+     * remain store-level concerns on the parent.
+     */
+    readonly withSignal: (signal: AbortSignal) => {
+        readonly dispatch: (...args: TArgs) => void;
+        readonly dispatchAsync: (...args: TArgs) => Promise<TResult>;
+    };
 };
 
 /**
- * Duck-type for objects that build a {@link ReactiveActionStore} on demand via a zero-argument
- * `reactiveStore()` method. Satisfied by `PendingRpcRequest<T>`. The `[]` argument tuple is
- * intentional — the operation's arguments are already baked into the pending request, so each
- * `dispatch()` re-fires the same call.
+ * Duck-type for objects that build a {@link ReactiveActionStore} on demand via `reactiveStore()`.
+ * Satisfied by `PendingRpcRequest<T>`. The `[]` argument tuple is intentional — the operation's
+ * arguments are already baked into the pending request, so each `dispatch()` re-fires the same
+ * call.
+ *
+ * The returned store is in the `idle` state — the caller is responsible for calling `dispatch()`
+ * to fire the first attempt. Attach a caller-provided cancellation source per dispatch via
+ * `store.withSignal(signal).dispatch(...)` — see {@link ReactiveActionStore.withSignal}.
  *
  * @typeParam T - The value type resolved by the wrapped operation.
  *
  * @example
  * ```ts
  * function bind<T>(source: ReactiveActionSource<T>) {
- *     return source.reactiveStore();
+ *     const store = source.reactiveStore();
+ *     // Per-attempt timeout, fresh signal per call:
+ *     store.withSignal(AbortSignal.timeout(30_000)).dispatch();
+ *     return store;
  * }
  * ```
  *
@@ -82,13 +112,16 @@ const IDLE_STATE: ReactiveActionState<never> = Object.freeze({
  * so only the most recent dispatch can mutate state.
  *
  * The wrapped function receives the `AbortSignal` as its first argument, followed by whatever
- * arguments were passed to `dispatch`.
+ * arguments were passed to `dispatch`. Callers attach their own cancellation source per-call via
+ * {@link ReactiveActionStore.withSignal} — `store.withSignal(signal).dispatch(...)`. The caller's
+ * signal is composed with the per-dispatch controller via `AbortSignal.any`, so aborting it
+ * cancels the in-flight call and surfaces the abort reason on state.
  *
  * @typeParam TArgs - Argument tuple forwarded from `dispatch` to `fn`.
  * @typeParam TResult - Resolved value type of `fn`.
  * @param fn - Async function to wrap. Receives an {@link AbortSignal} plus the dispatch arguments.
  * @return A {@link ReactiveActionStore} exposing `dispatch`, `dispatchAsync`, `getState`, `subscribe`,
- * and `reset`.
+ * `reset`, and `withSignal`.
  *
  * @example
  * ```ts
@@ -100,7 +133,10 @@ const IDLE_STATE: ReactiveActionState<never> = Object.freeze({
  * store.subscribe(() => console.log(store.getState()));
  * store.dispatch(someAccountId); // fire-and-forget; state is the source of truth
  *
- * // Or, when you need the resolved value imperatively:
+ * // Per-attempt timeout — fresh signal per call:
+ * store.withSignal(AbortSignal.timeout(30_000)).dispatch(someAccountId);
+ *
+ * // Imperative call with the resolved value:
  * const account = await store.dispatchAsync(someAccountId);
  * ```
  *
@@ -121,11 +157,16 @@ export function createReactiveActionStore<TArgs extends readonly unknown[], TRes
         listeners.forEach(listener => listener());
     }
 
-    const dispatchAsync = async (...args: TArgs): Promise<TResult> => {
+    const dispatchAsyncWithSignal = async (userSignal: AbortSignal | undefined, ...args: TArgs): Promise<TResult> => {
         currentController?.abort();
+        // If the caller's signal is already aborted, surface as error and bail.
+        if (userSignal?.aborted) {
+            setState({ data: state.data, error: userSignal.reason, status: 'error' });
+            throw userSignal.reason;
+        }
         const controller = new AbortController();
         currentController = controller;
-        const { signal } = controller;
+        const signal = userSignal ? AbortSignal.any([controller.signal, userSignal]) : controller.signal;
         const previousData = state.data;
         setState({ data: previousData, error: undefined, status: 'running' });
         try {
@@ -136,14 +177,19 @@ export function createReactiveActionStore<TArgs extends readonly unknown[], TRes
             setState({ data: result, error: undefined, status: 'success' });
             return result;
         } catch (error) {
-            if (signal.aborted) {
-                throw signal.reason;
+            // Superseded by a newer dispatch or `reset()` — drop silently so only the most recent
+            // dispatch mutates state, and reject with the abort reason rather than any underlying
+            // failure that happened to race the abort.
+            if (controller.signal.aborted) {
+                throw controller.signal.reason;
             }
+            // Real failure or the caller-provided signal firing — surface as error state.
             setState({ data: previousData, error, status: 'error' });
             throw error;
         }
     };
 
+    const dispatchAsync = (...args: TArgs): Promise<TResult> => dispatchAsyncWithSignal(undefined, ...args);
     const dispatch = (...args: TArgs): void => {
         dispatchAsync(...args).catch(() => {});
     };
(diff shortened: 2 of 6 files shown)


mcintyre94:
* **#1614** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1614?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1618](https://github.com/anza-xyz/kit/pull/1618) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1618?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1624](https://github.com/anza-xyz/kit/pull/1624) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1624?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1614?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/subscribable/src/__tests__/reactive-action-store-test.ts
```diff
@@ -174,6 +174,96 @@ describe('createReactiveActionStore', () => {
...
+            });
+            await store.dispatchAsync().catch(() => {});
+            expect(store.getState().status).toBe('error');
+            await store.dispatchAsync();
+            expect(store.getState()).toStrictEqual({
+                data: 'ok',
+                error: undefined,
+                status: 'success',
+            });
+        });
```
trevor-cortex: I think this test as written should fail. Tracing through:

1. First dispatch: `signals.shift()` returns `AbortSignal.abort(new Error('first'))` — already aborted.
2. `AbortSignal.any([controller.signal, abortedSignal])` returns an already-aborted combined `signal`.
3. `fn = () => Promise.resolve('ok')` — never reads `signal`, just resolves to `'ok'`.
4. Post-`await` check is `if (controller.signal.aborted)` — `controller` was just created and hasn't been aborted, so this is `false`.
5. Falls through to `setState({ ..., status: 'success' })`.

So the first `expect(...).toBe('error')` should be `'success'`. Either:
- the `fn` needs to actually honor the signal (e.g. `(signal) => signal.aborted ? Promise.reject(signal.reason) : Promise.resolve('ok')`), making this a real test of the recovery path, or
- the implementation should also check the combined `signal.aborted` post-await and surface that as an error (which would be a behavior change — currently only the per-dispatch controller can produce that throw).

Could you double-check this one runs green locally? If it does I'm missing something.

packages/subscribable/src/reactive-action-store.ts
```diff
@@ -49,25 +49,35 @@ export type ReactiveActionStore<TArgs extends readonly unknown[], TResult> = {
...
+ *     return source.reactiveStore({ perRequestSignal: () => AbortSignal.timeout(30_000) });
  * }
  * ```
  *
  * @see {@link ReactiveActionStore}
  * @see {@link ReactiveStreamSource}
  */
 export type ReactiveActionSource<T> = {
-    reactiveStore(): ReactiveActionStore<[], T>;
+    reactiveStore(options?: { perRequestSignal?: () => AbortSignal | undefined }): ReactiveActionStore<[], T>;
```
trevor-cortex: Minor consistency nit: `perRequestSignal` returns `AbortSignal | undefined`, while the stream-store's `perConnectionSignal` returns just `AbortSignal`. The action-store's `connect()` equivalent already handles `undefined` via the ternary, and the stream-store's `connect()` already handles it via `if (perConnection) signals.push(perConnection)` — so it'd be trivial to widen `perConnectionSignal` to `() => AbortSignal | undefined` for symmetry. Not blocking, but the asymmetric public surface stood out.

packages/subscribable/src/reactive-stream-store.ts
```diff
@@ -132,25 +159,43 @@ export type ReactiveStreamStore<T> = {
...
+                   * lets retries get fresh signals while still supporting kill-switch patterns.
+                   */
+                  abortSignal: AbortSignal;
+                  perConnectionSignal?: () => AbortSignal;
+              }
+            | {
+                  abortSignal?: AbortSignal;
+                  perConnectionSignal: () => AbortSignal;
+              },
+    ): ReactiveStreamStore<T>;
```
trevor-cortex: Worth noting in the review body rather than here, but anchoring near the type so it's findable: `PendingRpcSubscriptionsRequest.reactiveStore` (in `packages/rpc-subscriptions-spec/src/rpc-subscriptions.ts`) still uses the now-deprecated `abortSignal` option and hasn't been threaded with `perConnectionSignal`. So the end-to-end RPC-subscriptions ergonomics don't yet benefit from this PR — only `PendingRpcRequest.reactiveStore` was migrated. Intentional follow-up?

packages/subscribable/src/reactive-stream-store.ts
```diff
@@ -274,41 +319,42 @@ export function createReactiveStoreFromDataPublisher<TData>({
...
     function notify() {
         subscribers.forEach(cb => cb());
     }
 
     function connect() {
-        if (outerController.signal.aborted) return;
-        // Inner signal is passed to data publisher
+        if (longLivedAbortSignal?.aborted) return;
+        // Factory called per connection — fresh signal per attempt for `perConnectionSignal`,
+        // same kill-able signal returned on every call for the kill-switch pattern.
```
trevor-cortex: Sanity-checking the deprecated `abortSignal` path under the new wiring: previously, aborting the caller's `abortSignal` propagated through `outerController` → `forwardAbort` → `innerController`, which removed the publisher listeners (since they were `{ signal: innerController.signal }`). Now `longLivedAbortSignal` is one of the inputs to `AbortSignal.any`, so the combined `signal` aborts when the caller aborts, and the publisher listeners (registered with `{ signal }`) are removed the same way. ✓

One difference worth confirming: in the old code, aborting the caller's `abortSignal` would also propagate to `innerController` directly, so any code reading `innerController.signal.aborted` would see `true`. The new code reads the combined `signal` everywhere instead, which is correct — but if any future code path branches on `innerController.signal.aborted` specifically (e.g. "was this aborted by the error channel, vs by the caller?"), that signal stays `false` on caller-driven aborts now. Not a problem today, just a heads-up for whoever maintains this next.

packages/test-config/browser-environment.ts
```diff
@@ -10,5 +10,35 @@ export default class BrowserEnvironment extends TestEnvironment {
...
+                }
+                for (const inputSignal of signals) {
+                    inputSignal.addEventListener(
+                        'abort',
+                        function () {
+                            if (!controller.signal.aborted) controller.abort(this.reason);
+                        },
+                        { once: true, signal: controller.signal },
+                    );
+                }
```
trevor-cortex, line 39: Polyfill looks correct. Two small things:

1. Inside the listener, `this.reason` relies on `this` being bound to the input signal that fired the event. That works because `addEventListener` invokes listeners with `this === currentTarget`, but it's the kind of thing TypeScript-strict reviewers tend to flag. `event.currentTarget` (or capturing the input signal in the closure) is slightly more explicit. Style call.
2. `JsdomAbortController` is typed as `typeof globalThis.AbortController` — fine for the polyfill, but note that the `controller.signal` returned here is a jsdom `AbortSignal`. That's correct (we want it to pass jsdom's brand checks), just worth a mental note if anyone debugs realm issues later.

trevor-cortex (commented):
## Summary

Adds `perRequestSignal` / `perConnectionSignal` *factory* options to the reactive stores so callers can compose a fresh signal per dispatch / per connection (e.g. `() => AbortSignal.timeout(N)`) while still expressing a kill switch by returning the same cancellable signal each call. Internally rewires `reactive-stream-store.ts` to use `AbortSignal.any` for signal composition (replacing the manual outer/inner forwarder), preserves the existing `abortSignal` option with `@deprecated`, and polyfills `AbortSignal.any` in the shared jsdom test environment. Changeset is `minor` for both packages.

The overall shape is good — the factory API is a clean generalization over a single signal, and the action/stream stores end up with parallel mental models. The internal refactor of `connect()` is genuinely simpler than the previous `outerController` + `forwardAbort` dance.

## Key things to look at

- **Possible test bug in `reactive-action-store-test.ts` — `'lets later dispatches recover when the factory returns a fresh signal each call'`.** The first dispatch supplies an already-aborted `perRequestSignal`, but `fn = () => Promise.resolve('ok')` doesn't read the signal and resolves successfully. The post-`await` check is `if (controller.signal.aborted)` — only the per-dispatch controller, not the combined signal — so the dispatch falls into `setState({ status: 'success' })`. I'd expect the assertion `expect(store.getState().status).toBe('error')` to fail. Either the test needs an `fn` that actually rejects on `signal.aborted`, or the implementation should also surface aborts on the combined signal. Worth running this one in isolation to confirm. (Left an inline comment.)
- **Asymmetry between `perRequestSignal` and `perConnectionSignal` return types.** Action-store factory returns `AbortSignal | undefined`; stream-store factory returns `AbortSignal`. Both are defensible but it's inconsistent in the public surface — easy to unify on `AbortSignal | undefined` for both, with the stream-store's `connect()` already handling the `undefined` branch via `if (perConnection) signals.push(perConnection)`.
- **`PendingRpcSubscriptionsRequest.reactiveStore` isn't migrated.** The wiring in `packages/rpc-subscriptions-spec/src/rpc-subscriptions.ts` still passes `{ abortSignal }` through to `createReactiveStoreFromDataPublisherFactory`. Backwards-compatible, but the symmetric ergonomic win for subscription consumers (per-connection timeouts on RPC subs) isn't actually exposed end-to-end here — only the action-side `PendingRpcRequest.reactiveStore` got the new option. Intentional follow-up?
- **`@deprecated abortSignal` is still load-bearing inside the repo.** The deprecation is correct but worth being aware that the only first-party caller (rpc-subscriptions-spec) still uses it, so the deprecation warning will fire in our own builds the moment anyone enables `@typescript-eslint/no-deprecated` or similar.

- The behavioral change in the action-store catch block is subtle and worth a second look: previously `if (signal.aborted) throw signal.reason` covered both supersession *and* any external abort because `signal` was just `controller.signal`. Now `signal` is the *combined* signal, so the post-await check was correctly narrowed to `controller.signal.aborted` only — that's how `perRequestSignal` aborts get surfaced as `error` state rather than silently dropped. The comments document this well, just confirm the intent matches your mental model.
- The `FactoryConfig` discriminated union (`{ abortSignal; perConnectionSignal? } | { abortSignal?; perConnectionSignal }`) requires at least one of the two to be present at the type level, which is what we want. `{}` correctly fails to typecheck. Good.
- The jsdom `AbortSignal.any` polyfill is scoped to a single static-method patch on jsdom's existing class — this is the right call (cross-realm replacement would break jsdom's brand checks on `addEventListener({ signal })`, as the comment notes). The implementation looks correct, including the early-abort short-circuit and `{ once: true, signal: controller.signal }` cleanup so input listeners go away once the combined signal aborts.
- `AbortSignal.any` cleanup semantics: confirmed the `signals` array passed into `AbortSignal.any` doesn't retain listener references after the combined signal aborts — so retries don't leak listeners onto a long-lived `longLivedAbortSignal` across many `retry()` calls. That was the main concern with the old `forwardAbort` approach and it's resolved here.

packages/subscribable/src/__tests__/reactive-action-store-test.ts
mcintyre94: @trevor-cortex check the updated implementation of `dispatchAsync` in `createReactiveActionStore`: https://github.com/anza-xyz/kit/blob/a0c59f8f86b49433bd7643b8c756b7af7faa1071/packages/subscribable/src/reactive-action-store.ts#L148-L174

To clarify, the test does pass in CI and locally

Here's what I think is happening:

- Signal is set to an aborted signal because `perRequest` is aborted: `const signal = perRequest ? AbortSignal.any([controller.signal, perRequest]) : controller.signal;`
- The abortable promise with signal throws: `const result = await getAbortablePromise(fn(signal, ...args), signal);`
- In the catch block, we haven't aborted the controller (just per-request), so: `if (controller.signal.aborted)` is false
- We fall through to setting the error:
```ts
// Real failure or `perRequestSignal` firing — surface as error state.
setState({ data: previousData, error, status: 'error' });
throw error;
```

Can you double check I've got all this right?

trevor-cortex: You're right, my mistake — I missed that `fn(signal)` is wrapped in `getAbortablePromise(..., signal)`, which is what makes the already-aborted combined `signal` reject even though `fn` itself ignores it. Tracing again:

1. `signals.shift()` → already-aborted `AbortSignal.abort(new Error('first'))`.
2. `signal = AbortSignal.any([controller.signal, perRequest])` is already aborted.
3. `getAbortablePromise(fn(signal), signal)` rejects with `signal.reason` because `signal.aborted` is true — `fn` resolving to `'ok'` doesn't matter, the race is decided by the signal.
4. Catch block: `controller.signal.aborted` is false (only `perRequest` aborted), so we fall through to `setState({ status: 'error' })`. ✅

Second dispatch then gets a fresh non-aborted signal and succeeds. So the test is correct and exercises exactly the recovery path it claims to. Sorry for the noise — that was the piece I was missing.

packages/subscribable/src/reactive-stream-store.ts
mcintyre94: Pushing this to a follow up PR in the stack that will come before `useSubscription` to keep both PRs narrowly focused.

packages/subscribable/src/reactive-action-store.ts
mcintyre94: Moved all changes to reactive-stream-store to an upcoming upstack PR

packages/subscribable/src/reactive-stream-store.ts
mcintyre94: Moved all changes related to reactive-stream-store to an upstack PR

packages/subscribable/src/reactive-stream-store.ts
mcintyre94: This is intentional, they're now just treated as independent signals

mcintyre94:
This PR has been simplified to only add `perRequestSignal` to the `ReactiveActionStore`. Changes related to `ReactiveStreamStore` are moved to an upstack PR (not submitted yet)

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
LGTM (commenting since I can't approve). Scope tightened nicely — stream-store changes deferred to the follow-up, so this PR is now just the action-store `perRequestSignal` factory plus the wiring through `PendingRpcRequest.reactiveStore` and the jsdom `AbortSignal.any` polyfill. The factory shape is a clean generalization: per-attempt timeouts via `() => AbortSignal.timeout(N)` and kill-switch semantics via `() => killCtrl.signal` both fall out of the same primitive.

From my prior review, the three substantive points are resolved/scoped out:

- The test I flagged (`'lets later dispatches recover when the factory returns a fresh signal each call'`) is correct — `getAbortablePromise(fn(signal), signal)` races against the combined signal, so an already-aborted `perRequestSignal` rejects the await even when `fn` itself doesn't read the signal. Thanks for the walkthrough.
- The `perRequestSignal` / `perConnectionSignal` return-type asymmetry is moot now that `perConnectionSignal` isn't in this PR.
- `PendingRpcSubscriptionsRequest.reactiveStore` not being migrated is explicitly deferred to the upstack stream-store PR.

- The narrowing in the catch block from `signal.aborted` to `controller.signal.aborted` is the load-bearing behavior change: previously `signal` was just the per-dispatch controller, so it covered supersession only; now `signal` is the combined signal, and the post-await check is correctly narrowed to `controller.signal` so that `perRequestSignal` aborts surface as `error` state rather than being silently dropped as supersession. The inline comments document this well.
- The jsdom `AbortSignal.any` polyfill is scoped to a single static-method patch on jsdom's existing class (cross-realm replacement would break jsdom's brand checks for `addEventListener({ signal })`, as the comment notes). Implementation is correct: early-abort short-circuit, and `{ once: true, signal: controller.signal }` so input-signal listeners are cleaned up once the combined signal aborts — no listener leak on long-lived `perRequestSignal`s across many dispatches.
- Changeset is `minor` for both `@solana/subscribable` and `@solana/rpc-spec`; both are pre-1.0 so reviewers should treat that as breaking-equivalent per the repo's convention.

mcintyre94:
> Changeset is minor for both @solana/subscribable and @solana/rpc-spec; both are pre-1.0 so reviewers should treat that as breaking-equivalent per the repo's convention.

I think the line about pre-1.0 in our claude.md is confusing trevor, I've seen this locally too. But to clarify, minor because non breaking!

lorisleiva (commented):
Hmm I'm not a massive fan of how decoupled the `AbortSignal` is from the actual `dispatch` call. It feels like a significant indirection when ideally you'd want `dispatch(...args, { signal })`. Could you tell me what we'd be missing if we made each dispatchable function responsible for their own `AbortSignal`? That is, they must add them explicitly to their function signature.

mcintyre94:
@lorisleiva Agreed, thanks for the suggestion! This design is similar to what we have in the stream store ATM, which is an `abortSignal` passed in the factory config. My plan (in the later PR updating that) was to deprecate that and introduce the `perConnectionSignal` alongside it, which led to doing the same with `perRequestSignal` here.

But I think it's better to add `signal` as an optional options object on the calls to `dispatch` and `dispatchAsync`. We'd still internally combine that with a signal that is aborted when `reset` is called, but I agree it's more direct and clear. It also enables some interesting use cases, like different timeouts per dispatch, a per-call cancellation, and more composability.

The only downside is that if you do want the same kill-switch signal, or the same signal factory, then you need to repeat it on every dispatch. But it's definitely more obvious what's going on!

Will update this.

mcintyre94:
I went with the fluent helper idea to cleanly separate the signal from the args: `store.withSignal(AbortSignal.timeout(5_000).dispatch(...)`. When this is passed we use `AbortSignal.any` to abort if either this signal, or the store controller signal (aborted on eg `reset()`) is aborted. 

This is non-breaking, `store.dispatch(...)` dispatches without an additional signal as previously, and only aborts if the store controller signal is aborted.

The `.reactiveStore()` on `PendingRpcRequest` now takes an optional `{ signal }`: `const store = rpc.getLatestBlockhash().reactiveStore({ signal: AbortSignal.timeout(5000) })`. Subsequent calls do not use the same signal, they use the same `store.withSignal(newSignal).dispatch(...)` mechanism. I think this is potentially a little confusing, but on balance it's the best API. Using the same signal would be confusing, and re-introducing the factory concept here would have the same concerns about the signal being implicit and restrict the API.

trevor-cortex:
Worth noting the API has already moved in this direction since the PR description was written — the current head no longer exposes a `perRequestSignal` factory at the store level. It's now:

- `createReactiveActionStore(fn)` — `fn` receives the per-dispatch `AbortSignal` as its first positional arg (always plumbed through the internal controller).
- `store.dispatch(...args)` / `store.dispatchAsync(...args)` — bare dispatch, no caller signal.
- `store.withSignal(signal).dispatch(...args)` — per-call composition with the internal controller via `AbortSignal.any`.
- `source.reactiveStore({ signal })` — one-shot signal for the *implicit* initial dispatch that `reactiveStore()` fires on construction.

So modulo shape, this is pretty close to `dispatch(...args, { signal })` — the question is really about the `withSignal(signal).dispatch(...)` indirection vs. a trailing options bag. Callum will have the authoritative answer, but my read on the trade-offs:

**For `withSignal` over `dispatch(...args, { signal })`:**
- `dispatch` is variadic on `TArgs extends readonly unknown[]`. A trailing options object would either need runtime detection ("is the last arg an options bag?") or change the shape to `dispatch(args, options?)`, which loses the nice 1:1 with `fn`'s tail args. `withSignal` sidesteps that entirely.
- `withSignal` doubles as a binding primitive — `const killable = store.withSignal(killCtrl.signal)` lets you bind once and reuse the kill switch from many call sites without each one having to know about it. Per-call `{ signal }` forces every site to thread it.
- The wrapped `fn` already takes `signal` as its first positional arg. Keeping the caller-side signal positional-ish (`withSignal(s).dispatch(...)`) mirrors that.

**Against:**
- `withSignal(signal).dispatch(...)` is genuinely more typing and more indirection than `dispatch(...args, { signal })` at the most common call site (one-shot per-attempt timeout). The bind-once case is real but probably rarer than the one-shot.
- Two places to attach a signal (`reactiveStore({ signal })` for the initial fire, `withSignal` for subsequent ones) is a bit of a wart — but it falls out of the fact that the initial fire has no explicit call site, not out of the `withSignal` design itself.

**On the alternative you're floating** (each dispatchable function responsible for its own signal, no store-level plumbing): that works for the bare `createReactiveActionStore(fn)` case where you control `fn`'s signature, but it doesn't work for `PendingRpcRequest.reactiveStore()` — the args are baked in by the wrapper, so `dispatch()` is `() => void` and there's nowhere for the caller to inject a signal short of widening the API. You'd end up needing *something* like `withSignal` or `dispatch({ signal })` to attach one externally. The current design picks the former.

So I think the real question to weigh is whether the bind-once ergonomic + the variadic-tuple-friendliness of `withSignal` is worth the extra hop at the common call site, vs. moving to `dispatch(...args, { signal })` and accepting a slightly less clean kill-switch story. Both seem defensible to me.

mcintyre94:
With #1628 this PR is simplified, and no longer needs to change the `RpcPendingRequest.reactiveStore()` API, which no longer auto-dispatches and so doesn't need a signal when called.

@trevor-cortex Please do a fresh code review of this PR as-is. Earlier comments refer to a previous version of the API

trevor-cortex:
Fresh review — LGTM (commenting since I don't have write access). Scope is tight and the API is materially cleaner than where this started.

## What changed since the prior round

- Factory dropped; the surface is now `store.withSignal(signal).{ dispatch, dispatchAsync }`. Internally a single `dispatchAsyncWithSignal(userSignal, ...args)` core, with bare `dispatchAsync` just calling it with `undefined`. Minimal diff against the old code path — nice.
- `PendingRpcRequest.reactiveStore()` is now `() => ReactiveActionStore<[], TResponse>` with no options object, matching #1628 dropping auto-dispatch. Signal handling is now uniform: `reactiveStore().withSignal(sig).dispatch()`.
- Polyfill unchanged; still scoped to a single static-method patch on jsdom's `AbortSignal`.

## Behavior trace (for the record)

The two abort-check sites in `dispatchAsyncWithSignal` interact cleanly:

- **User signal aborts** → `AbortSignal.any` propagates → either the post-await `if (signal.aborted) throw signal.reason` fires, or `getAbortablePromise` rejects with the same reason → catch sees `controller.signal.aborted === false` → falls through to `setState({ status: 'error' })`. ✅
- **Controller aborts** (supersession or `reset()`) → combined signal aborts → catch sees `controller.signal.aborted === true` → re-throws `controller.signal.reason` and skips the `setState`, so only the *newer* dispatch's outcome reaches state. ✅
- **Already-aborted user signal at dispatch time** → `AbortSignal.any` returns already-aborted → `getAbortablePromise` rejects → user-signal path → `error` state. ✅ (matches the `'lets later dispatches recover when a prior call started with an already-aborted signal'` test.)

This was the load-bearing thing I asked about in the prior round and it's still correct under the simplified shape.

- **Wrapper is non-memoized** — each `store.withSignal(sig)` returns a fresh `{ dispatch, dispatchAsync }` object. That's by design and the docs explicitly tell kill-switch users to bind once (`const killable = store.withSignal(killCtrl.signal)`). Worth being aware of if anyone passes `store.withSignal(...)` directly through React render without memoization, but it's harmless — the wrapper isn't tracked as a dep anywhere meaningful.
- **`withSignal(signal: AbortSignal)` is non-optional.** Callers threading an optional signal through (e.g. `signal?: AbortSignal`) have to branch between `store.withSignal(sig).dispatch()` and `store.dispatch()`. Widening to `AbortSignal | undefined` and falling back to bare `dispatch` internally would remove that branch, but it's a minor ergonomics call — easy follow-up if it bites.
- **Doc wording nit on `withSignal`'s JSDoc** — says "aborting either cancels the in-flight call and surfaces the abort reason on state." Strictly true only for the *user* signal; the *internal* controller path (supersession / `reset()`) is intentionally silent so the newer dispatch owns state. The asymmetry is correct and documented on `dispatch` itself, just slightly elided here. Worth a one-line tweak; not blocking.
- **Changeset is `@solana/subscribable` minor only.** The `rpc-spec` change is JSDoc-only (no API or behavior change), so omitting it is correct. Pre-1.0 minor still treated as breaking per repo convention — reviewers should sanity-check downstream consumers of `reactiveStore()`'s old `(options)` signature, since that signature is gone now (per #1628).
- **Test coverage is solid** — seven cases over `withSignal`: signal-forwarded-to-`fn`, error transition, stale-data preservation, recovery after already-aborted, combined-signal observable by `fn`, varying signals across dispatches, and kill-switch reuse. Good spread; nothing obvious missing.
- **Polyfill listener cleanup** — `addEventListener('abort', ..., { once: true, signal: controller.signal })` removes input-signal listeners once the combined signal aborts. No leak when a long-lived user signal is composed into many short-lived dispatches.

packages/test-config/browser-environment.ts
```diff
@@ -10,5 +10,35 @@ export default class BrowserEnvironment extends TestEnvironment {
...
+        /**
+         * Polyfill `AbortSignal.any` on jsdom's `AbortSignal` class if missing. jsdom 22 doesn't
+         * implement it (added natively in jsdom 24, in Node 20.3, and shipped in all current
+         * browsers). Cross-realm replacement of the whole class would break jsdom's internal
+         * brand checks elsewhere, so this patches just the static method.
+         */
+        const JsdomAbortSignal = this.global.AbortSignal as typeof globalThis.AbortSignal & {
+            any?: typeof globalThis.AbortSignal.any;
+        };
+        const JsdomAbortController = this.global.AbortController as typeof globalThis.AbortController;
```
lorisleiva, line 22: Should we consider bumping jsdom to version 24 in a subsequent PR?

lorisleiva (approved):
Nice!

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1614).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1614) couldn't merge this pull request because a downstack PR #1607 failed to merge.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1614).
* **Jun 26, 2:00 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1614) couldn't merge this pull request because a downstack PR #1628 failed to merge.
* **Jun 26, 2:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1614).
* **Jun 26, 2:19 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1614) rebased this pull request as part of a merge.
* **Jun 26, 2:29 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1614).

PR #1618, opened 2026-05-11
Add `useRequest` for one-shot RPC reads

mcintyre94:
#### Summary of Changes

Fixes #

PR #1619, opened 2026-05-11
Add `useRequest` for one-shot RPC reads/data fetching

mcintyre94:
#### Summary of Changes

This PR adds the `useRequest` hook:

```typescript
function LatestBlockhash() {
    const client = useClient<ClientWithRpc<GetLatestBlockhashApi>>();
    const source = useMemo(() => client.rpc.getLatestBlockhash(), [client]);
    const { data, error, refresh, status } = useRequest(source);

    if (status === 'fetching') return <p>Loading…</p>;
    if (error) return <button onClick={refresh}>Retry</button>;
    return (
        <p>
            Blockhash: {data?.value.blockhash} <button onClick={refresh}>↻</button>
        </p>
    );
}
```

This uses the `ReactiveActionStore` state machine like `useAction`, but the request is immediately dispatched.

The hook was designed around `ReactiveActionSource`, ie the type `PendingRpcRequest` now satisfies with its `.reactiveStore()` function. But I realised that since it's using `ReactiveActionStore` it can actually work with any async function with hardly any extra code (all the logic is in the store):

```typescript
function Profile({ userId }: { userId: string }) {
    const fetcher = useCallback(
        (signal: AbortSignal) => fetch(`/api/users/${userId}`, { signal }).then(r => r.json()),
        [userId],
    );
    const { data, error, refresh } = useRequest(fetcher);
    if (error) return <button onClick={refresh}>Retry</button>;
    return <p>{data ? data.name : 'Loading…'}</p>;
}
```

I think this makes it much more useful for apps since it can cover much more of their data fetching requirements now. Apps using indexers/databases/SDKs etc in addition to RPC requests can transparently use this hook for everything, and it's all using the same state machine logic under the hood.

Note that we won't have the same flexibility with the subscription hooks since they build on `DataPublisher`, this works because the action store was always designed for arbitrary async functions.

The hook takes an optional `getAbortSignal` prop with a function to create an abort signal per request. This is also used to create the signal for the auto-dispatch request:

```tsx
function LatestBlockhash() {
    const client = useClient<ClientWithRpc<GetLatestBlockhashApi>>();
    const source = useMemo(() => client.rpc.getLatestBlockhash(), [client]);
    const { data, error, refresh, status } = useRequest(source, {
        getAbortSignal: () => AbortSignal.timeout(5_000),
    });

    if (status === 'fetching') return <p>Loading…</p>;
    if (error) return <button onClick={refresh}>Retry</button>;
    return (
        <p>
            Blockhash: {data?.value.blockhash} <button onClick={refresh}>↻</button>
        </p>
    );
}
```

The signal can optionally be overridden for `refresh` requests: `refresh({ abortSignal: userClickedCancel.signal })`.

### DIFF
--- a/.changeset/fresh-eggs-hear.md
+++ b/.changeset/fresh-eggs-hear.md
@@ -1,0 +1,24 @@
+---
+'@solana/react': minor
+---
+
+Add `useRequest` — a React hook for one-shot async reads. Pass either an async function `(signal) => Promise<T>` or a memoized `ReactiveActionSource<T>` (satisfied by `PendingRpcRequest`). The hook fires the call on mount, re-fires whenever the source identity changes, and aborts the in-flight call on cleanup.
+
+```tsx
+// `ReactiveActionSource` (e.g. `PendingRpcRequest`):
+const source = useMemo(() => client.rpc.getLatestBlockhash(), [client]);
+const { data, error, refresh } = useRequest(source);
+
+// Bare async function:
+const fetcher = useCallback(
+    (signal: AbortSignal) => fetch(`/api/users/${userId}`, { signal }).then(r => r.json()),
+    [userId],
+);
+const { data, error, refresh } = useRequest(fetcher);
+```
+
+The result reports `status` as one of `fetching | success | error | disabled`. A request in flight is always `fetching`; inspect `data` and `error` to know what stale content (if any) is available to render alongside a spinner — first attempt has neither, a refresh after a prior outcome carries one or both forward. Pass `null` for the source to gate the request off — useful while inputs aren't yet known. The result then reports `status: 'disabled'`.
+
+Optional `getAbortSignal: () => AbortSignal` is a factory invoked on every attempt (initial fire + every `refresh()`). Each attempt gets a fresh signal that's composed with the store's internal per-dispatch controller via `AbortSignal.any`. The natural use is per-attempt timeouts: `getAbortSignal: () => AbortSignal.timeout(5_000)` gives every attempt its own 5-second clock that resets on refresh. The factory is held in a ref synced to the latest render, so inline closures are fine — no `useCallback` needed. `refresh()` also accepts an optional `{ abortSignal }` override to replace the factory for one specific attempt.
+
+The new `RequestResult<T>` and `UseRequestOptions` types are exported alongside the hook so plugin hooks built on top can declare their return shape against them.
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -129,6 +129,76 @@ try {
 }
 ```
 
+### `useRequest(source, options?)`
+
+Fires a one-shot request on mount and re-fires whenever `source` changes identity. Returns `{ data, error, status, refresh }` where `status` is one of `'fetching' | 'success' | 'error' | 'disabled'`. Use it for RPC reads, or for any other one-shot async work an app needs (a `fetch`, a third-party SDK call, etc.).
+
+`source` is either an async function `(signal: AbortSignal) => Promise<T>` (most general), or any reactive store source `{ reactiveStore(): ReactiveActionStore<[], T> }` — `PendingRpcRequest` is the canonical implementation. Pass `null` to disable (the result reports `status: 'disabled'`).
+
+> Unlike `useAction`, `useRequest` needs the input to have stable identity across renders — it's how the hook knows when to re-fire. Memoize with `useMemo` (for a reactive store source) or `useCallback` (for a function), keyed on whatever inputs your call depends on.
+
+```tsx
+import { useClient, useRequest } from '@solana/react';
+import type { ClientWithRpc, GetLatestBlockhashApi } from '@solana/kit';
+
+function LatestBlockhash() {
+    const client = useClient<ClientWithRpc<GetLatestBlockhashApi>>();
+    const source = useMemo(() => client.rpc.getLatestBlockhash(), [client]);
+    const { data, error, refresh } = useRequest(source);
+    if (error) return <button onClick={refresh}>Retry</button>;
+    return <p>{data ? `Blockhash: ${data.value.blockhash}` : 'Loading…'}</p>;
+}
+```
+
+`refresh()` re-fires the request manually. While a refresh is in flight, `status` returns to `'fetching'` and `data` / `error` from the prior outcome stay populated until the new attempt resolves (stale-while-revalidate). On the first attempt both are `undefined`.
+
+```tsx
+function Balance({ address }: { address: Address | null }) {
+    const client = useClient<ClientWithRpc<GetBalanceApi>>();
+    // Disabled until an address is selected.
+    const source = useMemo(() => (address ? client.rpc.getBalance(address) : null), [client, address]);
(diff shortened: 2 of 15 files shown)


mcintyre94:
* **#1619** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1619?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1619?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds `useRequest`, a one-shot async-read hook on top of the existing `ReactiveActionStore` state machine. The hook accepts either a `ReactiveActionSource<T>` (the `{ reactiveStore() }` duck-type satisfied by `PendingRpcRequest`) or a bare `(signal) => Promise<T>` function, fires on mount, re-fires on identity change, and exposes a stable `refresh()`. It maps the store's `idle | running | success | error` onto a `disabled | loading | retrying | loaded | error` surface, plus a `perRequestSignal` factory escape hatch for per-attempt timeouts.

The extraction of `useIsomorphicLayoutEffect` from `useAction.ts` into its own module is a nice cleanup. The disabled-store pattern (`staticStores.ts`) is a clean way to keep the `idle` → `disabled` mapping internal to the bridge without complicating the store contract.

Nice docs and tests overall — README is thorough, the test file covers the lifecycle, the function/source shape split, source-identity transitions, abort propagation, `perRequestSignal` semantics, and the `disabled` ↔ active transitions.

## Things to look at

The main items below are two JSDoc inaccuracies on `RequestResult` (the `isLoading` and `status: 'retrying'` blurbs both overstate the conditions vs what the bridge actually does for the first-failed-call → refresh path) and a discussion point about doing the eager dispatch inside `useMemo`. Inline comments below.

- **Verify `useMemo` + eager dispatch under StrictMode / concurrent rendering.** The store is created (and dispatched) inside `useMemo`. In practice React keeps the memo unless deps change, but the docs explicitly say `useMemo` callbacks should be pure — worth a sanity check that the request doesn't fire more than once on mount under StrictMode, and that a dropped memo (if it ever happens) wouldn't leak an undispatched-but-also-not-reset store. The `useEffect(() => () => store.reset(), [store])` cleanup should catch the leak case, but the assumption is worth double-checking. See inline.
- **`useRequestResult.ts` switch is exhaustive over `idle | running | success | error`** — no `default` and no exhaustiveness check. Fine today because `ReactiveActionStatus` is closed in `@solana/subscribable`, but if a new status is ever added there, this bridge will silently fall through to `undefined`. A `default: status satisfies never` would be cheap insurance.
- **Type tests are sparse.** The function-shape inference path isn't covered (`useRequest((s: AbortSignal) => Promise<X>)` should infer `RequestResult<X>`), and there's no `@ts-expect-error` for invalid sources. Worth a follow-up if you want symmetry with other hooks in the package.
- **`disabledActionStore`'s `dispatch` is a silent no-op but `dispatchAsync` rejects with `AbortError`.** The hook only wires `refresh` to `dispatch`, so this is unobservable from `useRequest` today, but anyone reusing `disabledActionStore` elsewhere will see the asymmetry. Probably fine — flagging only because the rejection vs. no-op split isn't obvious from the JSDoc.

trevor-cortex:
Inline notes:

**`useRequest.ts` — `isLoading` JSDoc (around L24).** The comment says `isLoading` is `true` *only* on the very first `loading` state. But `useRequestResult` returns `isLoading: true` whenever it maps to `status: 'loading'`, which is **any** `running` state where `state.data === undefined`. So if the first dispatch errors (no prior `data` was ever set) and the user calls `refresh()`, the next `running` is still mapped to `loading` with `isLoading: true`. The "only on the very first" phrasing reads as if `isLoading` will never go true again after the first attempt finishes, which isn't what the code does. Suggest something like: "`true` whenever the request is in flight with no value yet to show — i.e. on the first call, and on any retry after an error that never produced data. `false` during `retrying` so spinners don't replace stale content."

**`useRequest.ts` — `status: 'retrying'` JSDoc (around L33).** Same root cause. `retrying` says "re-fire after a prior success *or error*", but the bridge only enters `retrying` when `state.data !== undefined` — i.e. only when there was a prior **success**. A refresh after a first-call error (no prior data) maps to `loading`, not `retrying`. Suggest dropping the "or error" wording.

**`useRequest.ts` — `useMemo` for store creation with eager dispatch (around L116–130).** `useMemo` builds the store and synchronously calls `actionStore.dispatch()` (function shape) or relies on `source.reactiveStore()` to dispatch internally (source shape — see the fake source in the test file). React's docs explicitly say `useMemo` callbacks should be pure and treated as a cache hint, not a guarantee. Two things worth checking:

1. Under StrictMode, does the request fire once or twice on mount? The `useEffect(() => () => store.reset(), [store])` cleanup runs between the two mounts, so the first store gets aborted — but the second mount creates a new store and dispatches again, which may be observable to consumers (two network round-trips in dev).
2. If React ever discards the memo and recomputes, a new store is created (and dispatched) without a `[source]` change. The effect's `[store]` dep will catch the cleanup of the old one when committed, but you'd briefly have two in-flight requests.

Neither is a correctness bug, but the patterns where this matters tend to surface late. If you want belt-and-braces, you can keep `useState` for the store identity (as `useAction` does) and rebuild on `source` change via a `useRef` + render-phase identity compare, falling back to a `useEffect` for the dispatch — at the cost of one tick of delay before the request fires. Probably not worth changing if StrictMode behaves correctly; mostly raising for awareness.

lorisleiva (approved):
Perfect! Love that we accept both a callback an a `ReactiveActionSource` so we can just plug in `client.rpc.method()`.

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1619).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1619) couldn't merge this pull request because a downstack PR #1607 failed to merge.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1619).
* **Jun 26, 2:00 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1619) couldn't merge this pull request because a downstack PR #1628 failed to merge.
* **Jun 26, 2:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1619).
* **Jun 26, 2:42 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1619) rebased this pull request as part of a merge.
* **Jun 26, 2:52 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1619).

PR #1629, opened 2026-05-13
ci: replace app-id with client-id

yihau:
#### Problem

`create-github-app-token` has deprecated `app-id` and recommends using `client-id` instead

#### Summary of Changes

replace app-id with client-id


mcintyre94 (approved):
Thanks!

PR #1630, opened 2026-05-13
ci: add github-actions to dependabot

yihau:
#### Problem

I noticed that we don't ask dependabot to check `github-actions`

#### Summary of Changes

add it to the process

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -55,3 +55,9 @@ updates:
         patterns:
           - 'undici'
           - 'undici-types'
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -55,3 +55,9 @@ updates:
         patterns:
           - 'undici'
           - 'undici-types'
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
+    cooldown:
+      default-days: 3
```
mcintyre94, line 63: Not related to this PR, but we should probably add a cooldown across the board on dependabot

mcintyre94 (approved):
Nice, TIL dependabot can do that!

PR #1646, opened 2026-05-14
Flatten the inferred type of clients built via `extendClient` and `withCleanup`

lorisleiva:
#### Problem

When chaining `.use()` calls on a `Client` built with `@solana/plugin-core`, the inferred type accumulates a deep `Omit<Omit<Omit<...>>>` chain because `extendClient` returns `Omit<TClient, keyof TAdditions> & TAdditions`. After five or six plugins, the hover tooltip and any error message referencing the client type become essentially unreadable. `withCleanup` suffers from the same problem with its `Disposable & TClient` return type. The runtime is unaffected — the chain only exists in the type display — but the developer experience for anyone using the plugin system degrades sharply as the chain grows.

#### Summary of Changes

Introduces a new exported `ExtendedClient<TClient, TAdditions>` helper type expressed as a homomorphic mapped type over `Omit<TClient, keyof TAdditions> & TAdditions`, intersected with `& {}` to force TypeScript to eagerly resolve and display the result as a flat object literal rather than retain the alias name. `extendClient` and `withCleanup` now return `ExtendedClient<...>`, so each step of a chain collapses to a single readable object type. Override semantics, optional (`?`) and `readonly` modifiers, symbol-keyed properties (including `Symbol.dispose`), function-valued properties, and branded primitives are all preserved — verified with new compile-time typetests covering modifier preservation in both directions, override behavior, empty-type edge cases, deep chaining (4 steps with an override mid-chain), and `withCleanup` composed after `extendClient`. The change is purely cosmetic at the type level: `ExtendedClient<X, Y>` is structurally equivalent to `Omit<X, keyof Y> & Y`, so all existing consumers remain valid without modification. Also exports `ExtendedClient` from the package so plugin authors can use it in their own custom merging helpers and get the same flattening guarantee.

##### Before
CleanShot 2026-05-14 at 15 46 22@2x
CleanShot 2026-05-14 at 15 46 13@2x

##### After
CleanShot 2026-05-14 at 15 43 07@2x
CleanShot 2026-05-14 at 15 43 30@2x

### DIFF
--- a/.changeset/empty-jars-double.md
+++ b/.changeset/empty-jars-double.md
@@ -1,0 +1,5 @@
+---
+'@solana/plugin-core': patch
+---
+
+Flatten the inferred return type of `extendClient` and `withCleanup` so that chained `.use()` calls on a `Client` no longer produce deeply nested `Omit<Omit<Omit<...>>>` types in editor tooltips and error messages. The inferred shape now displays as a single flat object literal at every step of the chain, while optional (`?`) and `readonly` modifiers, symbol keys, and override semantics are preserved exactly as before. Also exports the new `ExtendedClient<TClient, TAdditions>` helper type for plugin authors who write their own merging helpers and want the same flattening guarantee.
--- a/packages/plugin-core/src/__typetests__/client-typetest.ts
+++ b/packages/plugin-core/src/__typetests__/client-typetest.ts
@@ -1,8 +1,40 @@
 /* eslint-disable @typescript-eslint/no-floating-promises */
-import { type AsyncClient, type Client, type ClientPlugin, createClient, extendClient, withCleanup } from '../client';
+import {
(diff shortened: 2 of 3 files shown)


lorisleiva:
* **#1646** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1646?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1646?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94 (approved):
Awesome, that is so much better!

amilz:
🔥🔥🔥 thank you!

PR #1654, opened 2026-05-18
Add functions to estimate resource limits supporting v1 transactions

mcintyre94:
#### Problem

While testing v1 transactions I found that the `loadedAccountsDataSizeLimit` behaves the same as CU limit for them: it defaults to 0 and in practice needs to be set for almost all transactions.

#### Summary of Changes

Some notes on `loadedAccountsDataSizeLimit`:

- For legacy/v0 transactions, the server sets it to its max value 64 MiB. So transactions without this compute-budget instruction in practice have the max value.
- For v1 the server defaults it to 0, so transactions need to set it.
- It is returned from simulation, like CU limit
- For legacy/v0 transactions, it can be set to a lower value, and that could eg affect priority
- For helpers in Kit, we do not set it for legacy/v0 transactions. It can be manually set using `setTransactionMessageLoadedAccountsDataSizeLimit`. But because the implementation for these versions requires a compute-budget instruction, and it hasn't historically been set, I don't think it's likely to be worth it as a default. We always set it for v1 transactions. 

This PR adds new higher-level helpers for transaction resource limit estimates, mirroring the existing `compute-unit-limit-estimation` in the kit package.

It includes the following functions:

- `estimateResourceLimitsFactory`. Equivalent to `estimateComputeUnitLimitFactory`, but instead of just returning the CU estimate as a number, it returns an object: `{computeUnitLimit: number, loadedAccountsDataSizeLimit: number}`. For legacy and v0 transaction inputs, the `loadedAccountsDataSizeLimit` is optional, though in practice current RPCs seem to reliably return it. For v1 we throw if it is not returned (as we do for CU limit on all versions), because we need to set it for v1 transactions. No RPC should support v1 transactions without returning `loadedAccountsDataSizeLimit`. 

- `estimateAndSetResourceLimitsFactory`. Equivalent to `estimateAndSetComputeUnitLimitFactory`, but it takes the factory from `estimateResourceLimitsFactory` as input. For v1 transactions it additionally sets the `loadedAccountsDataSizeLimit` on the input transaction. Note that the CU limit continues to be overridden if it is currently set to either 0 (our provisory value) or its max value of 1.4M. For data size it is only overridden if it is currently set to 0 (our provisory value). This is under the assumption that existing code is not setting it to the max value, so if it has been set to max then that's intentional.

- `fillTransactionMessageProvisoryResourceLimits`. Equivalent to `fillTransactionMessageProvisoryComputeUnitLimit`, with the same signature. For v1 transactions it additionally sets the `loadedAccountsDataSizeLimit` config value. These limits do still increase the size of the v1 transaction so this is still relevant, but only by 4 bytes each. 

The functions in compute-unit-limit-estimation are deprecated in favour of these new functions.

The goal here is to provide an easy migration from compute-unit-limit functions to resource-limit functions, and to transparently support all transaction versions. For legacy and v0 transactions the behaviour should be identical.

Follow up work would be to modify the default transaction plan executor in kit-plugin-rpc to use these functions. This should allow users of kit-plugins to use v1 transactions without worrying about configuring this new resource limit manually by default.


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds resource-limit helpers (`estimateResourceLimitsFactory`, `estimateAndSetResourceLimitsFactory`, `fillTransactionMessageProvisoryResourceLimits`) in `@solana/kit` that mirror the existing compute-unit-limit-estimation API but additionally handle the `loadedAccountsDataSizeLimit` that v1 transactions require. The compute-unit-only counterparts are kept and marked `@deprecated`. Two new error codes are added to `@solana/errors`, plumbed through `codes.ts`, `context.ts`, `messages.ts`, and the `unwrapSimulationError` whitelist.

Overall this is a clean, well-scoped addition. The new file is a near-1:1 expansion of `compute-unit-limit-estimation.ts`, the test file follows the same patterns as the existing CU tests, and the conditional `ResourceLimitsEstimate<TTransactionMessage>` return type does the right thing for v1 vs the rest. Migration story in the changeset is excellent.

A few things worth a second look:

### 1. Error precedence when v1 simulation fails AND `loadedAccountsDataSize` is missing

In `estimateResourceLimits` the checks run in this order:

1. `unitsConsumed == null` → `FAILED_TO_ESTIMATE_COMPUTE_LIMIT`
2. `isDataSizeRequired && loadedAccountsDataSize == null` → `FAILED_TO_ESTIMATE_LOADED_ACCOUNTS_DATA_SIZE_LIMIT`
3. `transactionError` → `FAILED_WHEN_SIMULATING_TO_ESTIMATE_RESOURCE_LIMITS` (with the transaction error as `cause`)

For a v1 transaction that fails simulation, if the RPC also happens to omit `loadedAccountsDataSize` on failure (plausible — the field is populated from a successful execution), we'd surface a `FAILED_TO_ESTIMATE_LOADED_ACCOUNTS_DATA_SIZE_LIMIT` and swallow the underlying `transactionError`, which is the more informative signal for the caller. It mirrors the existing CU-limit ordering for `unitsConsumed`, so this is consistent and not a regression — but it's the kind of thing a user debugging "why does this v1 transaction fail estimation?" would appreciate. Worth considering either:

- Checking `transactionError` first so the cause is preserved, or
- Wrapping the data-size error with `cause: getSolanaErrorFromTransactionError(transactionError)` when both are present (and adding it to `SolanaErrorCodeWithCause`).

Not a blocker — happy to leave as-is if you'd rather keep the ordering symmetric with the existing CU path.

### 2. `fillTransactionMessageProvisoryResourceLimits` — reference-equality contract

The original `fillTransactionMessageProvisoryComputeUnitLimit` returns the exact input reference when no change is needed (the existing tests assert `expect(result).toBe(message)`). The new implementation preserves this — if both branches are skipped, `result === transactionMessage` — but the new tests don't assert it. Consider adding a `.toBe(message)` test for parity to lock the contract in; this matters for downstream memoization / equality checks.

### 3. Doc clarity on the `loadedAccountsDataSize` source

Minor: the docstring on `ResourceLimitsEstimate` says the data size is "returned when the RPC includes it in the simulation result." Worth tightening the type-level invariant in prose: for v1, missing → throw at estimation time; for legacy/v0, missing → field omitted from result; for legacy/v0, present → field passed through verbatim. The behavior is correct; just easier to scan.

- The error-code additions (`5663036`, `5663037`) extend the existing transaction range cleanly and are added to both `SolanaErrorCode` and `SolanaErrorCodeWithCause` (the second one — the resource-limits simulation error — carries a `cause`, mirroring its CU sibling).
- `unwrapSimulationError` includes the new `FAILED_WHEN_SIMULATING_TO_ESTIMATE_RESOURCE_LIMITS` code so callers using that helper transparently get the underlying transaction error. ✓
- `estimateAndSetResourceLimitsFactory` uses a runtime `'loadedAccountsDataSizeLimit' in estimate` check instead of just relying on the type. This is the right defensive choice given the factory accepts user-supplied estimators that may wrap/strip fields (for buffer logic) — flagging so future refactors don't "clean it up" away.
- The CU re-estimation behavior is unchanged: still triggers on unset / `0` / `1_400_000`. The data-size re-estimation only triggers on unset / `0`, deliberately leaving `64 MiB` alone. This asymmetry is documented in the changeset and tested — sensible given v0/legacy historically defaulted to max while v1 has no such legacy.
- The estimator still overrides the CU limit to max before simulation regardless of whether it's already explicitly set on the input, which is consistent with the existing behavior and necessary for getting a useful `unitsConsumed` reading. Worth keeping an eye on if anyone ever wants "simulate with the user's actual CU limit" semantics, but that's out of scope here.
- One minor pre-existing quirk preserved by the new code: when `unitsConsumed == null`, the thrown `FAILED_TO_ESTIMATE_COMPUTE_LIMIT` falls through the catch and gets re-wrapped as another `FAILED_TO_ESTIMATE_COMPUTE_LIMIT` with itself as `cause`. Not introduced here, but if anyone ever tidies the catch block, also add `FAILED_TO_ESTIMATE_COMPUTE_LIMIT` to the re-throw whitelist.

mcintyre94:
> 1. Error precedence when v1 simulation fails AND loadedAccountsDataSize is missing

This is intentional to mirror the existing CU limit function

> 2. fillTransactionMessageProvisoryResourceLimits — reference-equality contract

Good idea, added these tests that the message reference is unchanged

> 3. Doc clarity on the loadedAccountsDataSize source

I think this docstring is clear:

```ts
/**
 * The estimated resource limits for a transaction message.
 *
 * `computeUnitLimit` is always returned. For version 1 transaction messages,
 * `loadedAccountsDataSizeLimit` is also required. For legacy and version 0 messages, the
 * `loadedAccountsDataSizeLimit` is returned when the RPC includes it in the simulation result,
 * but is not required — callers that don't apply it can ignore the field.
 */
```

lorisleiva (approved):
Amazing, thank you for tackling this! I had no idea we had to explicitly set this for v1.

I think the new API for estimating and setting transaction "resources" is perfect.

I have two small icks about the API but I still think you've made the right choices:
- **Asymmetry between override flags**. CUs include 0 and max. Loaded Accounts Data Size (LADS? lol) only include 0. I think **0 only** is probably is the cleaner design but prioritising the CU migration path is the right choice.
- **Asymmetry between transaction versions**. CUs are set on all versions. LADS only on v1. I agree that the cost of adding a new instruction on legacy/v0 when you don't need it is not worth it so I think it's the right solution.

PR #1658, opened 2026-05-18
Update RPC tests for Agave 4.0

mcintyre94:
#### Problem

Our CI is now picking up Agave 4.0 as the latest stable version, which has broke a few tests

#### Summary of Changes

- `getClusterNodes`: Add the `clientId` field, deprecate the `tpu` and `tpuForward` fields that now return null on Agave 4.0. Also updated on the response type (not just tests),  with a minor changeset. `feat` is now hex (was decimal). 
- `getVersion`: `feat` is now hex (was decimal)
- `getVoteAccounts`: `epochVoteAccount` is no longer always true for our test fixtures, just using `expect.any(Boolean)` as other unpredictable fields in this test do 
- rpc-graphql: changes to token program representation on chain

Fixes: CI 🤞

### DIFF
--- a/.changeset/hip-mugs-sing.md
+++ b/.changeset/hip-mugs-sing.md
@@ -1,0 +1,5 @@
+---
+'@solana/rpc-api': minor
+---
+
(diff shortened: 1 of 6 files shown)


lorisleiva (approved):
Thanks!

joncinque:
I think with this, we can close https://github.com/anza-xyz/kit/pull/814, correct?

PR #1659, opened 2026-05-18
Allow branded base58 and base64 strings to satisfy the encoded bytes types

### LINKED ISSUE #950: We should accept Address as Base58EncodedBytes
## Overview

From twitter:

> Transitioning to gill and understanding the conventions. Things like converting an Address to Base58EncodedString to pass into a memcmp filter. I can get it to work but would like to know more best practices.

https://x.com/fzzyyti/status/1975327247231033718

## Steps to reproduce

Add this typetest in `rpc-api`:

```ts
{
    // address in memcmp
    const rpc = null as unknown as Rpc<GetProgramAccountsApi>;
    const programAddress = null as unknown as Address;
    const filterAddress = null as unknown as Address;

    rpc.getProgramAccounts(programAddress, {
        filters: [
            {
                memcmp: {
                    bytes: filterAddress,
                    encoding: 'base58',
                    offset: 0n,
                }
            }
        ]
    })
}
```

Type error on `bytes: filterAddress`. The only way to resolve this is to use `filterAddress as unknown as Base58EncodedBytes`

## Description of bug

I think we should allow `Address` to be used in the base58 memcmp type:

```ts
type ProgramNotificationsMemcmpFilterBase58 = Readonly<{
    /**
     * The bytes to match, as a base-58 encoded string.
     *
     * Data is limited to a maximum of 128 decoded bytes.
     */
    bytes: Base58EncodedBytes | Address; // add Address here
    /** The encoding to use when decoding the supplied byte string */
    encoding: 'base58';
    /** The byte offset into the account data from which to start the comparison */
    offset: bigint;
}>;
```

The docstring mentions the requirements: base58 encoded, max of 128 bytes. `Address` satisfies these. And since there's base64 support, addresses are probably the only common reason to use base58 memcmp at all. 

Note that `rpc-types` already has a dependency on `addresses` and it's widely used in RPC requests, so this doesn't add a new dependency between packages.

kh0ra:
## Summary

`Base58EncodedBytes`, `Base64EncodedBytes`, and `Base64EncodedZStdCompressedBytes` were declared as uniquely branded types. That prevented any other value carrying the same encoding tag, such as `Address`, `Signature`, or `Blockhash`, from being used wherever these types were expected, even though those values are by definition already a valid base 58 or base 64 encoded string. This change drops the brand wrapper so the three aliases become plain `EncodedString<...>` types over their respective encoding tags. The runtime representation is unchanged and the encoding tag still discriminates base 58 from base 64.

## Motivation

`#950` reports that passing an `Address` as the `bytes` of a `memcmp` filter in `getProgramAccounts` was a type error and could only be worked around with `as unknown as Base58EncodedBytes`. The maintainer suggestion from `#950` was to make the encoded bytes types "dumber" by dropping the brand, and `@mcintyre94` agreed in the same thread that the unique brand was not adding meaningful type safety because nothing was verifying the documented 128 byte limit at the type level.

## Before

```ts
export type Base58EncodedBytes = Brand<EncodedString<string, 'base58'>, 'Base58EncodedBytes'>;
export type Base64EncodedBytes = Brand<EncodedString<string, 'base64'>, 'Base64EncodedBytes'>;
export type Base64EncodedZStdCompressedBytes = Brand<
    EncodedString<CompressedData<string, 'zstd'>, 'base64'>,
    'Base64EncodedZStdCompressedBytes'
>;
```

```ts
const programAddress = null as unknown as Address;
const filterAddress = null as unknown as Address;

rpc.getProgramAccounts(programAddress, {
    filters: [
        {
            memcmp: {
                bytes: filterAddress, // Type error: Address is not Base58EncodedBytes
                encoding: 'base58',
                offset: 0n,
            },
        },
    ],
});
```

## After

```ts
export type Base58EncodedBytes = EncodedString<string, 'base58'>;
export type Base64EncodedBytes = EncodedString<string, 'base64'>;
export type Base64EncodedZStdCompressedBytes = EncodedString<CompressedData<string, 'zstd'>, 'base64'>;
```

The same call now compiles directly, with no cast required. `Base58EncodedBytes` and `Address` still remain distinguishable from `Base64EncodedBytes` because the encoding tag inside `EncodedString` is preserved.

## Tests

The existing typetests in `packages/rpc-types/src/__typetests__/encoded-bytes-typetests.ts` continue to assert that the three aliases satisfy the `EncodedString` brand over their respective encodings and that `Base64EncodedZStdCompressedBytes` also satisfies `CompressedData<string, 'zstd'>`.

The following additional cases are added:

- `Address` satisfies `Base58EncodedBytes`. This is the case originally reported in `#950`.
- A plain `string` does not satisfy `Base58EncodedBytes` or `Base64EncodedBytes`. The encoding tag is still enforced.
- `Base58EncodedBytes` does not satisfy `Base64EncodedBytes`. Encodings remain disjoint.
- An `Address` does not satisfy `Base64EncodedBytes`. Base 58 branded strings cannot pose as base 64.
- Plain `Base64EncodedBytes` does not satisfy `Base64EncodedZStdCompressedBytes`. The compression marker is still required.

A new typetest in `packages/rpc-api/src/__typetests__/get-program-accounts-typetest.ts` covers the original repro from `#950` end to end: `Address`, `Base58EncodedBytes`, and `Base64EncodedBytes` are each accepted by the `memcmp` filter at the appropriate encoding.

## Behavioral impact

None at runtime. The change is type only. Existing call sites that already use `Base58EncodedBytes`, `Base64EncodedBytes`, or `Base64EncodedZStdCompressedBytes` continue to type check without modification, because the brand wrapper was a strict superset of the encoding tagged type. Removing the wrapper widens what can flow into those positions but does not narrow anything.

## Closes

Closes #950

### DIFF
--- a/packages/rpc-api/src/__typetests__/get-program-accounts-typetest.ts
+++ b/packages/rpc-api/src/__typetests__/get-program-accounts-typetest.ts
@@ -1,0 +1,60 @@
+import type { Address } from '@solana/addresses';
+import type { PendingRpcRequest, Rpc } from '@solana/rpc-spec';
+import type { Base58EncodedBytes, Base64EncodedBytes } from '@solana/rpc-types';
+
+import type { GetProgramAccountsApi } from '../getProgramAccounts';
+
+const rpc = null as unknown as Rpc<GetProgramAccountsApi>;
+const programAddress = null as unknown as Address;
+const filterAddress = null as unknown as Address;
+
+// [DESCRIBE] The `memcmp` filter on `getProgramAccounts`.
+{
+    // It accepts an `Address` as the base58 filter bytes.
+    {
+        rpc.getProgramAccounts(programAddress, {
+            encoding: 'base64',
+            filters: [
+                {
+                    memcmp: {
+                        bytes: filterAddress,
+                        encoding: 'base58',
+                        offset: 0n,
+                    },
+                },
+            ],
+        }) satisfies PendingRpcRequest<unknown>;
+    }
+    // It accepts a `Base58EncodedBytes` value as the base58 filter bytes.
+    {
+        const base58Bytes = null as unknown as Base58EncodedBytes;
+        rpc.getProgramAccounts(programAddress, {
+            encoding: 'base64',
+            filters: [
+                {
+                    memcmp: {
+                        bytes: base58Bytes,
+                        encoding: 'base58',
+                        offset: 0n,
+                    },
+                },
+            ],
+        }) satisfies PendingRpcRequest<unknown>;
+    }
+    // It accepts a `Base64EncodedBytes` value as the base64 filter bytes.
+    {
+        const base64Bytes = null as unknown as Base64EncodedBytes;
+        rpc.getProgramAccounts(programAddress, {
+            encoding: 'base64',
+            filters: [
+                {
+                    memcmp: {
+                        bytes: base64Bytes,
+                        encoding: 'base64',
+                        offset: 0n,
+                    },
+                },
+            ],
+        }) satisfies PendingRpcRequest<unknown>;
+    }
+}
--- a/packages/rpc-api/src/__typetests__/get-program-accounts-typetest.ts
+++ b/packages/rpc-api/src/__typetests__/get-program-accounts-typetest.ts
@@ -1,6 +1,6 @@
 import type { Address } from '@solana/addresses';
 import type { PendingRpcRequest, Rpc } from '@solana/rpc-spec';
-import type { Base58EncodedBytes, Base64EncodedBytes } from '@solana/rpc-types';
+import type { Base58EncodedBytes, Base64EncodedBytes, GetProgramAccountsMemcmpFilter } from '@solana/rpc-types';
 
 import type { GetProgramAccountsApi } from '../getProgramAccounts';
 
@@ -56,5 +56,17 @@ const filterAddress = null as unknown as Address;
                 },
             ],
         }) satisfies PendingRpcRequest<unknown>;
+    }
+    // It rejects an `Address` as the bytes of a base64 memcmp filter.
+    {
+        const filter = {
+            memcmp: {
+                bytes: filterAddress,
+                encoding: 'base64' as const,
+                offset: 0n,
+            },
+        };
+        // @ts-expect-error An `Address` is base-58 encoded, so it cannot serve as the bytes of a base-64 filter.
+        filter satisfies GetProgramAccountsMemcmpFilter;
     }
 }
--- a/.changeset/puny-crabs-double.md
+++ b/.changeset/puny-crabs-double.md
@@ -1,0 +1,5 @@
+---
+'@solana/rpc-types': minor
+---
+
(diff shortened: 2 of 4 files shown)


packages/rpc-api/src/__typetests__/get-program-accounts-typetest.ts
```diff
@@ -0,0 +1,60 @@
...
+                {
+                    memcmp: {
+                        bytes: base64Bytes,
+                        encoding: 'base64',
+                        offset: 0n,
+                    },
+                },
+            ],
+        }) satisfies PendingRpcRequest<unknown>;
+    }
```
mcintyre94, line 59: Nit: a good typetest would be `Address` not accepted for base64 filter

mcintyre94 (approved):
Thanks, I think this is a good change and should make things more composeable.

packages/rpc-api/src/__typetests__/get-program-accounts-typetest.ts
kh0ra: Done in 594aaf4b. Thanks for catching that, the negative case makes the contract much clearer.

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Drops the outer `Brand<..., 'Base58EncodedBytes' | 'Base64EncodedBytes' | 'Base64EncodedZStdCompressedBytes'>` wrapper from the three encoded-bytes aliases so they become plain `EncodedString<...>` types tagged only by their encoding (and, for the zstd alias, also by `CompressedData<string, 'zstd'>`). The runtime representation is unchanged; this is a pure type-only widening. The motivation — letting an `Address` (which already carries the base58 `EncodedString` tag) flow into a `memcmp` filter's `bytes` without `as unknown as Base58EncodedBytes` — is exactly the maintainer-suggested fix from #950.

## Things that look right

- **Encodings stay disjoint.** Because the underlying `EncodedString<T, TEncoding>` uses a `NominalType<'stringEncoding', TEncoding>` tag, base58 and base64 remain non-interchangeable at the type level. The typetests assert this in both directions for plain strings and for `Address`.
- **Compression marker is preserved.** `Base64EncodedZStdCompressedBytes` keeps the `CompressedData<string, 'zstd'>` nominal layer, so plain base64 bytes still don't satisfy it. Good — the only piece of nominal information being dropped is the redundant outer brand, not the meaningful encoding/compression discriminators.
- **Strict superset relationship.** The old type was `Brand<EncodedString<...>, ...>`, i.e. a strict subtype of the new type. Existing callers that produce or consume `Base58EncodedBytes` continue to type-check unchanged — this is a widening of input positions only.
- **Changeset.** `minor` on `@solana/rpc-types` is appropriate per the repo's pre-1.0 convention; this is a user-visible additive type change, not a fix and not strictly breaking.
- **End-to-end repro covered.** The new `get-program-accounts-typetest.ts` exercises the exact `#950` scenario (`Address` directly as `memcmp.bytes`) plus the `Base58EncodedBytes`/`Base64EncodedBytes` paths, and includes a `@ts-expect-error` for `Address` as base-64 bytes. Nice coverage.

- **Asymmetry is intentional.** `Address` now satisfies `Base58EncodedBytes`, but `Base58EncodedBytes` does **not** satisfy `Address` (the `Address` brand still discriminates it from a generic base58 string). That's the desired direction — `Address` is strictly more specific.
- **Downstream brand consumers.** Worth a quick check that no internal code path uses `as Base58EncodedBytes` / `as Base64EncodedBytes` to assert a value as that specific brand for nominal narrowing — those would all still compile, but their meaning weakens (they now only assert the encoding tag, which the value usually already has). Nothing problematic, just a side effect of the widening.
- **Minor nit, not blocking:** the new typetests cover `Base58EncodedBytes` → not assignable to `Base64EncodedBytes`, but not the symmetric `Base64EncodedBytes` → not assignable to `Base58EncodedBytes`. Adding that one line would make the disjointness assertion fully symmetric. Small thing.

Overall this is a clean, well-motivated, well-tested change.

mcintyre94 (approved):
Thanks for this, looks great!

PR #1628, opened 2026-05-13
Drop auto-fire from `PendingRpcRequest.reactiveStore()`

mcintyre94:
#### Problem

I've realised that auto-dispatch on `RpcPendingRequest.reactiveStore()` was a mistake, it complicates the API when we add abort signals and is more opinionated/less flexible than the `ReactiveActionStore` itself.

#### Summary of Changes

This PR removes auto-dispatch from `RpcPendingRequest.reactiveStore()`. It now returns a store in the `idle` state, which can be dispatched as normal, at the caller's control.

Among other things will simplify adding abort signals to the store in the upstack PR, becasuse all dispatches now work the same, and `.reactiveStore()` is no longer responsible for a dispatch. 

Note: the "major" label just indicates the major changeset for the breaking change, just a note for myself managing the stack.

PR #1624, opened 2026-05-12
Action Store: Preserve last `error` through `running` for stale-while-revalidate

mcintyre94:
Improving SWR for reactive-action-store: it now maintains the `error` as well as `data` on a subsequent dispatch. This means that apps can display the previous error on a retrying state if they'd like to. 

Related to first note here: https://github.com/anza-xyz/kit/pull/1619#issuecomment-4421619031 - this means we can track state correctly on a running -> error -> running retry scenario.

PR #1662, opened 2026-05-19
Drop auto-connect from stream stores

mcintyre94:
#### Summary of Changes

This PR removes auto-connect from `ReactiveStreamStore`, such that it starts in an idle state and begins streaming only after `.connect()` is called. This mirrors the behaviour of the action store, and means that we can add per-connection signals (in a following PR) for subscriptions with a similar API to the action store.

We deprecate `.retry()`, which was previously only available after an error, in favour of `.connect()` which now also starts the stream for the first time.

The previously deprecated `createReactiveStoreFromDataPublisher`, which takes a `DataPublisher` rather than a `() => DataPublisher` function and therefore can't be called multiple times, is removed.

### DIFF
--- a/.changeset/bumpy-seas-melt.md
+++ b/.changeset/bumpy-seas-melt.md
@@ -1,0 +1,25 @@
+---
+'@solana/subscribable': major
+'@solana/rpc-subscriptions-spec': major
+'@solana/kit': major
+---
+
+Drop auto-connect from `ReactiveStreamStore`; callers explicitly invoke `connect()` to open the underlying stream. Mirrors the action store's caller-driven `dispatch()` pattern — the store is a state machine that callers orchestrate, not a self-starting subscription.
+
+The factory variant returned by `createReactiveStoreFromDataPublisherFactory` now starts in `status: 'idle'`. Call `store.connect()` to open the stream; from `idle`, the store transitions through `loading` → `loaded` (or `error`). A subsequent `connect()` from any non-idle status transitions through `retrying` while preserving the last known value. A new `reset()` method aborts the current connection and returns the store to `idle` without permanently killing it — natural for React effect cleanup.
+
+```ts
+const store = createReactiveStoreFromDataPublisherFactory({
+    abortSignal,
+    createDataPublisher,
+    dataChannelName: 'notification',
+    errorChannelName: 'error',
+});
+store.connect(); // opens the stream — previously this happened on construction
+```
+
+`retry()` is now deprecated; it remains as an error-only alias for `connect()`. Migrate to calling `connect()` directly. Code that previously relied on `retry()` being a no-op when the store was not in `error` state should add an explicit `if (status === 'error') store.connect();` guard at the call site.
+
+`createReactiveStoreFromDataPublisher` (the deprecated non-factory variant accepting a ready-made `DataPublisher`) is removed. Its only documented use was as a backwards-compatibility alias behind `PendingRpcSubscriptionsRequest.reactive()`, which is also removed in this release. Migrate to the factory variant — wrap a ready-made publisher in `() => Promise.resolve(publisher)` if needed — and use `reactiveStore()` for RPC subscriptions.
+
+`createReactiveStoreWithInitialValueAndSlotTracking` in `@solana/kit` no longer fires the RPC request on construction — call `store.connect()` to start it, or wrap in a `useEffect` that calls `connect()` on mount and `reset()` on cleanup. The store starts in `status: 'idle'` and follows the same lifecycle as the underlying stream store.
--- a/packages/subscribable/src/reactive-stream-store.ts
+++ b/packages/subscribable/src/reactive-stream-store.ts
@@ -1,41 +1,22 @@
-import { SOLANA_ERROR__SUBSCRIBABLE__RETRY_NOT_SUPPORTED, SolanaError } from '@solana/errors';
 import { AbortController } from '@solana/event-target-impl';
 
 import { DataPublisher } from './data-publisher';
 
-type Config = Readonly<{
+type FactoryConfig = Readonly<{
     /**
-     * Triggering this abort signal will cause the store to stop updating and will disconnect it from
-     * the underlying data publisher.
+     * Triggering this abort signal will cause the store to stop updating and permanently
+     * disconnect it from any active {@link DataPublisher}. Once the signal has fired,
+     * {@link ReactiveStreamStore.connect | `connect()`} becomes a no-op.
      */
     abortSignal: AbortSignal;
-    /**
-     * Messages from this channel of `dataPublisher` will be used to update the store's state.
-     */
-    dataChannelName: string;
-    // FIXME: It would be nice to be able to constrain the type of `dataPublisher` to one that
+    // FIXME: It would be nice to be able to constrain the type returned by `createDataPublisher` to one that
     //        definitely supports the `dataChannelName` and `errorChannelName` channels, and
     //        furthermore publishes `TData` on the `dataChannelName` channel. This is more difficult
     //        than it should be: https://tsplay.dev/NlZelW
-    dataPublisher: DataPublisher;
-    /**
-     * Messages from this channel of `dataPublisher` will cause subscribers to be notified without
-     * updating the state, so that they can respond to the error condition.
-     */
-    errorChannelName: string;
-}>;
-
-type FactoryConfig = Readonly<{
-    /**
-     * Triggering this abort signal will cause the store to stop updating and will disconnect it from
-     * any active {@link DataPublisher}. Subsequent calls to {@link ReactiveStreamStore.retry | `retry()`}
-     * are no-ops once this signal has fired.
-     */
-    abortSignal: AbortSignal;
     /**
      * An async factory that produces a fresh {@link DataPublisher} each time it is invoked. Called
-     * once on construction and again on every {@link ReactiveStreamStore.retry | `retry()`}. Rejections
-     * surface as a store error.
+     * on every {@link ReactiveStreamStore.connect | `connect()`}. Rejections surface as a store
+     * error.
      */
     createDataPublisher: () => Promise<DataPublisher>;
     /**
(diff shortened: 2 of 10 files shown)


mcintyre94:
* **#1662** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1662?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1662?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/subscribable/src/reactive-stream-store.ts
```diff
@@ -1,42 +1,25 @@
...
+     * on every {@link ReactiveStreamStore.connect | `connect()`}. Rejections surface as a store
+     * error.
+     *
+     *
      */
-    dataChannelName: string;
     // FIXME: It would be nice to be able to constrain the type of `dataPublisher` to one that
     //        definitely supports the `dataChannelName` and `errorChannelName` channels, and
     //        furthermore publishes `TData` on the `dataChannelName` channel. This is more difficult
     //        than it should be: https://tsplay.dev/NlZelW
```
trevor-cortex: Two small doc-comment cleanups in this block:

1. The JSDoc for `createDataPublisher` ends with two trailing blank `*` lines (`* error.\n *\n *\n */`) — looks like a leftover from the merge. Should be a single closing line.
2. The `// FIXME` block (lines 18–21 of this hunk) was originally above the now-removed `dataPublisher: DataPublisher;` field and refers to "dataPublisher" by name. It's now orphaned mid-declaration, sitting *between* the JSDoc for `createDataPublisher` and the `createDataPublisher` field itself. TypeDoc will probably either drop it or attach it to the wrong member. Either move the FIXME above the JSDoc (and reword to mention `createDataPublisher` / the factory return type), or delete it if the constraint no longer applies in the factory-only world.

packages/subscribable/src/reactive-stream-store.ts
```diff
@@ -53,43 +36,65 @@ type FactoryConfig = Readonly<{
...
+ *   {@link ReactiveStreamStore.reset | `reset()`}. Call
+ *   {@link ReactiveStreamStore.connect | `connect()`} to open the underlying stream.
+ * - `loading`: a first connection is in progress; no data has arrived yet.
+ * - `loaded`: a value has been received and no error is active.
+ * - `error`: the stream failed. `data` holds the last known value (or `undefined` if none ever
+ *   arrived) and `error` holds the failure.
+ * - `retrying`: a follow-up `connect()` is in progress after a previous outcome. `error` is
+ *   cleared; `data` is preserved from the previous connection if any.
  */
 export type ReactiveState<T> =
```
trevor-cortex, line 47: Minor semantic wrinkle: the doc says `retrying` is "in progress after a previous outcome", but `performConnect` will transition `loading → retrying` if `connect()` is called a second time before the first connection resolves (since the `else` branch in `performConnect` fires for any non-`idle` status, including `loading`). At that point `data` is still `undefined` and no outcome has occurred yet.

Two options worth considering:
- Tighten the doc to acknowledge that `retrying` can also be entered from in-flight `loading`, or
- Short-circuit `performConnect` so that re-entering from `loading` stays in `loading` (only swap the inner controller / factory invocation, no status transition).

Not blocking — just worth picking one explicitly so the state-machine docs and the implementation agree.

.changeset/bumpy-seas-melt.md
```diff
@@ -0,0 +1,22 @@
+---
+'@solana/subscribable': major
```
trevor-cortex, line 2: `createReactiveStoreWithInitialValueAndSlotTracking` in `@solana/kit` is also breaking — it no longer auto-connects, and every caller now needs `store.connect()` after construction. Kit will bump via the `fixed` config anyway, but the changelog entry won't surface the migration to consumers.

Suggest adding `'@solana/kit': major` to the frontmatter and a short paragraph explaining the migration (e.g. "`createReactiveStoreWithInitialValueAndSlotTracking` no longer fires the RPC request on construction — call `store.connect()` to start it, or wrap in a `useEffect` that calls `connect()` on mount and `reset()` on cleanup.").

trevor-cortex (commented):
## Summary

Drops auto-connect from `ReactiveStreamStore` so the store starts in `idle` and only opens its underlying stream when the caller invokes `connect()`. This mirrors the action-store pattern and sets up per-connection signals in a follow-up PR. `retry()` is deprecated in favour of `connect()` (which now both starts the stream and reconnects after an error/load), and a new `reset()` returns the store to `idle` without permanently killing it. The deprecated `createReactiveStoreFromDataPublisher` non-factory variant is removed, along with `PendingRpcSubscriptionsRequest.reactive()`. Tests across `@solana/subscribable`, `@solana/rpc-subscriptions-spec`, and `@solana/kit` are updated to call `connect()` explicitly.

State machine summary: `idle` → (`connect`) → `loading` → `loaded`/`error`; from `loaded`/`error` → (`connect`) → `retrying` → `loaded`/`error`; from anywhere → (`reset`) → `idle`; once `abortSignal` fires, all `connect()`s are no-ops.

## Things to watch out for

1. **Missing `@solana/kit` entry in the changeset.** `packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts` now starts in `idle` and requires an explicit `connect()` from callers — that's a breaking behavioural change for `@solana/kit` consumers, but the changeset only lists `@solana/subscribable` and `@solana/rpc-subscriptions-spec`. Kit will still bump via the `fixed` config, but the changelog won't tell consumers why their code stopped firing. Worth adding `'@solana/kit': major` and a short migration note (`store.connect()` after construction, or wrap in `useEffect`).

2. **Doc-comment artifact for `createDataPublisher` in `FactoryConfig`.** The JSDoc has two trailing blank `*` lines and the `// FIXME` block (originally above the now-removed `dataPublisher` field) is sandwiched between the JSDoc and `createDataPublisher: ...`. The FIXME also still refers to "dataPublisher" by name. See inline.

3. **`loading` → `retrying` with `data: undefined` when `connect()` is called during the initial connection.** Calling `connect()` while still in `loading` (factory hasn't resolved yet) transitions to `retrying` with `data: undefined`, even though no prior outcome has occurred. The state-machine doc comment says retrying is entered "after a previous outcome" — there isn't one in this case. This is a minor semantic wrinkle rather than a bug; worth confirming the intent and either tightening the doc, or short-circuiting `connect()` to leave the store in `loading` when called from `loading`. See inline.

- `SOLANA_ERROR__SUBSCRIBABLE__RETRY_NOT_SUPPORTED` (8195000) is now orphaned but correctly retained in `packages/errors/src/codes.ts` and the `SolanaErrorCode` union, per the repo's "never remove error codes" policy. The corresponding message in `messages.ts` should also still be present — worth a quick check that it wasn't removed.
- The `setState` equality short-circuit (`status`/`data`/`error` triple compare) prevents spurious subscriber notifications, which is a nice cleanup vs the previous direct `currentState = ...; notify();` pattern.
- `currentInnerController?.abort()` cleanly tears down the prior connection on every `connect()` and `reset()`, so listeners on the old `DataPublisher` are removed before the new connection wires up — covered by the `aborts the prior connection when called again before data arrives` test.
- In the slot-tracking variant, `lastUpdateSlot` is intentionally preserved across `connect()` reconnections (so the store never regresses) but reset on `reset()`. The comment in `performReset` calls this out; worth confirming this is the desired semantics, since a user using `reset()` to switch contexts vs to recover the same context might want different behaviours.
- `retry()` being kept as a deprecated alias is a reasonable migration ramp, but two semantically-different methods on the same store (`connect()` always reconnects; `retry()` only from `error`) may surprise consumers. The deprecation note is good. Worth confirming this is a one-release transitional state rather than a permanent dual-method shape.
- Test coverage on the new factory store is solid (idle/connect/reset/retry/abort all covered, including the in-flight-connect-then-reconnect case). The slot-tracking tests are still using the deprecated `retry()` rather than `connect()` for the error-recovery scenarios — not wrong (it's the deprecated alias path), but worth double-checking that the `connect()` recovery path is exercised at least once in that test file too.

packages/subscribable/src/reactive-stream-store.ts
mcintyre94: Still relevant, applies to the return of `createDataPublisher`. Reworded a bit and moved above the jsdoc. Intentionally not part of the jsdoc though.

packages/subscribable/src/reactive-stream-store.ts
mcintyre94: Agree with this, updated so it stays in `loading`.

.changeset/bumpy-seas-melt.md
mcintyre94: Fair, added `@solana/kit: major`

lorisleiva (approved):
Love it! Consistent APIs 🤤

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1662).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1662) couldn't merge this pull request because a downstack PR #1607 failed to merge.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1662).
* **Jun 26, 2:00 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1662) couldn't merge this pull request because a downstack PR #1628 failed to merge.
* **Jun 26, 2:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1662).
* **Jun 26, 2:54 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1662) rebased this pull request as part of a merge.
* **Jun 26, 3:04 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1662).

PR #1677, opened 2026-05-20
Collapse `retrying` into `loading` with stale-while-revalidate semantics

mcintyre94:
#### Summary of Changes

This PR modifies the states for the stream-store.

The `retrying` state is removed, and we now use `loading` for both first connects and subsequent ones. This matcheshow we structure the action store, and minimises states.

Additionally we preserve the last error (in addition to data) in the `loading` state, to improve SWR. Again this matches a change made to the action store.

PR #1663, opened 2026-05-19
Add `withSignal()` to `ReactiveStreamStore` for per-connection cancellation

mcintyre94:
#### Summary of Changes

This PR adds a per-connect abort signal to the reactive stream store. It uses the same pattern as the action store:

```ts
// no per-connection signal
store.connect();

// with a signal
store.withSignal(mySignal).connect();
```

Note that the previous PR removes auto-connect from the stream store, so this same pattern applies to all connects, including the first. 

The function to create a data publisher for the store is widened from `() => Promise<DataPublisher>` to `(signal: AbortSignal) => Promise<DataPublisher>`, ie the per-connect abort signal can be used as part of that factory. 

The `abortSignal` input to the stream store (and to `createReactiveStoreWithInitialValueAndSlotTracking`, and `PendingRpcSubscriptionsRequest.reactiveStore()`)  is removed in favour of this per-connect pattern.

PR #1678, opened 2026-05-20
Add `UnwrapRpcResponse<T>` + `isSolanaRpcResponse()` to `@solana/rpc-types`

mcintyre94:
#### Summary of Changes

This PR adds a new type `UnwrapRpcResponse` which unwraps `SolanaRpcResponse<T>` to `T` and leaves any other type unchanged. In practice this allows extracting the inner value type from an input that may or may not be a `SolanaRpcResponse`.

Alongside this it adds a new function `isSolanaRpcResponse` which takes an input that may be a `SolanaRpcResponse` and if it matches the `SolanaRpcResponse` shape, casts it to `SolanaRpcResponse<UnwrapRpcResponse<T>>`.

In practice this means we can write code like this:

```ts
if(isSolanaRpcResponse(myResponse)) {
    const slot = myResponse.context.slot;
    const { value } = myResponse;
}
```

So that we can more easily handle in a generic way responses/subscriptions that may or may not have the `SolanaRpcResponse` wrapper. 

These helpers are intended for processing RPC subscriptions in a react hook, but they apply to RPC requests with the `SolanaRpcResponse` wrapper too - and would be useful outside of react.

### DIFF
--- a/packages/rpc-types/src/__typetests__/rpc-api-typetest.ts
+++ b/packages/rpc-types/src/__typetests__/rpc-api-typetest.ts
@@ -1,0 +1,79 @@
+import { isSolanaRpcResponse, type SolanaRpcResponse, type UnwrapRpcResponse } from '../rpc-api';
+import type { Slot } from '../typed-numbers';
+
+// [DESCRIBE] UnwrapRpcResponse
+{
+    // Unwraps `SolanaRpcResponse<U>` to `U`
+    {
+        null as unknown as UnwrapRpcResponse<SolanaRpcResponse<{ lamports: bigint }>> satisfies {
+            lamports: bigint;
+        };
+    }
+
+    // Non-envelope types pass through unchanged
+    {
+        null as unknown as UnwrapRpcResponse<{ lamports: bigint }> satisfies { lamports: bigint };
+        null as unknown as UnwrapRpcResponse<number> satisfies number;
+        null as unknown as UnwrapRpcResponse<string> satisfies string;
+    }
+}
+
+// [DESCRIBE] isSolanaRpcResponse
+{
+    // Narrows an envelope-typed value to the same envelope shape.
+    {
+        const envelope = null as unknown as SolanaRpcResponse<{ lamports: bigint }>;
+        if (isSolanaRpcResponse(envelope)) {
+            envelope.context.slot satisfies Slot;
+            envelope.value satisfies { lamports: bigint };
+        }
+    }
+
+    // For a raw notification, the true-branch narrows to `SolanaRpcResponse<RawValue>`.
+    // (Reaching the true branch on a raw value is impossible at runtime — the guard
+    // validates the envelope shape — so the imprecision is benign.)
+    {
+        const raw = null as unknown as { parent: bigint; slot: bigint };
+        if (isSolanaRpcResponse(raw)) {
+            raw.context.slot satisfies Slot;
+            raw.value satisfies { parent: bigint; slot: bigint };
+        }
+    }
+
+    // For a union of envelope and raw, the true branch narrows to an envelope wrapping
+    // either inner type.
+    {
+        const mixed = null as unknown as SolanaRpcResponse<{ lamports: bigint }> | { lamports: bigint };
+        if (isSolanaRpcResponse(mixed)) {
+            mixed.context.slot satisfies Slot;
+            mixed.value satisfies { lamports: bigint };
+        } else {
+            mixed satisfies { lamports: bigint };
+        }
+    }
+
+    // `unknown` input narrows to `SolanaRpcResponse<unknown>` in the true branch.
+    {
+        const u = null as unknown;
+        if (isSolanaRpcResponse(u)) {
+            u.context.slot satisfies Slot;
+            u.value satisfies unknown;
+        }
+    }
+
+    // Generic-T call-site: the narrowing must work when invoked from inside a generic function
+    // with a parameter typed as `T | undefined`. This mirrors how `useSubscription` consumes
+    // the guard against a `ReactiveStreamStore`'s current value.
+    {
+        function generic<T>(value: T | undefined): {
+            slot: Slot | undefined;
+            value: UnwrapRpcResponse<T> | undefined;
+        } {
+            if (isSolanaRpcResponse(value)) {
+                return { slot: value.context.slot, value: value.value };
+            }
+            return { slot: undefined, value: value as UnwrapRpcResponse<T> | undefined };
+        }
+        void generic;
+    }
+}
--- a/packages/rpc-types/src/rpc-api.ts
+++ b/packages/rpc-types/src/rpc-api.ts
@@ -4,3 +4,56 @@ export type SolanaRpcResponse<TValue> = Readonly<{
     context: Readonly<{ slot: Slot }>;
     value: TValue;
 }>;
+
+/**
+ * Unwraps `SolanaRpcResponse<U>` → `U` at the type level so callers can surface
+ * the inner value without losing static type information. Values that are not
+ * wrapped in a `SolanaRpcResponse` envelope pass through unchanged.
+ *
+ * Pairs with {@link isSolanaRpcResponse} for runtime detection.
+ *
+ * @typeParam T - The raw notification shape.
+ *
+ * @example
+ * ```ts
+ * type AccountValue = UnwrapRpcResponse<SolanaRpcResponse<{ lamports: bigint }>>;
+ * //   ^? { lamports: bigint }
+ *
+ * type AccountValue = UnwrapRpcResponse<{ lamports: bigint }>;
+ * //   ^? { lamports: bigint }
+ * ```
+ */
+export type UnwrapRpcResponse<T> = T extends SolanaRpcResponse<infer U> ? U : T;
+
+/**
+ * Type-guards a notification as a {@link SolanaRpcResponse} envelope. Validates the shape by
+ * duck-typing for `context.slot: bigint` and the presence of `value`.
+ *
+ * The narrowed type is `SolanaRpcResponse<UnwrapRpcResponse<T>>`. In the false branch,
+ * `notification` retains its original type.
+ *
+ * @typeParam T - The notification shape, which may be a raw value, an envelope, or a union of the two.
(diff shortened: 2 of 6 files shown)


mcintyre94:
* **#1678** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1678?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds two helpers to `@solana/rpc-types`:

- `UnwrapRpcResponse<T>` — conditional type that unwraps `SolanaRpcResponse<U>` to `U`, passes other types through.
- `splitSolanaRpcResponse()` — runtime helper that duck-types the envelope shape (`'context' in x && 'value' in x`) and decomposes into `{ value, slot }`. Raw inputs pass through with `slot: undefined`; `undefined` returns both halves as `undefined`.

Four overloads cover the envelope, `undefined`, `SolanaRpcResponse<T> | undefined`, and raw `T` cases. Comes with a thorough unit test file and a small type test, plus a minor changeset.

## Things to watch out for / notes for subsequent reviewers

**1. Missing README update.** The package's `README.md` documents every existing public type and function but isn't updated here. Per the repo's TypeScript READMEs skill ("When adding a new public API, add or update the package's README"), this PR should add `UnwrapRpcResponse` under "Types" and `splitSolanaRpcResponse` under "Functions". The JSDoc is solid, but the README is the user-facing entry point.

**2. Type-level gap for genuinely mixed unions.** The PR description says these helpers handle "an input that may or may not be a `SolanaRpcResponse`". The overloads cover `SolanaRpcResponse<T>`, `undefined`, `SolanaRpcResponse<T> | undefined`, and raw `T` — but **not** `SolanaRpcResponse<T> | T` (i.e. the case where the call site genuinely doesn't know which it has). With a union like `SolanaRpcResponse<X> | RawX`, none of overloads 1–3 match (each arm fails them), so it falls through to overload 4 (`notification: T`) and you get `{ slot: undefined; value: SolanaRpcResponse<X> | RawX }` — slot is statically `undefined` even though at runtime the envelope arm would produce a real slot.

For the React-hook use case (a single subscription whose envelope-ness is statically known) this is fine. But if the intent is broader, consider adding an overload before overload 4 along the lines of:

```ts
export function splitSolanaRpcResponse<TRaw>(
    notification: SolanaRpcResponse<TRaw> | TRaw
): { slot: Slot | undefined; value: TRaw };
```

A matching type-test case would also be worth adding so the resolution doesn't silently regress.

**3. Overload ordering is correct, but fragile.** Overload 4 (`<T>(notification: T)`) is broad enough to swallow envelope inputs if TS ever picked it. It works today because earlier overloads come first and TS picks the first matching signature, but it's worth a comment explaining the ordering so a future contributor doesn't reorder them. Adding the union overload above would push overload 4 even further down — also fine, just keep it as the final fallback.

**4. Minor: `null` handling is implicit.** The runtime guard (`notification != null`) accepts `null`, but the public overloads don't list it; the test for `null` casts via `as unknown as undefined`. Either extend the `undefined` overload to `null | undefined` and add a real `null` overload, or have the test cover only what the type contract promises. The current state is slightly misleading.

**5. Naming.** `splitSolanaRpcResponse` is descriptive, but the function also unwraps — `unwrapSolanaRpcResponse` would arguably read better as a callsite verb (`const { value, slot } = unwrapSolanaRpcResponse(...)`). Subjective, not blocking.

Nothing here is a correctness issue — the runtime behaviour and tests are tight. Main asks are the README update and a decision on whether the mixed-union case is in scope.

mcintyre94:
> 1. Missing README update.

Updated this

> 2. Type-level gap for genuinely mixed unions.

Added an overload + type test. I don't think any of our code will use this, but easy to imagine app code handling multiple different shaped RPC responses etc

> 3. Overload ordering is correct, but fragile. 

I think typetests are fine here

> 4. Minor: null handling is implicit.

Removed the test using null, just leaving it out of the signature.

> 5. Naming. splitSolanaRpcResponse is descriptive, but the function also unwraps

I slightly prefer `split` because it returns `value` and `slot`. I think `unwrap` would imply just the value which is why I used `Unwrap` for the type that just does that, but not this function.

packages/rpc-types/src/rpc-api.ts
```diff
@@ -4,3 +4,78 @@ export type SolanaRpcResponse<TValue> = Readonly<{
...
+    value: UnwrapRpcResponse<T> | undefined;
+} {
+    if (
+        notification != null &&
+        typeof notification === 'object' &&
+        'context' in notification &&
+        'value' in notification
+    ) {
+        const envelope = notification as SolanaRpcResponse<unknown>;
+        return { slot: envelope.context.slot, value: envelope.value as UnwrapRpcResponse<T> };
```
lorisleiva: Just wondering, what if the RPC layer adds to the `context` object in the future? Should we future-proof this by making the return type `SolanaRpcResponse["context"] & Pick<SolanaRpcResponse, "value">`?

Or even better, we could just make this whole function something like `ensureIsSolanaRpcResponse` and if it's not it wraps the value in an empty context. That way, we're only working with the `SolanaRpcResponse` type dictated by the RPC server. Wdyt?

lorisleiva (approved):
I think it's fine as-is but I've made an alternative suggestion in case the context were to change in the future.

packages/rpc-types/src/rpc-api.ts
mcintyre94: Hmm that's a fair point. I wonder if maybe this should just be `isSolanaRpcResponse`, and then the react consumer branches based on that. I don't think `ensureIsSolanaRpcResponse` works, because `context` and `context.slot` are both non-optional, we'd have to use a placeholder value to match the type which feels hacky.

On the react side I'm using this to split `data` and `slot` so that `useSubscription` for example is:

```tsx
const { data, slot , /* other fields */ } = useSubscription(...)
```

I think I'd keep that API for now, rather than making it `context` and forcing you to use `context.slot` when there's nothing else there. And then if we add other fields to `context` they'd probably just be added additively, or maybe it'd be refactored at some point to group them.

That said this is already a little fiddly for the SWR and tanstack query versions, because both of those only provide a single `data` field. So I was considering swapping that out and having data with a `{slot?: Slot, value: T}` shape instead, so it'd be the same with and without a caching library.

In either case though, I don't think it makes sense to expose `context` and `context.slot` to the react consumers. But I think making this `isSolanaRpcResponse` and then letting the react consumer split the envelope is a more flexible/forward-compatible API.

mcintyre94: Refactored this to `isSolanaRpcResponse`, and updated the description.

mcintyre94: Thinking about this a bit more, I'm going to drop this from `useSubscription`. I'm going to have that just return its notification as-is, if you want the slot you use `data.context.slot` like all your other code. If we add something new to `SolanaRpcResponse` then it'll be automatically available like it would anywhere else. If a type doesn't have a `context` then you receive it as-is with no additional wrapping.

I will keep this PR in the stack because I think this functionality may be useful for apps, but I'm not going to use it for the react hooks any more.

packages/rpc-types/src/__typetests__/rpc-api-typetest.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// [DESCRIBE] isSolanaRpcResponse
+{
+    // Narrows an envelope-typed value to the same envelope shape.
+    const envelope = null as unknown as SolanaRpcResponse<{ lamports: bigint }>;
+    if (isSolanaRpcResponse(envelope)) {
+        envelope.context.slot satisfies Slot;
+        envelope.value satisfies { lamports: bigint };
+    }
```
lorisleiva, line 30: nit: We normally use a slightly different format for these typetests:

```ts
// [DESCRIBE] isSolanaRpcResponse
{
    // Test 1.
    {
        // ...
    }

    // Test 2.
    {
        // ...
    }
}
```

lorisleiva (approved):
Sounds good! I trust your judgement. 👌

packages/rpc-types/src/__typetests__/rpc-api-typetest.ts
mcintyre94: Agreed - these should all be fixed now + added a note to claude.md

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1678).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1678) couldn't merge this pull request because a downstack PR #1607 failed to merge.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1678).
* **Jun 26, 2:00 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1678) couldn't merge this pull request because a downstack PR #1628 failed to merge.
* **Jun 26, 2:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1678).
* **Jun 26, 3:06 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1678) couldn't merge this pull request because a downstack PR #1663 failed to merge.
* **Jun 26, 3:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1678).
* **Jun 26, 3:40 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1678) rebased this pull request as part of a merge.
* **Jun 26, 3:50 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1678).

PR #1679, opened 2026-05-20
Add NO_SLOT_HISTORY error

brooksprumo:
#### Problem

A new json rpc error was added to agave in https://github.com/anza-xyz/agave/pull/12598.

#### Summary of Changes

Add it to Kit.

PR #1681, opened 2026-05-21
Add FILTER_TRANSACTION_NOT_FOUND error

brooksprumo:
#### Problem

When working on https://github.com/anza-xyz/kit/pull/1679, I noticed Kit was a missing json rpc error. It was added to agave in https://github.com/anza-xyz/agave/pull/7937.

#### Summary of Changes

Add it to Kit.


packages/errors/src/messages.ts
```diff
@@ -570,6 +571,7 @@ export const SolanaErrorMessages: Readonly<{
     [SOLANA_ERROR__JSON_RPC__SERVER_ERROR_BLOCK_STATUS_NOT_AVAILABLE_YET]: '$__serverMessage',
     [SOLANA_ERROR__JSON_RPC__SERVER_ERROR_EPOCH_REWARDS_PERIOD_ACTIVE]:
         'Epoch rewards period still active at slot $slot',
+    [SOLANA_ERROR__JSON_RPC__SERVER_ERROR_FILTER_TRANSACTION_NOT_FOUND]: '$__serverMessage',
```
brooksprumo, line 574: I wasn't sure what to put here, so guessed based on similar json rpc errors that contained a String message and no data (e.g. `KEY_EXCLUDED_FROM_SECONDARY_INDEX`). Please double check 😸

brooksprumo (commented):
I also saw that a changeset was added to https://github.com/anza-xyz/kit/pull/1679, but I haven't done that here yet.

brooksprumo:
@lorisleiva @joncinque - tagging y'all again for a review request. Thanks in advance!

lorisleiva (changes requested):
Thanks! IIRC, you need to add this error code here to add the server message in the context:

https://github.com/anza-xyz/kit/blob/74b8d3d5166b4857ab722eae0ec5e2843e480a4b/packages/errors/src/json-rpc-error.ts#L121-L140

brooksprumo:
Oh, I just saw https://github.com/anza-xyz/kit/tree/main/packages/errors/#adding-a-new-error. Should I also add to `src/context.ts` as well?

Edit: I've added to `context.ts`.

packages/errors/src/context.ts
```diff
@@ -655,6 +655,9 @@ export type SolanaErrorContext = ReadonlyContextValue<
                 rewardsCompleteBlockHeight: bigint;
                 slot: bigint;
             };
+            [SOLANA_ERROR__JSON_RPC__SERVER_ERROR_FILTER_TRANSACTION_NOT_FOUND]: {
```
lorisleiva, line 658: Ah yeah good shout! I think you've just forgotten to import the constant in this file which is why CI is failing.

Btw, am I right to say your previous error code had no context? That is, no data attached to the error on the RPC server.

brooksprumo: Yep, good find. Should be fixed now.

And yes, the error does not have anything in the `data` field, only the `message` field.

```rust
            RpcCustomError::FilterTransactionNotFound { signature } => Self {
                code: ErrorCode::ServerError(JSON_RPC_SERVER_ERROR_FILTER_TRANSACTION_NOT_FOUND),
                message: format!("Transaction {signature} not found"),
                data: None,
            },
```

brooksprumo: Oh, did you mean the NO_SLOT_HISTORY error?

```rust
            RpcCustomError::NoSlotHistory => Self {
                code: ErrorCode::ServerError(JSON_RPC_SERVER_ERROR_NO_SLOT_HISTORY),
                message: "No slot history".to_string(),
                data: None,
            },
```

lorisleiva: Yeah that's all good then! The server message trick is to ensure we don't loose any dynamic data only present in the error message. In the case of `JSON_RPC_SERVER_ERROR_FILTER_TRANSACTION_NOT_FOUND`, that's the `signature` inside the message.

On that note, may I ask why we're not returning that `signature` as data? Then, consumers could access that signature directly instead of having to parse the error message.

brooksprumo: > On that note, may I ask why we're not returning that `signature` as data?

No idea. Probably because we didn't know to do that 😅

lorisleiva: Has this change landed already server side? Not a big deal if it has but it would be nice if moving forward we could make sure dynamic error information is provided as data instead of available in the error message only. Ideally we would even retroactively fix these errors (as per the following comment in Kit) but I realise this may be near impossible. ☺️

     // The server supplies no structured data, but rather a pre-formatted message. Put 
     // the server message in `context` so as not to completely lose the data. The long 
     // term fix for this is to add data to the server responses and modify the 
     // messages in `@solana/errors` to be actual format strings.

In any case, LGTM and we can always work on a subsequent PR if the server side can still be changed.

lorisleiva (approved):
Thank you! 🍺

packages/errors/src/context.ts
brooksprumo: Looks like the original agave PR was backported to v4.0, so it is going live now.

Re:
> dynamic error information is provided as data instead of available in the error message only

Yes, I agree.

PR #1683, opened 2026-05-21
fix: align codec branch size types

### LINKED ISSUE #1443: Incorrect type overloads for predicate/pattern match codecs

oritwoen:
#### Problem

`getPredicate*` and `getPatternMatch*` are wrappers over the union codec, but their types were more optimistic than runtime. They could expose `fixedSize` even when the selected branch may have a different size.

The union codec typing had the same missing split: same fixed size, known different fixed sizes, and fixed-size branches where TypeScript can't know if sizes match.

#### Summary of Changes

This follows the feedback on #1541. I fixed the union codec typing first, then aligned predicate and pattern-match with it.

- Same known fixed size stays `FixedSize*<..., S>`.
- Known different fixed sizes become `VariableSize*` where branch sizes are still preserved.
- Unknown or ambiguous sizes fall back to the broader `Encoder` / `Decoder` / `Codec` type.
- Typetests cover same-size branches, different-size branches, dynamic arrays, and ambiguous literal-union sizes.

Fixes #1443

### DIFF
--- a/.changeset/tidy-codecs-type-branches.md
+++ b/.changeset/tidy-codecs-type-branches.md
@@ -1,5 +1,5 @@
 ---
-'@solana/codecs-data-structures': patch
+'@solana/codecs-data-structures': major
 ---
 
-Correct union, predicate, and pattern-match codec return types so fixed-size results are only exposed when all known branch sizes match.
+Correct union, predicate, and pattern-match codec return types so fixed-size results are only exposed when all known branch sizes match. This is a type-only change with no runtime impact, but it is breaking: branches whose fixed sizes are unequal or not statically known now widen from `FixedSize*` to a plain `Encoder`/`Decoder`/`Codec`, so consumers that relied on the previous (unsound) fixed-size typing — e.g. reading `.fixedSize` or passing the result where a `FixedSize*` is required — will need to adjust.
--- a/.changeset/tidy-codecs-type-branches.md
+++ b/.changeset/tidy-codecs-type-branches.md
@@ -2,4 +2,9 @@
 '@solana/codecs-data-structures': major
 ---
 
-Correct union, predicate, and pattern-match codec return types so fixed-size results are only exposed when all known branch sizes match. This is a type-only change with no runtime impact, but it is breaking: branches whose fixed sizes are unequal or not statically known now widen from `FixedSize*` to a plain `Encoder`/`Decoder`/`Codec`, so consumers that relied on the previous (unsound) fixed-size typing — e.g. reading `.fixedSize` or passing the result where a `FixedSize*` is required — will need to adjust.
+Align the return types of the union, predicate, and pattern-match codecs so that a fixed-size result is only exposed when every branch is fixed-size _and_ of the same statically-known size. Branches whose sizes are unequal or not statically known now widen from `FixedSize*` to `VariableSize*` (when at least one branch is variable-size) or to a plain `Encoder`/`Decoder`/`Codec` (when the branches are fixed-size but their sizes are not statically comparable), matching what these codecs actually produce at runtime.
+
+This is a type-only change with no runtime impact, but it is breaking in two ways:
+
+- **Return types change.** Consumers that relied on the previous (unsound) fixed-size typing — e.g. reading `.fixedSize`, or passing the result where a `FixedSize*` is required — will need to adjust.
+- **The predicate and pattern-match signatures changed.** Their value/output type is now inferred from the branch encoders, decoders, or codecs rather than from an explicit type argument. Passing an explicit type argument no longer sets the branch domain: `getPatternMatchEncoder<MyType>([...])` (and the decoder/codec equivalents) now fails to compile, and `getPredicateEncoder<MyType>(...)` silently degrades its return to a plain `Encoder<MyType>`. Instead, type the branch predicates' parameters (e.g. `(value: MyType) => ...`) and let the return type be inferred.
--- a/.changeset/tidy-codecs-type-branches.md
+++ b/.changeset/tidy-codecs-type-branches.md
@@ -1,0 +1,5 @@
+---
+'@solana/codecs-data-structures': patch
+---
+
+Correct union, predicate, and pattern-match codec return types so fixed-size results are only exposed when all known branch sizes match.
--- a/packages/codecs-data-structures/src/__typetests__/pattern-match-typetest.ts
+++ b/packages/codecs-data-structures/src/__typetests__/pattern-match-typetest.ts
@@ -27,17 +27,35 @@ const stringTypePredicate = null as unknown as (value: number | string) => value
         // It returns an encoder for the same type as the inputs
         getPatternMatchEncoder([[numberValuePredicate, {} as Encoder<number>]]) satisfies Encoder<number>;
 
-        // It returns a FixedSizeEncoder if all encoders are FixedSizeEncoder
-        getPatternMatchEncoder([
-            [numberValuePredicate, {} as FixedSizeEncoder<number>],
-            [numberValuePredicate, {} as FixedSizeEncoder<number>],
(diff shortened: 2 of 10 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

This PR fixes the long-standing type inaccuracy where `getUnionCodec`, `getPredicateCodec`, and `getPatternMatchCodec` could expose `FixedSize*` even when the runtime helper would actually return a variable size (because the variant branches had different fixed sizes, or the sizes weren't statically knowable). It centralizes the branch-size-state computation in `utils.ts` (`GetFixedCodecSizeState` → `['fixed', S]` / `['variable']` / `['unknown']`) and routes union, predicate, and pattern-match through it. The result is a single, consistent set of rules:

- All branches share a known fixed size `S` → `FixedSize*<…, S>`
- All branches are fixed-size with known but unequal sizes → `VariableSize*`
- Fixed-size branches with `number`/literal-union sizes, or mixed with a non-fixed branch where the rest are unknown → fall back to the broader `Encoder` / `Decoder` / `Codec`
- Any known `VariableSize*` branch → `VariableSize*`

The typetests are thorough — same-size, different-known-size, unknown-size, ambiguous literal-union sizes, dynamic arrays, and the mixed fixed/variable case are all covered for encoder/decoder/codec across all three APIs. I worked through the type logic in `GetFixedCodecSizeState` / `GetFixedCodecSizeTupleState` and it lines up with what the typetests assert. Nicely done overall.

## Notes for subsequent reviewers / things to verify

- **Return-type change is technically a breaking type change.** Before this PR, two `FixedSizeEncoder<T>` branches (with unknown `TSize`) widened to `FixedSizeEncoder<T>`; now they widen to `Encoder<T>`. Same for the unequal-known-sizes case (was `FixedSize*`, now `VariableSize*`). This is correct — the old types were lying — but downstream code relying on the `.fixedSize` field being present at the type level on these results may stop type-checking. The changeset is marked `patch`; given Kit's pre-1.0 "minor may be treated as breaking" convention, it may be worth a quick sanity check with the maintainers on whether `patch` is the intended bump here, or whether to call this out more loudly in the changeset entry. Not blocking — just worth a conscious decision.
- **`predicate.ts` signature rewrite.** The old overloaded form (`<TFrom>` only) is replaced with a single generic form that captures `TIfTrue` / `TIfFalse` and threads them through `GetUnionEncoderType`. The `SameType<…>` constraint on `ifFalse` enforces that both branches share the same value type (the previous overloads achieved this via the shared `TFrom`). I traced inference for the common call shapes (annotated predicate, explicit `<TFrom>`, no-arg predicate) and it works out, but it's worth a careful read in review — it's the most behavior-changing piece of the PR.
- **`utils.ts` newly `export`s `GetUnionEncoderType` / `GetUnionDecoderType` / `GetUnionCodecType` / `GetEncoderTypeFromVariants` / `GetDecoderTypeFromVariants`.** Since `utils.ts` is not re-exported from `src/index.ts`, these stay internal to the package, which is the right call. Worth a glance to confirm no `kit`/`codecs` facade re-exports it transitively.
- **`IsUnion` is duplicated** between `pattern-match.ts` and `utils.ts` — minor; could be consolidated by importing the one from `utils.ts`.
- **`ContainsVariableSizeCodec` naming.** For dynamic arrays it actually checks "all elements are assignable to `TVariableSizeCodec`" rather than "contains one" (because `TItems[number]` is the union of all elements). The semantics are correct for the way it's used here, but the name reads slightly off — happy to leave it, just flagging.

No blockers from me. Inline comments below are minor.

mcintyre94 (approved):
Thanks for this! I've pushed some extra commits, to tighten up the pattern match a bit further and add some more typetests.

I've also updated the consumers in `transaction-messages` that originally motivated the issue. I opened https://github.com/anza-xyz/kit/issues/1738 because the config one still hits a wall, but it's not because of this typing any more.

I've changed the changeset to major (changes signatures + return types), and will merge this together with the other currently pending `major` changesets.

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Re-reviewed end-to-end. The follow-up commits address everything I raised last pass and then some — LGTM, no blockers. (Submitting as a comment because I can't formally approve on this repo.)

## What changed since my last review

- **Changeset bumped to `major`** with an explicit breakdown of the two breaking dimensions (return types widening from `FixedSize*`, and the predicate/pattern-match signatures no longer accepting an explicit value-type argument). The note about `getPredicateEncoder<MyType>(...)` silently degrading to `Encoder<MyType>` is accurate — with the new `TIfTrue = Encoder<TFrom>` default, an explicit `<MyType>` collapses variant inference rather than producing an error. Calling that out in the changeset is exactly right.
- **`pattern-match.ts` rewritten to the `<const TPatterns>` shape** that mirrors `union.ts`, with `GetPatternMatchEncoders` / `Decoders` / `Codecs` tuple extractors and a `TPatterns & readonly PatternMatchEncoderEntry<GetEncoderTypeFromVariants<...>>[]` intersection enforcing that every branch agrees on the value type. The new `// @ts-expect-error` test for mixing number/string branches confirms the constraint fires correctly. The duplicated `IsUnion` I flagged last time is now gone — `pattern-match.ts` imports from `utils.ts` and the local copy is removed.
- **New pattern-match typetests** cover same-size, different-known-size, unknown-size, ambiguous literal-union sizes, mixed fixed/variable, and entirely-unknown — symmetric across encoder/decoder/codec and across boolean vs. type-guard predicates. The inline-arrow test mirroring `getCompiledTransactionMessageEncoder` (distinct per-branch encoders for `Legacy | V0`) is the right shape to lock in real-world call sites.
- **New union typetests** cover the dynamic-array edge cases (`FixedSizeEncoder<…, 4>[]`, `FixedSizeEncoder<…, 1 | 4>[]`, `VariableSizeEncoder<…>[]`) and verify the union widens to `Encoder` rather than promising a size that may not hold for empty arrays. This also retroactively justifies the `ContainsVariableSizeCodec` naming I'd nudged on last time: the check is deliberately conservative (returns `false` for dynamic arrays because they might be empty and therefore fixed-size 0), which is correct behavior — the name reads less weird once you see it in context.
- **`transaction-messages` consumers updated.** `message.ts` is a clean conversion from the old `<…>` type argument to per-branch arrow predicates. `v1/config.ts` swaps the stale `TODO #1143` comment for an accurate explanation of why a cast is still needed — `getStructEncoder` types its `fixedSize` as the broad `number`, so both branches look like "all fixed, unknown size" and the union widens to a plain `Encoder` even though it's genuinely variable-size at runtime. Linking to #1738 as the follow-up is the right call.

## Things to verify before merge

- The `v1/config.ts` site now wraps each `getStructEncoder` in a `transformEncoder(…, (value: U32Value) => value)` no-op narrowing. The intent is to give each branch a distinct input type (`U32Value` vs `U64Value`) so the inferred union value type lands on `CompiledTransactionConfigValue`. Worth a sanity check that this is the cleanest path — an explicit `as Encoder<U32Value>` cast would be shorter — but the `transformEncoder` form is type-safe and doesn't require widening the encoder type, so I think it's the better choice.
- Confirm CI passes `tsc` on this PR *and* on whichever stack PR rolls this together with the other pending `major` changesets — the consumer surface here is narrow but the type-only break could show up downstream.

No blockers, no inline comments. Nice work.

PR #1700, opened 2026-05-26
Stop sending a manual Content-Length from the HTTP transport

pgtls:
#### Problem

undici fetch already sets `content-length` headers, and this library's content-length headers had the incorrect values. This mismatch caused this to break on undici.

#### Summary of Changes

The fix removes the manual header.

### DIFF
--- a/packages/rpc-transport-http/src/__tests__/http-transport-content-length-test.node.ts
+++ b/packages/rpc-transport-http/src/__tests__/http-transport-content-length-test.node.ts
@@ -1,0 +1,72 @@
+import { createServer, IncomingMessage, Server, ServerResponse } from 'node:http';
+import { AddressInfo } from 'node:net';
+
+import { createHttpTransport } from '../http-transport';
+
+// Integration test exercising the real `fetch` (undici on Node, wired up by setup-undici-fetch.ts)
+// rather than a mocked spy. The bug being guarded against is that the transport used to send a
+// manual `Content-Length` header equal to `body.length` (the JS string's UTF-16 code-unit count),
+// which underreports the byte count for any payload containing non-ASCII characters. Servers then
+// read only that many bytes from the socket and stall waiting for the rest of the request, hanging
+// the client until a timeout fires.
+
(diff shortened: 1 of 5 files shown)


packages/rpc-transport-http/src/__tests__/http-transport-content-length-test.node.ts
```diff
@@ -0,0 +1,72 @@
...
+// rather than a mocked spy. The bug being guarded against is that the transport used to send a
+// manual `Content-Length` header equal to `body.length` (the JS string's UTF-16 code-unit count),
+// which underreports the byte count for any payload containing non-ASCII characters. Servers then
+// read only that many bytes from the socket and stall waiting for the rest of the request, hanging
+// the client until a timeout fires.
+
+describe('createHttpTransport (real fetch against a local server)', () => {
+    // Short per-request abort timeout so the test fails fast and cleanly (with an AbortError) if
+    // a regression makes the request hang. Comfortably long for a localhost roundtrip.
+    const REQUEST_TIMEOUT_MS = 2_000;
```
mcintyre94, line 16: Noting that the happy path here is the test passes within a few ms, this timeout is only triggered if the request hangs. So this isn't adding a 2s long test or anything like that.

packages/rpc-transport-http/src/__tests__/http-transport-headers-test.ts
```diff
@@ -111,21 +111,6 @@ describe('createHttpRequest with custom headers', () => {
...
-        makeHttpRequest({ payload: 123 }).catch(() => {});
-        expect(fetchSpy).toHaveBeenCalledWith(
-            expect.anything(),
-            expect.objectContaining({
-                headers: expect.objectContaining({
-                    'content-length': '3',
-                }),
-            }),
-        );
-    });
```
mcintyre94, line 128: With this PR a caller can now set `content-length` if they want to and it'll be passed through unchanged. I think that's the right behaviour.

mcintyre94 (approved):
Thanks for this! I agree that we shouldn't be setting this header.

PR #1702, opened 2026-05-26
Add `useSubscription` for rendering based on a subscription

mcintyre94:
#### Summary of Changes

This PR adds the `useSubscription` hook:

```ts
    const client = useClient<ClientWithRpcSubscriptions<AccountNotificationsApi>>();
    const source = useMemo(() => client.rpcSubscriptions.accountNotifications(address), [client, address]);
    const { data, error, reconnect } = useSubscription(source);
     if (error) return <button onClick={reconnect}>Reconnect</button>;
    return <p>{data ? `${data.value.lamports} lamports at slot ${data.context.slot}` : 'Connecting…'}</p>;
}
```

This is a subscription analog to `useRequest`. It uses the `reactiveStreamStore` state machine, and immediately initialises the subscription on mount.

The expected input is a `ReactiveStreamSource`, the current implementation in Kit is a pending RPC subscriptions object. But eg. plugins could also publish their own `ReactiveStreamSource` of the shape `reactiveStore(): ReactiveStreamStore` and that would also work with this.

As with `useRequest` the hook takes an optional `getAbortSignal` which creates an abort signal per connect (initial and calls to `reconnect`). The call to `reconnect` can optionally override this per call.

### DIFF
--- a/packages/react/src/__typetests__/useSubscription-typetest.ts
+++ b/packages/react/src/__typetests__/useSubscription-typetest.ts
@@ -1,0 +1,56 @@
+/* eslint-disable react-hooks/rules-of-hooks */
+
+import { SolanaRpcResponse } from '@solana/rpc-types';
+import { ReactiveStreamSource } from '@solana/subscribable';
+
+import { SubscriptionResult, useSubscription } from '../useSubscription';
+
+const accountSource = null as unknown as ReactiveStreamSource<SolanaRpcResponse<{ lamports: bigint }>>;
+const slotSource = null as unknown as ReactiveStreamSource<{ slot: bigint }>;
+
+// [DESCRIBE] useSubscription
+{
+    // Envelope sources surface the envelope as-is — callers read `data.value` and
+    // `data.context.slot` directly.
+    {
+        const account = useSubscription(accountSource);
+        account satisfies SubscriptionResult<SolanaRpcResponse<{ lamports: bigint }>>;
+        account.data satisfies SolanaRpcResponse<{ lamports: bigint }> | undefined;
+        account.data?.value satisfies { lamports: bigint } | undefined;
+        account.data?.context.slot satisfies bigint | undefined;
+    }
+
+    // Raw notifications pass through unchanged.
+    {
+        useSubscription(slotSource) satisfies SubscriptionResult<{ slot: bigint }>;
+    }
+
+    // Source argument accepts null
+    {
+        useSubscription<{ slot: bigint }>(null) satisfies SubscriptionResult<{ slot: bigint }>;
+    }
+
+    // Options accept a `getAbortSignal` factory
+    {
+        useSubscription(slotSource, { getAbortSignal: () => AbortSignal.timeout(30_000) });
+    }
+
+    // `reconnect` accepts no args (uses the factory), an `{ abortSignal }` override, or
+    // `{ abortSignal: undefined }` to opt out of the factory entirely.
+    {
+        const { reconnect } = useSubscription(slotSource);
+        reconnect();
+        reconnect({ abortSignal: AbortSignal.timeout(1_000) });
+        reconnect({ abortSignal: undefined });
+        // @ts-expect-error - abortSignal must be an AbortSignal (or undefined)
+        reconnect({ abortSignal: 'nope' });
+    }
+
+    // Status is a discriminated string, not a generic string
+    {
+        const { status } = useSubscription(slotSource);
+        status satisfies 'disabled' | 'error' | 'loaded' | 'loading';
+        // @ts-expect-error - 'success' is the action-store vocabulary, not stream
+        status satisfies 'success';
+    }
+}
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -9,6 +9,7 @@ export * from './useClient';
 export * from './useClientCapability';
 export * from './useRequest';
 export * from './useSignAndSendTransaction';
+export * from './useSubscription';
 export * from './useSignIn';
 export * from './useSignMessage';
 export * from './useSignTransaction';
--- a/packages/react/src/useSubscription.ts
+++ b/packages/react/src/useSubscription.ts
@@ -1,0 +1,136 @@
+import { ReactiveStreamSource, ReactiveStreamStore } from '@solana/subscribable';
+import { useMemo } from 'react';
+
+import { disabledStreamStore } from './staticStores';
+import { useReactiveStoreLifecycle } from './useReactiveStoreLifecycle';
+import { useSubscriptionResult } from './useSubscriptionResult';
+
+// Module-level so it's a stable reference — keeps the `reconnect` callback from
+// `useReactiveStoreLifecycle` referentially stable across renders.
+function fireConnect<T>(store: ReactiveStreamStore<T>, signal: AbortSignal | undefined): void {
+    if (signal) store.withSignal(signal).connect();
+    else store.connect();
+}
+
+/**
+ * Reactive state for a subscription managed by {@link useSubscription} (and other stream-store
+ * hooks built on top of it).
+ *
+ * Lifecycle: starts at `loading` (or `disabled` when the source is `null`) and opens the
+ * underlying stream on mount; transitions to `loaded` on the first notification or `error` on
+ * failure. `reconnect()` re-opens the stream — while a reconnect is in flight, `status` returns
+ * to `loading` and the stale `data` and/or `error` from the prior connection remain populated
+ * (stale-while-revalidate).
+ *
+ * @typeParam T - The notification type emitted by the underlying source.
+ */
+export type SubscriptionResult<T> = {
+    /**
+     * The latest notification. `undefined` on the first load and while disabled. On `loading`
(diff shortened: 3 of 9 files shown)


mcintyre94:
* **#1702** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1702?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1702?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/react/src/index.ts
```diff
@@ -9,6 +9,7 @@ export * from './useClient';
 export * from './useClientCapability';
 export * from './useRequest';
 export * from './useSignAndSendTransaction';
+export * from './useSubscription';
```
trevor-cortex, line 12: Ordering nit: this is out of alphabetical order with the `useSign*` cluster — `useSubscription` should come after `useSignTransaction` (line 14), not between `useSignAndSendTransaction` and `useSignIn`. The rest of the file is alphabetical for the `use*` exports.

packages/react/src/useSubscription.ts
```diff
@@ -0,0 +1,164 @@
...
+    error: unknown;
+    /**
+     * Re-open the stream. By default each call mints a fresh signal from `getAbortSignal` (if
+     * configured) and threads it through the underlying store's `withSignal(signal).connect()`.
+     * Pass `{ abortSignal }` to override the configured factory for just this attempt. Pass
+     * `{ abortSignal: undefined }` to opt out of the factory entirely for this attempt and open
+     * with no caller-provided signal.
+     *
+     * Stable reference. Safe to put in `onClick` handlers or effect deps — typically wired up
+     * to a "Reconnect" button when `status === 'error'`. Calls `store.connect()` under the
```
trevor-cortex, line 48: Minor: the wording "or when only an error has arrived" implies `slot` is `undefined` whenever `status === 'error'`. In practice, after a `loaded` → `error` transition the stale envelope is preserved in `state.data` and `splitSolanaRpcResponse` will continue returning the stale `slot`. Consider clarifying that `slot` follows `data` — i.e. it's stale during `error` after a prior `loaded`, and only `undefined` when no envelope value has ever arrived.

packages/react/src/useSubscriptionResult.ts
```diff
@@ -0,0 +1,43 @@
...
+ *   `useSubscription` to signal the null-source case.
+ *
+ * @internal
+ */
+export function useSubscriptionResult<T>(
+    store: ReactiveStreamStore<T>,
+    reconnect: (options?: { abortSignal?: AbortSignal | undefined }) => void,
+    disabled: boolean,
+): SubscriptionResult<UnwrapRpcResponse<T>> {
+    const state = useSyncExternalStore(store.subscribe, store.getUnifiedState, store.getUnifiedState);
```
trevor-cortex: Worth noting that this passes `store.getUnifiedState` as the server snapshot (3rd arg) — `useRequestResult` doesn't. The reason this is correct here is that idle state is a frozen singleton with stable identity, so the server and client snapshots match on first paint and `useSyncExternalStore` doesn't warn. Probably worth a follow-up to bring `useRequestResult` in line (its current 2-arg form likely triggers a React SSR warning in dev). Not a blocker for this PR.

trevor-cortex (commented):
Adds a `useSubscription` hook that is a near-perfect mirror of `useRequest` for `ReactiveStreamSource`s, plus a `disabledStreamStore` companion for the null-source case. Composition is clean: `useSubscription` owns lifecycle (memo store, effect connect+reset, ref-synced `getAbortSignal`, stable `reconnect`), `useSubscriptionResult` does the `useSyncExternalStore` bridge and the `SolanaRpcResponse` unwrap via `splitSolanaRpcResponse`. The pattern matches `useRequest` / `useRequestResult` so closely that someone who already understands one can read the other immediately — which is exactly the right outcome here.

Things to watch out for:

- **`index.ts` export ordering is broken.** `useSubscription` is inserted between `useSignAndSendTransaction` and `useSignIn`, which puts it out of alphabetical order with the `useSign*` cluster. Should sit after `useSignTransaction` instead. Trivial fix; flagging inline.
- **SSR snapshot getter passed to `useSyncExternalStore`.** `useSubscriptionResult` passes `store.getUnifiedState` as both the client *and* server snapshot getter — that's what makes the SSR test pass without React warning about a missing server snapshot, and it's only correct because `IDLE_STATE` / `IDLE_STREAM_STATE` are frozen singletons with stable identity. Worth noting that `useRequestResult` doesn't currently pass a server snapshot at all, so there's a small inconsistency between the two bridges — not a blocker for this PR, but probably worth a follow-up to bring `useRequestResult` in line (or document why it doesn't need it).
- **`slot` after `loaded` → `error`.** The docblock for `SubscriptionResult.slot` lists when it's `undefined` (loading / disabled / server / error-only). After a successful `loaded` that transitions to `error`, the stale envelope is preserved in `state.data` and `splitSolanaRpcResponse` will keep returning the stale `slot`. That's the right SWR behaviour, and the docs aren't wrong, but the "when only an error has arrived" qualifier is doing more work than it looks — readers may assume `error` always nukes `slot`. Optional: a short note that `slot` follows `data` (stale on `error` after a prior `loaded`).

- The fake subscription helper in `useSubscription-test.browser.tsx` relies on the store binding both the `data` and `error` channels via `.on()` on every `connect()` — it counts to 2 before resolving `publisherReady`. If `createReactiveStoreFromDataPublisherFactory`'s binding order or channel count ever changes, that handshake silently deadlocks. Worth a peek if you're touching the underlying store.
- The `useEffect([store])` cleanup ordering matters under StrictMode and source identity changes: cleanup runs `store.reset()` on the *previous* store closure, then the new store's `connect()` fires. The fake's listener-array reset inside `createDataPublisher` is what makes the "aborts the prior connection's listeners when the source identity changes" test pass cleanly — confirmed by reading through; just calling out where the invariant lives in case it's not obvious.
- The `disabledStreamStore` tests are thorough (frozen state, no-op `connect`/`reset`/`retry`/`withSignal`, no-notify `subscribe`). Coverage of the disabled path is solid.
- Typetest covers the unwrap on both envelope and raw shapes, the null-source case, the presence-based `reconnect` API, and asserts the status union doesn't leak the action-store `'success'` vocabulary. Good signal-to-noise.

packages/react/src/useSubscriptionResult.ts
mcintyre94: Updated `useAction` and `useRequest` to use the same pattern in their PRs

lorisleiva (approved):
LGTM!

packages/react/src/__typetests__/useSubscription-typetest.ts
```diff
@@ -0,0 +1,43 @@
...
+
+// [DESCRIBE] useSubscription
+{
+    // Envelope sources surface the envelope as-is — callers read `data.value` and
+    // `data.context.slot` directly.
+    const account = useSubscription(accountSource);
+    account satisfies SubscriptionResult<SolanaRpcResponse<{ lamports: bigint }>>;
+    account.data satisfies SolanaRpcResponse<{ lamports: bigint }> | undefined;
+    account.data?.value satisfies { lamports: bigint } | undefined;
+    account.data?.context.slot satisfies bigint | undefined;
```
lorisleiva: Same nit on typetests here actually. I won't mention them again in I see that in the rest of the stack but maybe worth a quick Claude check on the whole stack for consistency.

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1702).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1702) couldn't merge this pull request because a downstack PR #1607 failed to merge.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1702).
* **Jun 26, 2:00 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1702) couldn't merge this pull request because a downstack PR #1628 failed to merge.
* **Jun 26, 2:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1702).
* **Jun 26, 3:06 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1702) couldn't merge this pull request because a downstack PR #1663 failed to merge.
* **Jun 26, 3:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1702).
* **Jun 26, 3:52 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1702) rebased this pull request as part of a merge.
* **Jun 26, 4:01 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1702).

PR #1703, opened 2026-05-26
Adds support for retrieving the slot in which migration to alpenglow happened

akhi3030:
#### Problem

In https://github.com/anza-xyz/agave/pull/12735, we are introducing a new rpc endpoint to retrieve the slot in which migration to alpenglow took place.

#### Summary of Changes

This adds support for the endpoint in this repo

### DIFF
--- a/packages/rpc-api/src/getAlpenglowMigrationSlot.ts
+++ b/packages/rpc-api/src/getAlpenglowMigrationSlot.ts
@@ -1,0 +1,12 @@
+import type { Slot } from '@solana/rpc-types';
+
+type GetAlpenglowMigrationSlotApiResponse = Slot | null;
+
+export type GetAlpenglowMigrationSlotApi = {
+    /**
+     * Returns the slot in which the cluster will migrate to Alpenglow.
+     *
+     * @see https://solana.com/docs/rpc/http/getalpenglowmigrationslot
+     */
+    getAlpenglowMigrationSlot(config?: Readonly<object>): GetAlpenglowMigrationSlotApiResponse;
+};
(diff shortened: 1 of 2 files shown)


packages/rpc-api/src/getAlpenglowMigrationSlot.ts
```diff
@@ -0,0 +1,12 @@
...
+
+type GetAlpenglowMigrationSlotApiResponse = Slot | null;
+
+export type GetAlpenglowMigrationSlotApi = {
+    /**
+     * Returns the slot in which the cluster will migrate to Alpenglow.
+     *
+     * @see https://solana.com/docs/rpc/http/getalpenglowmigrationslot
+     */
+    getAlpenglowMigrationSlot(config?: Readonly<object>): GetAlpenglowMigrationSlotApiResponse;
```
mcintyre94, line 11: If the request doesn't take any config (which it looks like from the Agave PR), we can just do:

```suggestion
    getAlpenglowMigrationSlot(): GetAlpenglowMigrationSlotApiResponse;
```

mcintyre94 (commented):
Thanks for this! I'll hold on merging until this is closer to release to avoid confusion, but will keep it open.

It'd also be worth us adding a unit test closer to the time, as we have for other RPC methods.

akhi3030:
Thanks for the review @mcintyre94.  Looks like we will be making changes to https://github.com/anza-xyz/agave/pull/12735 so we should also wait till we finalise what the exact API will be.

mcintyre94:
Marking draft for now. The method landed as `getAgGenesisCert` with no params. If https://github.com/anza-xyz/agave/pull/13170 lands (moving RPC response shape to camel case matching other RPC methods), then the RPC response shape will be:

```json
{
  "jsonrpc": "2.0",
  "result": null || {
    "bitmap": [1,2,3,...],
    "certType": {
      "genesis": {
        "blockId": [4,5,6,...],
        "slot": 0
      }
    },
    "signature": [7,8,9,...]
  },
  "id": 1
}
```

mcintyre94:
Closing, superseded by #2016

PR #1706, opened 2026-05-27
Migrate `@solana/react` to import from `@solana/kit`

mcintyre94:
#### Summary of Changes

This PR changes the imports of `@solana/react` to use `@solana/kit` instead of the individual sub-packages (addresses, transactions, etc) wherever possible

This explicitly makes `@solana/kit` a dependency of `@solana/react`. This is ok because we will never want to export `@solana/react` from `@solana/kit` because of its react dependency.

We do this because a following PR will use `createReactiveStoreWithInitialValueAndSlotTracking` from `@solana/kit`. This is a separate PR to make reviewing easier and keep that one focused.

We also export `@solana/subscribable` from Kit. I think the new types (ReactiveActionStore etc) justify this.

PR #1707, opened 2026-05-27
Add `useTrackedData` for an RPC subscription seeded by an RPC fetch

mcintyre94:
#### Summary of Changes

This PR adds the `useTrackedData` react hook. This is similar to `useSubscription`, but its reactive stream store is constructed using `createReactiveStoreWithInitialValueAndSlotTracking`. This means that it combines a `RpcSendable` and a `RpcSubscribable` and provides the latest value based on slot. In practice the usual behaviour will be that an RPC request provides an initial value before a subscription, but later messages from the subscription update the value.

Examples where this is useful are things like displaying an account's balance, or its latest data. RPC subscriptions don't vend data immediately on subscribing, so they need to be paired with a request for the initial value. This is what `createReactiveStoreWithInitialValueAndSlotTracking` handles, this hook just exposes it for react. 

```tsx
    const client = useClient<ClientWithRpc<GetBalanceApi> & ClientWithRpcSubscriptions<AccountNotificationsApi>>();
    const spec = useMemo(
        () => ({
            rpcRequest: client.rpc.getBalance(address),
            rpcSubscriptionRequest: client.rpcSubscriptions.accountNotifications(address),
            rpcValueMapper: (lamports: bigint) => lamports,
            rpcSubscriptionValueMapper: ({ lamports }: { lamports: bigint }) => lamports,
        }),
        [client, address],
    );
    const { data, slot, error, refresh } = useTrackedData(spec);
    if (error) return <button onClick={refresh}>Retry</button>;
    return <p>{data !== undefined ? `${data} lamports at slot ${slot}` : 'Loading…'}</p>;
}
```

This hook is more opinionated than the others - both the request and subscription must use the `SolanaRpcResponse` shape so that the slot is known.

### DIFF
--- a/packages/react/src/__tests__/useTrackedData-test.browser.tsx
+++ b/packages/react/src/__tests__/useTrackedData-test.browser.tsx
@@ -1,0 +1,432 @@
+import {
+    createReactiveActionStore,
+    createReactiveStoreFromDataPublisherFactory,
+    type ReactiveActionSource,
+    type ReactiveStreamSource,
+    type SolanaRpcResponse,
+} from '@solana/kit';
+import { act } from '@testing-library/react';
+import React from 'react';
+import { renderToString } from 'react-dom/server';
+
+import { renderHook } from '../__test-utils__/render';
+import { TrackedDataSpec, useTrackedData } from '../useTrackedData';
+
+type TestValue = { count: number };
+
+function rpcResponse(slot: number, value: TestValue): SolanaRpcResponse<TestValue> {
+    return { context: { slot: BigInt(slot) }, value };
+}
+
+// Backs `initialValueSource` with a real `ReactiveActionStore`. The wrapped function hands out a
+// fresh controllable promise per dispatch (so a `refresh()` gets its own) and captures the
+// per-dispatch signal so tests can assert it is forwarded and aborted.
+function createMockInitialValueSource(): {
+    fn: jest.Mock;
+    reject(error: unknown): void;
+    resolve(response: SolanaRpcResponse<TestValue>): void;
+    signals: (AbortSignal | undefined)[];
+    source: ReactiveActionSource<SolanaRpcResponse<TestValue>>;
+} {
+    const instances: { reject(error: unknown): void; resolve(response: SolanaRpcResponse<TestValue>): void }[] = [];
+    const signals: (AbortSignal | undefined)[] = [];
+    const fn = jest.fn().mockImplementation((signal?: AbortSignal) => {
+        const { promise, resolve, reject } = Promise.withResolvers<SolanaRpcResponse<TestValue>>();
+        instances.push({ reject, resolve });
+        signals.push(signal);
+        return promise;
+    });
+    return {
+        fn,
+        reject(error) {
+            instances[instances.length - 1]?.reject(error);
+        },
+        resolve(response) {
+            instances[instances.length - 1]?.resolve(response);
+        },
+        signals,
+        source: { reactiveStore: () => createReactiveActionStore(fn) },
+    };
+}
+
+// Backs `streamSource` with a real `ReactiveStreamStore` built from a mock `DataPublisher` factory.
+// The publisher buffers events delivered before the store subscribes (mirroring the previous
+// async-iterable mock) so notifications are never dropped on a timing technicality.
+function createMockStreamSource(): {
+    createDataPublisher: jest.Mock;
+    error(err: unknown): void;
+    pushNotification(notification: SolanaRpcResponse<TestValue>): void;
+    source: ReactiveStreamSource<SolanaRpcResponse<TestValue>>;
+} {
+    const publishers: { publish(channel: string, payload: unknown): void }[] = [];
+    const createDataPublisher = jest.fn().mockImplementation(() => {
+        const listeners: {
+            channel: string;
+            listener: (payload: unknown) => void;
+            options?: { signal?: AbortSignal };
+        }[] = [];
+        const buffered: { channel: string; payload: unknown }[] = [];
+        const publisher = {
+            on(channel: string, listener: (payload: unknown) => void, options?: { signal?: AbortSignal }) {
+                listeners.push({ channel, listener, options });
+                buffered
+                    .filter(b => b.channel === channel)
+                    .forEach(b => {
+                        if (!options?.signal?.aborted) listener(b.payload);
+                    });
+                return function unsubscribe() {};
+            },
+        };
+        publishers.push({
+            publish(channel, payload) {
+                const matched = listeners.filter(l => l.channel === channel && !l.options?.signal?.aborted);
+                if (matched.length === 0) {
+                    buffered.push({ channel, payload });
+                } else {
+                    matched.forEach(l => l.listener(payload));
+                }
+            },
+        });
+        return Promise.resolve(publisher);
+    });
+    return {
+        createDataPublisher,
+        error(err) {
+            publishers[publishers.length - 1]?.publish('error', err);
+        },
+        pushNotification(notification) {
+            publishers[publishers.length - 1]?.publish('data', notification);
+        },
+        source: {
+            reactiveStore: () =>
+                createReactiveStoreFromDataPublisherFactory<SolanaRpcResponse<TestValue>>({
+                    createDataPublisher,
+                    dataChannelName: 'data',
+                    errorChannelName: 'error',
+                }),
+        },
+    };
+}
+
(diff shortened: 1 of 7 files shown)


mcintyre94:
* **#1707** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1707?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1709](https://github.com/anza-xyz/kit/pull/1709) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1709?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1713](https://github.com/anza-xyz/kit/pull/1713) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1713?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1707?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/react/src/__tests__/useTrackedData-test.browser.tsx
```diff
@@ -0,0 +1,395 @@
...
+function makeSpec(): {
+    error: (err: unknown) => void;
+    pushNotification: (notification: SolanaRpcResponse<TestValue>) => void;
+    rejectRpc: (error: unknown) => void;
+    resolveRpc: (response: SolanaRpcResponse<TestValue>) => void;
+    rpcSendCalls: () => number;
+    spec: Spec;
+    subscribeCalls: () => number;
+} {
+    const rpc = createMockRpcRequest();
```
trevor-cortex: `error` is plumbed all the way through `makeSpec()`'s return shape but no test ever calls it. Either add a test for the subscription-side error path (status → `error`, stale data preserved) — the existing error-path test only covers `rejectRpc` — or drop the field to avoid the dead-code smell.

packages/react/src/useTrackedData.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    // One store per `spec`. Both creation paths return an `idle` store; the initial connect
+    // lives in the effect below so the memo body stays pure (StrictMode's dev double-render, and
+    // any future render-discard, won't fire a network request from a discarded render).
+    const store = useMemo(() => {
+        if (spec == null) return disabledStreamStore<SolanaRpcResponse<TItem>>();
+        return createReactiveStoreWithInitialValueAndSlotTracking(spec);
+    }, [spec]);
+
+    // Initial connect on commit + teardown on store change / unmount. `disabledStreamStore`
```
trevor-cortex: Minor: when `spec` is `null`, this still invokes `getAbortSignalRef.current?.()`, which for an `AbortSignal.timeout(30_000)` factory arms a timer for an "attempt" that's about to no-op against `disabledStreamStore`. Cheap nit, not a bug — and `useSubscription` does the same thing today, so leaving it as-is for consistency is fine. Flagging in case you want to gate the factory call on `spec != null` here (and mirror in `useSubscription`).

trevor-cortex (commented):
## Summary

Adds `useTrackedData`, the React hook counterpart of the Kit primitive `createReactiveStoreWithInitialValueAndSlotTracking`. It pairs a one-shot RPC request with a subscription, slot-deduping between the two so an out-of-order arrival never regresses the surfaced value. The React surface mirrors `useSubscription` almost line-for-line: the same memoized-store-per-input pattern, the same `getAbortSignal` factory held in a ref, the same presence-based `{ abortSignal }` override on `refresh()`, the same `idle → loading`/`disabled` disambiguation in the result bridge, and the same StrictMode-safe split between the pure memo body and the connect effect. A dedicated `useTrackedDataResult` bridge is introduced because the envelope shape is known to be `SolanaRpcResponse<T>` here — letting `data` and `slot` come straight off `state.data.value` / `state.data.context.slot` rather than running `splitSolanaRpcResponse`.

The changeset is a `minor` on `@solana/react`, which matches policy (new public API, additive). The README section is well-written and the worked example is consistent with the surrounding `useSubscription` / `useRequest` blocks. LGTM overall — commenting rather than approving since I don't have approve permission on this repo.

## Things to watch out for

- **Strong parallel with `useSubscription`.** This is a feature, not a bug — the two hooks should be mentally interchangeable from an API ergonomics standpoint, and they now are. But it does mean future changes to one are very likely to need a mirror change in the other (e.g. if `useSubscription` ever grows a new option). Worth keeping in mind on subsequent PRs.
- **No test for a subscription-side failure.** The error-path test exercises `rejectRpc` (the initial RPC throwing); there's no equivalent test for the subscription stream erroring. The underlying primitive handles both, and `useSubscription`'s tests cover the subscription-error path on the same store machinery, so this isn't a correctness gap — just a coverage gap on this specific hook. `makeSpec()` even exposes an `error` helper that no test currently uses — either wire it up to a test, or drop it from the returned shape.
- **`getAbortSignal` is invoked even when the spec is `null`.** In the connect effect, the factory is called unconditionally before the (no-op) `store.connect()` runs against the disabled stream store. For `AbortSignal.timeout(...)` factories this is harmless but slightly wasteful — a timer is armed for an attempt that never happens. Could be tightened to `if (spec != null)` around the factory call, but it's a micro-optimization and matches `useSubscription`'s current behaviour, so probably fine to leave for consistency.

- Verify the `TrackedDataSpec` re-export is intentional as a *type alias* rather than just letting users import `CreateReactiveStoreWithInitialValueAndSlotTrackingConfig` directly. The type tests assert mutual assignability, so there's no risk of drift; this is purely a naming-ergonomics choice that's worth a sanity check.
- The `useTrackedDataResult` bridge is marked `@internal` and not re-exported from `index.ts` — only `useTrackedData.ts` is. Good.
- SSR coverage exists (server renders `loading` without firing the RPC/subscription, and `disabled` when spec is null), and StrictMode is enforced via the package-local `renderHook`. Both are aligned with the repo guidance in `CLAUDE.md`.

packages/react/src/__tests__/useTrackedData-test.browser.tsx
mcintyre94: Added a test for this (transitions to error when the subscription throws, preserving stale data if any)

packages/react/src/useTrackedData.ts
mcintyre94: Agree with this, but leaving for now. I don't think this is worth changing.

lorisleiva (approved):
Love it!

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707) couldn't merge this pull request because a downstack PR #1607 failed to merge.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707).
* **Jun 26, 2:00 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707) couldn't merge this pull request because a downstack PR #1628 failed to merge.
* **Jun 26, 2:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707).
* **Jun 26, 3:06 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707) couldn't merge this pull request because a downstack PR #1663 failed to merge.
* **Jun 26, 3:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707).
* **Jun 26, 4:22 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707) couldn't merge this pull request because a downstack PR #1708 failed to merge.
* **Jun 26, 6:12 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707).
* **Jun 26, 6:26 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707) rebased this pull request as part of a merge.
* **Jun 26, 6:36 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1707).

PR #1708, opened 2026-05-27
Refactor createReactiveStoreWithInitialValueAndSlotTracking to use store inputs

mcintyre94:
Note: this PR has been rewritten, first review comments refer to an old version

#### Problem

`createReactiveStoreWithInitialValueAndSlotTracking` currently takes as input a `PendingRpcRequest` and a `PendingRpcSubscriptionsRequest`. Internally it is using the `send()` and `subscribe()` functions, and not the `reactiveStore()` that is used by the react hooks. 

#### Summary of Changes

`createReactiveStoreWithInitialValueAndSlotTracking` is refactored to take a `ReactiveActionStore` and a `ReactiveStreamStore`, and its internals are rewritten to produce its store from these inputs.

Its config now has the shape:

```ts
{
    initialValueSource: ReactiveActionSource<SolanaRpcResponse<TInitialValue>>;
    initialValueMapper: (value: TInitialValue) => TItem;
    streamSource: ReactiveStreamSource<SolanaRpcResponse<TStreamValue>>;
    streamValueMapper: (value: TStreamValue) => TItem;
}
```

Its behaviour is unchanged - it combines the initial value and the stream and always provides the latest available data (based on slot) from either source

The following PR adds a react hook that exposes `createReactiveStoreWithInitialValueAndSlotTracking`, and it will now share the same input types with the other react hooks.

### DIFF
--- a/packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts
+++ b/packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts
@@ -1,40 +1,44 @@
-import type { PendingRpcRequest } from '@solana/rpc';
-import type { PendingRpcSubscriptionsRequest } from '@solana/rpc-subscriptions';
 import type { SolanaRpcResponse } from '@solana/rpc-types';
-import type { ReactiveState, ReactiveStreamStore } from '@solana/subscribable';
+import type {
+    ReactiveActionSource,
+    ReactiveState,
+    ReactiveStreamSource,
+    ReactiveStreamStore,
+} from '@solana/subscribable';
 
 /**
  * Configuration for {@link createReactiveStoreWithInitialValueAndSlotTracking}. Pairs a one-shot
- * RPC fetch with an ongoing subscription so the resulting store can hydrate from the initial
- * response and keep up to date with notifications, slot-deduplicating the two streams.
+ * initial-value source with an ongoing stream source so the resulting store can hydrate from the
+ * initial response and keep up to date with notifications, slot-deduplicating the two sources.
  *
- * @typeParam TRpcValue - The value type returned by `rpcRequest` (inside the {@link SolanaRpcResponse} envelope).
- * @typeParam TSubscriptionValue - The value type emitted by `rpcSubscriptionRequest` (inside the {@link SolanaRpcResponse} envelope).
+ * @typeParam TInitialValue - The value type produced by `initialValueSource` (inside the {@link SolanaRpcResponse} envelope).
+ * @typeParam TStreamValue - The value type emitted by `streamSource` (inside the {@link SolanaRpcResponse} envelope).
  * @typeParam TItem - The unified item type the store holds, produced by the two value mappers.
  *
  * @see {@link createReactiveStoreWithInitialValueAndSlotTracking}
  */
-export type CreateReactiveStoreWithInitialValueAndSlotTrackingConfig<TRpcValue, TSubscriptionValue, TItem> = Readonly<{
+export type CreateReactiveStoreWithInitialValueAndSlotTrackingConfig<TInitialValue, TStreamValue, TItem> = Readonly<{
     /**
-     * A pending RPC request whose response will be used to set the store's initial state.
-     * The response must be a {@link SolanaRpcResponse} so that its slot can be compared with
-     * subscription notifications.
+     * Maps the value from the initial-value source's response to the item type stored in the
+     * reactive store.
      */
-    rpcRequest: PendingRpcRequest<SolanaRpcResponse<TRpcValue>>;
+    initialValueMapper: (value: TInitialValue) => TItem;
     /**
-     * A pending RPC subscription request whose notifications will be used to keep the store
-     * up to date. Each notification must be a {@link SolanaRpcResponse} so that its slot can be
-     * compared with the initial RPC response and other notifications.
+     * A reactive action source whose dispatched value will be used to set the store's initial
+     * state. The value must be a {@link SolanaRpcResponse} so that its slot can be compared with
+     * stream notifications. Satisfied by `PendingRpcRequest`.
      */
-    rpcSubscriptionRequest: PendingRpcSubscriptionsRequest<SolanaRpcResponse<TSubscriptionValue>>;
+    initialValueSource: ReactiveActionSource<SolanaRpcResponse<TInitialValue>>;
     /**
-     * Maps the value from a subscription notification to the item type stored in the reactive store.
+     * A reactive stream source whose notifications will be used to keep the store up to date. Each
+     * notification must be a {@link SolanaRpcResponse} so that its slot can be compared with the
+     * initial value and with other notifications. Satisfied by `PendingRpcSubscriptionsRequest`.
      */
-    rpcSubscriptionValueMapper: (value: TSubscriptionValue) => TItem;
+    streamSource: ReactiveStreamSource<SolanaRpcResponse<TStreamValue>>;
     /**
-     * Maps the value from the RPC response to the item type stored in the reactive store.
+     * Maps the value from a stream notification to the item type stored in the reactive store.
      */
-    rpcValueMapper: (value: TRpcValue) => TItem;
+    streamValueMapper: (value: TStreamValue) => TItem;
 }>;
 
 const IDLE_STATE: ReactiveState<never> = Object.freeze({
@@ -44,27 +48,31 @@ const IDLE_STATE: ReactiveState<never> = Object.freeze({
 });
 
 /**
- * Creates a {@link ReactiveStreamStore} that combines an initial RPC fetch with an ongoing subscription
- * to keep its state up to date.
+ * Creates a {@link ReactiveStreamStore} that combines an initial one-shot fetch with an ongoing
+ * stream to keep its state up to date.
  *
  * The store uses slot-based comparison to ensure that only the most recent value is kept,
- * regardless of whether it came from the initial RPC response or a subscription notification.
- * This prevents stale data from overwriting newer data when the RPC response and subscription
- * notifications arrive out of order.
+ * regardless of whether it came from the initial value source or a stream notification. This
+ * prevents stale data from overwriting newer data when the two sources arrive out of order.
+ *
+ * The two sources are consumed via their {@link ReactiveActionSource.reactiveStore | `reactiveStore()`}
+ * methods rather than by calling `send()` / `subscribe()` directly, so any object satisfying the
+ * {@link ReactiveActionSource} / {@link ReactiveStreamSource} duck-types works — including
+ * `PendingRpcRequest` / `PendingRpcSubscriptionsRequest` and plugin-authored wrappers.
  *
  * Things to note:
  *
  * - The returned store starts in `status: 'idle'`. Call
- *   {@link ReactiveStreamStore.connect | `connect()`} to fire the RPC request and open the
- *   subscription.
- * - The store transitions through `loading` until the first response or notification arrives,
+ *   {@link ReactiveStreamStore.connect | `connect()`} to dispatch the initial-value source and open
+ *   the stream.
+ * - The store transitions through `loading` until the first value or notification arrives,
  *   then to `loaded` with a {@link SolanaRpcResponse} containing the value and the slot context
  *   at which it was observed.
  * - On error from either source, the store transitions to `status: 'error'` preserving the last
  *   known value. Only the first error per connection window is captured.
  * - A subsequent `connect()` aborts the current connection, transitions back to
  *   `status: 'loading'` (preserving the last known `data` and `error` for stale-while-revalidate),
- *   and re-fires the RPC request and subscription with a fresh inner abort signal.
+ *   and re-builds both inner stores with a fresh inner abort signal.
  * - {@link ReactiveStreamStore.reset | `reset()`} aborts the current connection and returns the
  *   store to `idle`, clearing `data` and `error`.
  * - Attach a caller-provided cancellation source via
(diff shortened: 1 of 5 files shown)


mcintyre94:
* **#1708** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1708?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1708?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
LGTM (commenting because I can't formally approve without explicit repo access).

## Summary

Adds two structural duck-types — `RpcSendable<TResponse>` (just `send(options?)`) in `@solana/rpc-spec` and `RpcSubscribable<TNotification>` (just `subscribe(options)`) in `@solana/rpc-subscriptions-spec` — and uses them to loosen the `rpcRequest` / `rpcSubscriptionRequest` fields on the configs for `createReactiveStoreWithInitialValueAndSlotTracking` and `createAsyncGeneratorWithInitialValueAndSlotTracking`. Both primitives only ever call `.send()` / `.subscribe()` on those inputs, so the narrower types are sufficient and existing callers continue to work because `PendingRpcRequest` / `PendingRpcSubscriptionsRequest` structurally satisfy the new types.

Mirrors the prior-art pattern from `ReactiveActionSource` / `ReactiveStreamSource`, and makes life easier for plugin authors / test mocks that don't want to fabricate a full `{ reactiveStore, send }` object.

## What I verified

- The new `RpcSendable.send` signature is identical to `PendingRpcRequest.send` (`send(options?: RpcSendOptions): Promise<TResponse>`), and `RpcSubscribable.subscribe` is identical to `PendingRpcSubscriptionsRequest.subscribe` (`subscribe(options: RpcSubscribeOptions): Promise<AsyncIterable<TNotification>>`) — so structural compatibility actually holds and existing call sites compile unchanged.
- The new types flow through to `@solana/kit` via the existing `export *` chains (`rpc-spec` → `rpc` → `kit`, `rpc-subscriptions-spec` → `rpc-subscriptions` → `kit`), matching the changeset's documented `import type { RpcSendable, RpcSubscribable } from '@solana/kit'` example.
- Both implementation files only consume `.send()` / `.subscribe()` on these inputs, so loosening the type is purely additive at runtime.
- Docblock copy on the two config fields was updated from "A pending RPC request" → "A one-shot RPC request" / "An RPC subscription request", which more accurately describes the looser contract.

## Changeset

`minor` across `@solana/rpc-spec`, `@solana/rpc-subscriptions-spec`, and `@solana/kit` looks right — purely additive new exports plus a contravariant input loosening, no breaking change. Fixed-version lockstep handles the umbrella `@solana/rpc` / `@solana/rpc-subscriptions` packages automatically.

- Worth a quick sanity check that no downstream consumer is relying on the specific `PendingRpcRequest` / `PendingRpcSubscriptionsRequest` shape on the config (e.g. calling `.reactiveStore()` on the passed-in object). A grep over `rpcRequest.reactiveStore` / `rpcSubscriptionRequest.reactiveStore` in this repo and `kit-plugins` should be enough.
- No tests changed. Given this is type-only and the structural relationship is mechanically obvious, that's fine — but if there's a `__typetests__` directory for these primitives, a one-liner asserting that a `Pending*Request` is assignable to the new config (and that a bare `{ send }` / `{ subscribe }` object also is) would lock in the contract cheaply.

lorisleiva (commented):
I wonder if we even need these new types if we already have `ReactiveActionSource` and `ReactiveStreamSource`.

Since the goal is to create a `ReactiveStore` from two sources anyway, could we not use `ReactiveActionSource` for the RPC request and `ReactiveStreamSource` for the RPC subscription? That way, we leave `send` and `subscribe` alone in this reactive universe. Or am I missing something?

mcintyre94:
@lorisleiva Yep agreed, we can refactor `createReactiveStoreWithInitialValueAndSlotTracking` to use stores as inputs and that'll make the react API more cohesive too. Plugins would only need to worry about producing stores if they want to interop.

We can't make the same change to `createAsyncGeneratorWithInitialValueAndSlotTracking` because that yields all notifications, but I'll just leave that restricted to its current type for now. If anyone needs it to be more flexible then we'd add something like these types and use them there.

I'll swap this PR out for a refactor of `createReactiveStoreWithInitialValueAndSlotTracking` to use stores, moving to draft for now.

packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts
```diff
@@ -175,37 +183,56 @@ export function createReactiveStoreWithInitialValueAndSlotTracking<TRpcValue, TS
...
         }
 
-        function handleValue(value: SolanaRpcResponse<TItem>) {
-            setState({ data: value, error: undefined, status: 'loaded' });
+        // `lastUpdateSlot` persists across reconnects so the store never regresses. If a source
+        // delivers a value at a slot older than one we've already seen, we keep waiting for
+        // something newer before leaving `loading`.
+        function handleSlottedValue({ context: { slot }, value }: SolanaRpcResponse<TItem>) {
+            if (signal.aborted) return;
+            if (slot < lastUpdateSlot) return;
```
trevor-cortex, line 191: Nit: this comment is a leftover from the previous `send()` / `subscribe()`-based implementation and reads as if the helper is still wrapping those calls directly. Now that we're consuming the reactive-store contract instead, something like:

> Build fresh inner stores for this connection window and forward their state into the unified store. Drive both with the composed signal so they tear down when this connection window does.

would more accurately describe what the block is doing. Non-blocking.

(Anchoring on the last line of the comment block — confirm the new wording lines up with whatever you settle on.)

packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts
```diff
@@ -175,37 +183,56 @@ export function createReactiveStoreWithInitialValueAndSlotTracking<TRpcValue, TS
...
+            } else if (state.status === 'error') {
+                handleError(state.error);
+            }
+        });
+        const unsubscribeStream = streamStore.subscribe(() => {
+            const state = streamStore.getUnifiedState();
+            if (state.status === 'loaded') {
+                const { context, value } = state.data;
+                handleSlottedValue({ context, value: streamValueMapper(value) });
+            } else if (state.status === 'error') {
```
trevor-cortex, line 216: Nit: worth a one-line elaboration here so a future reader doesn't try to “improve” this by calling `actionStore.reset()` / `streamStore.reset()` and accidentally clobber the caller-signal abort reason. Something like:

> Don't call `reset()` on the inner stores — that would set them to `idle` synchronously, which races the abort-reason error transition already in flight on the composed signal and can drop the caller's abort reason on the floor.

(or whatever the actual failure mode you saw was — that's my best guess from reading the surrounding code). Non-blocking.

trevor-cortex (commented):
## Re-review (rewritten PR)

LGTM on the new shape — this is the right pivot in response to Loris's comment, and the design holds together cleanly. Commenting because I can't formally approve.

## What changed since last review

The PR no longer adds `RpcSendable` / `RpcSubscribable` to `rpc-spec` / `rpc-subscriptions-spec`. Instead `createReactiveStoreWithInitialValueAndSlotTracking` is rewritten to consume two **reactive sources** — `ReactiveActionSource` for the one-shot fetch and `ReactiveStreamSource` for the ongoing stream — calling `reactiveStore()` on each per connection window and forwarding their state into the unified store. The previous `rpcRequest` / `rpcSubscriptionRequest` (+ matching mappers) config fields are renamed to `initialValueSource` / `streamSource` (+ `initialValueMapper` / `streamValueMapper`). Behaviour is preserved.

Net effect: this primitive now lives entirely in the reactive-store universe (matching the React hooks), and `send()` / `subscribe()` are no longer called directly anywhere in the slot-tracking helper.

## What I verified

- **`PendingRpcRequest` / `PendingRpcSubscriptionsRequest` still slot in unchanged.** Both already expose `reactiveStore()`, so the changeset's claim that callers can still pass `rpc.getBalance(addr)` / `rpcSubscriptions.accountNotifications(addr)` directly holds. The example app diff confirms this — the only call-site change is the field renames (plus an explicit `withSignal(...).connect()`, which is the new required entry point).
- **Lifecycle plumbing is sound.** `performConnect` builds fresh inner action + stream stores on every (re)connect, subscribes to both, registers a single `innerSignal` `'abort'` listener that unsubscribes both, and then kicks them via `withSignal(signal).connect()` / `.dispatch()` where `signal` is the composed inner+caller signal. On supersede/reset/caller-abort, the inner controller fires, the unsubscribe listener runs, and the inner stores are orphaned with their own connections cancelled by the same signal — no listener leak, no double state writes (the existing `handleError` `signal.aborted` guard + `currentState.status === 'error'` guard cover the late-error race).
- **Slot tracking centralised.** `handleSlottedValue` now applies the `slot < lastUpdateSlot` skip to both sources symmetrically (previously the RPC path and subscription path each had their own copy), and `lastUpdateSlot` still resets in `performReset`. The “older RPC arrives after newer subscription” / “older subscription arrives after newer RPC” tests still pass against this unified path.
- **Inner-store API choice is correct.** The action subscriber reads `actionStore.getState()` (which returns the discriminated `ReactiveActionState`), and the stream subscriber reads `streamStore.getUnifiedState()` (the non-deprecated stream API). The deprecated `getState()` / `getError()` on `ReactiveStreamStore` aren't used internally.
- **Tests rewritten against real inner stores.** The mocks now back `initialValueSource` with a real `createReactiveActionStore` over a controllable promise factory, and `streamSource` with a real `createReactiveStoreFromDataPublisherFactory` over a mock `DataPublisher`. That's a strictly better test than the previous hand-rolled `send` / `subscribe` mocks — it exercises the actual contract instead of asserting against a fake.
- **`expect.assertions(n)` only on async tests**, fake timers + `await jest.runAllTimersAsync()` for flushing, `Promise.withResolvers` — all matches `CLAUDE.md`'s testing conventions.
- **Changeset bump.** `major` for `@solana/kit` is right: `rpcRequest` / `rpcSubscriptionRequest` (and their matching mappers) are gone from the config, so any existing caller breaks at compile time. Fixed-version lockstep will propagate the bump to the rest of the publishable packages automatically — no need to list them individually.
- **README + example app + JSDoc** all consistently use the new field names and the explicit `withSignal(...).connect()` entry point. The example app's `withSignal(abortController.signal).connect()` line is the one behavioural call-site change worth flagging to next reviewers — previously the abort signal was implicitly carried by the `PendingRpc*Request`; now it has to be threaded explicitly.

## Minor nits (non-blocking)

- The block comment above `const actionStore = initialValueSource.reactiveStore()` (in `create-reactive-store-with-initial-value-and-slot-tracking.ts`, around the new L188–L191) still says “Both stores only ever expose `send()` / `subscribe()`-equivalent work through their reactive contract.” That's a leftover from the prior approach and reads as if the helper is still wrapping `send` / `subscribe` directly. Could be simplified to something like: *“Build fresh inner stores for this connection window and forward their state into the unified store. Drive both with the composed signal so they tear down when this connection window does.”* Inline comment below.
- The cleanup comment `// Note: don't call reset here, causes a race with the abort reason` is slightly cryptic — worth a one-line elaboration so a future reader doesn't try to “fix” it by adding `reset()` calls and break the abort-reason propagation. Inline comment below.
- Test file imports `ReactiveActionSource` / `ReactiveStreamSource` as values rather than `import type`. They're only used in type positions in the test file, so `import type` would be marginally cleaner — but `verbatimModuleSyntax` isn't on, so this is purely stylistic. Non-blocking.

- **Worth thinking about**: there's now no public test that exercises a `PendingRpcRequest` / `PendingRpcSubscriptionsRequest` actually being passed in as the source. The test suite uses synthetic `ReactiveActionSource` / `ReactiveStreamSource` objects. The structural relationship is mechanical (both pending-request types expose `reactiveStore(): ReactiveActionStore<[], T>` / `reactiveStore(): ReactiveStreamStore<T>`), but a one-liner `__typetests__` asserting `PendingRpcRequest<SolanaRpcResponse<T>> satisfies ReactiveActionSource<SolanaRpcResponse<T>>` (and the subscription equivalent) on the config type would lock the duck-type contract in cheaply and survive any future signature drift on `PendingRpc*Request`.
- **Behavioural call-site delta to call out in release notes**: prior callers that relied on the implicit caller-signal propagation through `PendingRpc*Request` now need an explicit `withSignal(signal).connect()`. The example app diff shows exactly this. The changeset entry is clear about the API rename but doesn't explicitly call out the connect-now-required-to-fire bit — might be worth a sentence.

packages/kit/src/create-reactive-store-with-initial-value-and-slot-tracking.ts
mcintyre94: I think the comment this is referring to is: "// Note: don't call reset here, causes a race with the abort reason"

I think this is fine and enough to tell someone not to call `reset`, we don't need to detail how the race happens

lorisleiva (approved):
Thanks for the changes!

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708) couldn't merge this pull request because a downstack PR #1607 failed to merge.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708).
* **Jun 26, 2:00 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708) couldn't merge this pull request because a downstack PR #1628 failed to merge.
* **Jun 26, 2:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708).
* **Jun 26, 3:06 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708) couldn't merge this pull request because a downstack PR #1663 failed to merge.
* **Jun 26, 3:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708).
* **Jun 26, 4:15 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708) rebased this pull request as part of a merge.
* **Jun 26, 4:22 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708) couldn't merge this PR because it was not satisfying all requirements (Failed CI: 'Build Documentation Preview').
* **Jun 26, 6:12 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708).
* **Jun 26, 6:12 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1708).

PR #1709, opened 2026-05-28
Add `@solana/react/swr` subpath with `useRequestSwr`

mcintyre94:
#### Summary of Changes

This PR adds the `@solana/react/swr` subpath, with an optional `swr` dependency.

We add the first hook to this subpath, `useRequestSWR`. This mirrors `useRequest` but provides the shape of `useSWR`:

```tsx
function LatestBlockhash() {
    const client = useClient<ClientWithRpc<GetLatestBlockhashApi>>();
    const { data, error, isLoading, mutate } = useRequestSWR(['latestBlockhash'], client.rpc.getLatestBlockhash());
    if (error) return <button onClick={() => mutate()}>Retry</button>;
    if (isLoading) return <p>Loading…</p>;
    return <p>Blockhash: {data!.value.blockhash}</p>;
}
```

This enables all the niceties of SWR: shared cache (based on key) across components, dedupe, automatic refetching, global config, etc

A third optional param allows passing any SWR configuration (same as `useSWR`), as well as the `getAbortSignal(): AbortSignal` factory supported by `useRequest`. While SWR doesn't have specific abort signal handling, the signal is passed to each fetch. Note that SWR config passed will be merged (by SWR) with any global config in the normal way.

Like `useRequest` it can take either a `ReactiveActionSource` such as `PendingRpcRequest`, or an async function `(signal: AbortSignal) => Promise<T>`. The function version is included for symmetry with `useRequest`, but other than when used with `getAbortSignal` is identical to (and is interoperable with) plain `useSWR`.

Like `useRequest` it does not fetch when the source is null. SWR disables automatically if the key is null, so we disable by setting the key to null. This removes the need to test against null too. 

I've included a note in the readme that we will not have a `useActionSWR` hook. This would provide nothing over `useSWRMutation` and I think it'd be confusing.

Following PRs will add SWR versions of `useSubscription` and `useTrackedData`.


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1709?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1711** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1711?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1710** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1710?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1709** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1709?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1709?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1707** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1707?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1713](https://github.com/anza-xyz/kit/pull/1713) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1713?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/react/src/swr/useRequestSWR.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+
+    const fetcher = useCallback(async (): Promise<T> => {
+        const userSignal = getAbortSignalRef.current?.();
+        const current = sourceRef.current!;
+        if (typeof current === 'function') {
+            // The source's signature requires an `AbortSignal` — hand it the user's signal if
+            // one was configured, otherwise a fresh never-aborted signal so the type fits.
+            return await current(userSignal ?? new AbortController().signal);
+        }
```
trevor-cortex, line 81: Small correctness question on the `sourceRef.current!` non-null assertion.

If `source` transitions `non-null → null` between SWR's decision to fire (which sees the old key + old fetcher closure) and the fetcher's async body actually running (after the layout effect has updated `sourceRef.current` to `null`), the `typeof current === 'function'` branch is skipped and `current.reactiveStore()` throws a TypeError. SWR would surface that as a transient `error`, which is mildly confusing for what should be a clean "disabled" transition.

The window is genuinely narrow — SWR re-evaluates the key on every render so the disable usually wins — but since the assertion is doing real work I'd lean toward an explicit guard instead:

```ts
const current = sourceRef.current;
if (current == null) {
    // Source was nulled out between SWR firing and us reading the ref. SWR will
    // settle into the disabled state on the next render; bail with `undefined` so
    // we don't synthesize a spurious error.
    return undefined as T;
}
```

(or, if you'd rather not lie about the type, throw a sentinel and let SWR drop it — though `undefined as T` matches what SWR's `data` is anyway while loading.)

Not blocking — happy to be told this is impossible in practice.

packages/react/src/swr/useRequestSWR.ts
```diff
@@ -0,0 +1,80 @@
...
+        getAbortSignalRef.current = getAbortSignal;
+    });
+
+    const fetcher = useCallback(async (): Promise<T> => {
+        const userSignal = getAbortSignalRef.current?.();
+        const current = sourceRef.current!;
+        if (typeof current === 'function') {
+            // The source's signature requires an `AbortSignal` — hand it the user's signal if
+            // one was configured, otherwise a fresh never-aborted signal so the type fits.
+            return await current(userSignal ?? new AbortController().signal);
```
trevor-cortex: Two tiny things on the function-source branch:

1. `new AbortController().signal` allocates a controller every fetch just to throw it away. Since this branch only runs when no user signal was supplied and the signal is intentionally never aborted, you could hoist a module-level `const NEVER_ABORTED_SIGNAL = new AbortController().signal;` and reuse it. Micro-optimization, but it also makes the intent more obvious at the call site (`return await current(userSignal ?? NEVER_ABORTED_SIGNAL);`).
2. The comment says "a fresh never-aborted signal so the type fits" — worth calling out explicitly that this is **not** wired to component unmount or SWR supersession (the README does say this, but the code comment could mirror it so anyone reading just this file doesn't have to go check).

packages/react/src/swr/useRequestSWR.ts
```diff
@@ -0,0 +1,80 @@
...
+        }
+        if (userSignal) {
+            return await current.reactiveStore().withSignal(userSignal).dispatchAsync();
+        }
+        return await current.reactiveStore().dispatchAsync();
+    }, []);
+
+    // Force the key to `null` when there's no source to disable
+    return useSWR<T, TError>(source == null ? null : key, fetcher, swrConfig);
```
trevor-cortex, line 90: Minor: the trailing comment `// Force the key to \`null\` when there's no source to disable` reads a little ambiguously ("to disable" can parse as part of the noun phrase). Maybe:

```ts
// When `source` is null, force the SWR key to null so the fetch is disabled
// regardless of whether the caller passed a non-null key.
```

Matches the README phrasing ("Pass `null` for either `key` or `source` to disable") and removes the parse ambiguity.

packages/react/README.md
```diff
@@ -301,6 +301,61 @@ refresh({ abortSignal: undefined }); // no abort signal for this attempt
...
+        getAbortSignal: () => AbortSignal.timeout(5_000),
+    });
+    if (error) return <button onClick={() => mutate()}>Retry</button>;
+    if (isLoading) return <p>Loading…</p>;
+    return <p>{data!.name}</p>;
+}
+```
+
+When `getAbortSignal` isn't configured the signal is a fresh never-aborting `AbortSignal` (so the function's signature is satisfied) — it does **not** fire on unmount or when SWR supersedes the request. SWR's model is to discard the stale result rather than cancel the network call.
+
```
trevor-cortex, line 356: Worth a sentence here on how `getAbortSignal` interacts with `mutate()`. `useRequest.refresh()` has an explicit `{ abortSignal }` per-attempt override; SWR's `mutate()` doesn't take options that thread through to the fetcher, so the only way to influence the signal for a manual revalidation is via the `getAbortSignal` factory itself. Users coming from `useRequest` will likely look for the override and not find it. Something like:

> The factory is invoked on every attempt SWR makes — initial fire, focus/reconnect/poll revalidation, and `mutate()`. SWR's `mutate()` has no per-attempt override; if you need to change the signal for a single fire (e.g. a longer timeout for a manual retry), update the factory closure before calling `mutate()`.

packages/build-scripts/getBaseConfig.ts
```diff
@@ -14,11 +14,15 @@ type Platform =
...
+    // Packages with additional subpath entry points list them here so each subpath gets its own
+    // platform-specific bundle (`dist/<entry>.<platform>.<format>`). Consumers import the subpath
+    // via the package.json `exports` map.
     const moduleEntry =
         env.npm_package_name === '@solana/kit'
             ? ['./src/index.ts', './src/program-client-core.ts']
-            : ['./src/index.ts'];
+            : env.npm_package_name === '@solana/react'
+              ? ['./src/index.ts', './src/swr.ts']
+              : ['./src/index.ts'];
```
trevor-cortex, line 25: Optional: as the per-package entry list grows the nested ternary will get harder to read. A small lookup map keyed on `env.npm_package_name` reads better and scales to the upcoming SWR-adjacent packages (or future subpaths in other packages):

```ts
const SUBPATH_ENTRIES: Record<string, readonly string[]> = {
    '@solana/kit': ['./src/index.ts', './src/program-client-core.ts'],
    '@solana/react': ['./src/index.ts', './src/swr.ts'],
};
const moduleEntry = SUBPATH_ENTRIES[env.npm_package_name ?? ''] ?? ['./src/index.ts'];
```

Non-blocking — happy with the ternary if you'd rather defer until there's a third entry.

packages/react/src/swr/__tests__/useRequestSWR-test.browser.tsx
```diff
@@ -0,0 +1,348 @@
...
+
+        it('surfaces the rejection as `error`', async () => {
+            const boom = new Error('boom');
+            const req = makeFakeSource<string>();
+            const { result } = renderHook(() => useRequestSWR(['source-error'], req.source), { wrapper });
+            await waitFor(() => expect(req.fn).toHaveBeenCalledTimes(1));
+            expect(result.current.error).toBeUndefined();
+
+            await act(async () => req.rejectLatest(boom));
+            await waitFor(() => expect(result.current.error).toBe(boom));
```
trevor-cortex, line 226: Optional: the `as unknown as ReturnType<ReactiveActionSource<string>['reactiveStore']>` casts in these two tests are doing a lot of work to stub just `dispatchAsync` and `withSignal`. If `ReactiveActionStore<[], T>` exports a clean shape, a small typed factory (or a `Partial<...>` cast helper) at the top of this describe block would tighten the noise. Not blocking — these tests are clear enough as-is, just noting in case the same pattern shows up in the follow-up SWR tests.

packages/react/src/swr/__typetests__/useRequestSWR-typetest.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    // ReactiveActionSource — returns SWRResponse with the source's value type.
+    useRequestSWR(['balance'], reactiveSource) satisfies SWRResponse<{ slot: bigint }>;
+
+    // `data` is `T | undefined` until the first successful fetch.
+    const { data } = useRequestSWR(['epochInfo'], fnSource);
+    data satisfies { slot: bigint } | undefined;
+
+    // Null key is accepted (disables the fetch).
+    useRequestSWR(null, fnSource) satisfies SWRResponse<{ slot: bigint }>;
```
trevor-cortex: Nice that this case is covered — when `source` is `null`, `T` can't be inferred so the explicit generic is mandatory. Might be worth a one-line comment noting that ("explicit `T` required when `source` is `null` — no value to infer from") so future maintainers don't see it as a stylistic inconsistency and "fix" it.

trevor-cortex (commented):
## Summary

Adds an opt-in `@solana/react/swr` subpath exporting `useRequestSWR` — the SWR-backed counterpart to `useRequest`. Same source contract (`ReactiveActionSource<T>` or `(signal) => Promise<T>` or `null`), but returns SWR's native `SWRResponse<T>` so users get the shared cache, dedupe, focus/poll revalidation, etc., for free. `swr@^2` is wired up as an optional peer dependency, the build picks up the new entry via `getBaseConfig` + a `./swr` exports map, and the package's `tsconfig.declarations.json` now also emits types for `src/swr.ts`.

The implementation itself is small: split off the Kit-only `getAbortSignal` option, ref-sync `source` + `getAbortSignal` so inline closures don't churn the fetcher, build a stable fetcher that reads the latest source on each fire, and force `key` to `null` when `source` is `null` so the disable semantics match `useRequest`.

Overall this looks great — the surface area mirrors `useRequest` closely, the README is clear about why there's no `useActionSWR`, the test coverage is thorough on both source shapes (success/error/signal-threading/null-key/null-source/key-transitions/`mutate()`-uses-latest), and the SSR test guards against accidentally firing during `renderToString`. The type tests cover the important shapes (default `TError = Error`, override via generic, function vs reactive source, null disable). I have one small correctness question, a docs/symmetry nit, and a couple of optional polish suggestions inline.

## Things to watch out for

- The fetcher reads `sourceRef.current!` with a non-null assertion. SWR is supposed to be disabled when `source == null` (we force the key to `null`), but the ref can transition `non-null → null` between the moment SWR kicks off a fetch and the moment the async fetcher body actually runs. See inline.
- The per-fetch `current.reactiveStore()` call intentionally builds a fresh store on every revalidation — that's the right behavior given SWR owns the caching, but it's a real divergence from `useRequest`, which builds one store per source. Worth keeping in mind when porting the upcoming `useSubscriptionSWR` / `useTrackedDataSWR` hooks.
- `getAbortSignal` semantics around `mutate()` aren't documented. With `useRequest.refresh` there's an explicit per-attempt override (`{ abortSignal }`); with SWR's `mutate()` there isn't one. Worth a sentence in the README so users don't expect `result.mutate({ abortSignal })`.

- Worth eyeballing the `package.json` `exports` map — the previous root-level shape is now nested under `.` and the new `./swr` subpath mirrors it. The `browser` field also gains the two `swr.node.*` → `swr.browser.*` mappings.
- BundleMon: the new entry will register as a new artifact. First-CI baseline is expected.
- This PR sets up the subpath plumbing that `useSubscriptionSWR` and `useTrackedDataSWR` will reuse. The ref-sync pattern + `source == null ? null : key` disable trick are the bits that'll get copied; might be worth a tiny `swr/internal.ts` helper in the follow-up PRs if the pattern repeats verbatim.

packages/react/src/swr/useRequestSWR.ts
mcintyre94: Updated this

packages/react/src/swr/useRequestSWR.ts
mcintyre94: Added `NEVER_ABORTED_SIGNAL` with a comment

packages/react/README.md
mcintyre94: Added a note that `mutate` doesn't allow passing a per-request signal.

packages/build-scripts/getBaseConfig.ts
mcintyre94: Leaving this for now, agree we should do something like that if we add more

packages/react/src/swr/__tests__/useRequestSWR-test.browser.tsx
mcintyre94: Added a helper to stub a partial action store for tests, since this will come up again in the other PRs

mcintyre94:
Graphite has messed this PR up, I'm going to close and recreate it

PR #1713, opened 2026-05-30
Add `@solana/react/swr` subpath with `useRequestSWR`

mcintyre94:
#### Summary of Changes

This PR adds the `@solana/react/swr` subpath, with an optional `swr` dependency.

We add the first hook to this subpath, `useRequestSWR`. This mirrors `useRequest` but provides the shape of `useSWR`:

```tsx
function LatestBlockhash() {
    const client = useClient<ClientWithRpc<GetLatestBlockhashApi>>();
    const { data, error, isLoading, mutate } = useRequestSWR(['latestBlockhash'], client.rpc.getLatestBlockhash());
    if (error) return <button onClick={() => mutate()}>Retry</button>;
    if (isLoading) return <p>Loading…</p>;
    return <p>Blockhash: {data!.value.blockhash}</p>;
}
```

This enables all the niceties of SWR: shared cache (based on key) across components, dedupe, automatic refetching, global config, etc

A third optional param allows passing any SWR configuration (same as `useSWR`), as well as the `getAbortSignal(): AbortSignal` factory supported by `useRequest`. While SWR doesn't have specific abort signal handling, the signal is passed to each fetch. Note that SWR config passed will be merged (by SWR) with any global config in the normal way.

Like `useRequest` it can take either a `ReactiveActionSource` such as `PendingRpcRequest`, or an async function `(signal: AbortSignal) => Promise<T>`. The function version is included for symmetry with `useRequest`, but other than when used with `getAbortSignal` is identical to (and is interoperable with) plain `useSWR`.

Like `useRequest` it does not fetch when the source is null. SWR disables automatically if the key is null, so we disable by setting the key to null. This removes the need to test against null too. 

I've included a note in the readme that we will not have a `useActionSWR` hook. This would provide nothing over `useSWRMutation` and I think it'd be confusing.

Following PRs will add SWR versions of `useSubscription` and `useTrackedData`.

PR #1719, opened 2026-06-03
Add useSubscriptionSWR hook

mcintyre94:
#### Summary of Changes

This PR adds `useSubscriptionSWR`, the SWR shaped version of `useSubscription`.

```tsx
    const client = useClient<ClientWithRpcSubscriptions<AccountNotificationsApi>>();
    const { data, error } = useSubscriptionSWR(
        address ? ['account', address] : null,
        address ? client.rpcSubscriptions.accountNotifications(address) : null,
    );
    if (error) return <p>Failed to connect.</p>;
    if (!data) return <p>Connecting…</p>;
    return (
        <p>
            {data.value.lamports} lamports at slot {data.context.slot}
        </p>
    );
}
```

Similarly to `useRequestSWR` the third param is an `SWRConfiguration`, with our usual `getAbortSignal` factory to add a custom signal. It also disables when either key or source are null. Note that SWR subscriptions don't include the `reconnect` function from `useSubscription`. 

Note that under the hood this uses `useSWRSubscription`, which handles cleaning up the subscription and resetting the store.


mcintyre94:
* **#1786** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1786?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1782** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1782?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1780** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1780?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1779** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1779?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1769** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1769?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1760** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1760?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1759** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1759?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1758** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1758?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1754** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1754?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1747** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1747?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1727** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1727?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1719** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1719?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1719?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1713** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1713?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1707** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1707?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1709](https://github.com/anza-xyz/kit/pull/1709) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1709?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/react/src/swr/useSubscriptionSWR.ts
```diff
@@ -0,0 +1,77 @@
...
+    const getAbortSignalRef = useRef(getAbortSignal);
+    useIsomorphicLayoutEffect(() => {
+        sourceRef.current = source;
+        getAbortSignalRef.current = getAbortSignal;
+    });
+
+    const subscribe = useCallback((_key: SWRKey, { next }: SWRSubscriptionOptions<T, TError>) => {
+        const current = sourceRef.current!;
+        const store = current.reactiveStore();
+        const unsubscribe = store.subscribe(() => {
```
trevor-cortex: Minor robustness nit: `sourceRef.current!` uses a non-null assertion that's safe *today* because the SWR key is forced to `null` when `source == null` (so `subscribe` only fires when source was non-null at render time, and the layout effect keeps the ref in sync). But it's a forwards-fragile invariant — if the key-gating logic on line 76 is ever changed, this would blow up silently. Consider a defensive early-return:

```ts
const current = sourceRef.current;
if (!current) return () => {};
```

Not blocking — just defensive.

packages/react/README.md
```diff
@@ -354,6 +354,27 @@ function Profile({ userId }: { userId: string }) {
 
 When `getAbortSignal` isn't configured the signal is a fresh never-aborting `AbortSignal` (so the function's signature is satisfied) — it does **not** fire on unmount or when SWR supersedes the request. SWR's model is to discard the stale result rather than cancel the network call.
 
+### `useSubscriptionSWR(key, source, options?)`
+
+SWR-backed counterpart to `useSubscription`. Routes a `ReactiveStreamSource<T>` through SWR's subscription cache (`useSWRSubscription`). Returns SWR's native `{ data, error }` shape — `data` is the notification exactly as the source emits it. Pass `null` for either `key` or `source` to disable. Options accept SWR's config plus `getAbortSignal` for per-connection signals.
```
trevor-cortex: Two small doc points worth considering:

1. **"per-connection signals"** is a bit misleading here. Unlike `useSubscription`, there's no `reconnect()`, so `getAbortSignal` is invoked once per `subscribe` call (i.e. once per unique key, deduped across components), not once per connection attempt. Maybe "for an abort signal scoped to the subscription's lifetime" or similar.

2. It would be worth a sentence noting that — unlike `useSubscription` — there's no `reconnect()` affordance; the only way to re-open is to change the SWR key. Users picking between the two hooks will hit this.

Neither is blocking, but the second point in particular is a real behavioural difference that's easy to miss.

packages/react/src/swr/useSubscriptionSWR.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    // Ref-sync the source and the abort-signal factory so an inline value passed each render
+    // doesn't change the `subscribe` callback's identity. `subscribe` reads the latest values
+    // from the refs when SWR invokes it.
+    const sourceRef = useRef(source);
+    const getAbortSignalRef = useRef(getAbortSignal);
+    useIsomorphicLayoutEffect(() => {
+        sourceRef.current = source;
+        getAbortSignalRef.current = getAbortSignal;
+    });
```
trevor-cortex: Optional: consider seeding from `store.getUnifiedState()` immediately after subscribing, in case the store is ever in a non-`idle` state at the moment we attach (e.g. a future change where stores can be pre-warmed via a cache). Today every `source.reactiveStore()` returns an `idle` store so this is a no-op, but it would make the bridge resilient to that assumption changing. Not blocking.

trevor-cortex (commented):
## Summary

Adds `useSubscriptionSWR` — the SWR-backed counterpart to `useSubscription`, mirroring the shape and conventions of `useRequestSWR`. It wraps `useSWRSubscription`, accepts a `ReactiveStreamSource<T>` and SWR key (either nullable to disable), splits off the Kit-specific `getAbortSignal` factory from the rest of the SWR config, ref-syncs source/factory so inline values don't churn the `subscribe` callback's identity, and tears down the underlying store on cleanup via `unsubscribe()` + `store.reset()`. Multiple components with the same key share one underlying subscription (SWR dedupes), and either-null gating works by forcing the SWR key to null when `source` is null. Ships with a thorough browser test suite (envelope vs raw notifications, error channel, both-null gating, transitions in/out of null, key changes, dedup, abort-signal forwarding, SSR), a typetest covering envelope/raw shapes, default-`unknown` error type and overrides, options forwarding, and a minor changeset.

## Things to watch out for

- **No `reconnect()` affordance.** `useSubscription` exposes a `reconnect()` callback as part of its result, but `SWRSubscriptionResponse` doesn't have an equivalent and there's no way to force `useSWRSubscription` to re-open a stream while the key is stable (SWR's `mutate` doesn't reopen subscriptions). In practice the only way to reconnect is to change the SWR key. This is an inherent limitation of `useSWRSubscription`, not a bug, but it's a meaningful semantic difference from `useSubscription` that the README doesn't currently call out — worth a sentence so users picking between the two hooks aren't surprised.
- **"Per-connection signals" wording in the README is slightly off.** Because there's no `reconnect()`, `getAbortSignal` is invoked once per `subscribe` invocation (i.e. once per unique key, deduped across components), not once per connection in the `useSubscription` sense. Minor doc nit.
- **Bump level.** `minor` for a new public hook is correct per the repo's `CLAUDE.md` (pre-1.0, new feature).

- The `subscribe` callback subscribes to the store *before* calling `connect()`, which is the right order to avoid missing synchronous state transitions. The listener only forwards `loaded`/`error` states to `next`; `loading`/`idle` are correctly ignored so SWR's `data` retains its prior value across reconnect-style transitions (stale-while-revalidate is preserved at the kit-store layer, even though `useSWRSubscription` itself doesn't have that concept).
- The shared-subscription test (`'shares one active underlying subscription across components with the same key'`) relies on SWR's per-key dedup: only one `subscribe` invocation fires per key, so `sourceRef.current` from whichever render wins is what gets used. Worth being aware of if two callers ever pass *different* sources under the same key — the later mount won't re-subscribe.
- The `useCallback(subscribe, [])` + ref-sync pattern matches `useRequestSWR` exactly. Good consistency.
- Cleanup calls both `unsubscribe()` (removes our listener) and `store.reset()` (aborts the connection). Since the store is created locally per `subscribe` invocation and has no other listeners, this is the correct teardown.
- The abort-signal test is genuinely exercising the kit store's abort handling (not the fake publisher's listener removal) — the store transitions to `error` with the abort reason, which the subscribe listener then forwards via `next(reason)`. Good coverage.

packages/react/README.md
mcintyre94: Updated comment/readme

packages/react/src/swr/useSubscriptionSWR.ts
mcintyre94: Agree with this, and I think this is quite important because a store defined by a plugin can have whatever behaviour that plugin wants. This makes the hook a bit more complex, but is worth it IMO.

Note that other hooks don't have this issue because `useSyncExternalStore` handles the initial state.

mcintyre94: Actually this is wrong and I've reverted the change. `.reactiveStore()` is documented as returning an idle store, and this is an important part of the design that we settled on. So `useSubscriptionSWR` can trust the state to start idle and only change after it has subscribed and connected.

lorisleiva (approved):
Thanks!

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719) couldn't merge this pull request because a downstack PR #1607 failed to merge.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719).
* **Jun 26, 2:00 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719) couldn't merge this pull request because a downstack PR #1628 failed to merge.
* **Jun 26, 2:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719).
* **Jun 26, 3:06 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719) couldn't merge this pull request because a downstack PR #1663 failed to merge.
* **Jun 26, 3:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719).
* **Jun 26, 4:22 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719) couldn't merge this pull request because a downstack PR #1708 failed to merge.
* **Jun 26, 6:12 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719).
* **Jun 26, 6:50 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719) rebased this pull request as part of a merge.
* **Jun 26, 7:00 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1719).

PR #1723, opened 2026-06-03
Add configurable max instructions per transaction to transaction planner

mcintyre94:
#### Problem

Transactions are limited to 64 instructions, including inner instructions. We enforce the limit of 64 top-level instructions already, but we can't know how many CPIs/inner instructions there will be when the instructions are executed. With v1 transactions we can sometimes pack many more instructions into a transaction, increasing the probability of exceeding 64 total instructions including CPIs.

#### Summary of Changes

This PR changes transaction planning to by default pack only up to 16 instructions per transaction. This is based on an average of 3 inner instructions per instruction, for a total of 64. If you know that the instructions in your transaction use more or less CPIs than this default, then you can configure the max instructions allowed per transaction. 

This can be over-ridden in two ways:

- `createTransactionPlanner` now takes a `maxInstructionsPerTransaction` input, which sets the default for all plans created by the resulting planner:

```ts
const transactionPlanner = createTransactionPlanner({
    createTransactionMessage,
    maxInstructionsPerTransaction: 32,
});
```

- It can also be over-ridden per call to the planner:

```ts
const transactionPlan = await transactionPlanner(instructionPlan, {
    maxInstructionsPerTransaction: 8,
});
```

The hard limit of 64 top-level instructions is still enforced, and we throw if `maxInstructionsPerTransaction` is greater than 64. 

Note that the changeset is major because this would be a breaking change for apps that are currently planning transactions from 17+ instructions, which will now by default receive a plan with multiple transactions instead of one. They will need to set `maxInstructionsPerTransaction` to an appropriate value.

To reduce churn considering the major changesets in the react PRs, I've made this a draft PR and suggest releasing it alongside that stack.


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Introduces a configurable `maxInstructionsPerTransaction` ceiling on the transaction planner (and a matching `maxInstructions` option on `MessagePacker.packMessageToCapacity`), defaulting to **16** instead of the hard 64-instruction Solana limit. The intent is to leave headroom for CPIs, since Kit can't see inner instructions at planning time and v1 transactions make it easier to pack many top-level instructions. Overrides are plumbed in three places: planner construction, individual planner calls, and direct calls to the message packer.

Implementation looks clean and self-consistent: the same `??` chain (`per-call ?? planner-config ?? DEFAULT`) is used for resolution, and `assertMaxInstructionsPerTransaction` is added at all the places where a message gets newly populated or extended in `transaction-planner.ts`. The `getMessagePackerInstructionPlanFromInstructions` rewrite to use a `nextMessage` shadow before committing is correct and avoids mutating `message` past the limit. The new test coverage is solid for the divisible-sequential happy paths, the `createTransactionMessage` and `onTransactionMessageUpdated` accounting, and the message packer integration.

### 1. The 64-instruction hard cap is enforced indirectly

The PR description and the changeset both say “passing `maxInstructionsPerTransaction` greater than 64 is equivalent to passing 64”, and the test with `maxInstructionsPerTransaction: 65` does split at 64. But that behavior comes from the underlying transaction encoder (`getTransactionMessageSize`) throwing `TOO_MANY_INSTRUCTIONS` at 65 instructions, which `selectAndMutateCandidate` happens to catch — nothing in this PR actually clamps the user-provided value to 64. Worth either:

- explicitly clamping with `Math.min(maxInstructionsPerTransaction, 64)` at resolution time in `createTransactionPlanner` (clearer to readers, decouples this contract from the encoder's internals), or
- adding a comment at the resolution site documenting that the hard cap is enforced downstream.

The relevant code is the `??` chain in `createTransactionPlanner` (around the new `maxInstructionsPerTransaction:` line in `transaction-planner.ts`).

### 2. `0` and negative values aren't guarded

The `??` chain treats `0` as a valid override, which would make planning impossible (every message would already be at/above the limit). Likely fine to leave as undefined behavior, but a positive-integer guard or a JSDoc note that the value must be `> 0` would be a low-cost improvement.

### 3. Naming inconsistency: `maxInstructionsPerTransaction` vs `maxInstructions`

The planner uses `maxInstructionsPerTransaction` and the `MessagePacker.packMessageToCapacity` config uses `maxInstructions`. I get why — a message packer doesn't really think in terms of transactions — but having two names for the same conceptual knob is awkward, especially when the packer is the lower-level primitive the planner forwards to. Not a blocker; worth being deliberate before this ships as a major.

### 4. Error code reuse may obscure the source of the limit

`SOLANA_ERROR__TRANSACTION__TOO_MANY_INSTRUCTIONS` is now thrown for what is effectively a configurable soft limit. When a user sees `maxAllowed: 16` in production, it'll look like Solana's hard limit — which could be confusing to debug (“wait, Solana allows 64, why is this saying 16?”). Consider whether a distinct error code (e.g. `SOLANA_ERROR__INSTRUCTION_PLANS__MAX_INSTRUCTIONS_PER_TRANSACTION_EXCEEDED`) would tell a better story.

There's a real tradeoff here: reusing the existing code means `selectAndMutateCandidate`'s existing `isSolanaError` catch chain Just Works. If a new code is introduced, it would need to be added to that list (and to anything else that swallows `TOO_MANY_INSTRUCTIONS` during planning).

### 5. README not updated

Per the package guidance in `CLAUDE.md` (“When adding a new public API, add or update the package's README”), the new `maxInstructionsPerTransaction` / `maxInstructions` options should probably get a mention in the `@solana/instruction-plans` README — they're a meaningful new piece of the public surface.

### 6. JSDoc nits on the new `MessagePacker.packMessageToCapacity` config

- The new `@throws {@link SOLANA_ERROR__TRANSACTION__TOO_MANY_INSTRUCTIONS}` reads “if the provided transaction message cannot fit more instructions.” Worth being explicit that this means the configured `maxInstructions` ceiling has been reached (not the byte-size limit), e.g. “if the provided transaction message has already reached `maxInstructions` instructions.”
- `Defaults to \`16\`` (in both the `MessagePacker` config docblock and the new `maxInstructionsPerTransaction` docblocks in `transaction-planner.ts`) is a magic number that will silently drift if `DEFAULT_MAX_INSTRUCTIONS_PER_TRANSACTION` ever changes. Consider `{@link DEFAULT_MAX_INSTRUCTIONS_PER_TRANSACTION}` or just referencing the constant by name.

### 7. Three near-duplicate “too many instructions” throw sites

The same `SOLANA_ERROR__TRANSACTION__TOO_MANY_INSTRUCTIONS` is now constructed in three places with subtly different context shapes:

- `assertMaxInstructionsPerTransaction` in `transaction-planner.ts` — `actualCount: message.instructions.length` (no `+1`)
- `assertCanFitAdditionalInstruction` in `instruction-plan.ts` — `actualCount: message.instructions.length + 1`
- The inline throw inside `getMessagePackerInstructionPlanFromInstructions` — `actualCount: nextMessage.instructions.length`

Not strictly wrong (each is reporting from a slightly different vantage point), but they're easy to drift out of sync. Might be worth consolidating into a single shared helper — also helps if you ever swap the error code per point 4.

**Coverage gaps worth filling in.** All the new transaction-planner tests are divisible-sequential. The max-instructions limit also flows through:

- `traverseParallel` — a parallel plan whose siblings collectively exceed the limit should split into separate transactions rather than re-using a candidate. Not exercised today.
- `fitEntirePlanInsideMessage` for the non-divisible-sequential-inside-parallel case — this is the path that calls `assertMaxInstructionsPerTransaction` after appending each instruction; it's only exercised indirectly.
- A message-packer plan nested inside a `parallelInstructionPlan` (or non-divisible sequential), which exercises the `fitEntirePlanInsideMessage` → `packMessageToCapacity` branch in `instruction-plan.ts`.
- `getLinearMessagePackerInstructionPlan` — the change adds `assertCanFitAdditionalInstruction` but there's no test covering the case where the incoming message already has `>= maxInstructions`.

**On the existing test at `transaction-planner-test.ts` for the 65-instructions case** — the test name (“does not let max instruction overrides exceed the hard 64-instruction transaction limit”) implies the planner is doing the clamping, but as noted in point 1, the planner isn't — the encoder is. If you add the explicit clamp, this becomes a real unit test of that logic; otherwise consider renaming to something like “still respects the encoder's 64-instruction limit when overrides exceed it.”

**Major bump rationale.** The changeset correctly calls out that apps currently planning 17–64-instruction transactions will start getting split plans. The migration snippet (`maxInstructionsPerTransaction: 64`) is the right escape hatch — nice that it's spelled out.

mcintyre94:
> The 64-instruction hard cap is enforced indirectly

This is now clamped to 64, by calling a new function `getTransactionMessageInstructionCountLimit` in `@solana/transactions`. This mirrors `getTransactionMessageSizeLimit`, and takes a `TransactionMessage` as input. Currently the message is ignored and it always returns 64. This is under the assumption that a future version may change the limit, and avoids a breaking change if this happens.

> 0 and negative values aren't guarded

Updated the comments here. We typically don't add runtime checks for these 

> Naming inconsistency: maxInstructionsPerTransaction vs maxInstructions

This is intentional - `maxInstructionsPerTransaction` is used for the planner because it handles multiple transactions, `maxInstructions` for the packer because it's packing only one

> Error code reuse may obscure the source of the limit

Agree with this, it now uses a new error code `SOLANA_ERROR__INSTRUCTION_PLANS__MAX_INSTRUCTIONS_PER_TRANSACTION_EXCEEDED`

> README not updated

Agree with this, but the readme is currently just pointing to the docs. We probably should expand it, but it doesn't make sense to pull writing the whole readme into this PR. 

> JSDoc nits on the new MessagePacker.packMessageToCapacity config

Updated

> Three near-duplicate “too many instructions” throw sites

Agree with this, consolidated to one `assertMaxInstructionsPerTransaction(numInstructions: number, maxInstructions: number)` function.

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary of changes since the last review

The revision addresses essentially everything from the prior pass:

- **Explicit hard-cap clamping.** `resolveMaxInstructions` now does `Math.min(configured, getTransactionMessageInstructionCountLimit(message))`, replacing the previous reliance on the encoder throwing.
- **Dedicated error code.** `SOLANA_ERROR__INSTRUCTION_PLANS__MAX_INSTRUCTIONS_PER_TRANSACTION_EXCEEDED` (7618010) is introduced and threaded through the `selectAndMutateCandidate` catch list, so the soft limit is now clearly distinguishable from the format-enforced 64 in errors. Nice.
- **Consolidated assertions.** New `max-instructions.ts` centralizes `resolveMaxInstructions` and `assertMaxInstructionsPerTransaction`; the three previously-duplicated throw sites are gone.
- **Versioned hard limit.** `getTransactionMessageInstructionCountLimit(message)` is a forward-looking single source of truth that takes the message — leaves room for a per-version limit later without a breaking change. `LEGACY_TRANSACTION_INSTRUCTIONS_LIMIT` joins the size constants in the renamed `transaction-limits.ts`.
- **Coverage filled in.** New tests cover the `maxInstructionsPerTransaction: 65` clamp, the 16-default split, both override surfaces (config + per-call), counting against `createTransactionMessage`-provided instructions, counting against `onTransactionMessageUpdated` instructions, the message-packer-inside-planner path, and the "fresh message already over the limit" rejection. The `getMessagePackerInstructionPlanFromInstructions` and `getLinearMessagePackerInstructionPlan` already-full-message cases are also covered.
- **JSDoc.** Each `maxInstructionsPerTransaction` site now references `DEFAULT_MAX_INSTRUCTIONS_PER_TRANSACTION` rather than hard-coding `16`, the new error is documented on `packMessageToCapacity`, and `getTransactionMessageInstructionCountLimit` carries a good `@example` + `@see`.
- **Changeset.** Major for `instruction-plans`, minor for `transactions` (new public `getTransactionMessageInstructionCountLimit`), patch for `errors` — looks right. The breaking-change snippet is spelled out and the migration is unambiguous.

Net result: the contract is cleaner, the error model is more honest, and the test surface matches the new behavior. Implementation looks good to me — would approve if I had write access; flagging COMMENT in lieu.

## Small notes (non-blocking)

### 1. Double assertion after `packMessageToCapacity`

In `transaction-planner.ts`, `fitEntirePlanInsideMessage`'s `messagePacker` branch calls `packMessageToCapacity` (which now asserts internally and only returns a within-limit message) and then re-asserts the same condition on the returned message. The same harmless redundancy exists in `createNewMessage` / `selectAndMutateCandidate` when the predicate is `packMessageToCapacity`. Not a bug — purely defensive — but worth a comment noting the assertion is belt-and-braces, or trusting the packer's contract and dropping the outer assert. Leaning toward dropping it.

### 2. Resolution chain re-resolves per call site

`resolveMaxInstructions(context.maxInstructionsPerTransaction, message)` is invoked at every assertion site (`selectAndMutateCandidate`, `createNewMessage`, `fitEntirePlanInsideMessage`, and inside both message packers). That's fine while `getTransactionMessageInstructionCountLimit` is `O(1)` and message-independent, but it's a tiny footgun the day the limit actually becomes version-dependent and someone wants to cache it. Not worth changing now — flagging only because it's the natural follow-up when the version-dependent limit lands.

### 3. README is a 3-line pointer to solanakit.com

I raised this last time per the skills-inject `TypeScript READMEs` rule. Looking now, the `@solana/instruction-plans` README has no API surface at all — it's just a link to the official docs. Adding usage for `maxInstructionsPerTransaction` here would be inconsistent with the rest of the file. The right place is the external docs site; treating this as out-of-scope for the PR seems fine. Withdrawing the prior nit.

### 4. `transaction-limits.ts` file header comment

Minor: the new `transaction-limits.ts` file-level comment still says "used by both transaction-size and transaction-message-limits" but now also includes `LEGACY_TRANSACTION_INSTRUCTIONS_LIMIT`, which is additionally consumed by `@solana/instruction-plans` via the public `getTransactionMessageInstructionCountLimit`. The "intentionally not exported" note is about the raw constants — which is still true — so the comment is technically accurate, just slightly out of date in feel. Optional.

- **Bump levels** look right: `instruction-plans` major (default behavior change), `transactions` minor (new public function), `errors` patch (additive code). The fixed-version policy means the `transactions` minor will pull everything up regardless, but the per-package bumps are still individually correct.
- The new `getTransactionMessageInstructionCountLimit` function is a small but deliberate piece of public API. Worth confirming the signature (`(message: TransactionMessage) => number`) is what we want long-term — it's the version-aware-but-not-yet-version-dependent shape, which seems right.
- The `selectAndMutateCandidate` catch list now swallows the new soft-limit error during candidate search, which is what enables the "move on to the next candidate / open a new transaction" behavior. Worth double-checking this matches the intent for the parallel/message-packer paths in particular: a parallel sibling that overflows the per-transaction instruction count should *open a new transaction*, not silently get dropped. The new "uses the max instruction limit when packing message packer plans" test confirms this for the message-packer path; the parallel-with-overflow path is covered indirectly by the existing parallel tests but isn't called out as a dedicated instruction-count test.

mcintyre94:
> Double assertion after packMessageToCapacity

This is true here, but not for `createNewMessage` / `selectAndMutateCandidate`. Those are asserting after ` context.onTransactionMessageUpdated`, which can add arbitrary instructions after the packer. We can't rely on the packer's assertion here. It is true that we don't need the assertion in `fitEntirePlanInsideMessage` but I think it's clearer to keep it.

> Resolution chain re-resolves per call site

This is intentional for now, we might want to adjust slightly if max instructions is ever not a constant. I'm deliberately avoiding hardcoding the count without it being derived from a message.

> transaction-limits.ts file header comment

Updated this comment a bit

> The selectAndMutateCandidate catch list now swallows the new soft-limit error during candidate search, [...] a parallel sibling that overflows the per-transaction instruction count should open a new transaction, not silently get dropped.

Added a new test for this ("opens a new transaction for a parallel sibling that overflows the max instruction limit")

mcintyre94:
I've simplified this PR by dropping the changes to `@solana/transactions`, which included a speculative function to get the max instructions for a transaction message (currently always 64, speculatively future versions might increase it).

It now just hardcodes a check against 64, and we can revisit that (as we did with transaction size) if a new version increases it.

The check against 64 isn't necessarily required, the transaction planner handles the limit of 64 instructions internally anyway. But the message packer path doesn't currently need to do that, and is kept simpler by just checking `maxInstructions`.

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary of changes since the last review

The contract has shifted in a meaningful way: instead of clamping `maxInstructionsPerTransaction > 64` to 64, the planner now **throws** `SOLANA_ERROR__INSTRUCTION_PLANS__INVALID_MAX_INSTRUCTIONS_PER_TRANSACTION` up front. This is a stronger and clearer signal — a configured maximum the format could never satisfy is a programmer error, not something to silently coerce — and it matches the updated PR description (“we throw if `maxInstructionsPerTransaction` is greater than 64”).

Mechanically:

- **New helper module** `max-instructions.ts` centralizes `DEFAULT_MAX_INSTRUCTIONS_PER_TRANSACTION` (16), the private `TRANSACTION_INSTRUCTION_LIMIT` (64), `resolveMaxInstructions`, `assertValidMaxInstructionsPerTransaction` (the new up-front check), and `assertMaxInstructionsPerTransaction` (the per-message check). Clean, small, focused.
- **`createTransactionPlanner`** validates the resolved `maxInstructionsPerTransaction` once at planner entry; per-call overrides override planner config; both flow through the same `??` chain.
- **`MessagePacker.packMessageToCapacity`** now takes `{ maxInstructions }` and validates it inside each call (both `getLinearMessagePackerInstructionPlan` and `getMessagePackerInstructionPlanFromInstructions`).
- **Two new error codes**: `INVALID_MAX_INSTRUCTIONS_PER_TRANSACTION` (7618011) for the misconfiguration, `MAX_INSTRUCTIONS_PER_TRANSACTION_EXCEEDED` (7618010) for the runtime soft-limit hit. The latter is added to `selectAndMutateCandidate`'s catch list, so an overflowing parallel sibling correctly rolls over into a new transaction rather than getting silently dropped.
- **`getMessagePackerInstructionPlanFromInstructions`** rewritten to check `message.instructions.length >= maxInstructions` *before* appending, with a precondition assert on entry. Avoids ever compiling a message that exceeds the limit. Comments explain the shape well.
- **Test coverage** now covers: the new invalid-max-up-front throw (planner config + per-call + message-packer plans), the 16-default split, both override surfaces, instructions added by `createTransactionMessage`, instructions added by `onTransactionMessageUpdated`, message-packer-in-planner, packer-cap-at-64, the parallel-overflow path opening a new transaction (the dedicated test I asked for last time), and the “already-full message” rejection in both packers. Comprehensive.

## Notes on the contract change

For what it's worth, I think throwing on `> 64` is the better choice than clamping. It makes the misconfiguration loud, is consistent with how Kit treats other format constraints (compile-time-knowable errors should fail loudly), and the error message (`Provide a maxInstructionsPerTransaction value no greater than 64`) is actionable. The fact that it's validated once up front in `createTransactionPlanner` — before any traversal — is exactly right.

Dropping the public `getTransactionMessageInstructionCountLimit` function and the `transaction-limits.ts` plumbing also simplifies the surface area, at the cost of needing to update `TRANSACTION_INSTRUCTION_LIMIT` in two places (`@solana/transactions` already has its own internal limit constant) if the format ever raises the cap. The docblock explicitly calls this trade-off out (“intentionally duplicated here”), which is the right thing to do.

## Small notes (non-blocking)

### 1. `DEFAULT_MAX_INSTRUCTIONS_PER_TRANSACTION` is referenced from public JSDoc but not exported

`max-instructions.ts` is not in `index.ts`, so the constant is private. The JSDoc on both `TransactionPlanner` (per-call config), `TransactionPlannerConfig`, `createTransactionPlanner`, and `MessagePacker.packMessageToCapacity` all reference `\`DEFAULT_MAX_INSTRUCTIONS_PER_TRANSACTION\`` by name (and one says `{@link DEFAULT_MAX_INSTRUCTIONS_PER_TRANSACTION}` would resolve if it were public, though I see the current text just uses the backtick form). Two options:

- **Export it** from `index.ts` so consumers can reference the same default in their own code (e.g. “give me default+8”), and so the JSDoc `{@link …}` resolves. This is consistent with how the package exposes other intentional public constants.
- **Drop the constant name from public docs** and just say “defaults to 16 (subject to change in future versions)”, keeping the constant private.

My preference is the first — the value is useful for app code that wants to reason about Kit's default. Either is defensible; the current half-state (referenced but unreachable) is the awkward one.

### 2. `resolveMaxInstructions` accepts `undefined` but `assertValidMaxInstructionsPerTransaction` also does

Minor: `resolveMaxInstructions(undefined)` returns the default, and `assertValidMaxInstructionsPerTransaction(undefined)` is a no-op. That's a deliberate pair, but `assertMaxInstructionsPerTransaction` (no `Valid`) takes a number, not undefined — and is always called with a *resolved* value. The naming is fine, but a one-line note on `resolveMaxInstructions` saying “the returned value is always defined” (or typing it explicitly with a non-null return) would make the relationship between the three helpers slightly more obvious.

Also: `resolveMaxInstructions` no longer takes a `message` parameter (previous revision passed `message` to `getTransactionMessageInstructionCountLimit(message)`). That's correct now — the limit no longer depends on the message — but if a future version-dependent limit lands, this signature will need to grow back. Not actionable today.

### 3. Zero / negative values still aren't rejected

`assertValidMaxInstructionsPerTransaction` only checks `> TRANSACTION_INSTRUCTION_LIMIT`. `0`, negatives, and non-integers (`1.5`) are accepted. The runtime behavior with `0` is “throw on the very first instruction added”, which is at least observable, but the failure mode is `MAX_INSTRUCTIONS_PER_TRANSACTION_EXCEEDED` from the first instruction — not `INVALID_MAX_INSTRUCTIONS_PER_TRANSACTION`, which is what you'd expect for a misconfiguration.

Given we're now in the business of validating this value up front anyway, extending the check to `maxInstructions <= 0 || !Number.isInteger(maxInstructions) || maxInstructions > TRANSACTION_INSTRUCTION_LIMIT` would be a low-cost completeness win. The error message would need a small tweak (“must be a positive integer no greater than 64”). Up to you — also fine to defer.

### 4. Double assertion in `fitEntirePlanInsideMessage`'s `messagePacker` branch

In `transaction-planner.ts`'s `fitEntirePlanInsideMessage`, the `messagePacker` case calls `packMessageToCapacity` (which now asserts internally on entry) and then re-asserts `assertMaxInstructionsPerTransaction(newMessage.instructions.length, …)` on the returned message. Belt-and-braces — the packer's contract should guarantee the post-condition. Same as I flagged last time; just noting it's still there. Easy to drop one of the two, or leave a `// defensive` comment.

Same pattern exists in `createNewMessage` and `selectAndMutateCandidate` when the predicate is `packMessageToCapacity`, but those are more defensible because the predicate is opaque to the caller.

### 5. Per-call assertion site count

`resolveMaxInstructions(context.maxInstructionsPerTransaction)` is called at every assertion site (`selectAndMutateCandidate`, `createNewMessage`, `fitEntirePlanInsideMessage`'s `single` and `messagePacker` branches, and inside both message packers). Fine while it's `O(1)`; mild duplication. Could be resolved once into `context.maxInstructionsPerTransaction` at `createTransactionPlanner` entry and stored as a definite `number` on the context, eliminating the runtime `??` chains. Not worth changing now — noting for follow-up.

- **Bump levels** are now `instruction-plans: major` and `errors: patch`. The previous `transactions: minor` is gone since `getTransactionMessageInstructionCountLimit` was removed. Correct.
- **Behavior change for direct `MessagePacker.packMessageToCapacity` callers**: the changeset captures this in the breaking-changes section with the `{ maxInstructions: 64 }` escape hatch. Worth double-checking that all in-repo callers of `packMessageToCapacity` either pass a max or are happy with the 16 default — the planner forwards `context.maxInstructionsPerTransaction` so that's covered.
- **The `traverseParallel` overflow test** (`opens a new transaction for a parallel sibling that overflows the max instruction limit`) is the one I asked for last time — it does verify that an overflowing sibling rolls into a new transaction rather than getting silently dropped, which is the load-bearing behavior of adding `MAX_INSTRUCTIONS_PER_TRANSACTION_EXCEEDED` to the `selectAndMutateCandidate` catch list. Good coverage.
- **`max-instructions.ts` is internal.** If a downstream package ever needs to reuse `assertMaxInstructionsPerTransaction` or `resolveMaxInstructions` (e.g. an external planner implementation), they're not exported. Probably fine for now; flagging in case the surface needs to grow.

Implementation still looks good to me — would approve if I had write access. The pivot to throw-instead-of-clamp is the right call.

mcintyre94:
> DEFAULT_MAX_INSTRUCTIONS_PER_TRANSACTION is referenced from public JSDoc but not exported

Agree, just changed these to say 16 

> resolveMaxInstructions accepts undefined but assertValidMaxInstructionsPerTransaction also does [...] The naming is fine, but a one-line note on resolveMaxInstructions saying “the returned value is always defined” (or typing it explicitly with a non-null return)

It's typed as returning `number`, I think this is fine

> Zero / negative values still aren't rejected

Agreed, since we're validating for <= 64 now. Added validation for positive integer only.

> Double assertion in fitEntirePlanInsideMessage's messagePacker branch

As above, intentional because it's required in the cases that use `onTransactionMessageUpdated`

> Per-call assertion site count

Intentional for now

> Behavior change for direct MessagePacker.packMessageToCapacity callers

All callers are in `instruction-plans` package, I think this is fine

lorisleiva (approved):
Amazing, thank you!

PR #1730, opened 2026-06-06
Remove redundant bigint downcast from the default Solana RPC request transformer

### LINKED ISSUE #366: Investigate whether we can remove the `downcastNodeToNumberIfBigint` _request_ transform

plutohan:
As #366 asked, I checked what Agave's RPC does with a JSON number above Number.MAX_SAFE_INTEGER. The downcast turns out to be unnecessary: the transport already serializes bigints losslessly via stringifyJsonWithBigInts, and a local solana-test-validator parses the full u64 range without loss — 2^53 and 2^53+1 echo back distinct (no precision loss), and only values past u64::MAX are rejected, cleanly:

| slot sent | Agave response |
| --- | --- |
| 9007199254740992 (2^53) | Block not available for slot 9007199254740992 |
| 9007199254740993 (2^53 + 1) | Block not available for slot 9007199254740993 |
| 18446744073709551616 (u64::MAX + 1) | Invalid params: invalid type: floating point, expected u64 |

So this removes getBigIntDowncastRequestTransformer from the default request transformer and adds a regression test; the export is deprecated rather than deleted to stay non-breaking.

Fixes #366

### DIFF
--- a/.changeset/cyan-phones-joke.md
+++ b/.changeset/cyan-phones-joke.md
@@ -1,0 +1,7 @@
+---
+'@solana/rpc-transformers': patch
+---
+
+Stop downcasting `bigint` request params to `number` in the default Solana RPC request transformer
+
+`getDefaultRequestTransformerForSolanaRpc` no longer runs `getBigIntDowncastRequestTransformer`. The Solana RPC transport already serializes `bigint` values losslessly as large integer literals (via `stringifyJsonWithBigInts`), and Agave parses JSON integers across the full `u64` range without precision loss, so the lossy `bigint`->`number` downcast was redundant. Removing it also fixes silent truncation for RPC APIs configured without an `onIntegerOverflow` handler. `getBigIntDowncastRequestTransformer` is now deprecated and slated for removal in a future major version.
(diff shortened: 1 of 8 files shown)


mcintyre94 (dismissed):
Looks good, thankyou!

mcintyre94:
It looks like the test errors here are legitimate, we have some tests passing u64 MAX and they now get the correct number back instead of the previous hardcoded cap.

plutohan:
@mcintyre94 Good catch, thanks! Pushed a fix. Those two messages echoed 9223372036854776000 because the old downcast rounded 2^63-1 to a double and with it gone the exact 9223372036854775807 goes over the wire, so I updated both assertions. I also verified against a local test validator.

mcintyre94 (approved):
Thankyou!

PR #1731, opened 2026-06-08
Add createDependentStructDecoder for structs whose fields depend on earlier values

### LINKED ISSUE #1448: Add support for building a struct decoder where later fields can depend on earlier ones
## Motivation

The v1 transaction codec makes heavy use of dependencies between fields. `numInstructions` is encoded early on, and later is required to know how to decode instruction headers (encoded as an array with no size prefix). The instruction header for instruction N is required to decode the instruction payload.

Currently this can't be expressed as a struct decoder, because we don't have access to the decoded values in those earlier fields when writing the decoder for the later ones. 

## Example use case

For v1 transactions it would be great to be able to write something like (pseudocode, simplified):

```ts
getStructDecoder([
  ['numInstructions', getU8Decoder()],
  ['numStaticAccounts', getU8Decoder()],
  ['instructionHeaders', (fields) => getArrayDecoder(getInstructionHeaderDecoder(), { size: fields.numInstructions }),
  // more fields
])
```

## Details

I think this would be very difficult to do with the current `getStructDecoder` fields array, because there's no type-safe way to access the earlier fields when defining that array.

However, it could be done using a builder pattern, where we add `.field(name, decoder)` one at a time:

```ts
const decoder = structDecoderBuilder()
  .field('numInstructions', getU8Decoder())
  .field('numStaticAccounts', getU8Decoder())
  .field('instructionHeaders', fields => getArrayDecoder(getInstructionHeaderDecoder(), { size: fields.numInstructions }))
  // more .field calls
  .build();
```

This would allow `fields` to be defined at each stage as an object containing all previously decoded values, meaning that we can write the mappings from already decoded `fields` to a later field decoder in a type-safe way.

Claude coded prototype with tests: https://github.com/anza-xyz/kit/compare/main...struct-codec-dependencies

kh0ra:
## Summary

Adds `createDependentStructDecoder`, a fluent builder for a struct decoder whose later fields may depend on the decoded values of earlier ones. This addresses the case described in #1448 where a struct field's decoder needs access to the value of an earlier field, such as the per-instruction headers in a v1 transaction message.

The builder is intentionally additive. It does not replace or modify the existing `getStructDecoder`, which remains the right choice when every field's decoder is independent.

## API

```ts
const decoder = createDependentStructDecoder()
    .field('count', getU8Decoder())
    .field('values', fields => getArrayDecoder(getU32Decoder(), { size: fields.count }))
    .finish();
```

Each `field` call adds a named entry and returns a new builder whose accumulated type is widened by the newly added field. The second argument can be either a static `Decoder` or a factory that receives a frozen snapshot of every field decoded so far, in declaration order. Calling `finish` returns a `VariableSizeDecoder` for the accumulated struct.

Two type level guarantees are encoded directly in the builder signature:

1. Adding a field with a name that already exists on the builder is a compile time error.
2. A factory only sees the fields that have been added before it; fields added later are not present on its `fields` parameter.

## Design notes

I followed @mcintyre94's builder pseudocode from the issue rather than the array form, because the array form makes it difficult to surface previously declared fields to a per-field factory in a type safe way. The builder form lets each `field` call extend the accumulator type directly, so the factory's `fields` parameter narrows naturally without conditional types.

Only the decoder side is provided. The encoder direction already has access to the entire value when serialising, so the existing `getStructEncoder` already supports the use cases the issue describes. A symmetrical encoder builder can be added in a follow up if a concrete need arises.

The resulting decoder is always variable size, since the byte length of any factory-provided field cannot be known ahead of time.

## Tests

Twelve unit tests cover the behaviour:

- A struct with only static fields decodes the same way `getStructDecoder` would.
- A field factory drives the size of a later array from an earlier `u8`.
- A factory receives exactly the fields decoded so far, never more.
- A factory can pick between two decoders based on a discriminator, exercised for two distinct discriminator values.
- `read` honours a non-zero starting offset and returns the new offset.
- An empty builder finishes to a decoder for `{}`.
- The decoded object preserves declaration order in `Object.keys`.
- The result is a `VariableSizeDecoder` (no `fixedSize` property).
- The builder is immutable; calling `field` returns a new builder without mutating the receiver.
- Independent `finish` calls on the same builder produce distinct decoder instances.
- The new decoder composes with other variable size decoders such as `fixDecoderSize(getUtf8Decoder(), n)` and `getArrayDecoder`.

Eight typetests cover the type level contract using only `satisfies`, matching the style of the existing `struct-typetest.ts`:

- The empty builder finishes to `VariableSizeDecoder<Record<never, never>>`.
- The accumulator widens with each `field` call.
- A factory's `fields` parameter is exactly `Readonly<{ ...priorFields }>`.
- A factory cannot read fields that have not been added yet (`@ts-expect-error`).
- Adding the same field name twice is a type error (`@ts-expect-error`).
- The finished decoder is assignable to a generic `Decoder<T>`.
- Both a static `Decoder` argument and a factory argument type check.
- The result is always a `VariableSizeDecoder`, never a `FixedSizeDecoder`.

## Behavioral impact

Purely additive. No existing public type or call site changes.

## Closes

Closes #1448

### DIFF
--- a/.changeset/beige-kids-stop.md
+++ b/.changeset/beige-kids-stop.md
@@ -2,7 +2,7 @@
 '@solana/codecs-data-structures': minor
 ---
 
-Add `createDependentStructDecoder`, a fluent builder for a struct decoder whose later fields may depend on the decoded values of earlier ones. Each call to `field` adds a name and either a static `Decoder` or a factory that receives a snapshot of the fields decoded so far. Calling `finish` produces a `VariableSizeDecoder` for the accumulated struct.
+Add `createDependentStructDecoder`, a fluent builder for a struct decoder whose later fields may depend on the decoded values of earlier ones. Each call to `field` adds a name and either a static `Decoder` or a factory that receives a frozen snapshot of the fields decoded so far. Calling `build` produces a `FixedSizeDecoder` when every field added to the builder is itself a `FixedSizeDecoder`, and a `VariableSizeDecoder` otherwise.
 
 This is useful for binary formats where a count, version, or discriminator that appears near the start of the struct controls how a later field must be parsed, such as the per-instruction headers in a v1 transaction message.
 
@@ -10,5 +10,5 @@ This is useful for binary formats where a count, version, or discriminator that 
 const decoder = createDependentStructDecoder()
     .field('count', getU8Decoder())
     .field('values', fields => getArrayDecoder(getU32Decoder(), { size: fields.count }))
-    .finish();
+    .build();
 ```
--- a/.changeset/beige-kids-stop.md
+++ b/.changeset/beige-kids-stop.md
@@ -1,0 +1,14 @@
+---
+'@solana/codecs-data-structures': minor
+---
+
+Add `createDependentStructDecoder`, a fluent builder for a struct decoder whose later fields may depend on the decoded values of earlier ones. Each call to `field` adds a name and either a static `Decoder` or a factory that receives a snapshot of the fields decoded so far. Calling `finish` produces a `VariableSizeDecoder` for the accumulated struct.
+
+This is useful for binary formats where a count, version, or discriminator that appears near the start of the struct controls how a later field must be parsed, such as the per-instruction headers in a v1 transaction message.
+
+```ts
+const decoder = createDependentStructDecoder()
+    .field('count', getU8Decoder())
+    .field('values', fields => getArrayDecoder(getU32Decoder(), { size: fields.count }))
+    .finish();
+```
--- a/packages/codecs-data-structures/src/dependent-struct.ts
+++ b/packages/codecs-data-structures/src/dependent-struct.ts
@@ -1,0 +1,152 @@
+import { createDecoder, Decoder, ReadonlyUint8Array, VariableSizeDecoder } from '@solana/codecs-core';
+
+import { DrainOuterGeneric } from './utils';
+
+/**
+ * A function that builds a {@link Decoder} for a struct field whose shape
+ * depends on the values of previously decoded fields in the same struct.
+ *
+ * The function receives a frozen snapshot of all fields that have been decoded
+ * so far, in declaration order, and must return the {@link Decoder} that should
+ * be used to read the current field from the byte stream.
+ *
+ * @typeParam TPriorFields - The shape of the fields that have already been
+ *     decoded by the time this factory is invoked.
+ * @typeParam TValue - The type of the value produced by the returned decoder.
+ *
+ * @see {@link createDependentStructDecoder}
+ */
+export type DependentStructDecoderFieldFactory<TPriorFields extends Record<string, unknown>, TValue> = (
+    fields: Readonly<TPriorFields>,
+) => Decoder<TValue>;
+
+/**
+ * A fluent builder that accumulates field decoders for a struct whose later
+ * fields may depend on the values of earlier ones.
+ *
+ * Each call to {@link DependentStructDecoderBuilder.field | `field`} returns a
+ * new builder whose accumulated field type is widened by the newly added field.
+ * Call {@link DependentStructDecoderBuilder.finish | `finish`} to obtain the
+ * final {@link Decoder} once every field has been declared.
+ *
+ * Instances of this type are immutable. Calling `field` does not mutate the
+ * receiver; it returns a new builder.
+ *
+ * @typeParam TFields - The shape of the struct that has been accumulated so far.
+ *
+ * @see {@link createDependentStructDecoder}
+ */
+export type DependentStructDecoderBuilder<TFields extends Record<string, unknown>> = {
+    /**
+     * Adds a field to the struct whose decoder is either static, or built from
+     * the values of previously declared fields.
+     *
+     * Adding a field that has already been declared on this builder is a
+     * compile time error.
+     *
+     * @typeParam TName - The name of the field being added. Must not collide
+     *     with the name of a field already declared on this builder.
+     * @typeParam TValue - The type of the value produced by the field's decoder.
+     *
+     * @param name - The unique name of the field within the struct.
+     * @param decoderOrFactory - Either a {@link Decoder} for the field, or a
+     *     {@link DependentStructDecoderFieldFactory} that produces one from
+     *     the previously decoded fields.
+     *
+     * @returns A new builder whose accumulated field type includes the newly
+     *     added field.
+     */
+    field<TName extends string, TValue>(
+        name: TName extends keyof TFields ? never : TName,
+        decoderOrFactory: Decoder<TValue> | DependentStructDecoderFieldFactory<TFields, TValue>,
+    ): DependentStructDecoderBuilder<DrainOuterGeneric<TFields & { [K in TName]: TValue }>>;
+
+    /**
+     * Finalizes the builder and returns a {@link VariableSizeDecoder} that
+     * decodes each declared field in turn, in the order they were added.
+     *
+     * The resulting decoder is always variable size because the size of any
+     * field declared with a factory cannot be determined ahead of time.
+     */
+    finish(): VariableSizeDecoder<DrainOuterGeneric<TFields>>;
+};
+
+/**
+ * Creates a fluent builder for a struct decoder whose later fields may depend
+ * on the decoded values of earlier ones.
+ *
+ * Unlike {@link getStructDecoder}, which accepts a fixed array of named
+ * decoders, this builder lets each field provide a factory that receives the
+ * values that have already been decoded. This is useful for binary formats
+ * where a count, version, or discriminator decoded near the start of the
+ * struct controls how a later field must be parsed.
+ *
+ * The returned builder is immutable; each {@link DependentStructDecoderBuilder.field | `field`}
+ * call returns a new builder whose accumulated field type is widened by the
+ * newly added field. Call {@link DependentStructDecoderBuilder.finish | `finish`}
+ * to produce the final decoder.
+ *
+ * @example
+ * Decoding a struct whose array length is read from an earlier field.
+ * ```ts
+ * import { getArrayDecoder } from '@solana/codecs-data-structures';
+ * import { getU8Decoder, getU32Decoder } from '@solana/codecs-numbers';
+ *
+ * const decoder = createDependentStructDecoder()
+ *     .field('count', getU8Decoder())
+ *     .field('values', fields => getArrayDecoder(getU32Decoder(), { size: fields.count }))
+ *     .finish();
+ *
+ * decoder.decode(new Uint8Array([0x02, 0x01, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00]));
+ * // { count: 2, values: [1, 2] }
+ * ```
+ *
+ * @example
+ * Mixing static and dependent fields.
+ * ```ts
+ * const decoder = createDependentStructDecoder()
+ *     .field('version', getU8Decoder())
+ *     .field('header', fields => fields.version === 0 ? getU16Decoder() : getU32Decoder())
+ *     .finish();
+ * ```
+ *
+ * @see {@link getStructDecoder}
+ */
+export function createDependentStructDecoder(): DependentStructDecoderBuilder<Record<never, never>> {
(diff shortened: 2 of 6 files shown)


kh0ra:
@mcintyre94 @lorisleiva @steveluscher this is a follow up on #1448. Pinging because the PR has been sitting and I want to make sure it has the right shape before it lands. Happy to iterate on the API ergonomics if you would prefer something different.

What landed in `a96a7310` on top of the original commit:

**Tests (18 total, six new)**
- Threads decoded values across more than two dependent fields, modelled on a v1 transaction header where `numStaticAccounts` and `numInstructions` each drive a later array.
- Nests one `createDependentStructDecoder` inside another; the outer factory returns the inner decoder.
- Asserts each factory is invoked exactly once per `decode` call, so successive decodes use the per-call snapshot rather than a cached one.
- Asserts the snapshot a factory sees matches the field values that the final decoded object will contain.
- Exercises error propagation when the dependent decoder runs past the buffer.
- Models a v0 transaction message header style struct with three dependent counts and one dependent array.

**Typetests (12 total, four new)**
- Six field chain to confirm the accumulator widens cleanly.
- A `@ts-expect-error` on `fields.first = 0` confirms the factory's parameter is read only.
- A factory pre-typed as `(fields: Readonly<{ version: number }>) => Decoder<...>` still composes with the builder.
- A `@ts-expect-error` on `value.b` confirms keys that were never declared are absent from the finished decoder's output type.

**Docs**
- README now has a short when-to-use table comparing `createDependentStructDecoder` against `getStructDecoder`, and documents the `combineCodec(getStructEncoder(...), createDependentStructDecoder()....finish())` pairing for callers that want a full codec.
- The JSDoc on `createDependentStructDecoder` has a `@remarks` block explaining when not to reach for it, and a third `@example` showing the `combineCodec` pairing.

The original commit (`10c57118`) is unchanged.

A few API decisions worth flagging for review:

1. **Builder vs. array form.** I went with the builder pseudocode from the issue rather than the array form, because the array form makes it difficult to narrow each factory's `fields` parameter to the fields that precede it. The builder lets each `field` call extend the accumulator directly. I am open to switching to the array form if you would prefer to keep the API surface uniform with `getStructDecoder`.
2. **Decoder only.** The encoder direction already has access to the entire value at write time, so I did not introduce a symmetrical encoder builder. Callers that need a full `Codec` pair the new decoder with the existing `getStructEncoder` and `combineCodec`. If you would like a symmetrical encoder builder I can add it in this PR.
3. **`finish()` vs `build()`.** The original issue pseudocode used `build()`. I picked `finish` to avoid shadowing the meaning of the codec's underlying `read`/`write` build step, but I am happy to rename if `build` is preferred.

CI is green here. Thanks for taking a look.

mcintyre94:
Thanks for taking a look at this! On the API decisions:

> Builder vs. array form. I went with the builder pseudocode from the issue rather than the array form, because the array form makes it difficult to narrow each factory's fields parameter to the fields that precede it. The builder lets each field call extend the accumulator directly. I am open to switching to the array form if you would prefer to keep the API surface uniform with getStructDecoder.

Builder form is correct, I don't think we can get the dependent types from the array form

> Decoder only. The encoder direction already has access to the entire value at write time, so I did not introduce a symmetrical encoder builder. Callers that need a full Codec pair the new decoder with the existing getStructEncoder and combineCodec. If you would like a symmetrical encoder builder I can add it in this PR.

Agree with this, no need for an encoder here

> finish() vs build(). The original issue pseudocode used build(). I picked finish to avoid shadowing the meaning of the codec's underlying read/write build step, but I am happy to rename if build is preferred.

I think `build()` is probably better, but I don't understand this comment. Can you explain what you mean by "avoid shadowing the meaning of the codec's underlying read/write build step"? The core `codec.ts` file doesn't mention "build" anywhere.

.changeset/beige-kids-stop.md
```diff
@@ -0,0 +1,14 @@
...
+
+Add `createDependentStructDecoder`, a fluent builder for a struct decoder whose later fields may depend on the decoded values of earlier ones. Each call to `field` adds a name and either a static `Decoder` or a factory that receives a snapshot of the fields decoded so far. Calling `finish` produces a `VariableSizeDecoder` for the accumulated struct.
+
+This is useful for binary formats where a count, version, or discriminator that appears near the start of the struct controls how a later field must be parsed, such as the per-instruction headers in a v1 transaction message.
+
+```ts
+const decoder = createDependentStructDecoder()
+    .field('count', getU8Decoder())
+    .field('values', fields => getArrayDecoder(getU32Decoder(), { size: fields.count }))
+    .finish();
```
mcintyre94: As I mentioned I'd prefer `build()`, but don't understand your explanation for why you didn't use it yet. So I might be wrong.

packages/codecs-data-structures/src/dependent-struct.ts
```diff
@@ -0,0 +1,181 @@
...
+    ): DependentStructDecoderBuilder<DrainOuterGeneric<TFields & { [K in TName]: TValue }>>;
+
+    /**
+     * Finalizes the builder and returns a {@link VariableSizeDecoder} that
+     * decodes each declared field in turn, in the order they were added.
+     *
+     * The resulting decoder is always variable size because the size of any
+     * field declared with a factory cannot be determined ahead of time.
+     */
+    finish(): VariableSizeDecoder<DrainOuterGeneric<TFields>>;
```
mcintyre94: Always returning `VariableSize` is a regression over the main struct decoder, which gives a `FixedSizeDecoder` if all the fields are `FixedSizeDecoder`.

I think we could do this, by initialising to a `FixedSizeDecoder` form, ie `createDependentStructDecoder().finish()` gives a `FixedSizeDecoder`.

Then if you add a `FixedSizeDecoder` decoder or factory, then we keep it with that type.

If you add any other `Decoder` decoder or factory, then we change to a branch that has a variable size.

So it starts with FixedSize, it stays there while you keep adding only FixedSize fields, and if you add anything VariableSize then it drops to VariableSize and stays there.

I think this should be feasible.

packages/codecs-data-structures/src/dependent-struct.ts
```diff
@@ -0,0 +1,181 @@
...
+    entries: readonly InternalEntry[],
+): DependentStructDecoderBuilder<TFields> {
+    return {
+        field<TName extends string, TValue>(
+            name: TName extends keyof TFields ? never : TName,
+            decoderOrFactory: Decoder<TValue> | DependentStructDecoderFieldFactory<TFields, TValue>,
+        ): DependentStructDecoderBuilder<DrainOuterGeneric<TFields & { [K in TName]: TValue }>> {
+            const resolveDecoder: ResolvedDecoder =
+                typeof decoderOrFactory === 'function'
+                    ? fields => decoderOrFactory(fields as TFields)
```
mcintyre94: Nit: should probably do `Object.freeze` here, docs mention it receives a frozen snapshot and it makes sense that it should

packages/codecs-data-structures/src/dependent-struct.ts
```diff
@@ -0,0 +1,181 @@
...
+ *     .field('count', getU8Decoder())
+ *     .field('values', fields => getArrayDecoder(getU32Decoder(), { size: fields.count }))
+ *     .finish();
+ * const codec = combineCodec(encoder, decoder);
+ * ```
+ *
+ * @see {@link getStructDecoder}
+ * @see {@link getStructEncoder}
+ */
+export function createDependentStructDecoder(): DependentStructDecoderBuilder<Record<never, never>> {
```
mcintyre94: I think this naming is good - using `create` vs `get` because it's a builder API.

.changeset/beige-kids-stop.md
kh0ra: Yeah you were right. Renamed it to `build` everywhere in 3a310742 (source, tests, typetests, README, JSDoc, changeset). My pushback was weak, your original pseudocode reads better and now everything matches it.

packages/codecs-data-structures/src/dependent-struct.ts
kh0ra: Really good catch, thanks for laying out the mental model so clearly. I added a `TIsFixedSize extends boolean` to the builder type. Adding a `FixedSizeDecoder` keeps it `true` via overload 1, and adding a `VariableSizeDecoder` or a factory has its own overload that forces it to `false`. `build()` then returns the right variant. At runtime the total `fixedSize` is summed with `sumCodecSizes(staticDecoders.map(getFixedSize))` exactly the same way `getStructDecoder` does, so the two stay in lockstep. A few new tests pin it down: empty builder has `fixedSize: 0`, all-fixed builders sum cleanly, and the moment a variable size field or a factory shows up the result is `VariableSizeDecoder` for good. There's also a typetest that uses `@ts-expect-error` to confirm you cannot raise it back to `FixedSizeDecoder` once it has dropped.

packages/codecs-data-structures/src/dependent-struct.ts
kh0ra: Good call. The docs were lying about this. I said frozen snapshot but never actually froze anything. Now each factory call gets `Object.freeze({ ...decoded })` so the snapshot really is frozen. The internal accumulator stays mutable so the read loop can keep filling it. There's a test now (`passes a frozen snapshot to a factory`) that checks both `Object.isFrozen(fields)` and that an in-strict-mode mutation attempt throws.

kh0ra:
@mcintyre94 thanks for the thorough review, all three points made the PR genuinely better. The fixed vs variable size promotion in particular was something I should have caught on my own. Once you spelled out the parity with `getStructDecoder` it was obvious in hindsight. The `Object.freeze` nit was also a fair call, the docs were promising something the code was not actually doing.

Ready for another look when you have time. Happy to keep iterating if anything still feels off.

mcintyre94 (approved):
Thankyou, this looks good to me now! I think there are cases where the factory field form could maintain `FixedSize`, but I think that would probably be very unusual and we can improve that further later if necessary.

I'll kick off our AI review too in case I missed anything :)

mcintyre94:
@trevor-cortex

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Nice work — the API is tight, the type-level tests cover the contract well, and the runtime tests are thorough (the nested-dependence and frozen-snapshot tests in particular are good additions). Code reads cleanly and the docblocks are in good shape.

A few small observations, none blocking given Callum's approval:

**`maxSize` is not propagated.** When the builder produces a `VariableSizeDecoder`, the resulting decoder has no `maxSize`, even in cases where it could be computed (e.g. all fields are static and a subset are variable-size with a known `maxSize`). `getStructDecoder` propagates `maxSize` via `sumCodecSizes(fieldCodecs.map(getMaxSize))` and it would be straightforward to do the same here for the all-static case. Factories naturally can't contribute a `maxSize` (we don't have the inner decoder until decode time), so the rule would be: if every entry has a `staticDecoder` AND `sumCodecSizes(staticDecoders.map(getMaxSize))` isn't `null`, set `maxSize` on the result. Not a correctness issue — just a small metadata loss versus `getStructDecoder` that might matter to downstream tooling.

**Naming nit in `build()`.** `everyStatic` reads like "every field is fixed size," but it actually means "every field was added directly rather than via a factory." The real fixed-vs-variable decision still flows through `sumCodecSizes(...) === null`. Consider `everyDecoderProvidedDirectly` or just dropping the variable and inlining the check, since `staticDecoders` already gets walked by `sumCodecSizes(staticDecoders.map(getFixedSize))` regardless. Minor — just slowed me down briefly when reading.

**Test readability:** `b('02010000000200000000'.slice(0, 18))` (line ~21 of the test file) is doing a hex-string slice to drop the trailing `00` byte, which made me reach for a calculator. The same effect is achievable with `b('020100000002000000')` directly. Not wrong, just a small clarity win.

**Stale PR description.** The description still refers to a `finish` method and asserts "the result is always a `VariableSizeDecoder`," but the shipped API uses `build` and conditionally returns a `FixedSizeDecoder`. The README, changeset, and code all match the current shape — only the PR body is out of date. Cosmetic, but worth a quick edit before merge for future spelunkers reading the PR history.

Nothing here changes my assessment that this is ready to ship behind Callum's approval.

PR #1741, opened 2026-06-12
fix: include v0 preamble in offchain message size checks

### LINKED ISSUE #1726: v0 offchain message size checks exclude the preamble
## Overview

Found while looking into offchain-messages as part of the signOffchainMessage feature. We've decided to make that only support v1 transactions, so I'm just opening this as a bug. Doesn't block that feature because v1 doesn't enforce a max size in the spec.

## Description of bug

The checks for max size in eg https://github.com/anza-xyz/kit/blob/d32d88c07c0264d9cf5d294299c4cb78f502c23c/packages/offchain-messages/src/content.ts#L74 have checks like:

```ts
const length = getUtf8Encoder().getSizeFromValue(putativeContent.text);
if (length > MAX_BODY_BYTES_HARDWARE_WALLET_SIGNABLE)
```

That is, they are only checking the length of the text content against the maximum.

The [offchain message spec](https://docs.anza.xyz/proposals/off-chain-message-signing) specifies size limits as "Combined length of the message preamble and message body"

The kit functions should throw if the size of the text is smaller than the limit, but the size of the text + preamble is larger than the limit. This is not currently the case.

ChargingFoxSec:
#### Problem

The v0 1232-byte message size check only counted the message body. This allowed messages to pass validation when the body was under the limit, even if the full encoded message with the v0 preamble was over 1232 bytes.

#### Summary of Changes

Fixes #1726.

This updates the v0 offchain message validation to account for the preamble size when checking the 1232-byte formats.

It also adds encode and decode tests for the boundary case where the body alone fits, but the full v0 encoded message does not.

### DIFF
--- a/.changeset/slow-cameras-serve.md
+++ b/.changeset/slow-cameras-serve.md
@@ -1,0 +1,5 @@
+---
+'@solana/offchain-messages': patch
+---
+
+Include the v0 preamble bytes when enforcing the 1232-byte limit for hardware-wallet-signable offchain messages.
--- a/packages/offchain-messages/src/__tests__/message-v0-codec-test.ts
+++ b/packages/offchain-messages/src/__tests__/message-v0-codec-test.ts
@@ -52,6 +52,18 @@ const SIGNER_B_BYTES = new Uint8Array([
     0x94, 0x8f, 0xd8, 0xbb, 0x2c, 0x10, 0xa1, 0x01, 0x02, 0xce, 0x98, 0xb3, 0xa6,
 ]);
 
+const PREAMBLE_LENGTH_WITH_TWO_SIGNERS_BYTES =
+    OFFCHAIN_MESSAGE_SIGNING_DOMAIN_BYTES.length +
+    1 + // Version
+    APPLICATION_DOMAIN_BYTES.length +
+    1 + // Message format
+    1 + // Signer count
+    SIGNER_A_BYTES.length +
+    SIGNER_B_BYTES.length +
+    2; // Message length
+const MAX_BODY_LENGTH_WITH_TWO_SIGNERS_BYTES = 1232 - PREAMBLE_LENGTH_WITH_TWO_SIGNERS_BYTES;
+const BODY_LENGTH_EXCEEDING_TWO_SIGNER_PREAMBLE_BYTES = MAX_BODY_LENGTH_WITH_TWO_SIGNERS_BYTES + 1;
+
 describe('getOffchainMessageV0Decoder()', () => {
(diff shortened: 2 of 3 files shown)


mcintyre94 (approved):
LGTM, thankyou!

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Clean, well-targeted fix — LGTM (submitting as COMMENT since I don't have explicit approve permissions on this repo).

## Summary

The v0 1232-byte limit historically only counted the body, so a message whose body fit but whose body + preamble didn't could still pass validation and then be unsignable by a hardware wallet. This PR adds a new `assertFitsHardwareWalletSignableMessageByteLimit` helper that recomputes the available body budget as `1232 - preambleLength(requiredSignatories)` and wires it into the two `assertIsOffchainMessage…1232BytesMax` asserts. Because the codec in `codecs/message-v0.ts` already routes through those asserts on both encode and decode, the new check covers both directions for free. The `UTF8_65535_BYTES_MAX` format is correctly left alone.

The preamble-length math (16 + 1 + 32 + 1 + 1 + 2 + 32·signers) matches the on-wire layout, and the tests independently reconstruct the same calculation for the boundary case across both ASCII and UTF-8 formats, on both encoder and decoder paths.

## Things to watch out for

- **Two different `maxBytes` values can now surface for "too long"**: the existing content-level asserts in `content.ts` still throw `SOLANA_ERROR__OFFCHAIN_MESSAGE__MAXIMUM_LENGTH_EXCEEDED` with `maxBytes: 1232` when the body alone exceeds 1232, while the new outer check throws the same error code with a smaller `maxBytes` (1232 − preamble) when the body fits but the full message doesn't. Same error code, different bounds — worth flagging in case any downstream consumer was treating `maxBytes` as a fixed constant. Not a blocker; the new bound is strictly more correct.
- **Preamble-structure knowledge is now duplicated**: the byte-length constants here mirror the field layout encoded by `codecs/preamble-v0.ts`. If the v0 preamble ever grows a field, both places need updating in lockstep. The constants are at least clearly named so the duplication is easy to spot.

- Worth a quick sanity-check against `codecs/preamble-v0.ts` to confirm the field sizes (`SIGNING_DOMAIN_LENGTH_BYTES = 16`, `APPLICATION_DOMAIN_LENGTH_BYTES = 32`, etc.) match what the encoder actually writes. The numbers line up with the SRFC-3 spec and the test fixture.
- The fix is purely additive — no error code or public type changes — so the `patch` changeset bump is appropriate.

PR #1748, opened 2026-06-16
fix: avoid treating JS protocol hooks as RPC methods

### LINKED ISSUE #509: Check if we should be intercepting properties in Proxy objects
Our RPC and subscriptions objects use a proxy to support arbitrary JSON-RPC methods. Currently we don't special-case any properties, and everything is treated as defined to return the `async send()` function.

See https://github.com/anza-xyz/kit/pull/508 where this causes an issue with the special `then`, which makes JS interpret our proxy as a valid promise.

We should check if there are any other standard properties that might cause similar issues where our proxy gets interpreted as providing functionality that it does not.

ChargingFoxSec:
#### Problem

RPC and subscription proxy objects currently treat most property access as an RPC method name. After #508, `then` is handled specially, but other JavaScript protocol hooks can still be interpreted as methods.

For example, `JSON.stringify(rpc)` looks up `toJSON`, and primitive/iteration/inspection paths can look up symbols such as `Symbol.toPrimitive` and `Symbol.iterator`.

#### Summary of Changes

Fixes #509.

This adds a narrow non-RPC property check for RPC proxy objects and subscription proxy objects.

- Avoid treating JavaScript protocol hooks like `toJSON`, `Symbol.toPrimitive`, `Symbol.iterator`, `Symbol.asyncIterator`, `Symbol.toStringTag`, and Node's custom inspect symbol as RPC methods.
- Preserve normal Object prototype behavior for properties such as `toString` and `hasOwnProperty`.
- Apply the behavior to both outer RPC objects and the lower-level API proxy objects.
- Add regression tests for RPC and subscription proxies.

#### Testing

- `corepack pnpm --filter @solana/rpc-spec test:unit:node`
- `corepack pnpm --filter @solana/rpc-subscriptions-spec test:unit:node`
- `corepack pnpm --filter @solana/rpc-spec test:unit:browser`
- `corepack pnpm --filter @solana/rpc-subscriptions-spec test:unit:browser`
- `corepack pnpm --filter @solana/rpc-spec test:typecheck`
- `corepack pnpm --filter @solana/rpc-subscriptions-spec test:typecheck`
- `corepack pnpm --filter @solana/rpc-spec test:lint`
- `corepack pnpm --filter @solana/rpc-subscriptions-spec test:lint`
- `corepack pnpm --filter @solana/rpc-spec test:prettier`
- `corepack pnpm --filter @solana/rpc-subscriptions-spec test:prettier`
- `corepack pnpm --filter @solana/rpc-spec compile:js`
- `corepack pnpm --filter @solana/rpc-subscriptions-spec compile:js`


packages/rpc-spec/src/non-rpc-proxy-properties.ts
```diff
@@ -0,0 +1,37 @@
+const NODEJS_CUSTOM_INSPECT_SYMBOL = Symbol.for('nodejs.util.inspect.custom');
+
+const JAVASCRIPT_PROTOCOL_PROPERTY_NAMES = new Set<PropertyKey>([
+    'then',
+    'toJSON',
+    NODEJS_CUSTOM_INSPECT_SYMBOL,
+    Symbol.asyncIterator,
+    Symbol.iterator,
+    Symbol.toPrimitive,
+    Symbol.toStringTag,
```
mcintyre94: I think it'd be worth adding `dispose` and `asyncDispose`, just because accidentally doing `using rpc = ...` is slightly-plausible.

mcintyre94 (commented):
This is great, thankyou! My only concern is the duplication, given that we might need to update these in future. My instinct is to move it to the shared `rpc-spec-types`, which is already a dependency of both `rpc-spec` and `rpc-subscriptions-spec`. WDYT?

ChargingFoxSec:
Yeah, that makes sense. I’ll move the shared non-RPC property list into `rpc-spec-types` and add `dispose` and `asyncDispose` as well.

packages/rpc-spec/src/__tests__/rpc-api-test.ts
```diff
@@ -105,16 +115,14 @@ describe('createJsonRpcApi', () => {
...
         expect(api).not.toHaveProperty('toJSON');
-        expect(getUntypedProperty(api, Symbol.asyncIterator)).toBeUndefined();
-        expect(getUntypedProperty(api, Symbol.for('nodejs.util.inspect.custom'))).toBeUndefined();
-        expect(getUntypedProperty(api, Symbol.iterator)).toBeUndefined();
-        expect(getUntypedProperty(api, Symbol.toPrimitive)).toBeUndefined();
-        expect(getUntypedProperty(api, Symbol.toStringTag)).toBeUndefined();
+        JAVASCRIPT_PROTOCOL_SYMBOLS.forEach(symbol => {
+            expect(getUntypedProperty(api, symbol)).toBeUndefined();
+        });
     });
```
mcintyre94: Nit: a cleaner way to write these tests would be using `it.each`, for example used here: https://github.com/anza-xyz/kit/blob/ec4d3ef7192a437ecbdc7103aa1c95556dbd7f5f/packages/instructions/src/__tests__/roles-test.ts#L43 

This would give us much better feedback if any of the tests fails

packages/rpc-spec-types/src/non-rpc-proxy-properties.ts
```diff
@@ -1,16 +1,25 @@
...
-    NODEJS_CUSTOM_INSPECT_SYMBOL,
-    Symbol.asyncIterator,
-    Symbol.iterator,
-    Symbol.toPrimitive,
-    Symbol.toStringTag,
-]);
+const SymbolWithDispose = Symbol as typeof Symbol & {
+    asyncDispose?: symbol;
+    dispose?: symbol;
+};
```
mcintyre94: We shouldn't need this - `Symbol.dispose` and `Symbol.asyncDispose` should be available in the verison of Typescript we're using, and are available for me on main.

packages/rpc-spec/src/__tests__/rpc-test.ts
```diff
@@ -93,12 +103,10 @@ describe('JSON-RPC 2.0', () => {
...
-            expect.assertions(5);
-            expect(getUntypedProperty(rpc, Symbol.asyncIterator)).toBeUndefined();
-            expect(getUntypedProperty(rpc, Symbol.for('nodejs.util.inspect.custom'))).toBeUndefined();
-            expect(getUntypedProperty(rpc, Symbol.iterator)).toBeUndefined();
-            expect(getUntypedProperty(rpc, Symbol.toPrimitive)).toBeUndefined();
-            expect(getUntypedProperty(rpc, Symbol.toStringTag)).toBeUndefined();
+            expect.hasAssertions();
+            JAVASCRIPT_PROTOCOL_SYMBOLS.forEach(symbol => {
+                expect(getUntypedProperty(rpc, symbol)).toBeUndefined();
+            });
```
mcintyre94: Same comment re `it.each` here

mcintyre94 (changes requested):
Thanks, reducing the duplication is definitely nicer! :)

ChargingFoxSec:
Thanks, updated this.

I removed the custom Symbol typing and now use Symbol.dispose / Symbol.asyncDispose directly. Also changed the symbol checks to it.each so failures point at the specific symbol, and mirrored that in the subscription proxy tests.

Related node/browser tests, typecheck, lint, and Prettier checks are passing locally.

packages/rpc-spec-types/src/non-rpc-proxy-properties.ts
```diff
@@ -0,0 +1,41 @@
...
+export function getNonRpcPropertyValue(propertyName: string | symbol, receiver: unknown): unknown {
+    if (JAVASCRIPT_PROTOCOL_PROPERTY_NAMES.has(propertyName)) {
+        return undefined;
+    }
+    return Reflect.get(Object.prototype, propertyName, receiver);
+}
+
+export function isNonRpcPropertyName(propertyName: string | symbol): boolean {
+    return JAVASCRIPT_PROTOCOL_PROPERTY_NAMES.has(propertyName) || OBJECT_PROTOTYPE_PROPERTY_NAMES.has(propertyName);
+}
```
mcintyre94, line 53: Nit: these are both exported functions. Can you add a docblock for each (just a one liner and reference to the issue is fine), and also `@internal`? I don't think these should be part of our API surface, but exporting them from here is correct. Sorry, should have flagged that when I suggested moving them here.

mcintyre94 (changes requested):
One last nit and this is ready to go, thankyou!

packages/rpc-spec-types/src/non-rpc-proxy-properties.ts
ChargingFoxSec: Done, thanks! I added docblocks to both helpers, marked them as @internal, and linked #509.
I also fixed the tree-shakability failure by marking the module-level initializers as pure. The affected rpc-spec-types checks are passing locally.

mcintyre94:
@trevor-cortex

mcintyre94 (dismissed):

trevor-cortex (commented):
This looks ready to go (would approve, but I don't have approval permission on this repo). The shared `getNonRpcPropertyValue` / `isNonRpcPropertyName` helpers in `@solana/rpc-spec-types` resolve the duplication concern from earlier rounds, and all four proxy sites (`rpc.ts`, `rpc-api.ts`, `rpc-subscriptions.ts`, `rpc-subscriptions-api.ts`) now go through the same code path.

**What this does well:**

- The two-tier design is sound: protocol hooks (`then`, `toJSON`, `Symbol.toPrimitive`, etc.) return `undefined` so JS falls back to default behavior, while `Object.prototype` members are forwarded via `Reflect.get(Object.prototype, p, receiver)` so `String(rpc)`, `hasOwnProperty`, etc. behave like a plain object. Notably, `String(rpc)` works end-to-end because `Symbol.toPrimitive` and `Symbol.toStringTag` return `undefined` and `toString` resolves to `Object.prototype.toString` — and the tests verify exactly that chain.
- The `.filter(propertyName => propertyName != null)` guard on the Set entries (mirrored in the test fixtures) correctly handles runtimes where `Symbol.asyncDispose` / `Symbol.dispose` don't exist yet.
- `/* @__PURE__ */` annotations keep the module-level `Symbol.for` call and Sets tree-shakeable.
- Changeset is present with patch bumps for all three affected packages, which matches the bug-fix nature of the change.

- This is technically a behavior change: an RPC method literally named `toString`, `valueOf`, `constructor`, etc. can no longer be invoked via property access. No real Solana RPC method collides with these, so this seems like the right trade-off, but worth being aware of.
- The test fixture block (`getUntypedProperty` + `JAVASCRIPT_PROTOCOL_SYMBOLS`) is copied across the four test files. That's fine for self-contained tests and not worth blocking on; if it grows further it could move to a shared test helper.

mcintyre94 (approved):

mcintyre94:
Thankyou for this!

PR #1759, opened 2026-06-17
Add a /query subpath to react, add useRequestQuery

mcintyre94:
#### Summary of Changes

This PR adds the `@solana/react/query` subpath, with an optional `@tanstack/react-query` dependency.

We add the first hook to this subpath, `useRequestQuery`. This mirrors `useRequest` and `useRequestSWR` but provides the shape of `useQuery`:

```tsx
function LatestBlockhash() {
    const client = useClient<ClientWithRpc<GetLatestBlockhashApi>>();
    const { data, error, isLoading, refetch } = useRequestQuery(
        ['latestBlockhash'],
        client.rpc.getLatestBlockhash(),
    );
    if (error) return <button onClick={() => refetch()}>Retry</button>;
    if (isLoading) return <p>Loading…</p>;
    return <p>Blockhash: {data!.value.blockhash}</p>;
}
```

This enables all the niceties of Tanstack Query: shared cache (based on key) across components, dedupe, automatic refetching, global config, etc

A third optional param allows passing any Tansatck Query configuration (same as `useQuery`), as well as the `getAbortSignal(): AbortSignal` factory supported by `useRequest`. Query handles signals, so if passed this is combined with the Tanstack Query signal using `AbortSignal.any`. Note that Query config passed will be merged (by Query) with any global config in the normal way.

Like `useRequest` it can take either a `ReactiveActionSource` such as `PendingRpcRequest`, or an async function `(signal: AbortSignal) => Promise<T>`.

Like `useRequest` it does not fetch when the source is null. We treat a null source the same as passing `enabled: false` in the Query options. This removes the need to test against null too. 

Following PRs will add Tanstack Query versions of `useSubscription` and `useTrackedData`.

PR #1758, opened 2026-06-17
Refactor sol <> lamport conversion in example app

mcintyre94:
This PR refactors the Lamport -> SOL and SOL -> Lamport conversion in the react example app to use the new `@solana/fixed-points` package.

Unrelated to the react changes but added since I'm poking at the example app and noticed this code that can be refactored to demonstrate some of the new functionality.

PR #1754, opened 2026-06-16
Refactor the react example app to use new react hooks

mcintyre94:
Using this to validate parts of the stack so far. In general, much less code for the same functionality. This uses the new react hooks, does not yet refactor to use plugins.

- `AirdropButton` now uses `useAction`
- `Balance` is refactored to use `useTrackedDataSWR` , it previously used equivalent custom code with SWR
- `SlotIndicator` refactored to use `useSubscription`. Newly displays previous data on an error, and newly enables reconnecting after an error
- Sign Message, Partial Sign, Sign + Send, Sign panels, now all use `useAction`

PR #1747, opened 2026-06-15
Add test coverage to the React example app and align it with the current kit API

mcintyre94:
#### Adding tests for the existing behaviour of the react-app example

This will provide validation when parts of the react-app example are refactored to use the new hooks (and eventually wallet plugin hooks)

PR #1760, opened 2026-06-17
Add useSubscriptionQuery to react/query subpath

mcintyre94:
#### Summary of Changes

This PR adds `useSubscriptionQuery`, the Tanstack Query shaped version of `useSubscription`.

```tsx
function SlotHeight() {
    const client = useClient<ClientWithRpcSubscriptions<SlotNotificationsApi>>();
    const { data, error } = useSubscriptionQuery(['slot'], client.rpcSubscriptions.slotNotifications());
    if (error) return <p>Disconnected.</p>;
    if (!data) return <p>Connecting…</p>;
    return <p>Slot {String(data.slot)}</p>;
}
```

Similarly to `useRequestQuery` the third param is an `UseQueryOptions`, with our usual `getAbortSignal` factory to add a custom signal. It also disables when either key or source are null.

Note that under the hood this uses [`experimental_streamedQuery`](https://tanstack.com/query/latest/docs/reference/streamedQuery), which allows `useQuery` to work with a `queryFn` that streams an `AsyncIterable`. Unlike SWR this does not use a different subscription type, so we expose the same API as any other `useQuery`. Most notably this means that we have `refetch` which is missing from `useSubscriptionSWR`. 

We use a bridge function to bridge from our `ReactiveStreamStore` into an `AsyncIterable`, in the same way that we bridged to SWR's custom subscription interface.

As this hook does work with `AsyncIterable`, I've also made it accept a `signal => AsyncIterable` function, so that it can be used without wrapping an `AsyncIterable` in a store, just for us to bridge it back. Note that our other `useSubscription` hooks still don't work with `AsyncIterable` though and continue to require our store.

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -454,6 +454,41 @@ function Profile({ userId }: { userId: string }) {
 
 The function source is held in a ref synced to the latest render, so an inline closure recreated each render is fine — no `useCallback` needed. TanStack keys the cache off `key`, not the queryFn identity.
 
+### `useSubscriptionQuery(key, source, options?)`
+
+TanStack Query-backed counterpart to `useSubscription`, for streams that have no one-shot RPC fetch. This hook is built on `experimental_streamedQuery`: a long-lived query that stays `fetching` for the subscription's whole life, writing each notification to the cache as it arrives. Components reading the same `key` share one underlying connection and the stream shows up in TanStack Query's devtools.
+
+Returns TanStack's native `UseQueryResult<T>`. `data` is the raw notification, exactly as the source emits it, so for RPC subscriptions read `data.value` and `data.context.slot`. Pass `null` for `source` to disable (TanStack's `enabled: false`).
+
+```tsx
+import { useClient } from '@solana/react';
+import { useSubscriptionQuery } from '@solana/react/query';
+import type { ClientWithRpcSubscriptions, SlotNotificationsApi } from '@solana/kit';
+
+function SlotHeight() {
+    const client = useClient<ClientWithRpcSubscriptions<SlotNotificationsApi>>();
+    const { data, error } = useSubscriptionQuery(['slot'], client.rpcSubscriptions.slotNotifications());
+    if (error) return <p>Disconnected.</p>;
+    if (!data) return <p>Connecting…</p>;
+    return <p>Slot {String(data.slot)}</p>;
+}
+```
+
+Because the stream never settles, the query sits in `fetchStatus: 'fetching'` for the subscription's whole life — `isFetching` is permanently `true`, and `isLoading` flips false after the first notification. Read `data` / `error` / `status` and ignore `isFetching` for subscriptions. Note that `isLoading` only reports the _initial_ connect — it stays false across reconnects. `result.refetch()` is the reconnect verb: it aborts the current connection (resetting its store) and opens a fresh one. Fire and forget — for a never-ending stream the returned promise never resolves, so don't `await refetch()` or it will hang forever. Sensible defaults are applied and all overridable: `retry: false` (the underlying reactive store owns retry/backoff), and `staleTime: Infinity` + `refetchOnWindowFocus: false` so a focus revalidation doesn't tear down and re-open the socket — the subscription is what keeps the data fresh.
+
+The source matches `useSubscription`: a `ReactiveStreamSource<T>`. This hook also accepts a raw `(signal: AbortSignal) => AsyncIterable<T>` factory, as `experimental_streamedQuery` is built on `AsyncIterable`.
+
+```tsx
+function Ticker() {
+    const { data } = useSubscriptionQuery(['ticker'], (signal: AbortSignal) => streamPrices(signal));
+    return <p>{data ?? '…'}</p>;
+}
+```
+
+### Why no `useActionQuery`?
+
+It would just be a wrapper around Tanstack's built-in [`useMutation`](https://tanstack.com/query/latest/docs/framework/react/guides/mutations) with no additional functionality. Either use `useMutation` or, if you don't need the Tanstack integration, use `useAction`.
+
 ## Hooks
 
 ### `useSignIn(uiWalletAccount, chain)`
--- a/packages/react/src/query/__tests__/useSubscriptionQuery-test.browser.tsx
+++ b/packages/react/src/query/__tests__/useSubscriptionQuery-test.browser.tsx
@@ -1,0 +1,416 @@
+import { type ReactiveState, type ReactiveStreamSource, type ReactiveStreamStore } from '@solana/kit';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { act, screen, waitFor } from '@testing-library/react';
+import React from 'react';
+import { renderToString } from 'react-dom/server';
+
+import { render, renderHook } from '../../__test-utils__/render';
+import { useSubscriptionQuery } from '../useSubscriptionQuery';
+
+// A fresh `QueryClient` per render so cache state never leaks between tests. `gcTime` defaults to 0
+// so an unmounted query is collected immediately; the focus/reconnect refetch triggers are off so
+// the test environment doesn't accidentally re-fire queries during assertions.
+function createWrapper(gcTime = 0) {
+    const queryClient = new QueryClient({
+        defaultOptions: {
+            queries: {
+                gcTime,
+                refetchOnReconnect: false,
+                refetchOnWindowFocus: false,
+                retry: false,
+            },
+        },
+    });
+    return function wrapper({ children }: { children: React.ReactNode }) {
+        return <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>;
+    };
+}
+
+// A controllable push-based async iterable: values pushed before a pull are buffered, and `fail`
+// makes the iterator throw. Used to stand in for an arbitrary `(signal) => AsyncIterable<T>` source.
+function makePushStream<T>() {
+    const buffer: T[] = [];
+    let failure: { readonly error: unknown } | undefined;
+    let ended = false;
+    let waiter: PromiseWithResolvers<void> | undefined;
+    const wake = () => waiter?.resolve();
+    return {
+        end() {
+            ended = true;
+            wake();
+        },
+        fail(error: unknown) {
+            failure = { error };
+            wake();
+        },
+        iterable: {
+            async *[Symbol.asyncIterator]() {
+                while (true) {
+                    while (buffer.length) yield buffer.shift()!;
+                    if (failure) throw failure.error;
+                    if (ended) return;
+                    waiter = Promise.withResolvers<void>();
+                    await waiter.promise;
+                }
+            },
+        } as AsyncIterable<T>,
+        push(value: T) {
+            buffer.push(value);
+            wake();
+        },
+    };
+}
+
+// Wraps `makePushStream` in a `jest.fn` factory that mints a fresh stream per invocation (so a
+// StrictMode double-mount's discarded stream never steals values from the live one) and records the
+// latest stream + signal for the test to drive.
+function makeStreamFactory<T>() {
+    let latest: ReturnType<typeof makePushStream<T>> | undefined;
+    let latestSignal: AbortSignal | undefined;
+    const factory = jest.fn((signal: AbortSignal): AsyncIterable<T> => {
+        const stream = makePushStream<T>();
+        latest = stream;
+        latestSignal = signal;
+        // End the stream when its signal aborts so a discarded iterator tears down cleanly.
+        signal.addEventListener('abort', () => stream.end(), { once: true });
+        return stream.iterable;
+    });
+    return {
+        factory,
+        fail: (error: unknown) => latest!.fail(error),
+        push: (value: T) => latest!.push(value),
+        signal: () => latestSignal,
+    };
+}
+
+function createFakeStore<T>(): {
+    connectCount: () => number;
+    emit: (state: ReactiveState<T>) => void;
+    resetCount: () => number;
+    store: ReactiveStreamStore<T>;
+    withSignalArg: () => AbortSignal | undefined;
+} {
+    let state: ReactiveState<T> = { data: undefined, error: undefined, status: 'idle' };
+    const listeners = new Set<() => void>();
+    let connects = 0;
+    let resets = 0;
+    let withSignalArg: AbortSignal | undefined;
+    const store: ReactiveStreamStore<T> = {
+        connect: jest.fn().mockImplementation(() => {
+            throw new Error('not implemented');
+        }),
+        getError: jest.fn().mockImplementation(() => {
+            throw new Error('not implemented');
+        }),
+        getState: jest.fn().mockImplementation(() => {
+            throw new Error('not implemented');
+        }),
+        getUnifiedState: () => state,
+        reset: () => {
+            resets++;
+        },
+        retry: jest.fn().mockImplementation(() => {
+            throw new Error('not implemented');
+        }),
+        subscribe: (callback: () => void) => {
+            listeners.add(callback);
+            return () => {
+                listeners.delete(callback);
+            };
+        },
+        withSignal: (signal: AbortSignal) => {
+            withSignalArg = signal;
+            return {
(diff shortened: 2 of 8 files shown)


mcintyre94:
* **#1786** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1786?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1782** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1782?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1780** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1780?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1779** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1779?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1769** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1769?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1760** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1760?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1760?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1759** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1759?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1758** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1758?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1754** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1754?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1747** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1747?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1727** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1727?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1719** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1719?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1713** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1713?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1707** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1707?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1709](https://github.com/anza-xyz/kit/pull/1709) <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1709?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds `useSubscriptionQuery` to `@solana/react/query` — the TanStack Query-backed counterpart to `useSubscription`, built on `experimental_streamedQuery` so a long-lived stream lives in the TanStack cache, dedupes across consumers with the same key, and shows up in devtools. Accepts the same `ReactiveStreamSource` shape as `useSubscription`, plus a raw `(signal) => AsyncIterable<T>` factory (since `experimental_streamedQuery` is built on `AsyncIterable` natively). `data` is the raw notification — the `SolanaRpcResponse` envelope is not unwrapped — matching `useSubscription`.

The core pieces:
- `bridgeStoreToAsyncIterable` — pull-based async-iterable adapter for a `ReactiveStreamStore`, the TanStack twin of `bridgeStoreToSwr`. Latest-wins single-slot buffer; throws on store error (with the `SOLANA_ERROR__REACT__SUBSCRIPTION_CLOSED_WITHOUT_ERROR` sentinel for nullish errors); ends cleanly on signal abort. `finally` unsubscribes and `reset()`s the store.
- `useSubscriptionQuery` — wires `experimental_streamedQuery<T, T>` with `initialValue: undefined as T`, `reducer: (_, chunk) => chunk` (replace-latest scalar instead of array accumulation), `refetchMode: 'append'` (every chunk written to cache live; prior value persists across refetch). Threads TanStack's query-cancellation signal — combined with any `getAbortSignal`-supplied signal via `AbortSignal.any` — through to the factory. Defaults `retry: false`, `staleTime: Infinity`, `refetchOnWindowFocus: false`. Source `null` → `enabled: false`, combined with the caller's `enabled`.

Structure and conventions mirror `useRequestQuery` closely (ref-synced source, stable `queryFn`, options shape, null-source gating, error type generic). Overall this looks great — leaving as a comment review (no write access to approve), but nothing blocking.

## Things to watch out for

- **`experimental_streamedQuery` commitment.** The hook is built directly on a TanStack experimental API. The teardown-timing test (`gcTime: Infinity` + unmount → signal aborts) explicitly pins this as a regression guard against a silent TanStack bump. Worth a follow-up here when TanStack stabilizes the API or changes its semantics.
- **`refetch()` never settles for a stream.** A `streamedQuery` for an infinite source never resolves its `refetch()` promise. Tests intentionally `void result.current.refetch()` rather than awaiting it. This is a real usability footgun for consumers reaching for `await refetch()`; see the inline comment for a suggested doc tweak.
- **`initialValue: undefined as T`.** The cast lies about the type, but the reducer `(_previous, chunk) => chunk` replaces it on the first chunk so the lie is contained. The inline comment is clear about why; just flagging for reviewers that this is load-bearing — losing the replace-latest reducer would expose the cast.
- **Abort-wins ordering in the bridge.** The `signal.aborted` check runs before the `failure` check, so an abort that races with a store error ends cleanly rather than throwing. This is the documented intent (abort = teardown, not failure), and the abort-test pins it.
- **`data` is the raw notification, NOT the unwrapped value.** Matches `useSubscription` for consistency. The README's SlotHeight example reads `data.slot`, which is correct for `slotNotifications()` (no envelope); the contrast with account/balance notifications (where you'd read `data.value.lamports`) is only mentioned in passing. The type test `// @ts-expect-error - data is the raw notification, not the unwrapped value` pins it well.

- The test file is large (~400 lines). Worth a careful pass through `makePushStream` / `makeStreamFactory` / `makeStoreSource` — these are the load-bearing test helpers and most assertions flow through them.
- The `getError` / `getState` / `connect` / `retry` jest mocks throwing `'not implemented'` in `createFakeStore` follow the CLAUDE.md guidance for placeholder mocks. ✓
- No new errors added to `@solana/errors`; the bridge reuses `SOLANA_ERROR__REACT__SUBSCRIPTION_CLOSED_WITHOUT_ERROR` from the existing SWR bridge. ✓
- Changeset bump is `minor` for `@solana/react`, correct for a new public hook. ✓
- One small inline nit on the SSR test comment's claim about "queryFn fires inside an effect" — see below.

Nice work — well-tested, the docblocks and README pull their weight, and the design rationale for the `streamedQuery` config is captured in code comments instead of being folklore.

packages/react/src/query/useSubscriptionQuery.ts
```diff
@@ -0,0 +1,147 @@
...
+ * @typeParam TError - The error type TanStack Query will surface on failure.
+ */
+export type UseSubscriptionQueryOptions<T, TError = unknown> = Omit<
+    UseQueryOptions<T, TError, T, QueryKey>,
+    'queryFn' | 'queryKey'
+> &
+    UseRequestOptions;
+
+/**
+ * TanStack Query-backed counterpart to core's `useSubscription`, for streams that have no one-shot
```
trevor-cortex, line 34: Doc nit: worth surfacing the "don't `await refetch()`" gotcha to consumers in the JSDoc, not just in the test comments. A `streamedQuery` on an infinite source never resolves its `refetch()` promise, so a consumer writing the natural `await refetch()` in a click handler will hang forever. The hook tests explicitly `void result.current.refetch()` for this reason — a user reading only the docblock won't know to.

Something like:

> `result.refetch()` reconnects: TanStack aborts the old signal (…). Fire and forget — for a never-ending stream the returned promise never resolves, so don't `await` it.

(Same caveat would be worth a line in the README's "Because the stream never settles…" paragraph.)

packages/react/src/query/useSubscriptionQuery.ts
```diff
@@ -0,0 +1,147 @@
...
+                    return bridgeStoreToAsyncIterable(current.reactiveStore(), combinedSignal);
+                },
+            })(context),
+        [getAbortSignalRef, sourceRef],
+    );
+
+    return useQuery<T, TError, T, QueryKey>({
+        // Sensible defaults for a long-lived stream; all overridable by `queryOptions` below.
+        refetchOnWindowFocus: false,
+        retry: false,
```
trevor-cortex, line 145: Minor: `experimental_streamedQuery({...})` is rebuilt on every queryFn invocation. It's only called once per fetch (so the perf impact is nil), but the config is fully closed-over and never depends on `context` — you could hoist the streamedQuery factory call out of the curry and assign its return directly to `queryFn`:

```ts
const queryFn = useCallback<QueryFunction<T, QueryKey>>(
    experimental_streamedQuery<T, T>({
        initialValue: undefined as T,
        reducer: (_previous, chunk) => chunk,
        refetchMode: 'append',
        streamFn: ({ signal }) => { /* …sourceRef + getAbortSignalRef… */ },
    }),
    [getAbortSignalRef, sourceRef],
);
```

Reads a little cleaner and skips the per-fetch construction. Not worth blocking on — current shape is fine.

packages/react/src/query/bridgeStoreToAsyncIterable.ts
```diff
@@ -0,0 +1,95 @@
...
+            const onAbort = () => wake();
+            signal.addEventListener('abort', onAbort, { once: true });
+            const unsubscribe = store.subscribe(onChange);
+            // Bind the connection to `signal` so an abort tears down the underlying stream too.
+            store.withSignal(signal).connect();
+            try {
+                while (true) {
+                    // Abort wins over everything: an abort is teardown, so end cleanly without
+                    // surfacing the store's incidental abort-driven error state.
+                    if (signal.aborted) return;
```
trevor-cortex, line 80: Subtle: when an `error` lands after a `loaded` value is already buffered in `latest`, the loop checks `failure` before `latest`, so the buffered value is silently dropped in favor of the error. This is probably the right call (once errored, stop yielding) and matches how a real subscription tearing down behaves — but it's not currently asserted by a test, and the docblock doesn't call it out.

Worth either a one-liner in the JSDoc (“a pending value is dropped if an error arrives before the next pull”) or a small test pinning the behavior, so a future refactor doesn't accidentally invert the precedence.

packages/react/src/query/__tests__/useSubscriptionQuery-test.browser.tsx
```diff
@@ -0,0 +1,400 @@
...
+            function Component() {
+                const { data } = useSubscriptionQuery(['ssr'], f.factory);
+                return <p>{data ?? 'no-data'}</p>;
+            }
+            const html = renderToString(
+                <QueryClientProvider client={new QueryClient()}>
+                    <Component />
+                </QueryClientProvider>,
+            );
+            expect(html).toBe('<p>no-data</p>');
```
trevor-cortex, line 410: Comment nit: "TanStack fires the queryFn inside an effect — on the server, effects don't run" isn't quite accurate. TanStack fires the queryFn from `QueryObserver` subscription / `fetchOptimistic` logic, not from a React effect; the reason SSR doesn't fire it here is simply that no observer subscribes during `renderToString` (no client-side mount). Suggest:

```
// On the server, no observer subscribes during renderToString, so TanStack
// never schedules the queryFn.
```

packages/react/README.md
```diff
@@ -454,6 +454,41 @@ function Profile({ userId }: { userId: string }) {
...
+Returns TanStack's native `UseQueryResult<T>`. `data` is the raw notification, exactly as the source emits it, so for RPC subscriptions read `data.value` and `data.context.slot`. Pass `null` for `source` to disable (TanStack's `enabled: false`).
+
+```tsx
+import { useClient } from '@solana/react';
+import { useSubscriptionQuery } from '@solana/react/query';
+import type { ClientWithRpcSubscriptions, SlotNotificationsApi } from '@solana/kit';
+
+function SlotHeight() {
+    const client = useClient<ClientWithRpcSubscriptions<SlotNotificationsApi>>();
+    const { data, error } = useSubscriptionQuery(['slot'], client.rpcSubscriptions.slotNotifications());
```
trevor-cortex, line 470: Tiny doc nit: "Read `data` / `error` / `status` and ignore `isFetching` for subscriptions" is great advice — consider also flagging the corollary that `isLoading` only reports the *initial* connect (it flips false after the first notification and stays false across reconnects). A consumer reaching for `isLoading` to drive a "reconnecting…" spinner during a `refetch()` will be surprised — they'd want `status === 'pending'` semantics from a custom flag instead.

packages/react/src/query/useSubscriptionQuery.ts
mcintyre94: I don't think this is true - it'd be rebuilt on every render in `useCallback`. Currently should be just once per connect.

packages/react/src/query/bridgeStoreToAsyncIterable.ts
mcintyre94: Test added for this

packages/react/README.md
```diff
@@ -454,6 +454,41 @@ function Profile({ userId }: { userId: string }) {
...
+```tsx
+function Ticker() {
+    const { data } = useSubscriptionQuery(['ticker'], (signal: AbortSignal) => streamPrices(signal));
+    return <p>{data ?? '…'}</p>;
+}
+```
+
+### Why no `useActionQuery`?
+
+It would just be a wrapper around Tanstack's built-in [`useMutation`](https://tanstack.com/query/latest/docs/framework/react/guides/mutations) with no additional functionality. Either use `useMutation` or, if you don't need the Tanstack integration, use `useAction`.
```
lorisleiva, line 490: Makes sense, thanks for adding this to the docs.

lorisleiva (approved):
Thank you! We've got such a nice React toolkit now. 🥰

mcintyre94:
### Merge activity

* **Jun 26, 1:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760).
* **Jun 26, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760) couldn't merge this pull request because a downstack PR #1607 failed to merge.
* **Jun 26, 1:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760).
* **Jun 26, 2:00 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760) couldn't merge this pull request because a downstack PR #1628 failed to merge.
* **Jun 26, 2:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760).
* **Jun 26, 3:06 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760) couldn't merge this pull request because a downstack PR #1663 failed to merge.
* **Jun 26, 3:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760).
* **Jun 26, 4:22 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760) couldn't merge this pull request because a downstack PR #1708 failed to merge.
* **Jun 26, 6:12 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760).
* **Jun 26, 8:02 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760) rebased this pull request as part of a merge.
* **Jun 26, 8:12 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1760).

PR #1761, opened 2026-06-17
Infer a literal fixedSize for struct codecs

### LINKED ISSUE #1738: getStructEncoder/Decoder/Codec do not infer a literal fixedSize from their fields
Note: minor, may not be feasible/realistic to fix. Opening as part of addressing #1443 in #1683. We've also discussed this idea before, so worth having an issue for it. 

### Summary

The fixed-size overloads of `getStructEncoder`, `getStructDecoder`, and `getStructCodec` type
their result as `FixedSize*<T>` - i.e. with the default `TSize` of `number` - rather than
inferring the literal byte size from the sum of their fields. Even when every field has a
statically-known literal size, the struct's `fixedSize` is typed as the broad `number`.

### Current behaviour

```ts
import { getStructEncoder } from '@solana/codecs-data-structures';
import { getU32Encoder } from '@solana/codecs-numbers';

const e = getStructEncoder([['value', getU32Encoder()]]);
//    ^? FixedSizeEncoder<{ value: number }>   (size is `number`, not the literal `4`)
```

Compare to the U32Encoder:
```ts
const u32 = getU32Encoder();
//    ^? FixedSizeEncoder<number, 4>
```

Eg:
https://github.com/anza-xyz/kit/blob/93191af2fd088cd1c56cbed65b2ba1acd2a49ff6/packages/codecs-data-structures/src/struct.ts#L85-L87

### Why this matters

Size-aware combinators - getUnion*, getPredicate*, getPatternMatch* decide between a
FixedSize* and a VariableSize* result by comparing the literal sizes of their branches. When
two struct branches have different real sizes but both report `fixedSize: number`, these
combinators can't tell the sizes apart and conservatively widen to a plain Encoder/Decoder/
Codec. Consumers are then forced to cast. If we knew the sizes are different we could return `VariableSize*`. 

### Suggested fix

Add a `TSize` type parameter to the fixed-size overloads and compute it as the literal sum of the
field sizes (mirroring how the runtime already sums getFixedSize over the fields).

### Caveat / open question / why this might not make sense

Type-level addition in TypeScript is pretty hacky. We might be able to do it when the struct is very small?

The motivating example is v1 transaction config where the struct actually only has one field:

https://github.com/anza-xyz/kit/blob/93191af2fd088cd1c56cbed65b2ba1acd2a49ff6/packages/transaction-messages/src/codecs/v1/config.ts#L24-L29

We could avoid the type-level arithmetic by only setting the size if there's exactly one field with a fixed size. 

### Acceptance criteria (typetest)

```ts
getStructEncoder([
    ['a', getU32Encoder()],
    ['b', getU8Encoder()],
]) satisfies FixedSizeEncoder<{ a: number; b: number }, 5>;
```

plutohan:
Closes #1738.

Made `getStructEncoder`/`Decoder`/`Codec` infer a literal `fixedSize` from their fields instead of just falling back to `number`:

```ts
getStructEncoder([
  ['a', getU32Encoder()],
  ['b', getU8Encoder()],
]) satisfies FixedSizeEncoder<{ a: number; b: number }, 5>; // used to be `number`
```

So the size-aware combinators (`getUnion*`, `getPredicate*`, `getPatternMatch*`) can actually tell struct branches of different sizes apart now, which was the point in the issue.

How it works: it's just a type-level sum of each field's `fixedSize`. The catch is the addition builds a tuple of that length, and TS gives up with TS2589 once a tuple gets near ~1000 elements. So to make sure this never blows up in someone's code, I capped it (`MaxInferredStructSize`, 512 for now) — within the cap you get the literal sum, anything bigger or a non-literal field just stays `number` like before. The in-range check walks the cap tuple down instead of building a tuple of the field's size, so a huge single field doesn't trip TS2589 either.

Runtime is untouched — it already summed the sizes, this is purely the types.

Verified with `test:typecheck` (the acceptance criterion from the issue is in there as a typetest), plus lint/prettier and the unit tests on node + browser.

Couple of things I wasn't sure about:
- the cap value (512) — happy to change it. And do you want the size helpers in `struct.ts`, or moved into `codecs-core` so the union/predicate/pattern-match work (#1443, #1683) can reuse them?
- patch or minor changeset? I went patch since it's types-only and a literal size still satisfies the existing `FixedSizeEncoder<T>` tests.

### DIFF
--- a/packages/codecs-data-structures/src/struct.ts
+++ b/packages/codecs-data-structures/src/struct.ts
@@ -29,6 +29,58 @@ import { DrainOuterGeneric, getFixedSize, getMaxSize, sumCodecSizes } from './ut
  */
 type Fields<T> = readonly (readonly [string, T])[];
 
+/**
+ * The maximum total byte size for which {@link getStructEncoder}, {@link getStructDecoder} and
+ * {@link getStructCodec} infer a *literal* `fixedSize` from their fields. Structs whose fields sum
+ * to more than this (or that contain a single field larger than this) keep the broad `number` size.
+ * The cap keeps the type-level addition below TypeScript's recursion ceiling, so it can never
+ * surface as a compile error in a consumer's code.
+ */
+type MaxInferredStructSize = 512;
+
+// Builds a tuple of length `TLength` (tail-recursive accumulator).
+type BuildTuple<TLength extends number, TTuple extends unknown[] = []> = TTuple['length'] extends TLength
+    ? TTuple
+    : BuildTuple<TLength, [...TTuple, unknown]>;
+
+type StructSizeCapTuple = BuildTuple<MaxInferredStructSize>;
+
+// True when `0 <= TSize <= MaxInferredStructSize`. Walks the cap tuple down rather than
+// materialising `BuildTuple<TSize>`, which would blow the recursion limit for a large `TSize`.
+type IsStructSizeInRange<TSize extends number, TCap extends unknown[] = StructSizeCapTuple> = number extends TSize
+    ? false
+    : TSize extends TCap['length']
+      ? true
+      : TCap extends readonly [unknown, ...infer TRest]
+        ? IsStructSizeInRange<TSize, TRest>
+        : false;
+
+// The literal byte size of each field, in declaration order.
+type FieldFixedSizes<TFields extends Fields<{ readonly fixedSize: number }>> = {
+    [I in keyof TFields]: TFields[I][1] extends { readonly fixedSize: infer TSize extends number } ? TSize : never;
+};
+
+// Sums the field sizes into a literal, falling back to `number` if any field (or the running
+// total) is non-literal or exceeds the cap.
+type SumFieldSizes<TSizes extends readonly number[], TAcc extends unknown[] = []> = TSizes extends readonly [
+    infer THead extends number,
+    ...infer TTail extends readonly number[],
+]
+    ? IsStructSizeInRange<THead> extends true
+        ? [...TAcc, ...BuildTuple<THead>] extends infer TNext extends unknown[]
+            ? IsStructSizeInRange<TNext['length']> extends true
+                ? SumFieldSizes<TTail, TNext>
+                : number
+            : number
+        : number
+    : TAcc['length'];
+
+/**
+ * Infers the literal `fixedSize` of a struct as the sum of its fields' sizes, when that sum is
+ * statically known and within {@link MaxInferredStructSize}; otherwise resolves to `number`.
+ */
+type StructFixedSize<TFields extends Fields<{ readonly fixedSize: number }>> = SumFieldSizes<FieldFixedSizes<TFields>>;
+
 type ArrayIndices<T extends readonly unknown[]> = Exclude<Partial<T>['length'], T['length']> & number;
 
 /**
@@ -84,7 +136,7 @@ type GetDecoderTypeFromFields<TFields extends Fields<Decoder<any>>> = DrainOuter
  */
 export function getStructEncoder<const TFields extends Fields<FixedSizeEncoder<any>>>(
     fields: TFields,
-): FixedSizeEncoder<GetEncoderTypeFromFields<TFields>>;
+): FixedSizeEncoder<GetEncoderTypeFromFields<TFields>, StructFixedSize<TFields>>;
 export function getStructEncoder<const TFields extends Fields<Encoder<any>>>(
     fields: TFields,
 ): VariableSizeEncoder<GetEncoderTypeFromFields<TFields>>;
@@ -146,7 +198,7 @@ export function getStructEncoder<const TFields extends Fields<Encoder<any>>>(
  */
 export function getStructDecoder<const TFields extends Fields<FixedSizeDecoder<any>>>(
     fields: TFields,
-): FixedSizeDecoder<GetDecoderTypeFromFields<TFields>>;
+): FixedSizeDecoder<GetDecoderTypeFromFields<TFields>, StructFixedSize<TFields>>;
 export function getStructDecoder<const TFields extends Fields<Decoder<any>>>(
     fields: TFields,
 ): VariableSizeDecoder<GetDecoderTypeFromFields<TFields>>;
@@ -221,7 +273,8 @@ export function getStructCodec<const TFields extends Fields<FixedSizeCodec<any>>
     fields: TFields,
 ): FixedSizeCodec<
     GetEncoderTypeFromFields<TFields>,
-    GetDecoderTypeFromFields<TFields> & GetEncoderTypeFromFields<TFields>
+    GetDecoderTypeFromFields<TFields> & GetEncoderTypeFromFields<TFields>,
+    StructFixedSize<TFields>
 >;
 export function getStructCodec<const TFields extends Fields<Codec<any>>>(
     fields: TFields,
--- a/.changeset/struct-codec-literal-fixed-size.md
+++ b/.changeset/struct-codec-literal-fixed-size.md
@@ -1,0 +1,7 @@
+---
+'@solana/codecs-data-structures': patch
+---
+
+Infer a literal `fixedSize` for `getStructEncoder`, `getStructDecoder`, and `getStructCodec`
+
+When every field is fixed-size, the struct codecs now type their `fixedSize` as the literal sum of the field sizes (for example `5`) instead of the broad `number`. This lets size-aware combinators such as `getUnion*`, `getPredicate*`, and `getPatternMatch*` tell struct branches of different sizes apart. Structs whose fields sum to more than 512 bytes, or that contain a single field larger than that, keep the `number` size so the type-level addition can never exceed TypeScript's recursion limit. This is a type-only change; the runtime is unchanged.
--- a/packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
+++ b/packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
@@ -6,7 +6,14 @@ import {
     VariableSizeDecoder,
     VariableSizeEncoder,
 } from '@solana/codecs-core';
-import { getU32Codec, getU32Decoder, getU32Encoder } from '@solana/codecs-numbers';
+import {
+    getU8Codec,
+    getU8Decoder,
+    getU8Encoder,
+    getU32Codec,
+    getU32Decoder,
+    getU32Encoder,
+} from '@solana/codecs-numbers';
 import { getUtf8Codec, getUtf8Decoder, getUtf8Encoder } from '@solana/codecs-strings';
 
 import { getStructCodec, getStructDecoder, getStructEncoder } from '../struct';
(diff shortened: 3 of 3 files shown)


plutohan:
Hey @mcintyre94, this one is green and ready whenever you get a chance. I left a couple of open questions in the description (the 512 cap, and whether the size helpers should stay local or move into codecs-core for the union / predicate / pattern-match work) and I'm happy to go whichever way you prefer. No rush!

mcintyre94:
@plutohan Thanks for this, and sorry I didn't get back to you quicker. Part of the reason I opened this as an issue is because I think there's some research/feasibility/stress testing to do before committing to a solution here. I appreciate your PR as a demonstration of how the implementation will look if we land it, but I'm a bit unsure about committing to this Typescript complexity. We need to think a bit more about the cost/benefit here - I do think there's a real benefit for small structs and unions/predicates.

I've marked this do-not-close for now so it won't get closed before we make a decision.

WRT to your open questions, I think just having it in struct for now and a patch changeset are right, so no immediate changes. I think the approach is good too, it's just a decision of whether it's worth it.

plutohan:
Thanks @mcintyre94, no worries on the timing, and totally fair to weigh the complexity.

On the cost side: the type machinery is self-contained in struct.ts (about 40 lines) and the cap is the safety net. Because it is gated at 512 bytes and the range check walks a fixed tuple rather than building one from the field size, it can never produce a TS2589 in a consumer's build. Anything over the cap, or any non-literal field, just stays `number` exactly like today. So the downside is bounded, worst case it is a no-op and it cannot break anyone.

On the benefit: it is mainly the unions / predicates / pattern-match work (#1443, #1683) being able to tell struct branches of different sizes apart instead of widening, plus the v1 tx config case.

Happy to help with the stress testing you mentioned if useful, for example running it across the real struct codecs in the repo and checking the tsc compile-time impact. No rush either way, and feel free to park it.

mcintyre94:
@trevor-cortex

packages/codecs-data-structures/src/struct.ts
```diff
@@ -29,6 +29,58 @@ import { DrainOuterGeneric, getFixedSize, getMaxSize, sumCodecSizes } from './ut
...
+    ...infer TTail extends readonly number[],
+]
+    ? IsStructSizeInRange<THead> extends true
+        ? [...TAcc, ...BuildTuple<THead>] extends infer TNext extends unknown[]
+            ? IsStructSizeInRange<TNext['length']> extends true
+                ? SumFieldSizes<TTail, TNext>
+                : number
+            : number
+        : number
+    : TAcc['length'];
```
trevor-cortex, line 76: **Bug: non-tuple field arrays infer a literal `fixedSize` of `0`.**

When `TFields` isn't a `const`-inferred tuple — e.g. a variable typed as `Fields<FixedSizeEncoder<any>>` built dynamically — `FieldFixedSizes` produces an array type like `readonly number[]`, which doesn't match `readonly [infer THead, ...infer TTail]`, so this falls through to `TAcc['length']` and returns `0`:

```ts
declare const fields: Fields<FixedSizeEncoder<number>>;
getStructEncoder(fields);
// inferred as FixedSizeEncoder<..., 0> — should be `number`
```

The same fall-through hits tuples with rest elements (`readonly [['a', U32], ...(readonly ['b', U8])[]]` sums only the leading fixed part). A claimed literal `0` is worse than the old `number` — the size-aware combinators this PR is targeting would actively mis-classify these branches.

Suggested guard at the entry of the type (also catches the rest-element case on recursion, since `TTail` becomes the non-tuple array):

```ts
type SumFieldSizes<TSizes extends readonly number[], TAcc extends unknown[] = []> =
    number extends TSizes['length']
        ? number
        : TSizes extends readonly [infer THead extends number, ...infer TTail extends readonly number[]]
          ? /* existing logic */
          : TAcc['length'];
```

And a typetest, e.g.:

```ts
getStructEncoder({} as Fields<FixedSizeEncoder<number, 4>>) satisfies FixedSizeEncoder<Record<string, number>, number>;
```

trevor-cortex (commented):
## Summary

This PR makes `getStructEncoder`/`getStructDecoder`/`getStructCodec` infer a literal `fixedSize` by summing the fields' literal sizes at the type level, falling back to `number` when any field size is non-literal or the total exceeds a cap (`MaxInferredStructSize = 512`) chosen to stay clear of TS2589. Runtime is untouched; the acceptance criterion from #1738 is covered in the typetests. The overall approach is sound and the cap-tuple walk-down for the range check is a nice way to avoid materialising large tuples.

Two inference issues worth fixing before merge (details inline):

1. **Non-tuple field arrays infer a literal `0`.** When `TFields` is a plain `Fields<FixedSizeEncoder<any>>` array (not a `const`-inferred tuple), `SumFieldSizes` falls through to `TAcc['length']` and claims `fixedSize: 0`. Same failure mode for tuples with rest elements. This needs a guard and a typetest.
2. (Edge case) **Union field sizes resolve to the smallest member** instead of falling back to `number`, because the range check distributes and `BuildTuple` stops at the first matching union member. Suggested one-character-class fix inline.

## Answers to your open questions

- **Cap value**: 512 seems reasonable. One constraint worth documenting: `BuildTuple` relies on TS's tail-recursion elimination, which caps out around ~1000 iterations — so the cap can't be raised near or past that without restructuring `BuildTuple` (e.g. doubling). Also note `IsStructSizeInRange` costs O(cap) instantiations per field, so raising the cap has a typecheck-perf cost too.
- **Location**: keeping the helpers private in `struct.ts` is fine for this PR. If #1443/#1683 land on the same approach, moving `BuildTuple`/range-check/sum helpers into `codecs-core` then makes sense — but I'd let the union/predicate work drive that move rather than exporting speculatively. If a maintainer prefers to move them now to avoid churn, no objection.
- **Changeset bump**: `patch` looks defensible under the repo's policy — it's a types-only narrowing and a literal size still satisfies every existing `FixedSizeEncoder<T>`/`FixedSizeEncoder<T, number>` position. One could argue it's a `minor` (new capability), and since pre-1.0 `minor` may be treated as breaking here, `patch` is the conservative choice. Maintainer call; I wouldn't block on it either way.

- Worth sanity-checking typecheck time on a large downstream consumer (e.g. a Codama-generated client with many wide structs), since every fixed-size struct now pays the per-field range walk.
- The typetests cover the encoder fallback paths but only the happy path for decoder/codec — probably fine given all three share `StructFixedSize`, but if the fix for the non-tuple bug lands, one typetest per entry point for that case would be cheap insurance.
- Once the non-tuple guard is in, an empty-struct case (`getStructEncoder([]) satisfies FixedSizeEncoder<{}, 0>`) would pin down that `0` remains correct for the genuinely-empty tuple while plain arrays fall back to `number`.

*(Submitting as a comment rather than "request changes" — I don't have write access to this repo — but I'd treat point 1 above as blocking.)*

packages/codecs-data-structures/src/struct.ts
```diff
@@ -29,6 +29,58 @@ import { DrainOuterGeneric, getFixedSize, getMaxSize, sumCodecSizes } from './ut
...
+
+// True when `0 <= TSize <= MaxInferredStructSize`. Walks the cap tuple down rather than
+// materialising `BuildTuple<TSize>`, which would blow the recursion limit for a large `TSize`.
+type IsStructSizeInRange<TSize extends number, TCap extends unknown[] = StructSizeCapTuple> = number extends TSize
+    ? false
+    : TSize extends TCap['length']
+      ? true
+      : TCap extends readonly [unknown, ...infer TRest]
+        ? IsStructSizeInRange<TSize, TRest>
+        : false;
```
trevor-cortex, line 56: **Edge case: union sizes resolve to the smallest member instead of falling back to `number`.**

`TSize extends TCap['length']` is a distributive conditional, so for a field typed e.g. `FixedSizeEncoder<number, 4 | 8>` each member is checked independently and the result is `true`. Then `BuildTuple<4 | 8>` stops as soon as `TTuple['length'] extends 4 | 8` first succeeds — i.e. at length 4 — so the struct claims a literal total computed from the *smallest* union member, which is wrong for the `8` case.

Wrapping the check to make it non-distributive fixes it — union sizes then never match any single cap length, fail the range check, and fall back to `number` as they should:

```ts
: [TSize] extends [TCap['length']]
    ? true
```

Worth a typetest alongside: `getStructEncoder([['a', {} as FixedSizeEncoder<number, 4 | 8>]]) satisfies FixedSizeEncoder<{ a: number }, number>;`

packages/codecs-data-structures/src/struct.ts
```diff
@@ -29,6 +29,58 @@ import { DrainOuterGeneric, getFixedSize, getMaxSize, sumCodecSizes } from './ut
...
 type Fields<T> = readonly (readonly [string, T])[];
 
+/**
+ * The maximum total byte size for which {@link getStructEncoder}, {@link getStructDecoder} and
+ * {@link getStructCodec} infer a *literal* `fixedSize` from their fields. Structs whose fields sum
+ * to more than this (or that contain a single field larger than this) keep the broad `number` size.
+ * The cap keeps the type-level addition below TypeScript's recursion ceiling, so it can never
+ * surface as a compile error in a consumer's code.
+ */
+type MaxInferredStructSize = 512;
```
trevor-cortex, line 39: Nit: worth noting in this docblock that the cap must stay comfortably below ~1000 — `BuildTuple` depends on TS's tail-recursion elimination, which gives up around 1000 iterations, so bumping this constant past that would reintroduce the TS2589 this cap exists to prevent.

lorisleiva:
I've not read this whole thread so apologies in advance if I've missed something but this feels like an overkill solution for the issue it's tackling. I don't believe we should start doing TypeScript arithmetics (our type-system is already pretty complex as it is). My thought on the original issue is we just needed an easy way to identify a struct that only contains fixed-size fields and return a fixed-size struct (the actual size number does not need to be known by the type-system).

mcintyre94:
@lorisleiva The referenced issue here (#1738) is for the fact that it's currently `Number` and not the fixed size, but I'm definitely happy to just leave it as-is. I opened the issue speculatively, and it says it might not be worth addressing. I agree the complexity is a lot, and when I opened the issue I thought that would probably be the case.

The motivating example in the issue is:

```ts
 function getCompiledTransactionConfigValueEncoder(): VariableSizeEncoder<CompiledTransactionConfigValue> { 
     return getPatternMatchEncoder<CompiledTransactionConfigValue>([ 
         [value => value.kind === 'u32', getStructEncoder([['value', getU32Encoder()]])], 
         [value => value.kind === 'u64', getStructEncoder([['value', getU64Encoder()]])], 
     ]) as unknown as VariableSizeEncoder<CompiledTransactionConfigValue>; 
 } 
```

Here the two `getStructEncoder` paths both type as size `Number`, while each is one small fixed size field. If the type system knew their actual sizes (which it would if they weren't wrapped in a struct) we wouldn't need the cast to `VariableSizeEncoder`. Because it doesn't know if their sizes are the same or different, it widens to `Encoder`, which is sound but slightly annoying.

I think on balance it's probably not worth it though.

lorisleiva:
Gotcha! Yeah I don't think it's worth it. I know we use a lot of complex TypeScript features to offer good type-safety but I think we should draw the line at TypeScript arithmetics because it's not just complex it's also really finicky and has its own limitations.

One thing we could do is return a fixed-size codec of size `N` if the struct contains only a single fixed-size field of size `N`. For anything more complex, we can always type-cast the struct before feeding it to the next codec. Wdyt?

mcintyre94:
I think just supporting a single field is a nice middle ground. Will close this PR for now because that implementation can be much simpler. But thanks for the exploration @plutohan!

PR #1766, opened 2026-06-18
Update parseJsonWithBigInts to protect against overly large numbers

mcintyre94:
#### Problem

If an RPC returns a `$n` field then this gets parsed as a bigint. There was previously no bound on the number we'd attempt to convert to a bigint, nor validation that it was convertible to a bigint. This meant that a malformed RPC response could potentially freeze/crash JS.

#### Summary of Changes

Update our parsing, so it now validates the `$n` value is a number. It also validates that it materializes to a number with no more than 1000 digits, meaning it'll parse in a fraction of a millisecond. We don't currently have a usecase for RPC returning numbers any bigger than this.

Thanks to @nbelenkov and @EQSTLab for telling us about this issue!

### DIFF
--- a/.changeset/humble-words-shout.md
+++ b/.changeset/humble-words-shout.md
@@ -1,0 +1,5 @@
+---
+'@solana/rpc-spec-types': patch
+---
+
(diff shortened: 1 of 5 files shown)


lorisleiva (approved):
Perfect, thank you!

PR #1776, opened 2026-06-22
Add getTransactionsForAddress to RPC API

mcintyre94:
#### Problem

It's not part of the JSON-RPC spec, but most major RPC providers have `getTransactionsForAddress`. The types are complex and annoying to use without a Kit helper. 

#### Summary of Changes

This PR adds `getTransactionsForAddress` to the Kit RPC.

While this isn't part of the current spec or the Agave RPC, it's part of the new RPC spec and is already widely supported.

The refactor pulls the shared types from `getTransaction`.

Also added `meta.costUnits: bigint` to both `getTransaction` and `getTransactionsForAddress`, see https://solana.com/docs/rpc/http/gettransaction 

See:
- Source of truth, superbank RPC types: https://github.com/solana-rpc/superbank/blob/7db011aa36616034bf117e2ef1ee8a722689b33f/crates/superbank-rpc/src/handlers/types.rs#L103 
- Current Triton docs: https://docs.triton.one/chains/solana/gettransactionsforaddress
- Current Helius docs: https://www.helius.dev/docs/rpc/gettransactionsforaddress


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds `GetTransactionsForAddressApi` to `@solana/rpc-api`, exposing the new method that combines `getSignaturesForAddress`-style address-history discovery with `getTransaction`-style per-transaction fetching, plus server-side filtering, bidirectional sorting, and cursor-based pagination. As part of the change, the shared transaction-meta types (`TransactionMetaBase`, `TransactionJson`/`TransactionJsonParsed`, inner-instruction/loaded-address shapes, etc.) move out of `getTransaction.ts` into a new `transaction-meta.ts` module so they can be reused by both methods, and an optional `meta.costUnits: bigint` field is added — which now also surfaces on `getTransaction`.

The extraction itself is mechanical and looks clean: the moved types are byte-identical to the originals, just re-exported (now public types in `transaction-meta.ts`) and re-imported by `getTransaction.ts`. Index wiring (`SolanaRpcApiForAllClusters`, exports, numeric-keypath entries) is consistent with the existing patterns for `getTransaction` and `getSignaturesForAddress`.

## Things to watch out for

- **`version` is not surfaced on full-mode responses.** Compare `getTransaction`, which conditionally adds `version: TransactionVersion` when `maxSupportedTransactionVersion` is set. The full-mode return types here omit it entirely, even though the underlying response per the spec/Helius/Triton docs includes `version` for versioned transactions. Worth confirming whether this was intentional or an oversight — see inline comment.
- **`OptionalTransactionAddressTableLookups` and `Partial<TransactionMetaLoadedAddresses>` are always-optional**, regardless of `maxSupportedTransactionVersion`. `getTransaction` keys these off the type parameter (required when versioned, absent when legacy-only). This PR takes a more permissive stance — the fields are always optional. Defensible, but worth flagging as a deliberate divergence from the `getTransaction` modeling.
- **`limit` default doc may be incomplete.** Helius docs note a cap of 100 for `transactionDetails: 'full'` vs 1000 for signatures. The JSDoc currently states a single 1000 cap. If providers cap full-mode lower, callers may be surprised.
- **Overload resolution for invalid combinations.** The overloads correctly require `transactionDetails: 'full'` when `encoding` is one of the full encodings, but there's no typetest covering what happens if a caller passes `{ encoding: 'jsonParsed' }` without `transactionDetails`. Excess-property checks should catch this on object literals, but a `@ts-expect-error` test would lock the behavior in.
- **No runtime tests.** `getTransaction` has all-`it.todo` tests too, so this is the established pattern in this package — coverage is provided by the typetest. Calling that out so subsequent reviewers don't go looking for unit tests.

- The `transaction-meta.ts` extraction makes previously-private types **part of the package's public surface** (they're not exported from `index.ts`, but the module is now importable from the package by virtue of being a non-private `.ts` file). That's not necessarily wrong, but worth being aware of — any future changes to these shapes are now potentially observable.
- The `costUnits` field addition is a soft user-visible change on `getTransaction` too. The changeset (correctly) calls this out and the bump level is `minor`, which is appropriate.
- Allowed-numeric-keypath entries for `getTransactionsForAddress` are a straight port of the `getTransaction` entries with a `['data', KEYPATH_WILDCARD, ...]` prefix — looks correct given the response envelope shape.
- Worth a second pair of eyes on the spec mapping (filters, sortOrder enum, status enum values, paginationToken semantics) against the superbank source-of-truth linked in the PR description, since that's the canonical reference and providers differ in the details they document.

trevor-cortex:
**Missing `version` field on full-mode responses?**

`GetTransactionsForAddressFullBase` doesn't surface `version`, and none of the full-mode return types intersect it in either. Compare `GetTransactionApi`, which adds `version: TransactionVersion` to the response when `maxSupportedTransactionVersion` is set:

```ts
// getTransaction.ts
(TMaxSupportedTransactionVersion extends void
    ? Record<string, never>
    : { version: TransactionVersion })
```

The Helius/Triton/superbank docs all show `version` in the per-transaction payload for versioned transactions, so callers will likely want to read it. Was this an intentional omission, or should the full-mode results include `version: TransactionVersion` (possibly gated on `maxSupportedTransactionVersion` like `getTransaction`)?

(Anchoring here on `GetTransactionsForAddressFullBase`; same point applies to all four full-mode overloads further down.)

mcintyre94:
Excluding version was intentional because that's the current behaviour of Superbank and the Triton docs don't mention it, but Helius does include it. I've opened a PR to add it to Superbank: https://github.com/solana-rpc/superbank/pull/20

If that gets merged then we can add the `version` field, and it'll align with `getTransaction`. I'll hold this in draft for now.

mcintyre94:
Updated after https://github.com/solana-rpc/superbank/pull/20 was merged, we now include `version: TransactionVersion` when `maxSupportedTransactionVersion` is set.

Please re-review this PR @trevor-cortex

trevor-cortex (commented):
Re-review — LGTM (posting as COMMENT since I can't APPROVE without repo access).

The main open point from my prior pass was that `version` (and the other version-gated fields) needed to be conditional on `maxSupportedTransactionVersion`. That's now modelled correctly:

- Each `full` encoding has a paired legacy / version-specified overload. The version-specified overloads intersect in `GetTransactionsForAddressVersion` (`version: TransactionVersion`), `TransactionMetaLoadedAddresses`, and `TransactionAddressTableLookups`; the legacy overloads omit all three.
- The typetest asserts the presence/absence in both directions with `@ts-expect-error` on `version`, `addressTableLookups`, and `meta.loadedAddresses` for each of `json` / `jsonParsed` / `base58` / `base64`. Nice coverage.
- The `signatures` overload sits last and correctly rejects `encoding` and `processed` commitment via the negative tests.

Other things I re-checked and am happy with:

- **`limit` caps** are now documented (1000 signatures / 100 full).
- **`paginationToken` format** is documented as `"<slot>:<position>"`, matching the example in the typetest.
- **Response envelope** (`{ data, paginationToken }`) is intentionally bare (not `context`/`value`), and the typetest calls that out explicitly.
- **`transaction-meta.ts` extraction** is internal only — none of those types are re-exported from `packages/rpc-api/src/index.ts`, so the public surface remains just `GetTransactionsForAddressApi`. Good call keeping them un-exported.
- **Keypath registration** for `getTransactionsForAddress` mirrors `getTransaction`'s exactly (transposed under `['data', KEYPATH_WILDCARD, ...]`) and additionally covers `transactionIndex`, which is the non-obvious one.
- **`meta.costUnits` on `getTransaction`** is optional, so it's a non-breaking widening. Flagged in the changeset — good.
- **`Exclude<TransactionVersion, 'legacy'>`** for `maxSupportedTransactionVersion` is a nice constraint that lines up with the existing `getTransaction` semantics ("set to `0` to include version-0").

Two very small things, none blocking:

1. `getTransaction` uses a generic `TMaxSupportedTransactionVersion` parameter with conditional types; `getTransactionsForAddress` doubles the overloads (legacy + version-specified) with concrete configs. Both work — the doubled form gives clearer per-case JSDoc, so I don't think it's worth changing, but it does mean the two files now diverge in style. Worth being aware of if a third method ever wants to pick a house pattern.
2. `minContextSlot` is in the common config but I didn't verify against the superbank source-of-truth that it's actually a supported knob for this method (the linked Triton/Helius docs don't mention it either way). Worth a quick sanity check before merge, but low risk — extra fields are typically ignored server-side.

Happy to see this go in. Nice work on the version modelling and the typetest expansion — the negative assertions make the intent unambiguous.

mcintyre94:
* **#1814** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1814?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1776** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1776?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1776?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
Confirmed that `minContextSlot` is supported by the superbank source.

lorisleiva (approved):
Nice, thanks for that!

PR #1778, opened 2026-06-23
Refactor card-tabs to work in StrictMode

mcintyre94:
The /docs page would error under `pnpm dev` using the latest React/Next versions, because they do a StrictMode-style double render:

```markdown
## Error Message
CardTab at index 4 does not have a matching card definition. Make sure the number of CardTab children matches the number of cards.


    at CardTab (src/components/card-tabs.tsx:128:15)
    at _createMdxContent (content/docs/index.mdx.js?collection=docs&hash=1778160327373:27375:15)
    at MDXContent (content/docs/index.mdx.js?collection=docs&hash=1778160327373:40652:8)
    at Page (src/app/docs/[[...slug]]/page.tsx:31:17)

## Code Frame
  126 |
  127 |     if (!value) {
> 128 |         throw new Error(
      |               ^
  129 |             `CardTab at index ${index} does not have a matching card definition. ` +
  130 |                 `Make sure the number of CardTab children matches the number of cards.`,
  131 |         );
```

The cause of this was the `collectionRef.current` being iterated as a render side effect:

```typescript
    // Collection pattern: each CardTab calls register() during render
    // to get its index. The counter resets on each render cycle.
    const collectionRef = useMemo(() => ({ current: 0 }), []);
    collectionRef.current = 0;

    const ctx = useMemo(
        () => ({
            values,
            register: () => collectionRef.current++,
        }),
        [values, collectionRef],
```

This pattern is not StrictMode safe because the double render increments `collectionRef.current` past the number of panels.

The new version is deterministic, based on iterating the children, so does not rely on a render side effect.

Including in the React stack because it's low priority and was found while adding React docs.

### DIFF
--- a/docs/src/components/card-tabs.tsx
+++ b/docs/src/components/card-tabs.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { Tabs, TabsContent, TabsList, TabsTrigger } from 'fumadocs-ui/components/ui/tabs';
-import { createContext, useContext, useLayoutEffect, useMemo, useState, type ReactNode } from 'react';
+import { Children, cloneElement, isValidElement, useLayoutEffect, useMemo, useState, type ReactNode } from 'react';
 import { cn } from 'fumadocs-ui/utils/cn';
 import { FlaskConicalIcon, GlobeIcon, LaptopIcon, PuzzleIcon, type LucideIcon } from 'lucide-react';
 
@@ -29,11 +29,6 @@ function toValue(title: string): string {
     return title.toLowerCase().replace(/\s+/g, '-');
 }
 
-const CardTabContext = createContext<{ values: string[]; register: () => number }>({
-    values: [],
-    register: () => 0,
-});
-
 /**
  * Renders a tab list as a grid of selectable cards.
  *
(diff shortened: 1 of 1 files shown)


mcintyre94:
* **#1778** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1778?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1778?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Clean fix — looks good to me.

The previous version mutated `collectionRef.current` during render (both the `collectionRef.current = 0` reset and `register()` increments), which is exactly the kind of render-phase side effect that StrictMode's double render is designed to surface — the parent's reset ran once per parent render, but children rendered twice, so the counter walked off the end of `values` on the second pass.

The new approach is render-pure: the parent walks its children once, assigns each `CardTab` its `index` and `value` via `cloneElement`, and the children become dumb consumers of injected props. Deterministic across any number of render passes.

A few notes for follow-up reviewers:

- **Children shape**: `Children.map` iterates direct children only and does not descend into nested fragments. Current MDX usage in `docs/content/docs/index.mdx` passes `<CardTab>` elements directly under `<CardTabs>` with no wrapping fragments, so this is fine today. Worth keeping in mind if anyone ever wraps `CardTab`s in a fragment or conditional component — those would be skipped by `isValidElement<CardTabProps>` and silently render as-is.
- **Public type surface**: `index` and `value` are now optional props on `CardTabProps`, which technically means a consumer could pass them manually from MDX. The JSDoc warns against it, which is the right call for an internal docs component — not worth hiding behind a symbol or branded type.
- **No changeset needed**: docs-only change under `docs/`, consistent with the repo's changeset policy.
- **Minor**: the reformatted continuation inside the `throw new Error(...)` template (around line 123) dedents the second line to the same depth as the `+` operand, which reads a bit oddly compared to the previous extra-indented style. Not a blocker — the `docs` package doesn't gate on prettier in CI — but `pnpm style:fix` at the repo root may or may not normalize it.

lorisleiva (approved):

mcintyre94:
Note: Re-parented this on main, because it fixes `pnpm dev` and is useful outside of the react stack. Merging it separately.

PR #1788, opened 2026-06-24
docs: implement a custom toc only showing visible headers

mcintyre94:
#### Problem

The `/docs` landing page has a confusing table of contents, showing every title on the page, while the page only actually renders one tab:

Screenshot 2026-06-24 at 18 09 27

#### Summary of Changes

This PR fixes it by passing the docs a custom `tableOfContents` which renders our component `VisibleToc` which filters to only the items that are rendered, based on `querySelector`. The actual ToC rendering uses fumadocs components so it looks the same:

Screenshot 2026-06-24 at 19 19 29

The same is applied to the mobile popover ToC, which TIL about:

Screenshot 2026-06-24 at 19 27 26

Other pages have an unchanged ToC because they render all their content, but this should automatically work if we add any form of filtering/tabs etc to any other page in the future too.

### DIFF
--- a/docs/src/app/docs/[[...slug]]/page.tsx
+++ b/docs/src/app/docs/[[...slug]]/page.tsx
@@ -1,5 +1,6 @@
 import { mdxComponents } from '@/app/layout.config';
 import { LLMCopyButton, ViewOptions } from '@/components/page-actions';
+import { VisibleToc } from '@/components/visible-toc';
 import { overridenMdxComponents } from '@/lib/Overrides';
 import { docsSource } from '@/lib/source';
 import { getPageTreePeers } from 'fumadocs-core/server';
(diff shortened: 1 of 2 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Nicely scoped fix for a real UX paper-cut. The approach — wrap the desktop TOC in its own `AnchorProvider` seeded with a DOM-filtered subset, and reapply the same filter to the mobile popover via a side-effect — is pragmatic given Fumadocs doesn't expose the mobile TOC for replacement, and the docblocks do a great job explaining *why* each piece exists (the SSR-matching initial state, the `childList`-only observer, the nested provider). That context is exactly what a future reader will want.

A few things worth a second look, none blocking:

- **MutationObserver scope.** `subtree: true` on `<article>` fires on every DOM change in the body — code-highlight tweaks, animation class toggles, scroll-spy markers, etc. The `unchanged` reference check rightly prevents re-renders, but each fire still does `O(toc.length)` `querySelector` calls. Probably fine in practice, but if you ever see jank on heavy pages, debouncing the recompute (rAF or microtask) would be cheap insurance.
- **Mobile popover effect timing.** `document.getElementById('nd-tocnav')` is captured once when the effect runs and bails if not found. It should be present by the time the page mounts, but if Fumadocs ever lazy-mounts that nav (or if the structure changes in a future version) the filter would silently no-op. Not worth fixing now, just something to be aware of when bumping Fumadocs.
- **Initial flicker.** First client render uses the full `toc` (for SSR parity), then the effect trims it. On the landing page this means the TOC briefly shows all entries before settling. Likely imperceptible, but worth a quick eyeball on a slower device to confirm.

No changeset needed (docs-only). LGTM from me — can't formally approve on this repo, leaving as a comment review.

lorisleiva (approved):
Oh wow, I hadn't even noticed how horrible the original version looked like. Thank you for fixing this!

PR #1789, opened 2026-06-26
Add typetest for getHighestSnapSlot

atharva-patill:
#### Problem

A previous PR added typetest coverage for `getEpochInfo()`, but `getHighestSnapshotSlot()` still doesn't have a dedicated typetest.

#### Summary of Changes

This PR adds a typetest for `getHighestSnapshotSlot()`.

It verifies the inferred return type of:

```ts
const result = await rpc.getHighestSnapshotSlot().send();
```

and checks that it matches:

```ts
result satisfies Readonly<{
    full: Slot;
    incremental: Slot | null;
}>;
```

This helps ensure the no-argument RPC method continues to infer the expected response type, including preserving the branded `Slot` type and the nullable `incremental` field.

#### Testing

```bash
pnpm compile
pnpm --filter @solana/rpc-api test:typecheck
```

### DIFF
--- a/packages/rpc-api/src/__typetests__/get-highest-snapshot-slot-typetest.ts
+++ b/packages/rpc-api/src/__typetests__/get-highest-snapshot-slot-typetest.ts
@@ -1,0 +1,16 @@
+import type { Rpc } from '@solana/rpc-spec';
+import type { Slot } from '@solana/rpc-types';
+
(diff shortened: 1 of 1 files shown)


mcintyre94 (approved):
LGTM, thankyou!

PR #1795, opened 2026-06-29
RPC updates for Agave 4.1.0

mcintyre94:
#### Problem

CI has moved on to Agave 4.1.0, which has breaking changes to RPC 

#### Summary of Changes

- stake config is no longer parsed by the RPC
- parsed rent sysvar is now just `lamportsPerByte`, a different shape to previously 
- `warmupCooldownRate` removed from parsed stake delegation
- `inflationRewardsCommissionBps` added to `getVoteAccounts`
- also aligned `number` override fields with the server

### DIFF
--- a/.changeset/ripe-badgers-guess.md
+++ b/.changeset/ripe-badgers-guess.md
@@ -1,0 +1,32 @@
+---
+'@solana/rpc-parsed-types': major
+'@solana/rpc-transformers': major
+'@solana/rpc-api': major
+'@solana/rpc-graphql': minor
+---
+
+Update RPC and parsed-account types to match Agave 4.1.0, and surface basis-points commission and vote-latency fields as numbers instead of bigints.
+
+**BREAKING CHANGES**
+
+**Parsed vote-account commissions and vote latency are now `number` instead of `bigint`.** Agave returns these as small bounded integers (`u16`/`u8`), so kit no longer upcasts them. This affects `blockRevenueCommissionBps`, `inflationRewardsCommissionBps`, and each vote's `latency` on `JsonParsedVoteAccount`.
+
+```diff
+- const bps: bigint = voteAccount.inflationRewardsCommissionBps;
++ const bps: number = voteAccount.inflationRewardsCommissionBps;
+```
+
+**The parsed rent sysvar is now a union of the current and pre-4.1.0 shapes.** Agave 4.1.0 reshaped the rent sysvar from `{ burnPercent, exemptionThreshold, lamportsPerByteYear }` to `{ lamportsPerByte }`. The type is now a union of both, so consumers must narrow before accessing the legacy fields. Narrow on the presence of `lamportsPerByte` (current) versus `lamportsPerByteYear` (deprecated).
+
+```diff
+- const perByteYear = rent.info.lamportsPerByteYear;
++ const perByte = 'lamportsPerByte' in rent.info
++     ? rent.info.lamportsPerByte
++     : rent.info.lamportsPerByteYear;
+```
+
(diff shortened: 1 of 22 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Updates RPC types and tests to track Agave 4.1.0:

- **Stake config program** is no longer recognized by the RPC's JSON parser → `jsonParsed` requests fall back to annotated base64. The `JsonParsedStakeConfigAccount` type stays as `@deprecated` for older validators, and the test fixtures now assert the base64 fallback shape.
- **Parsed rent sysvar** was reshaped from `{ burnPercent, exemptionThreshold, lamportsPerByteYear }` to `{ lamportsPerByte }`. Modeled as a union (`JsonParsedRentAccount | JsonParsedRentAccount_DEPRECATED`) so callers narrowing on either field stay correct across validator versions.
- **Parsed stake delegation**: `warmupCooldownRate` removed in 4.1.0 → now optional + `@deprecated`.
- **`getVoteAccounts`**: new optional `inflationRewardsCommissionBps?: number` field, with both `current` and `delinquent` keypaths allow-listed so it's not upcast to `bigint`.
- **Parsed vote account**: `blockRevenueCommissionBps`, `inflationRewardsCommissionBps`, and per-vote `latency` retyped from `bigint` to `number` (small bounded ints on the Rust side) and added to the `jsonParsedAccountsConfigs` allow-list.
- **Tests**: `sendTransaction`/`simulateTransaction` error assertions relaxed to match only the error code + presence of `__serverMessage`, since the validator's deserialization wording shifts between Agave releases. Sensible change.

The approach throughout (additive unions, optional fields, `@deprecated` tags rather than hard replacements) lines up well with the new CLAUDE.md guidance on backward-compatible widenings, and the changeset accurately calls out the breaking surface and gives consumers a narrowing migration snippet. Overall this looks good — a few small things worth verifying below.

## Things to verify

- **Agave parity for `inflationRewardsCommissionBps` width.** Worth a quick check that the Rust field on `RpcVoteAccountInfo` in `agave` 4.1.0 is genuinely a small bounded integer (`u16`/basis points) and not anything that could overflow `2^53`. Same for the parsed vote account's `blockRevenueCommissionBps` and `latency` (`u8`). The 5000 value in the test reads like bps, so this is almost certainly correct — just flagging the source-of-truth check called out by CLAUDE.md.
- **`blockRevenueCommissionBps` on top-level `getVoteAccounts`.** The parsed vote account exposes both `blockRevenueCommissionBps` and `inflationRewardsCommissionBps`, but `getVoteAccounts`'s `VoteAccount<T>` only adds `inflationRewardsCommissionBps`. If Agave 4.1.0's `RpcVoteAccountInfo` also surfaces `blockRevenueCommissionBps`, that would be missing here (both the type and the allow-list keypath). If it doesn't, ignore.
- **`latency` optionality.** Retyped from `bigint` to `number` and stayed non-optional. If Agave's `RpcVote.latency` is `Option<u8>` / `#[serde(skip_serializing_if)]` (it was added later than the rest of the struct), it might be worth marking `latency?:` — but if it's always present on 4.1.0 output and pre-`latency` validators aren't a concern, the current shape is fine.

- Pre-existing allow-list entries for the legacy rent shape (`burnPercent`, `exemptionThreshold`) and stake-config (`slashPenalty`, `warmupCooldownRate`) are intentionally kept — they're still needed to keep those fields as `number` when older validators return them.
- The `lamportsPerByte` / `lamportsPerByteYear` fields are typed `StringifiedBigInt` (server returns them as strings), so no numeric-allow-list entry is needed for the new rent shape.
- The error-message relaxation in `send-transaction-test.ts` / `simulate-transaction-test.ts` switches from `toThrow(new SolanaError(...))` (which matches on the constructed error's message) to `expect.objectContaining({ context: ... })`. That works because `SolanaError` exposes `context` containing `__code` and `__serverMessage` — worth a quick sanity check that this is the matcher pattern used elsewhere in the suite for `SolanaError`.

PR #1811, opened 2026-07-01
Move bridgeStoreToAsyncIterable from react/query to subscribable/

mcintyre94:
#### Problem

When we added the Tanstack Query versions of `useSubscription` and `useTrackedData`, we added a bridge to an `AsyncIterable` from the `ReactiveStreamStore`. This is how TanStack's (experimental) `streamedQuery` is implemented.

This bridge can be useful for consumers outside of Tanstack Query, which want to consume an `AsyncIterable`. It's likely that more/future libraries will eventually standardise on this data structure too.

We intentionally didn't just use the `AsyncIterable` from `rpcSubscriptions.method().subscribe()` for this, for compatibility with `useSubscription` and also because the store lifecycle is better suited for UI, always providing the latest data and providing status etc.

So any other consumer looking for an `AsyncIterable` in reactive UI would benefit from this bridge, which was internal in `react/query`.

#### Summary of Changes

This PR moves `bridgeStoreToAsyncIterable` to `packages/subscribable` as a public export.

It's made less opinionated and now just observes the store, the caller still controls connect/reset. This aligns it with the rest of the `subscribable` API. The `react/query` consumers are updated accordingly.


mcintyre94:
@trevor-cortex

packages/react/src/query/useTrackedDataQuery.ts
```diff
@@ -152,6 +155,7 @@ export function useTrackedDataQuery<
                     // signal (when present) so aborting either tears the stream down.
                     const combinedSignal = userSignal ? AbortSignal.any([signal, userSignal]) : signal;
                     const store = createReactiveStoreWithInitialValueAndSlotTracking(current);
+                    store.withSignal(combinedSignal).connect();
```
trevor-cortex, line 159: Small consistency question, not a blocker: the docs and README for `bridgeStoreToAsyncIterable` establish the caller-owns-lifecycle pattern with a `try { for await … } finally { store.reset(); }` example. Here (and analogously in `useSubscriptionQuery.ts`) we `connect()` but never `reset()` — we rely on the signal abort tearing down the connection and the store being GC'd when the queryFn returns.

That's functionally fine (ephemeral per-attempt store, no consumers left), but it's a bit at odds with the pattern the docs are teaching. Options:

1. Actually `reset()` in a `finally` around the returned iterable (wrap the return to add teardown). Costs a small closure but matches the documented pattern exactly.
2. Add a short comment here ("per-attempt store; abort tears the connection down and the store is GC'd, so no explicit `reset()` is needed") so a future reader isn't confused by the docs/hook divergence.

I'd lean toward option 2 — the cost of option 1 is real (an extra async wrapper around the iterable) and the correctness case is airtight. But worth being explicit about the choice.

trevor-cortex (commented):
Nice refactor — LGTM overall (leaving as `COMMENT` since I don't have approval rights on this repo). Moving `bridgeStoreToAsyncIterable` to `@solana/subscribable` is the right call: it's a store adapter with no React/TanStack coupling, and the behavioural change (bridge only observes; caller owns `connect()`/`reset()`) aligns it with the rest of the package where a store does nothing until you connect it. The new "seeds from current snapshot" semantic falls out cleanly and is a nice quality-of-life improvement — order of connect vs iteration no longer matters.

Correctness spot-checks all pass:
- Seeding: `subscribe` then a synchronous `onChange()` is safe — latest-wins collapses any duplicate emission if the store also fires synchronously on subscribe.
- Already-aborted signal: the `while (true)` loop's first check is `signal.aborted`, so an entering-aborted state returns cleanly before touching `failure`/`latest`. Abort still wins over an incidental store error triggered by the abort itself.
- Already-errored store: `onChange()` seeds `failure` and the loop throws on first pull — covered by the new `throws an error already present when iteration begins` test.
- Hook call sites: both `useSubscriptionQuery` and `useTrackedDataQuery` connect *before* handing the store to the bridge, so the pending first pull will see either the seeded value or park correctly.

Error code addition is well-placed (sequential in the reserved 8195000–8195999 subscribable range, added to the union alphabetically, matching message). Retaining `SOLANA_ERROR__REACT__SUBSCRIPTION_CLOSED_WITHOUT_ERROR` for the SWR bridge is correct — you can't remove codes, and re-pointing SWR would be a breaking behavioural change. Worth a follow-up thought (not this PR): the two codes now describe the same condition ("stream closed without a payload"), so consumers checking one won't catch the other. If SWR ever migrates in a major bump it'd be nice to converge on the new subscribable code.

Docs (`reactive-stores.mdx`, `subscribable/README.md`) are excellent — the contrast against `createAsyncIterableFromDataPublisher` (queued, publisher-level) and the RPC subscription's own iterable (raw transport, no store lifecycle) is exactly the framing readers need. Docblock covers `@param`/`@throws`/`@see`/`@example` per the skill.

Test updates look right, and the switch to throwing stubs for `connect`/`withSignal`/`reset` in `createFakeStore` is textbook per the AGENTS.md placeholder-mock guidance — an accidental lifecycle call now fails loudly.

One consistency question for subsequent reviewers to weigh in on (see inline on `useTrackedDataQuery.ts`).

Changeset marks `@solana/subscribable`, `@solana/errors`, `@solana/kit`, `@solana/react` all `minor`. That looks right: new public export in subscribable+kit+errors; react's `useSubscriptionQuery`/`useTrackedDataQuery` change is behavioural (no more `reset()` on teardown) but the consumer-observable state is the same (connection torn down via signal abort, store GC'd), so `minor` is defensible pre-1.0.

packages/react/src/query/useTrackedDataQuery.ts
mcintyre94: Comment added

lorisleiva (approved):
Makes sense!

PR #1816, opened 2026-07-04
Add typetest for getMinimumBalanceForRentExemption

atharva-patill:
#### Summary

Adds a typetest for `getMinimumBalanceForRentExemption()`.

This continues the recent typetest coverage for RPC methods that take arguments by covering a method with a required `bigint` argument and a direct branded return type.


mcintyre94 (dismissed):
Thanks!

PR #1817, opened 2026-07-04
Add typetest for getInflationReward

atharva-patill:
#### Summary

Adds a typetest for `getInflationReward()`.

This continues the recent typetest coverage for RPC methods that take arguments by covering a method with a required array argument and nullable elements in the response.


mcintyre94 (approved):
Thanks!

PR #1818, opened 2026-07-04
Add typetest for getSlotLeaders

atharva-patill:
#### Summary

Adds a typetest for `getSlotLeaders()`.

This continues the recent typetest coverage for RPC methods that take arguments by covering a method with multiple required positional arguments and a direct `Address[]` return.


mcintyre94 (approved):
Thanks for adding all these!

atharva-patill:
> Thanks for adding all these!

Glad I could help!

PR #1820, opened 2026-07-06
Wrap long inline code inside callouts

### LINKED ISSUE #1802: Docs nit: callout doesn't work well with long error codes
Do something to make this look better:

Image

(in the new transaction-introspection page)

plutohan:
Long error codes in callouts overflow past the border on narrower screens. Allow inline code inside callouts to wrap anywhere so it stays inside the box.

before
before-660
after
after-660

Fixes #1802


lorisleiva (dismissed):
Looks good to me, thanks!

mcintyre94 (approved):
This shouldn't have got closed, thanks!

PR #1825, opened 2026-07-10
Update the react-app example to use the wallet plugin

mcintyre94:
This PR updates the react-app example to use the `walletSigner` plugin, and the react hooks from the wallet plugin. 

It creates a `Client` with only the `walletSigner` plugin, future iterations will add more plugins (namely `solanaRpc` and the system program plugin).

It maintains as much functionality from the previous react app as possible, I've flagged inline where the functionality differs.

This PR now uses 0.14.0 of the wallet plugin, which requires `client` as the first param of each hook.

The PR also bumps wallet-standard dependencies in `@solana/react` to match those in the wallet plugin (the latest versions).


mcintyre94:
* **#1825** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1825?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1825?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

examples/react-app/src/components/ConnectWalletMenu.tsx
```diff
@@ -18,49 +15,32 @@ type Props = Readonly<{
...
 export function ConnectWalletMenu({ children }: Props) {
     const { current: NO_ERROR } = useRef(Symbol());
-    const [selectedWalletAccount, setSelectedWalletAccount, wallets] = useSelectedWalletAccount();
+    const { displayName: currentChainName } = useContext(ChainContext);
+    const wallets = useWallets();
+    const connected = useConnectedWallet();
     const [error, setError] = useState(NO_ERROR);
     const [forceClose, setForceClose] = useState(false);
+    // Every wallet from `useWallets()` is pre-filtered by the plugin to those that support
+    // `standard:connect` on the active chain, so every item rendered here is connectable.
```
mcintyre94, line 28: Flagging this: the previous implementation displayed wallets that cannot be connected to, eg because they don't support the selected chain or don't have the `StandardConnect` feature. The plugin filters wallets to only those that support the selected chain and have the `StandardConnect` feature.

It is not possible, using only the plugin, to display wallets that have been filtered out. An app could do this by using the wallet-standard libraries directly, as the previous implementation did, but I haven't included that here.

In general I think this is the right default - the app only gets wallets that it can connect to.

Opened an issue in kit-plugins to add the filtered wallets if we want to bring this ability to display them back: https://github.com/anza-xyz/kit-plugins/issues/313

examples/react-app/src/components/Nav.tsx
```diff
@@ -52,8 +54,16 @@ export function Nav() {
...
+                <WalletReadyGate
+                    fallback={
+                        <Button disabled loading>
+                            Connect Wallet
+                        </Button>
+                    }
+                >
+                    <ConnectWalletMenu>Connect Wallet</ConnectWalletMenu>
+                    <SignInMenu>Sign In</SignInMenu>
+                </WalletReadyGate>
```
mcintyre94: Uses the `<WalletReadyGate>` so that it only renders the wallet-dependent UI after the wallet has resolved to a connected/disconnected state.

examples/react-app/src/components/SignInMenuItem.tsx
```diff
@@ -1,41 +1,41 @@
...
+        e.preventDefault();
+        try {
+            await signIn.dispatchAsync(wallet, {
+                domain: window.location.host,
+                statement: 'You will enjoy being signed in.',
+            });
+            onSignIn();
+        } catch (e) {
+            // Filter out abort error, which just means a later action superseded
+            if (!isAbortError(e)) {
```
mcintyre94, line 30: Example of the app using `isAbortError` to avoid throwing when an action is superseded, which previously required an install of `@solana/promises`.

examples/react-app/src/components/SolanaPartialSignTransactionFeaturePanel.tsx
```diff
@@ -85,10 +87,15 @@ export function SolanaPartialSignTransactionFeaturePanel({ account }: Props) {
         },
     };
 
+    // Render-time capability guard: throws so the surrounding `ErrorBoundary` renders
+    // `FeatureNotSupportedCallout` when the connected account can't sign transactions
+    // Also narrows the signer for the `useAction` below
+    assertCanSignTransactions(signer);
```
mcintyre94, line 95: Note that this is a bit different from the previous version, which used hooks like `useWalletAccountTransactionSigner`.

The app now receives a `signer: TransactionSigner`, and needs to check for (in this case) it having the partial sign feature. It does this by using an assertion which narrows the type. This is just a wrapper around functions we have in the `signers` package, but the app handles a null signer and has its own errors.

This `assert` throws if false, like the previous `useWalletAccountTransactionSigner` hook did, rendering the error boundary. But apps can also just branch on the signer type, etc, depending what makes sense for them. 

I'd also note that this example is a bit arbitrary - the instruction-plan and signer APIs offer higher-level helpers and mean apps generally don't need to know exactly which features are available.

examples/react-app/src/components/SolanaPartialSignTransactionFeaturePanel.tsx
```diff
@@ -106,7 +113,7 @@ export function SolanaPartialSignTransactionFeaturePanel({ account }: Props) {
                     getTransferSolInstruction({
                         amount,
                         destination: address(recipientAccount.address),
-                        source: transactionSigner,
+                        source: signer,
                     }),
                     m,
                 ),
```
mcintyre94, line 121: Just to highlight that this transaction construction and program instruction code will disappear in a future iteration that uses the instruction plans plugin and program plugins. This PR intentionally only adds wallet, and leaves everything RPC/transaction building/etc alone.

examples/react-app/src/context/ChainContext.tsx
```diff
@@ -1,11 +1,12 @@
...
 import { devnet } from '@solana/kit';
+import type { SolanaChain } from '@solana/wallet-standard-chains';
 import { createContext } from 'react';
 
 export type ChainContext = Readonly<{
-    chain: `solana:${string}`;
+    chain: SolanaChain;
     displayName: string;
-    setChain?(chain: `solana:${string}`): void;
+    setChain?(chain: SolanaChain): void;
```
mcintyre94, line 9: Not new in the plugin, but wallet-standard exports a `SolanaChain` type that we use throughout

examples/react-app/src/context/WalletClientProvider.tsx
```diff
@@ -0,0 +1,97 @@
...
+ * Client disposal (tearing down wallet-standard registry listeners) is split by who last owns the
+ * client: a published client is disposed when a successor replaces it (or the provider unmounts),
+ * while a client superseded *before* it was published is disposed by its own effect's cleanup. The
+ * latter must not live in the `whenReady()` continuation: a client stuck mid-warm-up never settles
+ * `whenReady` on its own — it is the *disposal* that settles it. Which is also why the `cancelled`
+ * guard in the continuation is load-bearing: once the cleanup disposes a superseded client, its
+ * `whenReady()` resolves and the continuation runs — the guard is what keeps the disposed client
+ * from being published.
+ */
+export function WalletClientProvider({ children }: Props) {
```
mcintyre94, line 32: This is the biggest red flag from this exercise

This is intentionally poking at an advanced use case, but it is realistic. When we change the chain, we build a new client (wallet plugin uses chain), but we avoid re-rendering using it until it is ready, using the `next.wallet.whenReady` construct. This means that the UI doesn't flash an unconnected state, and if the connected wallet remains valid on the new chain then there's no UI churn.

The rest of this is all juggling react lifecycle and the dispose mechanism.

My plan to fix this:

- Add `withWhenReady` as a function that plugins can use to register a `whenReady` promise. Currently only the wallet needs this. In a similar way to dispose, we will internally expose a `whenReady` on the client, that combines (using `Promise.allSettled` in this case) all registered promises from all plugins. The result is that an app will only need `client.whenReady` regardless which plugins are used
- Given that, this `WalletClientProvider` can be generalised to a `ManagedClientProvider` that takes a function to construct the client, and then handles the `client.whenReady` for you. Nothing wallet specific there. I will add that to the `@solana/react` package as a new provider.
- After that, apps that don't need to change their client will just use `ClientProvider` as they currently do. Apps that need to handle the client changing, and want to be able to wait until the new client has settled to re-render with it, will use `ManagedClientProvider`. All this complexity will disappear from the app while achieving the same UX.

examples/react-app/src/main.tsx
```diff
@@ -1,40 +1,42 @@
...
 import { RpcContextProvider } from './context/RpcContextProvider.tsx';
+import { WalletClientProvider } from './context/WalletClientProvider.tsx';
 import Root from './routes/root.tsx';
 
-const STORAGE_KEY = 'solana-wallet-standard-example-react:selected-wallet-and-address';
-const stateSync = {
-    deleteSelectedWallet: () => localStorage.removeItem(STORAGE_KEY),
-    getSelectedWallet: () => localStorage.getItem(STORAGE_KEY),
-    storeSelectedWallet: (accountKey: string) => localStorage.setItem(STORAGE_KEY, accountKey),
-};
```
mcintyre94, line 20: All this goes away, the wallet plugin handles persisting the connected wallet (using localStorage by default)

examples/react-app/src/main.tsx
```diff
@@ -1,40 +1,42 @@
...
-
 const rootNode = document.getElementById('root')!;
 const root = createRoot(rootNode);
 root.render(
     <StrictMode>
         <Theme>
             <ChainContextProvider>
-                <SelectedWalletAccountContextProvider filterWallets={(_: UiWallet) => true} stateSync={stateSync}>
-                    <RpcContextProvider>
+                <RpcContextProvider>
```
mcintyre94, line 20: The `RpcContextProvider` remains for now, because I haven't moved RPC into the client.

This does mean that the RPC (and the slot/balance rendered from it) doesn't participate in withholding the re-render until the client is ready. For a split second the app can be showing wallet data from the previous chain and RPC data from the new chain. This will go away when the RPC moves into the client and is re-rendered in the same tick.

examples/react-app/package.json
```diff
@@ -19,16 +19,18 @@
         "@radix-ui/themes": "3.3.0",
         "@solana-program/system": "^0.12.2",
         "@solana/kit": "workspace:*",
+        "@solana/kit-plugin-wallet": "link:/home/sol/projects/kit-plugins/packages/kit-plugin-wallet",
```
mcintyre94: Temporary, breaks CI

mcintyre94: This is now fixed, uses a published version of `kit-plugin-wallet`

mcintyre94:
@trevor-cortex

examples/react-app/src/components/SolanaPartialSignTransactionFeaturePanel.tsx
```diff
@@ -19,23 +20,25 @@ import {
     signTransactionMessageWithSigners,
     TransactionPartialSigner,
 } from '@solana/kit';
```
trevor-cortex, line 22: Nit: `ReadonlyUint8Array`, `SignatureBytes`, and `TransactionPartialSigner` are all type-only imports here — none of them are referenced as values. Sibling files use `import type { ReadonlyUint8Array } from '@solana/kit'` (e.g. `BaseSignMessageFeaturePanel.tsx` after this PR, and `Balance.tsx`), so this is inconsistent. Either split them into a separate `import type { … }` line, or add the `type` modifier per specifier (`import { …, type ReadonlyUint8Array, type SignatureBytes, type TransactionPartialSigner, … }`). Not blocking.

examples/react-app/src/components/SignInMenu.tsx
```diff
@@ -1,39 +1,30 @@
...
-                    }}
-                    onError={setError}
-                    wallet={wallet}
-                />
-            </ErrorBoundary>
+                onSignIn={() => setForceClose(true)}
+                onError={setError}
+                wallet={wallet}
+            />
         );
```
trevor-cortex, line 32: Behaviour change worth double-checking: previously each `SignInMenuItem` was wrapped in an `ErrorBoundary` that fell back to `UnconnectableWalletMenuItem`, so a wallet that advertised `solana:signIn` but crashed at hook-init render time still rendered as a disabled item rather than tearing down the menu. That safety net is gone here. Sign-in *invocation* errors still surface through `onError` → `ErrorDialog` (good), but a render-time throw from `useSignIn()` would now propagate up. In practice `useSignIn()` from the plugin shouldn't throw at render, so this is probably fine — flagging so it's an explicit decision rather than an accidental regression.

examples/react-app/src/context/WalletClientProvider.tsx
```diff
@@ -0,0 +1,103 @@
...
+                }
+            });
+        }
+        return () => {
+            cancelled = true;
+            if (activeClientRef.current !== next) {
+                // Superseded before it was published; nothing else owns it, so dispose it here.
+                next[Symbol.dispose]();
+            }
+        };
```
trevor-cortex: The lifecycle logic looks correct to me — walked through the StrictMode mount → cleanup → re-mount path and the chain-switch path and both hold together — but this is subtle enough that I want to name the invariants explicitly for future readers, in addition to the (already very good) prose docblock:

1. `activeClientRef.current` is always either `null` or a client that has been passed to `setClient` (never a disposed one).
2. A client built by an effect run has exactly one owner at any moment: either the ref (once published), or the layout-effect cleanup closure (if superseded before publish), or the empty-deps `useEffect` cleanup (on final unmount).
3. `next.wallet.whenReady()` also resolves when `next` gets disposed by the cleanup below, which is why `cancelled` is load-bearing rather than a defensive extra — a disposed client would otherwise slip through into `publish()`.

Since this provider might get copy-pasted into consumer apps, would it be worth extracting to a small utility in `@solana/kit-plugin-wallet` (or documenting the intended pattern there) in a follow-up? Not for this PR, but the reasoning here really shouldn't have to be rediscovered by every app author.

examples/react-app/src/walletCapability.ts
```diff
@@ -0,0 +1,60 @@
...
+ * A sending-only signer (an account with `solana:signAndSendTransaction` but not
+ * `solana:signTransaction`) deliberately fails this assert even though it satisfies
+ * `TransactionSigner`, because it doesn't work with those panels.
+ */
+export function assertCanSignTransactions(signer: WalletSigner | null): asserts signer is TransactionSigner {
+    if (!signer || !(isTransactionModifyingSigner(signer) || isTransactionPartialSigner(signer))) {
+        throw new Error('This account does not support signing transactions');
+    }
+}
+
```
trevor-cortex, line 34: Worth a passing note in the docblock (or here) that a `TransactionSendingSigner` also *satisfies* `TransactionSigner` structurally (a modifying signer that can also send), but this assert deliberately narrows to *partial-or-modifying-only* because the "Sign Transaction" / "Partial Sign Transaction" panels need to obtain the raw signed transaction bytes rather than have the wallet send it. The current docblock says "deliberately fails this assert" which is right — just calling out that the *reason* is UX-level (panel semantics), not that a sending signer isn't a valid `TransactionSigner`. Not blocking, but the comment might read more clearly if it said e.g. "…because those panels need the signed transaction bytes, not a submitted signature".

examples/react-app/vite.config.ts
```diff
@@ -17,5 +17,15 @@ export default defineConfig(({ mode }) => ({
...
         'process.env': process.env,
     },
+    optimizeDeps: {
+        // `@solana/react` is a linked workspace package, so Vite excludes it from dependency
+        // pre-bundling by default. When Vite then pre-bundles the npm dependency
+        // `@solana/kit-plugin-wallet` (which imports `@solana/react`), esbuild inlines a *second*
+        // copy of `@solana/react`, which causes mis-matched context.
+        // Forcing `@solana/react` to be pre-bundled makes it a single shared copy
+        // that both the app and the plugin reference.
+        // This is only an issue in our example because of the linked workspace.
```
trevor-cortex, line 27: Great comment — this is exactly the kind of thing that would burn hours to rediscover. One small suggestion: consider mentioning that the two-copies-of-`@solana/react` symptom typically surfaces as a *context mismatch* at runtime (e.g. `useClient()` from one copy reading `null` even though `ClientProvider` from the other copy is mounted above it), so a future reader hitting that symptom can search for it and land here. Optional.

trevor-cortex (commented):
## Summary

Migrates the `examples/react-app` demo from `@wallet-standard/react` (`useSelectedWalletAccount`, `useConnect`, `useDisconnect`, `useWalletAccount*Signer`) to the `@solana/kit-plugin-wallet` model — a Kit `Client` built with the `walletSigner` plugin, published via `ClientProvider`, and consumed through the plugin's React hooks (`useConnectedWallet`, `useConnect`, `useDisconnect`, `useSelectAccount`, `useSignIn`, `useSignMessage`, `useWallets`, `WalletReadyGate`). Also bumps `@wallet-standard/ui` → `^1.0.3` and `@wallet-standard/ui-registry` → `^1.1.1` in `packages/react` so a consumer that mixes `@solana/react` with `@solana/kit-plugin-wallet` resolves a single, shared registry singleton (well-explained in the changeset).

The example ports feature-for-feature: connect/disconnect/sign-in dropdowns, sign message / sign transaction / sign-and-send / partial sign panels, balance, chain switching, and the recipient picker. The former `SelectedWalletAccountContextProvider` + localStorage sync is replaced by the plugin's own connected-account state, and `UnconnectableWalletMenuItem`/`ErrorBoundary` wrapping is dropped because `useWallets()` already filters to wallets that support `standard:connect` on the active chain.

A new `WalletClientProvider` handles the chain-switch client rebuild (each wallet plugin is chain-bound) with hand-over-hand disposal and a `whenReady()` handoff so mid-switch the old client stays on screen. `WalletReadyGate` gates only the wallet-dependent UI so the chrome stays visible during the initial warm-up. Feature panels now call render-time `assertCan…` guards (`walletCapability.ts`) that throw into the existing `FeatureNotSupportedCallout` `ErrorBoundary`, replacing the previous static feature-list checks in the wallet-menu items.

Overall the migration is clean and the tricky bits (client lifecycle, StrictMode double-invoke, `whenReady()` deadlock avoidance, `optimizeDeps` for the linked workspace) are all called out in comments. This is a good reference for how the plugin is meant to be consumed from a React app.

- **`WalletClientProvider` lifecycle.** Worth re-reading the docblock and effect carefully — the invariants (who owns disposal at each point, why `cancelled` is load-bearing, why the whenReady continuation can't own disposal) are all correct as far as I can tell, but this is the kind of code that's easy to break in a follow-up. Suggest keeping the current comments verbatim as a regression guard, and considering a small test in a later PR that exercises the chain-switch and StrictMode double-invoke paths — an example app doesn't have to have one, but this provider is subtle enough that it might be worth extracting for reuse eventually.
- **Draft / follow-ups.** PR description flags (a) the pending bump to `@solana/kit-plugin-wallet@0.14.0` (currently open on `kit-plugins`) will require a small refactor, and (b) future iterations will add the `solanaRpc` and system-program plugins to the client. That's fine for a draft, just wanted to surface it here so it isn't lost.
- **`@wallet-standard/ui-registry` bump in `@solana/react`.** `^1.0.1` → `^1.1.1` is a minor bump, shipped as a `patch` changeset. The changeset justifies it as a backward-compatible superset (registry is a runtime singleton and this de-duplicates the shared instance), which reads reasonably to me — but flagging for a second pair of eyes since this is the only user-facing change in a published package.
- **`packages/react` still depends on `@wallet-standard/react`.** The example app dropped it, but `SelectedWalletAccountContextProvider` / `selectedWalletAccountContext` are still exported from `@solana/react` and still use it, so this is intentional. Just noting in case a follow-up wants to prune those exports now that the recommended flow no longer needs them.

- The `ErrorBoundary` around each feature panel is now doing double duty: it still catches thrown-from-render errors from the Kit signer hooks, and it now also catches the new `assertCan…` throws from `walletCapability.ts`. The message the user sees in `FeatureNotSupportedCallout` should be checked once against a wallet that doesn't support each feature (e.g. a wallet without `solana:signMessage`, or a sending-only wallet against the "Sign Transaction" panel) to make sure the copy matches expectations.
- The `SignInMenu` used to wrap each menu item in an `ErrorBoundary` → `UnconnectableWalletMenuItem`. That safety net is gone; sign-in failures now surface through `onError` → `ErrorDialog`. That's a behaviour change worth verifying — since `SolanaSignIn` is a per-feature filter (not a per-chain connect filter), any wallet that advertises the feature but errors on invocation should reach `onError` correctly, but a render-time throw from `useSignIn()` would now propagate. In practice it shouldn't happen, but worth a manual smoke test.
- Manually verifying under StrictMode is worth doing: the `WalletClientProvider` comment describes the StrictMode double-invoke path, and it's the most likely place for a subtle regression to hide.

examples/react-app/src/context/WalletClientProvider.tsx
mcintyre94: Issues opened for this: #1842, #1843

examples/react-app/src/context/WalletClientProvider.tsx
mcintyre94: This is planned, and issues are linked in my comment above. Resolving this one, will leave that open

examples/react-app/src/components/SignInMenu.tsx
mcintyre94: This is intentional - the plugin version doesn't call the sign-in function at render, only when the action is dispatched. So there's no render-time failure path here

lorisleiva (approved):

examples/react-app/src/context/WalletClientProvider.tsx
mcintyre94: I've reconsidered this design in ee8d403ef40bfc1c59b15966466fe74733cf2643

New design, app changes:

- The client provider now immediately creates and updates the client, does not wait for `whenReady`. All the complexity is gone, except dispose
- The UI now uses the hooks from the wallet plugin to granularly disable/dim UI that depends on the wallet state when it's in a not ready state

I am no longer suggesting standardising `whenReady` in Kit, and `ManagedClientProvider` may not be worth building - maybe for dispose. Everything we have in the wallet plugin remains, as it enables the UI to react to the readiness.

The issue with this design shows as we add more to the client - we want to immediately have access to eg the new RPC when the chain changes. UI that only depends on the RPC should not wait for the wallet to reconnect before updating. I don't have any examples of other plugins that would need this yet, but if there was one then UI that depends only on the wallet shouldn't wait for that plugin either. A client-level `whenReady` doesn't make sense IMO.

The app-level complexity remaining is that I have a hook `useDisplayedWallet` which captures the last seen `wallet.connected` state, and returns that alongside an `isStale` based on the wallet ready hook. This enables the app to display the previous connected wallet in a disabled/dimmed state during the reconnect. We could improve this by exposing the `UiWalletAccount` that is being reconnected to from the plugin when it is available, which would provide a better first load UX (the app can't currently know it until it's connected), and would avoid the app needing to store it in this hook. Captured as a plugin issue: https://github.com/anza-xyz/kit-plugins/issues/350 

Example of the current UX using a deliberately slowed down wallet reconnect path:

https://github.com/user-attachments/assets/0c47cb99-a39e-4405-a1fb-927c4edfc638

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Nice iteration. The 0.14 hook API (`client` as first param) is threaded through cleanly, and the new decomposition — `Dimmable`, `GatedRoot`, `SlotIndicatorPanel`, plus `useDisplayedWallet` / `useHasWalletSettled` — is a real readability win over the previous inline logic. The stale-vs-settled distinction is subtle, and pulling it into two small named hooks with docblocks that spell out *why* they're separate is the right call. New tests cover the tricky bits (StrictMode double-mount disposal, retained-vs-live connection, latching behavior) — that's exactly where I would have wanted them.

A few small notes below; none block. Given Loris already approved, treat these as follow-up thoughts rather than change requests.

**Things worth a second look:**

1. **Duplicated "busy button" styling between `ConnectWalletMenu` and `SignInMenu`.** Both apply the same `aria-busy`/`opacity: 0.5`/`pointerEvents: 'none'`/`transition: 150ms` block to their `<Button>` triggers. `Dimmable` can't wrap a `Trigger` child (Radix needs the button directly), but a tiny `<BusyButton>` component (or a `useBusyButtonProps(busy)` hook that returns the props bag) would keep the two triggers in lockstep and match the abstraction level of `Dimmable`.

2. **`ConnectWalletMenu` and `SignInMenu` use different signals for the same UX.** `ConnectWalletMenu` disables via `isStale` (from `useDisplayedWallet`), `SignInMenu` disables via `!isReady` (from `useIsWalletReady`). They're the same value today (`isStale === !isReady`), but a reader has to work that out. Deriving both from the same primitive — or using `useDisplayedWallet`'s `isStale` in both — would make the intent unambiguous.

3. **`SolanaSignMessageFeaturePanel` deliberately leaks a `WALLET_STANDARD_ERROR__*` up to the boundary** so `getErrorMessage` can format it consistently with the other panels. `walletCapability.ts` explains this, which is great — but a reader landing in `SolanaSignMessageFeaturePanel.tsx` first won't have that context. Consider a one-line comment on the `assertCanSignMessages(account)` call pointing at the docblock (or restating the "we intentionally let the wallet-standard error through here" bit).

**Notes for later reviewers:**

- The two custom hooks disable `react-hooks/refs` and mutate a ref during render. That's safe here because the mutation is a pure function of the props (`isReady`, `connected`), so StrictMode's double-render is idempotent — and there are tests exercising exactly that path via the StrictMode-wrapping `render` helper. Worth being aware of when reading them.
- `WalletClientProvider` publishes the client synchronously from a `useLayoutEffect` and disposes it in cleanup. Under StrictMode the effect runs twice, so the mount test's `mockPublishedClients` array will actually contain the disposed dev-double *and* the live client — `[length - 1]` correctly picks the live one, but if that test ever gets tightened it's worth asserting the disposed one was in fact disposed.
- Follow-ups the PR description already flags: `solanaRpc` + system-program plugins, and any hooks that currently accept `client` explicitly could later move to a `useClient()`-implicit style if the plugin gains one.

mcintyre94:
### Merge activity

* **Aug 4, 4:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1825).
* **Aug 4, 4:16 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1825) couldn't merge this pull request because a downstack PR #1841 failed to merge.
* **Aug 4, 5:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1825).
* **Aug 4, 5:12 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1825).
* **Aug 4, 5:12 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1825) couldn't merge this PR because it failed for an unknown reason.

PR #1821, opened 2026-07-06
Add transaction planning & sending hooks to `@solana/react`

mcintyre94:
#### Summary of Changes

This PR builds on our react hooks by adding hooks related to the instruction-plans plugin to `@solana/react`

This includes 4 hooks: `usePlanTransaction`, `usePlanTransactions`, `useSendTransaction`, and `useSendTransactions`

All are simple wrappers of `useAction` and `useClientCapability`. They require the `instruction-plans` plugin, or something shaped like it, to be installed on the client returned by `useClient`. 

This replaces a previous PR that added these hooks to the `instruction-plans` plugin: https://github.com/anza-xyz/kit-plugins/pull/301. We have the plugin interfaces in Kit, so `@solana/react` is able to wrap them directly. Plugins don't need to have their own `/react` subpath, with the exception of the wallet plugin which requires wallet-standard dependencies. And having these hooks in `@solana/react` makes it easier for developers to use them.

PR #1833, opened 2026-07-13
Add `useRpc` and `useRpcSubscriptions` hooks to `@solana/react`

mcintyre94:
This PR adds React hooks to read `rpc` and `rpcSubscriptions` from a Kit client

These are just wrappers around `useClientCapability` and require the rpc plugins, most likely using `solanaRpc`. Probably the simplest hooks we'll add, but I think justified by how much apps need to use the RPC.

They're added near the start of the readme, and the other hooks (useRequest etc) are updated to use them instead of using `useClient` directly.

If reading both `rpc` and `rpcSubscriptions` (as some examples in the readme do), this is now two hook calls instead of just one to `useClient[Capability]`. I think this tradeoff is worth it, and any app that needs the micro-optimisation can use `useClient[Capability]` instead. We could provide a hook that returns both in one call, but I don't think it's necessary.

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -92,6 +92,28 @@ function useRpc() {
 
 Pass an array of capability names when a hook needs more than one (e.g. `['rpc', 'rpcSubscriptions']`) — the same `providerHint` is surfaced for whichever is missing.
 
+### `useRpc()` / `useRpcSubscriptions()`
+
+First-class, capability-checked access to the client's `rpc` and `rpcSubscriptions` objects. Each is a thin `useClientCapability` wrapper: it asserts the capability at mount (throwing `SOLANA_ERROR__REACT__MISSING_CAPABILITY` when the plugin is absent) and returns the object directly. The generic is the method set, defaulting to `SolanaRpcApi` / `SolanaRpcSubscriptionsApi`; use it to match the client you built.
+
+```tsx
+import { useEffect, useState } from 'react';
+import { useRpc } from '@solana/react';
+
+function EpochBadge() {
+    const rpc = useRpc();
+    const [epoch, setEpoch] = useState<bigint>();
+    useEffect(() => {
+        rpc.getEpochInfo()
+            .send()
+            .then(info => setEpoch(info.epoch));
+    }, [rpc]);
+    return <span>{epoch !== undefined ? `Epoch ${epoch}` : '…'}</span>;
+}
+```
+
+Calling `.send()` (or `.subscribe()` on a subscription) yourself is the low-level path. In practice you'll usually hand the resulting `PendingRpcRequest` / `PendingRpcSubscriptionsRequest` to `useRequest`, `useSubscription`, or `useTrackedData` (below), which track the lifecycle as React state for you.
+
 ### `useAction(fn)`
 
 Bridges any async function into a tracked action with `dispatch` / `status` / `data` / `error` / `reset`. Each `dispatch(...)` runs `fn` with a fresh `AbortSignal` and tracks the lifecycle through React state; calling `dispatch` again while a prior call is in flight aborts the first.
@@ -138,12 +160,12 @@ Fires a one-shot request on mount and re-fires whenever `source` changes identit
 > Unlike `useAction`, `useRequest` needs the input to have stable identity across renders — it's how the hook knows when to re-fire. Memoize with `useMemo` (for a reactive store source) or `useCallback` (for a function), keyed on whatever inputs your call depends on.
 
 ```tsx
-import { useClient, useRequest } from '@solana/react';
-import type { ClientWithRpc, GetLatestBlockhashApi } from '@solana/kit';
+import { useRequest, useRpc } from '@solana/react';
+import type { GetLatestBlockhashApi } from '@solana/kit';
 
 function LatestBlockhash() {
-    const client = useClient<ClientWithRpc<GetLatestBlockhashApi>>();
-    const source = useMemo(() => client.rpc.getLatestBlockhash(), [client]);
+    const rpc = useRpc<GetLatestBlockhashApi>();
+    const source = useMemo(() => rpc.getLatestBlockhash(), [rpc]);
     const { data, error, refresh } = useRequest(source);
     if (error) return <button onClick={refresh}>Retry</button>;
     return <p>{data ? `Blockhash: ${data.value.blockhash}` : 'Loading…'}</p>;
@@ -154,9 +176,9 @@ function LatestBlockhash() {
 
 ```tsx
 function Balance({ address }: { address: Address | null }) {
-    const client = useClient<ClientWithRpc<GetBalanceApi>>();
+    const rpc = useRpc<GetBalanceApi>();
     // Disabled until an address is selected.
-    const source = useMemo(() => (address ? client.rpc.getBalance(address) : null), [client, address]);
+    const source = useMemo(() => (address ? rpc.getBalance(address) : null), [rpc, address]);
     const { data, status } = useRequest(source);
     if (status === 'disabled') return <p>Select an account to see its balance.</p>;
     return <p>{data?.value !== undefined ? `${data.value} lamports` : 'Loading…'}</p>;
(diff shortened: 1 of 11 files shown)


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1833?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1833** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1833?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1833?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1821** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1821?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Clean, well-scoped addition. Both hooks are the natural, minimal `useClientCapability` wrappers — matching almost verbatim the illustrative `useRpc` example already present in `useClientCapability`'s docblock, which is a nice consistency signal. Looks good to me overall.

**What this does**
- Adds `useRpc<TRpcMethods = SolanaRpcApi>()` and `useRpcSubscriptions<TRpcSubscriptionsMethods = SolanaRpcSubscriptionsApi>()` in `@solana/react`, each asserting the capability at mount and returning the client's `rpc` / `rpcSubscriptions` object directly.
- Migrates every downstream README example (`useRequest`, `useSubscription`, `useTrackedData`, plus the SWR/Query variants) off `useClient<ClientWithRpc<…>>()` to the new hooks, giving a consistent story across the file.
- Ships a `minor` changeset, a browser test per hook covering the three paths (capability present, missing, no provider), and typetests exercising the default + narrowed generics with `@ts-expect-error` on absent methods.

**Tradeoff acknowledged in the description** — reading both `rpc` and `rpcSubscriptions` is now two hook calls instead of one to `useClient[Capability]`. Agree that's fine: apps that want the single-call form still have `useClientCapability<ClientWithRpc<…> & ClientWithRpcSubscriptions<…>>({ capability: ['rpc', 'rpcSubscriptions'], … })` as an escape hatch, and the readability win on the common case is worth it.

**Things for subsequent reviewers to check**
- Nothing structural — the hooks are trivially small and the test coverage matches the pattern in `useClientCapability-test.browser.tsx` (which was the natural reference).
- Worth eyeballing the README diff to confirm all example blocks still typecheck against the new hook signatures. The migrations look correct on inspection (the `useMemo` dep arrays are updated in each case to depend on `rpc` / `rpcSubscriptions` instead of `client`), but a `pnpm turbo compile:typedefs` in `packages/react` is the cheap sanity check.

A couple of small optional nits inline — none blocking.

packages/react/src/__tests__/useRpc-test.browser.tsx
```diff
@@ -0,0 +1,57 @@
...
+    SOLANA_ERROR__REACT__MISSING_PROVIDER,
+    type SolanaRpcApi,
+} from '@solana/kit';
+import React from 'react';
+
+import { renderHook } from '../__test-utils__/render';
+import { ClientProvider } from '../ClientProvider';
+import { useRpc } from '../useRpc';
+
+type ClientWithRpc = { rpc: Rpc<SolanaRpcApi> };
```
trevor-cortex: Minor: `ClientWithRpc` is already exported from `@solana/kit` (and `useRpc.ts` itself imports it from there). You could drop this local alias and use the exported type directly with `SolanaRpcApi` as the generic — mirrors the production code's own types more closely. Same nit applies to the `ClientWithRpcSubscriptions` alias in the sibling test file. Non-blocking.

packages/react/src/__tests__/useRpc-test.browser.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+function wrapperFor<T extends object>(client: ReturnType<typeof createClient<T>>) {
```
trevor-cortex: Optional: this `wrapperFor` helper is identical to the one in `useClientCapability-test.browser.tsx` (and now duplicated again in `useRpcSubscriptions-test.browser.tsx`). If a fourth capability test lands, worth lifting into `__test-utils__/` as a shared `createClientWrapper` — but no need to do it in this PR.

packages/react/README.md
```diff
@@ -92,6 +92,30 @@ function useRpc() {
...
+    const rpc = useRpc();
+    const [epoch, setEpoch] = useState<bigint>();
+    useEffect(() => {
+        rpc.getEpochInfo()
+            .send()
+            .then(info => setEpoch(info.epoch));
+    }, [rpc]);
+    return <span>{epoch !== undefined ? `Epoch ${epoch}` : '…'}</span>;
+}
+```
```
trevor-cortex, line 113: Nit: the section header uses the generic-parameter form (`useRpc<TRpcMethods>()` / `useRpcSubscriptions<TRpcSubscriptionsMethods>()`) whereas the other hook sections in this file (`useAction(fn)`, `useRequest(source)`, `useSubscription(source)`, etc.) show the runtime call signature. Consider `useRpc()` / `useRpcSubscriptions()` for header consistency — the generic is fully documented in the paragraph directly below.

packages/react/src/__tests__/useRpc-test.browser.tsx
mcintyre94: Updated

packages/react/src/__tests__/useRpc-test.browser.tsx
mcintyre94: Updated to extract a `createClientWrapper`, might as well do it now because there will be more

packages/react/README.md
mcintyre94: Updated to drop the generic in the heading

packages/react/src/useRpc.ts
```diff
@@ -0,0 +1,38 @@
...
+ *     const rpc = useRpc();
+ *     // ...call rpc.getBalance(address).send() in an effect or with useRequest, etc.
+ * }
+ * ```
+ *
+ * @see {@link useRpcSubscriptions}
+ * @see {@link useClientCapability}
+ */
+export function useRpc<TRpcMethods = SolanaRpcApi>(): Rpc<TRpcMethods> {
+    const client = useClientCapability<ClientWithRpc<TRpcMethods>>({
```
lorisleiva, line 32: Isn't this going to lose any type information on the client itself and replace it with `SolanaRpcApi` all the time? 🤔

mcintyre94: Hmm yep our `useClient` and `useClientCapability` don't inherit any types currently, it's casting (for `useClient`) or runtime checks (for `useClientCapability`, these checks are only for the top-level capability though)

It looks like the clever thing other libraries do, which would be additive, is to ship a function `createClientContext<T extends Client>` that returns `{ClientProvider, useClient}` where both are typed for `T`.

Then you could add a `client: T extends Client & ClientWithRpc` param here, and at the call-site:

```
const client = useClient(); // the one from createClientContext, so has a known `ClientWithRpc<CustomMethods>
const rpc = useRpc(client);
```

We should be able to do this without breaking changes, we'd just need to add `createClientContext` and an override of each method to enable optionally passing in an already typed client 

Though maybe we should just require `client` to be passed to these hooks from the start, and add that `createClientContext` to get a typed one without casting. 

At that point though, `useRpc` is pointless, the plan/send hooks are probably still justified since they wrap `useAction`

This also becomes more tricky though when the `client` changes, and that potentially also changes its type. I'm struggling to think how we could model that through react context. So something like RPC methods varying based on chain, which we do have.

lorisleiva: I think the fact that `useRpc` becomes useless is a good sign. It's important that we keep the client type defined by the client as-is otherwise all the work we've done to keep everything typesafe on the plugins is wasted at the very last user layer which is a shame.

mcintyre94: I think that's fair. Do you think the right design is just to require a `Client & ClientWithThePlugin` as input to each react hook? The only ones we've shipped are the wallet ones, which currently use `useClientCapability` to runtime check the wallet plugin exists. So I'd add `client` to them, and also to the open PR here that adds instruction plan/send, and then just close this PR as unneeded. I'd also add the `createClientContext` function in `packages/react` here to get a typed `useClient`, and when I add the managed equivalent (which falls out of the prototype using the wallet plugin, and handles rebuilding the client for apps that need that) that would also get a typed `useClient`.

Typed `useClient` also gets us the improvements in the readme examples here, those become `client.rpc`

The only limitation I can see is that the managed version will need to have a fixed client type, so you'd lose type safety if you have a factory function that returns differently typed clients, because the actual client and its type then will only be known at runtime. This gets us almost all the way there though.

mcintyre94:
Draft for now - expect this to be closed per https://github.com/anza-xyz/kit/pull/1833#discussion_r3576878877

mcintyre94:
Closing. Open PRs require a type on `useClient` and document the expected usage. For RPC it would be:

```ts
// any component/hook
const client = useClient<AppClient>()
const rpc = client.rpc // precisely typed
```

No need for a hook here

PR #1835, opened 2026-07-14
feat: expose codecs subpath from kit

### LINKED ISSUE #1834: feat: expose codecs subpath from kit
## Motivation

 We want @solana/web3.js v3 to re-export @solana/codecs. Since web3.js already depends on @solana/kit, we want to re-export them directly from the kit (via @solana/kit/codecs) instead of adding a direct @solana/codecs dependency to web3.js to avoid extra maintenance burden. 
[Issue](https://github.com/solana-foundation/solana-web3.js/issues/3843)

## Example use case

Allow consumers or downstream packages (like @solana/web3.js) to import codecs like this:

```ts
import { getU32Codec, getStructCodec } from '@solana/kit/codecs';
```

## Details

Expose a ./codecs subpath in @solana/kit's package.json exports map that re-exports everything from the internal @solana/codecs package.

HkSolDev:
#### Problem

Expose a ./codecs subpath to allow downstream consumers (like @solana/web3.js) to import codecs via @solana/kit/codecs.

#### Summary of Changes

- Created src/codecs.ts to re-export everything from @solana/codecs.
- Exposed ./codecs subpath in the kit's package.json exports.
- Configured build scripts and tsconfig to build the new entrypoint and generate type declarations.
- Added a changeset.

Fixes #1834

### DIFF
--- a/.changeset/swift-humans-grin.md
+++ b/.changeset/swift-humans-grin.md
@@ -1,0 +1,5 @@
+---
+"@solana/kit": minor
+---
+
+Expose `@solana/codecs` from the `@solana/kit/codecs` subpath.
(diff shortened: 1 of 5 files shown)


lorisleiva:
See https://github.com/anza-xyz/kit/issues/1834#issuecomment-4966440849

PR #1838, opened 2026-07-14
Expose codecs subpath from kit

### LINKED ISSUE #1834: feat: expose codecs subpath from kit

oxy-Op:
#### Problem

`@solana/web3.js` v3 wants to consume codecs through `@solana/kit/codecs`, but `@solana/kit` currently only exposes codecs from the root package export. That means downstream packages need to add or maintain a direct `@solana/codecs` dependency instead of importing through Kit.

#### Summary of Changes

Adds a dedicated `@solana/kit/codecs` subpath export that re-exports `@solana/codecs`.

This PR also:
- Adds the `src/codecs.ts` entrypoint.
- Adds the `./codecs` package export for Node, browser, React Native, workerd, and edge-light.
- Includes the codecs entrypoint in JS and declaration builds.
- Extends the existing subpath export test.
- Adds a changeset for `@solana/kit`.

Fixes #1834

### DIFF
--- a/.changeset/khaki-codecs-exist.md
+++ b/.changeset/khaki-codecs-exist.md
@@ -1,0 +1,5 @@
+---
+'@solana/kit': minor
+---
+
+Expose `@solana/kit/codecs` as a dedicated subpath export that re-exports `@solana/codecs`.
--- a/packages/build-scripts/getBaseConfig.ts
+++ b/packages/build-scripts/getBaseConfig.ts
@@ -19,7 +19,7 @@ export function getBaseConfig(platform: Platform, formats: Format[], _options: O
     // via the package.json `exports` map.
     const moduleEntry =
         env.npm_package_name === '@solana/kit'
-            ? ['./src/index.ts', './src/program-client-core.ts']
+            ? ['./src/index.ts', './src/codecs.ts', './src/program-client-core.ts']
             : env.npm_package_name === '@solana/react'
               ? ['./src/index.ts', './src/swr.ts', './src/query.ts']
               : ['./src/index.ts'];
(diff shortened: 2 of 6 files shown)


mcintyre94:
I'm not clear why you need this, the codecs are already available to import from `@solana/kit` which you said you have as a dependency.

[Typescript Playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgcwKYwKoGYBMBhCAE1QGMAaFdAZRigFcSYDiS4BfOAMyghDgHIAAgGcIAGwCGAOwkB6ANbAY-ANwAoNSQhTh8OrjgBeSplzNSACgCU6-TgB0qKVuIWAjAAYPNoA)

```ts
import { getU32Codec, getStructCodec } from '@solana/kit';

const u32 = getU32Codec();
u32.encode(100);
```

What am I missing?

oxy-Op:
Thanks for checking. I missed that #1834 was closed as not planned.

Given codecs are already available from the root `@solana/kit` export and the subpath direction is not preferred, I’ll close this PR.

PR #1840, opened 2026-07-14
Rework the react docs to prioritise `useClient<AppClient>`

mcintyre94:
#### Problem

Initially I planned to write the react hooks around `useClientCapability`, which does a runtime check for the capability on the client under `ClientProvider`.

[As discussed here though](https://github.com/anza-xyz/kit/pull/1833#discussion_r3572581679), this reduces type safety which plugins otherwise maintain all the way through the stack.

We therefore plan to focus on a pattern where type safety is maintained for react apps using a typed client

We also plan to have hooks that do depend on a plugin, eg the wallet hooks, take the client with that plugin as a param: https://github.com/anza-xyz/kit-plugins/pull/326. This enables type safety across the whole app.

#### Summary of Changes

- We prioritise a pattern where apps export `AppClient`, the type of their built client. Each component uses this in a `useClient<AppClient>()` 
- `useClientCapability` is de-emphasised as a runtime escape hook, not the intended way to build client-dependent plugins
- All docs are updated to use this `<AppClient>` type, instead of manually typing a `useClient` call. 

This all fits the existing API, it's just a shift in documentation and in how we intend to build the hooks that depend on the client. 

Note that I decided not to add a `createClientContext` that creates a `ClientProvider` and a typed `useClient`, I don't think that additional API surface is warranted and I think it would be more confusing.

Also note that we plan to make the `useClient` type non-optional, this will be a follow up.

### DIFF
--- a/docs/content/docs/guides/react/core-hooks.mdx
+++ b/docs/content/docs/guides/react/core-hooks.mdx
@@ -9,36 +9,56 @@ Every hook on this page must be rendered under a [`ClientProvider`](/docs/guides
 
 ### `useClient`
 
-Reads the Kit client published by the nearest `ClientProvider`. It defaults to the base `Client` shape; when you know a plugin is installed, narrow the type through the generic. This is a pure type-cast with no runtime check - reach for `useClientCapability` instead when a missing plugin should fail. Throws `SOLANA_ERROR__REACT__MISSING_PROVIDER` if no provider is mounted above it.
+Reads the Kit client published by the nearest `ClientProvider`. Throws `SOLANA_ERROR__REACT__MISSING_PROVIDER` if no provider is mounted above it.
+
+Pass your client's type through the generic so every capability you installed is typed at the call site. The simplest way is to reuse the type of the [client you already built](/docs/guides/react) - export it once alongside the client - so the hook's type always matches your real plugin composition:
 
 ```tsx twoslash
-import { ClientWithRpc, GetEpochInfoApi } from '@solana/kit';
+// client.ts — where you build the client
+import { createClient } from '@solana/kit';
+import { solanaDevnetRpc } from '@solana/kit-plugin-rpc';
+import { generatedSigner } from '@solana/kit-plugin-signer';
+
+export const client = createClient().use(generatedSigner()).use(solanaDevnetRpc());
+export type AppClient = Awaited<typeof client>;
+
+// any component
 import { useClient } from '@solana/react';
 
 function FetchEpochButton() {
-    const client = useClient<ClientWithRpc<GetEpochInfoApi>>();
+    const client = useClient<AppClient>(); // rpc, signers, … all typed
     return <button onClick={() => client.rpc.getEpochInfo().send()}>Fetch epoch</button>;
 }
 ```
 
+`AppClient` uses `Awaited<typeof client>` so it resolves the promise when a plugin is async, for a fully-synchronous client `Awaited<…>` is a no-op and `typeof client` alone would do.
+
+`useClient` is a pure type-cast with no runtime check.
+
 ### `useClientCapability`
 
-Reads the client and asserts at mount that a capability is installed, narrowing the return type via the generic. If the capability is absent it throws `SOLANA_ERROR__REACT__MISSING_CAPABILITY`, including the `hookName` and `providerHint` you supply so the mistake is easy to locate. This is the building block for plugin-specific hooks.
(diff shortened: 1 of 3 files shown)


mcintyre94:
* **#1840** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1840?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Docs-only rework that reorients the React guide + README around `useClient<AppClient>()` as the primary pattern, with `useClientCapability` reframed as a runtime escape hatch. The `AppClient = Awaited<typeof client>` idiom is a nice ergonomic default and it drops a lot of the noisy per-call `ClientWithRpc<...> & ClientWithRpcSubscriptions<...>` narrowing from the examples. No changeset needed — docs-only.

A few things worth a look before merging:

**1. The `useClientCapability` example undercuts the new narrative.** Both files now open the section with "you rarely need this — when you type your client with `useClient<AppClient>()` the compiler already guarantees the capability is present," and then immediately show a hand-rolled `useCheckedRpc` wrapper around `useClientCapability<ClientWithRpc<GetEpochInfoApi>>`. That's exactly the shape of code the prose just said you shouldn't be writing, and the example doesn't illustrate the motivating scenario the prose describes ("a loosely-typed `Client` read from context, say"). Consider replacing it with an example that actually shows the escape-hatch case — e.g. a helper that takes a bare `Client` from somewhere and asserts a capability — or at minimum a comment on the example framing it as "if you find yourself with an untyped client, here's how you'd guard". Right now the example reads as "here's the pattern" while the prose says "don't do this."

**2. Please verify the `generatedSigner()` async claim in `packages/react/README.md`.** The new `client.ts` snippet has the comment `` `Awaited<…>` resolves the promise when any plugin is async (e.g. `generatedSigner()`); for a fully-synchronous client it is simply `typeof client` ``. That parenthetical asserts `generatedSigner()` is async, but the same composition is used elsewhere in `docs/content/docs/guides/react/index.mdx` at module scope as a synchronous `const client = createClient()...` and the async example there wraps it in a separate `createClientWithAsyncPlugins()` helper. If `generatedSigner()` returns synchronously, the parenthetical is misleading — either drop the example or pick a plugin that really is async.

**3. Two comma splices worth cleaning up:**
- `docs/content/docs/guides/react/core-hooks.mdx`: "`AppClient` uses `Awaited<typeof client>` so it resolves the promise when a plugin is async, for a fully-synchronous client `Awaited<…>` is a no-op and `typeof client` alone would do." — the comma before "for a fully-synchronous client" should be a period or semicolon.
- `packages/react/README.md` `useClient` intro: "It defaults to the base `Client` shape, pass your client's type through the generic to get the capabilities you installed typed at the call site." — same pattern, wants a period or semicolon before "pass".

**4. Code-block structure in `core-hooks.mdx`.** The `useClient` example is a single twoslash block that starts as `client.ts` (defines and exports `client` + `AppClient`), then has an `import { useClient } from '@solana/react'` in the middle, then "any component" with a `FetchEpochButton`. Reads a bit oddly as one file. The README handles the same content with two separate code blocks (one for `client.ts`, one for the component importing `AppClient` from `./client`), which is much clearer. Would be nice to align — either two twoslash blocks with a shared prelude, or a stronger visual break inside the single block. Not blocking, but the README version is easier to follow.

- Worth checking whether the reframe is consistent with any other places that reference `useClientCapability` in docs (e.g. the `ClientProvider` intro in the README still lists `useClientCapability` in the same breath as `useClient` as "Required for…"; that's fine, but if the intent is really to demote it, it may deserve a lighter mention).
- The `Choosing a hook` table in `index.mdx` now describes `useClientCapability` as "Runtime-guard a capability on an untyped client" — nice framing that matches the new narrative.
- Sanity-check the twoslash snippets actually compile with the new imports (`@solana/kit-plugin-rpc`, `@solana/kit-plugin-signer`) — these look right based on the existing `index.mdx`, but a local `pnpm docs:build` (or however docs are built) would confirm.

mcintyre94:
> `The useClientCapability example undercuts the new narrative.`

Updated this, now just specifies that the `Client` is untyped. 

> Please verify the generatedSigner() async claim

It's async

> Two comma splices worth cleaning up

I think these read fine

> Code-block structure in core-hooks.mdx

This is intentional, to show `AppClient` before the first time we use it on this page.

lorisleiva (approved):

mcintyre94:
### Merge activity

* **Jul 31, 3:40 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1840).
* **Jul 31, 3:40 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1840).
* **Jul 31, 3:41 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1840) couldn't merge this PR because it failed for an unknown reason.

Issue #1843, opened 2026-07-14
Add `ManagedClientProvider` to `@solana/react` (manage runtime client swaps without UI flash)


mcintyre94:
## Summary

Add a new `ManagedClientProvider` component to `@solana/react` that moves the runtime client-swap lifecycle — build → await readiness → hold → swap → dispose — out of apps and into the library. An app that rebuilds its Kit client on a config change (e.g. switching chains) passes **only** a `create` callback; the provider handles readiness, holding the old client until the new one settles, disposing on replace, and StrictMode safety.

It exposes **two read channels** over that one lifecycle so components can opt into the behavior they need: `useClient()` (the ready/warmed-up client, for wait-until-ready data fetching — the existing default) and a new `useLatestClient()` (the latest/warming client, for optimistic/immediate rendering). See [Two client channels](#two-client-channels-ready-vs-warming-per-1844) below — this was prompted by [#1844](https://github.com/anza-xyz/kit/pull/1844).

> **Depends on #1842** (`withWhenReady` in `@solana/plugin-core`). `ManagedClientProvider` consumes the client's aggregated `whenReady()` capability (duck-typed — `@solana/react` never imports the wallet plugin), so that primitive should land first.

## Motivation

Each wallet plugin is bound to a single chain, so an app that lets the user switch chains has to rebuild the Kit client at runtime. A freshly built client runs a silent auto-reconnect that briefly passes through `'pending'` / `'reconnecting'`; naively swapping the new client in flashes a "reconnecting" state on every switch.

The working solution shipped in the react-app example pushes all of that lifecycle ceremony into the app itself — a hand-rolled `WalletClientProvider` (build → `await client.wallet.whenReady()` → hold-current-until-ready → swap → dispose-on-replace) plus a `WalletReadyGate` to keep chrome visible during cold-start warm-up. It works, but every app that swaps clients would have to copy it. That contradicts the north star: *plugins should let you build a good app with clean code.*

Full context and rationale in the review comment: https://github.com/anza-xyz/kit/pull/1825/changes#r3561003685

We also proved (three separate attempts) that Suspense + `startTransition` **cannot** hold the UI across a client *swap* — the transition commits the new transient client with `isPending: false`, then suspends post-commit, producing the flash. So the swap must be handled imperatively. This issue is about giving that imperative machinery one home so apps don't repeat it.

## Proposed API

[...]

PR #1841, opened 2026-07-14
Make the type param in useClient non-optional

mcintyre94:
#### Problem

- `useClient()` gives you `Client<object>` which is basically useless
- We're now documenting a pattern of using `useClient<AppClient>()` for type safety

#### Summary of Changes

Make the `TClient` type param in `useClient()` required, so that the type error is there instead of when you try to access `client.rpc` etc

This is a breaking change, but we are using a minor changeset because `useClient()` is not useful as-is and this is worth fixing. It's unlikely that any code is using `useClient()` as-is.

Note that `useClientCapability`, the runtime checked variant, already required its type param.

PR #1844, opened 2026-07-14
Add the RPC plugin to the react-app example

mcintyre94:
This PR adds the `solanaRpc` plugin to the example app, and generalises its `WalletClientProvider` to just `ClientProvider`. It replaces use of `rpc` and `rpcSubscriptions` to come from the client, and removes the old `RpcContextProvider` that manually handled the same state.

It uses our newly documented pattern of `useClient<AppClient>()`

This PR does not yet use any of the other features of `solanaRpc`, such as the instruction plans plugin, but intentionally installs `solanaRpc` so that future PRs can.

An issue this caused is that the `rpc` now waits to update until the wallet is ready, since it updates with the client. This is I think correct - the balance to display depends on the connected wallet. Technically slot doesn't need to wait though. This means that the selected chain from `ChainContext` can be out of sync with the `rpc`. We use `extendClient` to add the chain to the client, and update callers that should depend on the active chain to read chain from the client. The select UI still reads the chain from the `ChainContext` so that it updates immediately.

This leads to 2 points:

- We could consider adding a chain plugin, that exposes `chain` and `setChain`, and we could add a `useChain` react hook
- If we did this, we'd need a concept of selected (updates immediately) and active (updates when the client is ready and updates). This might point to us needing a way to access the 'warming up' client immediately when we add `ManagedClientProvider` (#1843) so that components can decide whether to render immediately or wait for the client to be ready.


mcintyre94:
* **#1844** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1844?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1844?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Solid refactor. The move to `solanaRpc` + `extendClient({ chain })` collapses two overlapping context providers into one and delivers a nice property: the `rpc` and the `chain` a component reads are now guaranteed to be in lockstep, because both come off the same client reference. The `Balance` refetch test is a great codification of the invariant.

### Main thing to look at

**The two chain sources aren't consistently applied across components.** The PR intentionally moves `Balance` and `Root` to read `chain` from the client (with excellent comments explaining why — active vs. selected), but the sibling components (`AirdropButton`, `SolanaSignAndSendTransactionFeaturePanel`, `SolanaSignTransactionFeaturePanel`, `SolanaPartialSignTransactionFeaturePanel`, `SlotIndicator`) silently keep reading `chain` / `solanaExplorerClusterName` from `ChainContext`. Some of these uses are genuinely fine — for example the account-filtering `Select`s want the eagerly-updated selection so the UI reflects the user's choice immediately, matching the same rationale as the chain switcher itself. Others are more subtle:

- `AirdropButton`: `isMainnet` (from `ChainContext`) disables the button based on the selected chain, but the airdrop actually runs against `client.rpc`. In practice this fails safe (the mainnet rpc has no `requestAirdrop`, and the mainnet→devnet direction just hits a devnet rpc that supports it), but the eager/lagging split is worth being deliberate about.
- `solanaExplorerClusterName` in `SlotIndicator`, `AirdropButton`, and the transaction panels updates eagerly on chain switch. During the brief warm-up window, a slot / signature that was actually fetched from the previous network can be rendered inside an explorer link pointing at the newly-selected network. It's a narrow window, but the whole point of the client-derived chain in `Balance` was to close exactly this class of window.

Either pattern is defensible per-site, but there's currently no principle documented for choosing between them. Your own follow-up idea — a `useChain()` hook that returns both `selected` and `active` (and matching `useExplorerClusterName()` variants) — feels like the right shape, since the two are semantically different values. Even without that hook landing in this PR, it might be worth a short comment at the top of `ChainContext.tsx` (or in `ClientProvider.tsx`) documenting the split so future contributors don't have to reverse-engineer which one to reach for.

### Nice touches

- The comments in `Balance.tsx` and `routes/root.tsx` explaining why the chain is now read from the client (rather than `ChainContext`) are exactly the kind of thing that saves future readers a lot of head-scratching. Please keep this quality of comment on the sibling components when they migrate.
- The new `refetches against the new client when the network (client) switches` test is a great regression guard for the SWR-key-derived-from-client-chain invariant, and the comment inside it spells out precisely what the test is protecting against.
- Extending the client with `chain` via `extendClient` (rather than adding a whole plugin just for that) is the right weight of solution for the example app.

- The `as unknown as AppClient` casts in the two test wrappers are minimal stubs — `ClientProvider` from `@solana/react` accepts `Client<object>`, and `useClient<AppClient>()` narrows on the consumer side, so a plain object with the fields the component reads is sufficient. Worth being aware of if these tests later cover code paths that touch more of the real client surface (plugin capabilities, dispose, etc.) — the stubs would silently satisfy TypeScript while the runtime shape diverges.
- No changeset needed — examples-only changes.
- The comment in `SlotIndicator-test.browser.tsx` correctly keeps `ChainContext.Provider` because `SlotIndicator` still reads `solanaExplorerClusterName` from it. Consistent with the file it's testing.
- The docblock on `ClientProvider.tsx` still `{@link}`s `WalletReadyGate`, which isn't exported from this file. Pre-existing, not introduced here.

lorisleiva (approved):
In my mind a chain change would trigger a new client creation and hydrate the components accordingly but you've got a lot more context than me on that one so I trust your judgement. We can also chat more about it next week IRL. ☺️

mcintyre94:
### Merge activity

* **Aug 4, 4:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1844).
* **Aug 4, 4:16 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1844) couldn't merge this pull request because a downstack PR #1841 failed to merge.
* **Aug 4, 5:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1844).
* **Aug 4, 5:12 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1844) couldn't merge this pull request because a downstack PR #1825 failed to merge.
* **Aug 5, 7:47 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1844).
* **Aug 5, 7:48 AM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1844).

PR #1851, opened 2026-07-20
fix: narrow transaction message signer types

### LINKED ISSUE #1036: Narrowing the signer in `TransactionMessageWithSigners<TAddress, TSigner>` doesn't work
## Overview

I'm looking into whether we can make the `signTransactionMessageWithSigners` method return an unmodified specific lifetime (instead of the looser `TransactionWithLifetime`) in cases where the signers in a transaction message are non-modifying

I think this can be accomplished with some clever overloads, making use of the `TransactionMessageWithSigners<TAddress, TSigners>` type. As documented, this should allow us to narrow the type of signers permitted in the transaction.

But there's currently a bug where this type does not actually narrow the allowed signers. 

## Steps to reproduce

```ts
import { Address } from "@solana/addresses";
import { pipe } from "@solana/functional";
import { TransactionMessageWithSigners, TransactionModifyingSigner, TransactionPartialSigner, setTransactionMessageFeePayerSigner } from "@solana/signers";
import { createTransactionMessage } from "@solana/transaction-messages";

{
    const signer = null as unknown as TransactionModifyingSigner;

    const transactionMessage = pipe(
        tx => setTransactionMessageFeePayerSigner(signer, tx),
    )

    // no type error
    transactionMessage satisfies TransactionMessageWithSigners<Address, TransactionPartialSigner>;
}
```

## Description of bug

The issue is that the type is defined to allow eg a non-signer fee payer, or a signer fee payer. `TransactionModifyingSigner satisfies {feePayer: Address}` since the signer is a superset of the fee payer `{address: Address}`. This means that the type narrowing on `TransactionMessageWithFeePayerSigner<TAddress, TSigner>>` is not applied.

The same logic affects instructions, where an instruction with a `TransactionModifyingSigner` will satisfy the `Instruction` part of that check and again the narrowing is never applied.

If I'm understanding this correctly, the signer-related types in this function are never actually used currently, because the non-signer subsets are always satisfied by all signers.

We will need to define types like `TransactionMessageWithNonSignerFeePayer` that signers cannot match, so that the narrowed signer checks are applied when a signer is the fee payer/in an instruction.

rajanpanth:
## Summary

Make `TransactionMessageWithSigners<TAddress, TSigner>` enforce its signer type parameter for both fee payers and instruction account metas.

## Problem

Signer objects structurally satisfy the plain fee-payer and account-meta branches because their extra methods and `signer` property do not make them incompatible. As a result, a transaction containing a `TransactionModifyingSigner` can satisfy `TransactionMessageWithSigners<Address, TransactionPartialSigner>`.

## Changes

- Distinguish plain account metas from account metas containing signers.
- Distinguish plain fee payers from all transaction signer interfaces.
- Preserve signer extraction with the narrowed union types.
- Add regression typetests for matching, mismatched, and non-signer cases.
- Add a patch changeset for `@solana/signers`.

## Testing

- Signer typecheck
- Focused JS and declaration build
- Node signer unit suite: 138 tests
- Browser signer unit suite: 135 tests
- Full signer ESLint suite
- Direct Prettier validation

## Before

The narrowed signer parameter can be bypassed through the plain structural union branches.

## After

Mismatched fee-payer and instruction signers are rejected, while matching signers and plain non-signer account shapes remain accepted.

## Related issue

Fixes #1036

### DIFF
--- a/packages/signers/src/__typetests__/transaction-message-with-signers-typetest.ts
+++ b/packages/signers/src/__typetests__/transaction-message-with-signers-typetest.ts
@@ -1,0 +1,85 @@
+import { Address } from '@solana/addresses';
+import { AccountRole } from '@solana/instructions';
+import { createTransactionMessage } from '@solana/transaction-messages';
+
+import { TransactionMessageWithSigners } from '../account-signer-meta';
+import { setTransactionMessageFeePayerSigner } from '../fee-payer-signer';
+import { TransactionModifyingSigner } from '../transaction-modifying-signer';
+import { TransactionPartialSigner } from '../transaction-partial-signer';
+
+// [DESCRIBE] TransactionMessageWithSigners
+{
+    // It rejects a fee payer whose signer type is outside the requested signer type
+    {
+        const signer = null as unknown as TransactionModifyingSigner;
+        const transactionMessage = setTransactionMessageFeePayerSigner(
+            signer,
+            createTransactionMessage({ version: 0 }),
+        );
+
+        // @ts-expect-error The fee payer is not a TransactionPartialSigner.
+        transactionMessage satisfies TransactionMessageWithSigners<Address, TransactionPartialSigner>;
+    }
+
+    // It accepts plain fee payers and account metas without signers
+    {
+        const accountAddress = null as unknown as Address;
+        ({
+            feePayer: { address: accountAddress },
+            instructions: [
+                {
+                    accounts: [
+                        {
+                            address: accountAddress,
+                            role: AccountRole.READONLY,
+                        },
+                    ],
+                    programAddress: accountAddress,
+                },
+            ],
+        }) satisfies TransactionMessageWithSigners<Address, TransactionPartialSigner>;
+    }
+
+    // It accepts fee payer and instruction signers matching the requested signer type
+    {
+        const signer = null as unknown as TransactionPartialSigner;
+        const transactionMessage = {
+            feePayer: signer,
+            instructions: [
+                {
+                    accounts: [
+                        {
+                            address: signer.address,
+                            role: AccountRole.READONLY_SIGNER,
+                            signer,
+                        },
+                    ],
+                    programAddress: signer.address,
+                },
+            ],
+        } as const;
+        transactionMessage satisfies TransactionMessageWithSigners<Address, TransactionPartialSigner>;
+    }
+
+    // It rejects an instruction signer outside the requested signer type
+    {
+        const signer = null as unknown as TransactionModifyingSigner;
+        const transactionMessage = {
+            instructions: [
+                {
+                    accounts: [
+                        {
+                            address: signer.address,
+                            role: AccountRole.READONLY_SIGNER,
+                            signer,
+                        },
+                    ],
+                    programAddress: signer.address,
+                },
+            ],
+        } as const;
+
+        // @ts-expect-error The instruction signer is not a TransactionPartialSigner.
+        transactionMessage satisfies TransactionMessageWithSigners<Address, TransactionPartialSigner>;
+    }
+}
--- a/packages/signers/src/__typetests__/transaction-message-with-signers-typetest.ts
+++ b/packages/signers/src/__typetests__/transaction-message-with-signers-typetest.ts
@@ -2,10 +2,11 @@ import { Address } from '@solana/addresses';
 import { AccountRole } from '@solana/instructions';
 import { createTransactionMessage } from '@solana/transaction-messages';
 
-import { TransactionMessageWithSigners } from '../account-signer-meta';
+import { InstructionWithSigners, TransactionMessageWithSigners } from '../account-signer-meta';
 import { setTransactionMessageFeePayerSigner } from '../fee-payer-signer';
 import { TransactionModifyingSigner } from '../transaction-modifying-signer';
 import { TransactionPartialSigner } from '../transaction-partial-signer';
+import { TransactionSendingSigner } from '../transaction-sending-signer';
 
 // [DESCRIBE] TransactionMessageWithSigners
 {
@@ -81,5 +82,25 @@ import { TransactionPartialSigner } from '../transaction-partial-signer';
 
         // @ts-expect-error The instruction signer is not a TransactionPartialSigner.
         transactionMessage satisfies TransactionMessageWithSigners<Address, TransactionPartialSigner>;
+    }
+
+    // It accepts mixed signer kinds when using the default signer type parameter
+    {
+        const feePayer = null as unknown as TransactionModifyingSigner;
+        const instructionSigner = null as unknown as TransactionSendingSigner;
+        const instruction = {
+            accounts: [
+                {
+                    address: instructionSigner.address,
+                    role: AccountRole.READONLY_SIGNER,
+                    signer: instructionSigner,
+                },
+            ],
+            programAddress: instructionSigner.address,
+        } as const satisfies InstructionWithSigners & { programAddress: Address };
+        ({
+            feePayer,
+            instructions: [instruction],
+        }) satisfies TransactionMessageWithSigners;
     }
 }
--- a/packages/signers/src/account-signer-meta.ts
+++ b/packages/signers/src/account-signer-meta.ts
@@ -38,12 +38,22 @@ export interface AccountSignerMeta<
     readonly signer: TSigner;
 }
 
+type AccountMetaWithoutSigner = Readonly<{ signer?: never }> & (AccountLookupMeta | AccountMeta);
+
+type TransactionMessageWithNonSignerFeePayer<TAddress extends string = string> = Readonly<{
+    feePayer: Readonly<{
+        modifyAndSignTransactions?: never;
+        signAndSendTransactions?: never;
+        signTransactions?: never;
+    }> &
+        TransactionMessageWithFeePayer<TAddress>['feePayer'];
+}>;
+
 /**
  * A union type that supports base account metas as well as {@link AccountSignerMeta | signer account metas}.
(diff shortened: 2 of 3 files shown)


lorisleiva:
That's not a bad design. @mcintyre94 wdyt?

packages/signers/src/account-signer-meta.ts
```diff
@@ -38,12 +38,22 @@ export interface AccountSignerMeta<
...
 
+type AccountMetaWithoutSigner = Readonly<{ signer?: never }> & (AccountLookupMeta | AccountMeta);
+
+type TransactionMessageWithNonSignerFeePayer<TAddress extends string = string> = Readonly<{
+    feePayer: Readonly<{
+        modifyAndSignTransactions?: never;
+        signAndSendTransactions?: never;
+        signTransactions?: never;
+    }> &
+        TransactionMessageWithFeePayer<TAddress>['feePayer'];
```
mcintyre94: Something we might like to do here is to make this defensive against a future signer type, by deriving the methods of `TransactionSigner`. I think we could do something like this:

```ts
type UnionKeys<T> = T extends unknown ? keyof T : never;
type TransactionSignerMethodKeys = Exclude<UnionKeys<TransactionSigner>, 'address'>;
type WithoutSignerMethods = { readonly [K in TransactionSignerMethodKeys]?: never };

type TransactionMessageWithNonSignerFeePayer<TAddress extends string = string> = Readonly<{
    feePayer: TransactionMessageWithFeePayer<TAddress>['feePayer'] & WithoutSignerMethods;
}>;
```

[Typescript playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgFSgQwHYGc0GMbAQYCyApljgOakDqwMAFgGKmkAKaAnqVADQrpseAkQDKwShh5wAvnABmUCCDgByAAJYIAG0xoA9AGt6qgNwAoczE5hScAKoZCGANKlOWADzIAfHAC8KHCkAB4wpBgAJlhwAK4YhhgQAO4YcAD8cIbuEPJBAFxwUgBuPBbWtgKYOPjO4pI8ZIwQkW4eAXAAoiG42rGRpJ6Ozm1eqNXCdRJSUD78qmiRkVDkWKo+5TZ2dM2xMPUzTQwtMYFIK4tE2pxwANoucMBp40K1YtONpM2t7lgAuulCiVpDILFYtlVXiISKs0NQdgwAHLvBpQFjsLg8bwAQSWKwowTCEWicCwMCgT0oHTJFIwlD8gQASqRLhhrp4EOY4NyFKwONwoIUXjVoWQKHDaPRmHzMVAcXjVj5bqp5DKBao-nAAGRwBEQPYHT7fLAWGQbcxAA)

mcintyre94 (dismissed):
I think this looks good, definitely think it was a mistake to accept the other signer types here. Just a small suggestion to make the `feePayer` exclusion more future proof.

rajanpanth:
Friendly follow-up on this PR 👋

When you have a moment, could you please take a look? I’m happy to update anything if needed.

Thanks!

rajanpanth:
Thanks for the earlier suggestion on future-proofing eePayer exclusions. I implemented it by deriving signer method keys from TransactionSigner (UnionKeys + Exclude<..., 'address'>) so new signer methods are automatically excluded for non-signer fee payers.

mcintyre94:
@trevor-cortex

packages/signers/src/account-signer-meta.ts
```diff
@@ -38,12 +38,23 @@ export interface AccountSignerMeta<
     readonly signer: TSigner;
 }
 
+type AccountMetaWithoutSigner = Readonly<{ signer?: never }> & (AccountLookupMeta | AccountMeta);
+
+// Derived from `TransactionSigner`'s member types rather than hardcoded, so that a future signer
+// variant with a new method automatically excludes that method from a non-signer fee payer too.
+type UnionKeys<T> = T extends unknown ? keyof T : never;
+type TransactionSignerMethodKeys = Exclude<UnionKeys<TransactionSigner>, 'address'>;
+type WithoutSignerMethods = Readonly<{ [K in TransactionSignerMethodKeys]?: never }>;
```
trevor-cortex: Nice — deriving the keys from the `TransactionSigner` union addresses the future-proofing concern from the last round. One optional step further: instead of hardcoding `'address'`, you could exclude every key of the plain fee-payer shape:

```ts
type TransactionSignerMethodKeys = Exclude<
    UnionKeys<TransactionSigner>,
    keyof TransactionMessageWithFeePayer['feePayer']
>;
```

Today both spell `'address'`, but if a property is ever added to both the signer interfaces and the plain fee-payer shape, the hardcoded variant would mark it `never` and make plain fee payers unrepresentable. Non-blocking.

packages/signers/src/__typetests__/transaction-message-with-signers-typetest.ts
```diff
@@ -0,0 +1,85 @@
...
+                    ],
+                    programAddress: signer.address,
+                },
+            ],
+        } as const;
+
+        // @ts-expect-error The instruction signer is not a TransactionPartialSigner.
+        transactionMessage satisfies TransactionMessageWithSigners<Address, TransactionPartialSigner>;
+    }
+}
```
trevor-cortex, line 106: Consider adding one more case covering the default type parameter, e.g. a message mixing a `TransactionModifyingSigner` fee payer with a `TransactionSendingSigner` instruction signer satisfying plain `TransactionMessageWithSigners` (no explicit `TSigner`). All four current cases pin `TransactionPartialSigner`, so nothing here would catch a future refactor that accidentally over-narrows the wide/default case — which is the most common way this type is consumed.

trevor-cortex (commented):
## Summary

This PR closes a structural typing hole in `TransactionMessageWithSigners<TAddress, TSigner>`: because signer objects are structurally compatible with the plain fee-payer and account-meta branches of the union, a message carrying e.g. a `TransactionModifyingSigner` could satisfy `TransactionMessageWithSigners<Address, TransactionPartialSigner>`. The fix adds explicit exclusions on the non-signer branches:

- `AccountMetaWithoutSigner` adds `signer?: never` so a mismatched `AccountSignerMeta` can no longer sneak in through the plain meta branch.
- `TransactionMessageWithNonSignerFeePayer` intersects the plain fee-payer shape with `?: never` for every `TransactionSigner` method key, derived from the union rather than hardcoded — which also addresses the future-proofing point from my earlier review round.
- Regression typetests cover the mismatched, matching, and plain non-signer cases, plus a patch changeset.

The approach is sound and this looks good to me (I can't formally approve on this repo, so leaving this as a comment review). The two branches use different exclusion mechanisms (property-key `signer?: never` for metas, method-key exclusion for fee payers), which is the right call given the fee payer *is* the signer while metas wrap one.

## Notes

- The `getSignersFromInstruction` change from `'signer' in account` to a truthiness check is a slight runtime behavior change, but strictly an improvement: an explicitly-`undefined` `signer` (which `signer?: never` permits) previously flowed into `deduplicateSigners` and would have thrown on `signer.address`; now it's skipped.
- For subsequent reviewers: this ships as a `patch`, so external consumers who (likely unknowingly) relied on the unsound assignability — passing a wider signer kind where a narrower one was requested — will start getting type errors on upgrade. That's exactly the point of the fix, but it's the kind of change that can surface in downstream repos (e.g. Codama-generated clients, wallet adapters) that this repo's CI can't see. Worth keeping an eye on issue reports after release.
- CI (`pnpm build`) will confirm no other workspace package tripped over the narrowed union; the PR description reports the signer suites and typechecks passing.

rajanpanth:
Thanks for the follow-ups — I pushed a commit addressing both points:\n\n- switched signer-method exclusion to derive shared keys from \TransactionMessageWithFeePayer['feePayer']\ (avoids hardcoding \'address'\),\n- added a default-parameter typetest covering mixed signer kinds.\n\nValidation run: \pnpm --dir kit --filter @solana/signers test:typecheck\.

rajanpanth:
I addressed the future-proofing follow-up by deriving the excluded signer keys from the signer union, and added coverage for the default signer type parameter. The current checks are green. Could a maintainer please re-review the latest commit when convenient?

trevor-cortex (commented):
## Re-review

Both points from my previous review have been addressed:

1. **Derived fee-payer exclusion keys** — `TransactionSignerMethodKeys` now uses `Exclude<UnionKeys<TransactionSigner>, keyof TransactionMessageWithFeePayer['feePayer']>` instead of hardcoding `'address'`, with a comment explaining the rationale. A field added to both the plain fee-payer shape and the signer interfaces in the future will automatically drop out of the `?: never` set, so plain fee payers stay representable. ✅
2. **Default `TSigner` typetest** — the new "accepts mixed signer kinds when using the default signer type parameter" block covers the wide case (`TransactionModifyingSigner` fee payer + `TransactionSendingSigner` instruction signer with no type arguments), guarding against a future over-narrowing regression. ✅

Only other delta since my last pass is the explicit type arguments on `deduplicateSigners<TSigner>` and `getSignersFromInstruction<TSigner>` in the extraction functions — behavior-neutral inference aids for the narrowed union, no concerns.

Everything else is unchanged from what I already reviewed: the `signer?: never` exclusion on plain account metas, the fee-payer branch swap in `TransactionMessageWithSigners`, and the (strictly safer) truthiness check in `getSignersFromInstruction` all still look correct.

LGTM — no further comments. (Leaving this as a comment review since I don't have approval permissions on this repo.)

lorisleiva (approved):
Thanks! Would love @mcintyre94's approval here as well though. 🙏

mcintyre94 (approved):
Nice, thank you!

PR #1867, opened 2026-07-29
Simplify the react-app example's chain handling

mcintyre94:
#### Summary of Changes

As a next step in refactoring the example app to use plugins, this PR drastically simplifies the way the example app handles chains to make it work better with the Kit client.

We remove the chain provider/context, now no app components read the chain from here

It's replaced by a small `usePersistedChain` hook that just stores chain as state, reads it from localstorage on load, writes it on change. Same behaviour as before. The same file has functions to map a chain to a display name / explorer URL / etc. This is just app logic.

The `ClientProvider` is the only consumer of this chain state, and is unchanged - it re-builds the client when the chain changes, and uses `extendClient` to add the chain to the client.

The rest of the app now reads the chain from the client, meaning that everything is synced with the client changing. This means that the RPC and chain displayed are always aligned. There's no longer any confusion about which context the chain should be read from - it's always the client.

Other stuff:

- I also renamed the environment variable that enables mainnet to `REACT_EXAMPLE_APP_MAINNET_URL`, and this is now set to an endpoint instead of `true`. The default mainnet-beta endpoint doesn't work for this app, makes more sense to pass your own URL if you want to use mainnet.

- I noticed a bug where the sign-then-send panel would keep its transaction in state if you changed chain (or account), and that transaction can't send after changing chain. The panels have other dependent state, like filtering recipient accounts by chain. We now re-render these panels when the chain changes which I think makes more sense. This also simplifies their error boundary, which no longer needs its own re-render key.


mcintyre94:
* **#1867** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1867?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1867?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

examples/react-app/src/routes/root.tsx
```diff
@@ -78,24 +75,29 @@ function Root() {
                 </Flex>
             </Flex>
             <Dimmable busy={isStale}>
-                <DataList.Root orientation={{ initial: 'vertical', sm: 'horizontal' }} size="3">
+                {/* Re-render the panels when account or chain changes */}
+                <DataList.Root
+                    key={`${getUiWalletAccountStorageKey(connected.account)}:${chain}`}
+                    orientation={{ initial: 'vertical', sm: 'horizontal' }}
+                    size="3"
```
trevor-cortex, line 82: Worth a quick sanity check that this behaviour change is intentional.

Previously each `ErrorBoundary` got `resetKeys={[chain, accountKey]}`, which only tells the error boundary to *retry rendering* when those keys change — the panel components themselves stay mounted and keep their local state (recipient selection, lamports input, dialog open/error state, etc.).

With the new `key={\`${accountKey}:${chain}\`}` on `<DataList.Root>`, the entire subtree unmounts and remounts on any chain **or** account change, so all that per-panel local state is wiped. That's a broader behaviour change than dropping `ChainContext` implies, though it's arguably the more correct behaviour on a chain switch (a devnet recipient isn't meaningful on testnet). If intentional, no action needed — just wanted it on the record.

examples/react-app/src/chain.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/**
+ * The RPC HTTP endpoint for a chain, branded to that cluster. `solanaRpc` derives the WebSocket
+ * (subscriptions) URL from this by swapping the protocol to `ws`/`wss`, so only the HTTP URL is
+ * needed here. Mainnet uses `REACT_EXAMPLE_APP_MAINNET_URL` when set (see {@link SUPPORTED_CHAINS}).
+ */
+export function getChainRpcUrl(chain: SolanaChain): ClusterUrl {
+    switch (chain) {
+        case 'solana:mainnet':
```
trevor-cortex, line 51: Nit: the `?? 'https://api.mainnet-beta.solana.com'` fallback is effectively unreachable via the UI — `SUPPORTED_CHAINS` only includes `'solana:mainnet'` when `REACT_EXAMPLE_APP_MAINNET_URL` is set, so by the time this branch runs the env var is guaranteed non-empty. Either drop the fallback and non-null-assert (or throw with a clearer message), or leave a one-liner comment noting the fallback is defence-in-depth for direct callers. Not blocking.

examples/react-app/src/chain.ts
```diff
@@ -0,0 +1,88 @@
...
+        if (stored != null) {
+            localStorage.removeItem(STORAGE_KEY);
+        }
+        return DEFAULT_CHAIN;
+    });
+    function setChain(nextChain: SolanaChain) {
+        localStorage.setItem(STORAGE_KEY, nextChain);
+        setChainState(nextChain);
+    }
+    return [chain, setChain];
```
trevor-cortex, line 87: Minor: `setChain` is recreated on every render. It's only handed to `Nav` (which forwards it to a `DropdownMenu.RadioGroup.onValueChange`), so there's no memoisation-sensitive consumer today and this is a non-issue in practice — but if you ever pass it deeper or into a `useEffect` dep array, wrap it in `useCallback` (or return a `useMemo`'d tuple) to keep the reference stable. Feel free to ignore.

examples/react-app/src/chain.ts
```diff
@@ -0,0 +1,88 @@
...
+    const [chain, setChainState] = useState<SolanaChain>(() => {
+        const stored = localStorage.getItem(STORAGE_KEY);
+        if (stored != null && isSupportedChain(stored)) {
+            return stored;
+        }
+        if (stored != null) {
+            localStorage.removeItem(STORAGE_KEY);
+        }
+        return DEFAULT_CHAIN;
+    });
```
trevor-cortex, line 82: Small behaviour change from the old `ChainContextProvider`: previously an unrecognised stored value that happened to be `'solana:mainnet'` (with the enable flag off) would silently fall through to testnet **without clearing** storage, so re-enabling mainnet later would restore the selection. Now any unrecognised value — including a stashed `'solana:mainnet'` when the mainnet URL env var isn't set — gets wiped from localStorage. Probably fine (and arguably cleaner), just flagging.

trevor-cortex (commented):
## Summary

This PR removes the `ChainContext` / `ChainContextProvider` pair and replaces it with:

- A tiny `usePersistedChain` hook in the new `src/chain.ts` (state + `localStorage` read/write, same behaviour as before).
- Standalone `getChainDisplayName` / `getExplorerClusterName` / `getChainRpcUrl` helpers and a `SUPPORTED_CHAINS` constant, colocated in the same file.
- A new top-level `<App />` component that owns the selected chain, passes it to `ClientProvider`, and hands the setter to `Nav`.
- Every downstream consumer now reads the *active* chain off `useClient()` instead of the context, so chain + rpc + subscriptions always move as one unit.

As a follow-on, `REACT_EXAMPLE_APP_ENABLE_MAINNET` is replaced by `REACT_EXAMPLE_APP_MAINNET_URL` (the URL doubles as the enable flag).

The refactor is a clear net win: one source of truth for the active chain, no more “selected vs active” skew across contexts, and the mainnet URL is finally configurable per-deployment instead of hardcoded. Tests were updated accordingly and the previously-brittle “lag render” Balance test is now replaced by a simpler client-swap assertion, which is appropriate since the lag is architecturally gone.

## Things to watch out for

**One user-visible behaviour change worth confirming is intentional.** In `src/routes/root.tsx`, the old code reset the per-panel `ErrorBoundary`s via `resetKeys={[chain, ...account]}` — that retries the render without unmounting the panels themselves. The new code hoists a `key={\`${accountKey}:${chain}\`}` onto the parent `<DataList.Root>`, which **remounts the entire panels subtree** on any chain or account change. That means all local state inside the feature panels (recipient dropdown selection, lamports input, success/error dialog state, etc.) is now wiped on a chain switch, whereas before it was preserved. This is probably desirable (a chain switch arguably invalidates the recipient context), but it's a broader behaviour change than “simplify chain handling” implies — flagging it so it's a conscious call. See inline comment for detail.

**Mainnet env-var rename is a breaking change for anyone with existing dev scripts / CI env** setting `REACT_EXAMPLE_APP_ENABLE_MAINNET=true`. It'll silently no-op after this PR (mainnet will just be hidden). The README covers the new variable, which is the important thing; just calling it out.

- Worth eyeballing `usePersistedChain` — no `try/catch` around `localStorage` access, but that matched the previous implementation, so no regression.
- The `getChainRpcUrl` mainnet fallback (`?? 'https://api.mainnet-beta.solana.com'`) is unreachable via the UI path (mainnet is only in `SUPPORTED_CHAINS` when the env var is set), but harmless. See inline nit.
- `ClientProvider-test.browser.tsx` now stubs `mainnet` and `testnet` on the `@solana/kit` mock in addition to `devnet`, which matches the new import surface in `ClientProvider` (transitively via `chain.ts`). Good catch.

examples/react-app/src/routes/root.tsx
mcintyre94: Forgot this in the PR description but yep intentional. Fixes a bug where you can sign on one chain, change chain, still in send state -> fails.

examples/react-app/src/chain.ts
mcintyre94: Made it a `useCallback` - worth using good practices in our example app

lorisleiva:
@trevor-cortex

trevor-cortex:
```
Anthropic Error [400]: invalid_request_error - messages.3.content.0.tool_use.input: Input should be an object
```

lorisleiva:
@trevor-cortex

trevor-cortex:
```
Anthropic Error [400]: invalid_request_error - messages.4.content.0.tool_use.input: Input should be an object
```

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Re-review

I re-fetched all diffs. The points from my previous pass have all been resolved or explicitly ratified, and I found nothing new worth blocking on. This would be an approval, but I don't have approve permissions on this repo — treat this as LGTM.

### Prior points, now settled

- **Panels remount on chain/account switch** (`root.tsx` `key` on `DataList.Root`): the updated PR description now documents this as a deliberate bug fix — the sign-then-send panel previously kept a stale, unsendable transaction in state across a chain switch, and the recipient list is chain-dependent anyway. Wiping panel-local state on switch is the right call, and it's now a conscious, documented decision rather than an implicit side effect. Resolved.
- **`setChain` memoisation**: `usePersistedChain` now wraps the setter in `useCallback`. Resolved.
- **Unrecognised stored chain values**: the `useState` initializer now clears stale `localStorage` entries (`removeItem` on unrecognised values), matching the old provider's behaviour. Nice touch that this also handles the "selected mainnet, then removed `REACT_EXAMPLE_APP_MAINNET_URL`" case gracefully — the stored `solana:mainnet` fails the `SUPPORTED_CHAINS` check and falls back to devnet. Resolved.
- **Env-var rename** (`REACT_EXAMPLE_APP_ENABLE_MAINNET` → `REACT_EXAMPLE_APP_MAINNET_URL`): now called out in the PR description and covered by the README. Fine for an example app.
- **Unreachable mainnet `??` fallback in `getChainRpcUrl`**: still present, but it's harmless defense-in-depth and I won't re-raise it.

### Remaining observations (non-blocking, no action needed)

- The `removeItem` call inside the `useState` initializer is technically a render-phase side effect (it runs twice under StrictMode), but it's idempotent and the old `ChainContextProvider` did the equivalent inside a `useMemo`, so this is no worse than before.
- `SolanaChain` includes `solana:localnet`, which all three helpers silently map to devnet via their `default` branches. Unreachable through the UI since `SUPPORTED_CHAINS` never offers it — just noting for anyone extending the example later.

Good simplification overall — one source of truth for the active chain, panels re-keyed off it, and the tests read much more directly now that the "lag render" gymnastics are gone. Ship it.

lorisleiva (approved):

mcintyre94:
### Merge activity

* **Aug 4, 4:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1867).
* **Aug 4, 4:16 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1867) couldn't merge this pull request because a downstack PR #1841 failed to merge.
* **Aug 4, 5:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1867).
* **Aug 4, 5:12 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1867) couldn't merge this pull request because a downstack PR #1825 failed to merge.
* **Aug 5, 7:47 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1867).
* **Aug 5, 7:49 AM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1867) rebased this pull request as part of a merge.
* **Aug 5, 8:03 AM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1867).

PR #1870, opened 2026-07-30
Bump examples to latest @solana-program/ libs, fix deprecations

mcintyre94:
I noticed while using our examples to test v1 transactions that some of them are using older versions of our program clients and deprecated functions.

- update all `@solana-program/` dependencies to the latest version
- fix deprecations in token-airdrop, new resource limit functions + new estimator + `getCreateMintInstructionPlan` updated to take a client as first param
- Update explorer links for localhost (they require http:// now)

### DIFF
--- a/examples/deserialize-transaction/package.json
+++ b/examples/deserialize-transaction/package.json
@@ -12,9 +12,9 @@
         "test:typecheck": "tsc"
     },
     "dependencies": {
-        "@solana-program/address-lookup-table": "^0.12.1",
-        "@solana-program/memo": "^0.11.2",
-        "@solana-program/system": "^0.12.2",
+        "@solana-program/address-lookup-table": "^0.13.0",
+        "@solana-program/memo": "^0.12.0",
+        "@solana-program/system": "^0.13.0",
         "@solana/example-utils": "workspace:*",
         "@solana/kit": "workspace:*"
     },
--- a/examples/deserialize-transaction/src/example.ts
+++ b/examples/deserialize-transaction/src/example.ts
@@ -370,12 +370,7 @@ if (identifiedInstruction === SystemInstruction.TransferSol) {
     // Narrow the type again, the instruction must have accounts to be parsed as a transfer SOL instruction
     assertIsInstructionWithAccounts(firstInstruction);
 
-    // TODO: This can just be `parseTransferSolInstruction(firstInstruction)` when the client is updated
-    // with the `@solana/kit` version that changes the instruction data type to `ReadonlyUint8Array`
-    const parsedFirstInstruction = parseTransferSolInstruction({
-        ...firstInstruction,
-        data: firstInstruction.data as unknown as Uint8Array,
-    });
+    const parsedFirstInstruction = parseTransferSolInstruction(firstInstruction);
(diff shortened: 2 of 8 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Bumps the example apps to the latest `@solana-program/*` releases (`address-lookup-table` 0.12→0.13, `memo` 0.11→0.12, `system` 0.12→0.13, `token` 0.14→0.15) and migrates `token-airdrop` and `deserialize-transaction` to the corresponding Kit v7 APIs. The bumps are actually **required**, not just nice-to-have — every new `@solana-program/*` release declares `@solana/kit ^7.0.0` as a peer, and this repo is on `@solana/kit@7.0.0`, so the old pins are on the wrong major.

### Changes verified

- **`token-airdrop`**: migrated from the deprecated `estimateComputeUnitLimitFactory` / `estimateAndSetComputeUnitLimitFactory` / `fillTransactionMessageProvisoryComputeUnitLimit` trio to the new `…ResourceLimits…` equivalents. The `estimateWithMultiplier` wrapper was correctly rewritten to accept a `TransactionMessage & TransactionMessageWithFeePayer` generic and return a full `ResourceLimitsEstimate<T>` (preserving other estimated fields like `loadedAccountsDataSizeLimit`, only multiplying `computeUnitLimit`). The 10% CU buffer behavior is preserved.
- **`token-airdrop`**: new `ClientWithGetMinimumBalance` adapter passed as the first arg to `getCreateMintInstructionPlan`, matching the v7 removal of the standalone `getMinimumBalanceForRentExemption` helper. Nice touch adding the `MINIMUM BALANCE CLIENT` section with an explanatory docblock rather than just inlining it.
- **`deserialize-transaction`**: the `data: firstInstruction.data as unknown as Uint8Array` casts are gone now that `@solana-program/memo@0.12` and `@solana-program/system@0.13` accept `ReadonlyUint8Array` — the resolved TODO comments match the change.
- **package.json bumps** in the four example dirs plus `pnpm-lock.yaml` line up cleanly; no unrelated churn.

### Notes for reviewers

- **On the Copilot bot's Node engines comment**: the concern about `@solana-program/token@0.15.0` (and `@solana-program/address-lookup-table@0.13.0`) declaring `engines.node >=24.0.0` while the repo requires `node >=20.18.0` — this is **pre-existing**, not introduced by this PR. Both `@solana-program/address-lookup-table@0.12.1` and `@solana-program/token@0.14.0` already declared the same `>=24.0.0` engine before this bump. Whatever CI arrangement (or ignored engine warning) makes the current state work will continue to work after this bump. Worth tracking separately if `>=24` is genuinely a problem, but not a reason to block this PR.
- **Changeset**: none needed — all four modified `package.json`s are `examples/*` importers (private, unpublished), and there are no changes to any `packages/*` publishable package.
- **Worth smoke-testing** `pnpm start` in `examples/token-airdrop` against the local test validator, since the resource-limits estimator and the async `getCreateMintInstructionPlan` signature are both new surfaces here. The rest of the changes are essentially version bumps with no code changes.

LGTM overall — clean, focused maintenance PR.

lorisleiva (approved):
Nice, thanks for the cleanup.

PR #1869, opened 2026-07-30
Add react hooks for transaction-planning and -sending

mcintyre94:
#### Summary of Changes

This PR adds hooks for the functions from `ClientWithTransactionPlanning` and `ClientWithTransactionSending` to the `@solana/react` package

- `usePlanTransaction`
- `usePlanTransactions`
- `useSendTransaction`
- `useSendTransactions`

All are just `useAction`-shaped wrappers around these client functions. As with the wallet hooks, they take a `client` with the required plugin as a param.

PR #1872, opened 2026-07-30
Update the example app to use transaction plan/send

mcintyre94:
This PR updates the react example app to use our transaction plan/send functions, instead of hand-crafting transactions. Note that in a previous PR I added `solanaRpc` to the client, so it already has transaction plan/send functions available.

- The RPC plugin is updated to the latest version, with its improved CU calculation
- The sign + send panel is refactored to  use `useSendTransaction`
- The other panels are refactored to use `client.planTransaction` to construct their transaction

### DIFF
--- a/examples/react-app/src/components/SolanaSignTransactionFeaturePanel.tsx
+++ b/examples/react-app/src/components/SolanaSignTransactionFeaturePanel.tsx
@@ -1,14 +1,13 @@
 import { Blockquote, Box, Button, Dialog, Flex, Link, Select, Text, TextField } from '@radix-ui/themes';
 import {
     address,
-    appendTransactionMessageInstruction,
     assertIsSendableTransaction,
     assertIsTransactionWithBlockhashLifetime,
-    createTransactionMessage,
+    estimateAndSetResourceLimitsFactory,
+    estimateResourceLimitsFactory,
     getSignatureFromTransaction,
     pipe,
     sendAndConfirmTransactionFactory,
-    setTransactionMessageFeePayerSigner,
     setTransactionMessageLifetimeUsingBlockhash,
     signTransactionMessageWithSigners,
 } from '@solana/kit';
@@ -38,6 +37,11 @@ export function SolanaSignTransactionFeaturePanel({ signer }: Props) {
         () => sendAndConfirmTransactionFactory({ rpc, rpcSubscriptions }),
         [rpc, rpcSubscriptions],
     );
+    const estimateResourceLimits = useMemo(() => estimateResourceLimitsFactory({ rpc: client.rpc }), [client.rpc]);
+    const estimateAndSetResourceLimits = useMemo(() => estimateAndSetResourceLimitsFactory(async (tx, config) => {
+        const resourceLimits = await estimateResourceLimits(tx, config);
+        return { ...resourceLimits, computeUnitLimit: resourceLimits.computeUnitLimit + 300 };
+    }), [estimateResourceLimits]);
     const wallets = useWallets(client);
     const [solQuantityString, setSolQuantityString] = useState<string>('');
     const [recipientAccountStorageKey, setRecipientAccountStorageKey] = useState<string | undefined>();
@@ -68,22 +72,25 @@ export function SolanaSignTransactionFeaturePanel({ signer }: Props) {
         if (!recipientAccount) {
             throw new Error('The address of the recipient could not be found');
         }
+        const planned = await client.planTransaction(
+            getTransferSolInstruction({
+                amount,
+                destination: address(recipientAccount.address),
+                source: signer,
+            }),
+            { abortSignal: signal },
+        );
+        // `planTransaction` builds the message but does not set a lifetime or add CUs,
+        // these are part of the default transaction executor.
         const { value: latestBlockhash } = await rpc
             .getLatestBlockhash({ commitment: 'confirmed' })
             .send({ abortSignal: signal });
-        const message = pipe(
-            createTransactionMessage({ version: 0 }),
-            m => setTransactionMessageFeePayerSigner(signer, m),
-            m => setTransactionMessageLifetimeUsingBlockhash(latestBlockhash, m),
-            m =>
-                appendTransactionMessageInstruction(
-                    getTransferSolInstruction({
-                        amount,
-                        destination: address(recipientAccount.address),
-                        source: signer,
-                    }),
-                    m,
-                ),
+        const message = await pipe(
+            planned,
+            // Set the lifetime to the latest blockhash  
+            tx => setTransactionMessageLifetimeUsingBlockhash(latestBlockhash, tx),
+            // Estimate and set the resource limits for the transaction
+            async tx => await estimateAndSetResourceLimits(tx, { abortSignal: signal }),
         );
         const transaction = await signTransactionMessageWithSigners(message);
         assertIsSendableTransaction(transaction);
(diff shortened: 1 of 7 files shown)


mcintyre94:
* **#1872** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1872?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1872?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Refactors the three transfer panels in the react-app example to use the new client-level `planTransaction` / `useSendTransaction` helpers instead of hand-rolling transaction messages, and bumps `@solana/kit-plugin-rpc` from `0.13.0` → `0.15.0` to pick up the improved CU calculation. Net result is a substantial simplification: `SolanaSignAndSendTransactionFeaturePanel` loses ~44 lines, and `SolanaSignTransactionFeaturePanel` / `SolanaPartialSignTransactionFeaturePanel` each drop the `pipe(...)` + blockhash-fetch boilerplate. `assertCanSignAndSendTransactions` and its now-unused imports are removed from `walletCapability.ts`, and the surrounding `ErrorBoundary` around the sign-and-send panel is dropped in `root.tsx`.

Example-app-only, private package — no changeset needed.

## Things to watch out for

**1. Behavioral change in sign-and-send: `source` is now `client.payer` instead of the connected wallet's `signer`.** These should be equivalent here, because `walletSigner({ chain })` is registered before `solanaRpc(...)` in `ClientProvider.tsx` and sets `client.payer` from the connected wallet — but it's worth double-checking that the panel behaves the same when there are multiple connected wallets or when the connected account is changed mid-session. If `client.payer` ever diverges from "the currently selected wallet signer" in this app, the transfer's `source` would silently follow `client.payer` rather than the account the user thinks they're operating from.

**2. Removed capability guard for sign-and-send.** The old code called `assertCanSignAndSendTransactions(signer)` at render time inside an `ErrorBoundary` with `FeatureNotSupportedCallout`, so an account that doesn't expose `solana:signAndSendTransaction` got a nice "not supported" UI. With this PR, the panel is unconditionally rendered and any incompatibility only surfaces at dispatch time via `useSendTransaction`'s `error` field (which routes to `ErrorDialog`). That's a minor UX regression for accounts on the connected wallet that only support `signTransaction` — worth confirming whether `useSendTransaction` (or `client.sendTransaction` under the hood) picks a compatible signer automatically, or whether this now fails at click-time for those accounts. If it does fail at click time, consider keeping a lightweight capability check so the button is disabled rather than throwing after the user tries.

**3. Partial-sign panel: fee-payer swap after planning.** In `SolanaPartialSignTransactionFeaturePanel`, `client.planTransaction(...)` runs with the wallet as the implicit payer (via `client.payer`), then `setTransactionMessageFeePayerSigner(feePayerSigner, planned)` swaps the fee payer to the mock server signer. This works, and the comment on the line is helpful, but note that the plugin-rpc 0.15 CU calculation is done against the wallet-as-payer rather than the actual mock fee payer. In practice the CU estimate should be identical for a simple SOL transfer, but if this pattern is ever copied for a transaction where the fee payer materially changes account access (e.g. affects which accounts are writable), the estimate could drift. Not a blocker for the example — flagging for future readers.

- Verify with Callum whether `useSendTransaction` gracefully handles the case where the connected wallet account cannot sign-and-send (point 2 above). If it doesn't, either the `ErrorBoundary`/capability guard should stay, or the submit button should be disabled based on the connected account's capabilities.
- The lockfile change is just the `kit-plugin-rpc` bump; `@solana/kit-plugin-instruction-plan` is still pinned to `0.13.0` as a transitive — nothing to do, but worth eyeballing that this is intentional and not a version-skew concern for the plugin-rpc author.
- The formatting change in `root.tsx` around the `ErrorBoundary` for `Balance` is cosmetic (line wrap) and unrelated to the feature work — fine to keep.

mcintyre94:
Added back the capability check and `ErrorBoundary`, these still make sense for sign + send. We check for `assertCanSignTransactions` because that's the capability the transaction executor uses.

examples/react-app/src/components/SolanaSignTransactionFeaturePanel.tsx
```diff
@@ -68,22 +72,25 @@ export function SolanaSignTransactionFeaturePanel({ signer }: Props) {
...
-                ),
+        const message = await pipe(
+            planned,
+            // Set the lifetime to the latest blockhash  
+            tx => setTransactionMessageLifetimeUsingBlockhash(latestBlockhash, tx),
+            // Estimate and set the resource limits for the transaction
+            async tx => await estimateAndSetResourceLimits(tx, { abortSignal: signal }),
         );
         const transaction = await signTransactionMessageWithSigners(message);
         assertIsSendableTransaction(transaction);
```
mcintyre94, line 96: Highlighting this, I think this may be an API gap

We refactor to use `client.planTransaction` which I think is much better, but our default transaction planner intentionally does not set the lifetime nor the CUs, both of which are left to the default transaction executor. See https://github.com/anza-xyz/kit-plugins/blob/497b55fcdd7c1df905083e7eda160e06b220d555/packages/kit-plugin-rpc/src/transaction-plan-executor.ts#L143

This means that the app needs to handle adding the blockhash, and setting CUs. The default executor now has a quite sophisticated approach to CUs, but I don't think an app can easily use it outside of the executor.

My instinct here is that it may make sense to split out a `client.signTransaction` from `client.sendTransaction`. This would return a fully signed transaction that is ready to send but has not been sent.

The default implementation of this in our `solanaRpc` plugin would be the current default executor, except without the call to `sendAndConfirm`, it'd just return the transaction after signing it.

This panel would then collapse to simply `client.signTransaction(transferInstruction)` here, and then whatever send logic the app chooses to use as the second part. This would also be a clean migration from the wallet-adapter `signTransaction` API.

This needs a bit more design thought, so leaving as-is in this PR - this demonstrates the small amount of transaction logic an app will currently still need for the sign-then-send flow.

lorisleiva (approved):

mcintyre94:
### Merge activity

* **Aug 4, 4:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1872).
* **Aug 4, 4:16 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1872) couldn't merge this pull request because a downstack PR #1841 failed to merge.
* **Aug 4, 5:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1872).
* **Aug 4, 5:12 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1872) couldn't merge this pull request because a downstack PR #1825 failed to merge.
* **Aug 5, 7:47 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1872).
* **Aug 5, 8:16 AM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1872) couldn't merge this pull request because a downstack PR #1869 failed to merge.
* **Aug 5, 4:53 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1872).
* **Aug 5, 4:56 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1872) rebased this pull request as part of a merge.
* **Aug 5, 5:08 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1872).

PR #1874, opened 2026-07-31
Restrict compressTransactionMessageUsingAddressLookupTables to reject v1 messages

mcintyre94:
#### Problem

Spotted while testing our examples with v1 transactions, the `compressTransactionMessageUsingAddressLookupTables` was typed to reject legacy transactions but not v1. v1 transactions do not support lookup tables.

#### Summary of Changes

Updated to only allow v0 transactions

I've also checked the rest of the transaction-messages function and higher-level APIs, and can't find anywhere else we missed.

### DIFF
--- a/.changeset/sixty-ducks-count.md
+++ b/.changeset/sixty-ducks-count.md
@@ -1,0 +1,5 @@
+---
+'@solana/transaction-messages': patch
+---
+
+Update type of compressTransactionMessageUsingAddressLookupTables to reject v1 transactions
--- a/packages/transaction-messages/src/__typetests__/compress-transaction-message-typetest.ts
+++ b/packages/transaction-messages/src/__typetests__/compress-transaction-message-typetest.ts
@@ -19,11 +19,18 @@ const addressesByLookupTableAddress = null as unknown as AddressesByLookupTableA
         result satisfies v0TransactionMessage;
     }
 
-    // It accepts only v0 transaction messages
+    // It rejects legacy transaction messages
     {
-        const invalidMessage = null as unknown as Exclude<TransactionMessage, { version: 0 }>;
-        // @ts-expect-error Only v0 messages are accepted.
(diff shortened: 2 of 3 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Tightens the type signature of `compressTransactionMessageUsingAddressLookupTables` so it only accepts v0 transaction messages. Previously it excluded `legacy` but still accepted `v1`, which is incorrect since v1 messages don't support address lookup tables (their instruction type is `InstructionWithoutLookupTables`, same as legacy — confirmed in `packages/transaction-messages/src/transaction-message.ts`).

The change is a straightforward swap from `Exclude<TransactionMessage, { version: 'legacy' }>` to `Extract<TransactionMessage, { version: 0 }>`, which more accurately expresses the intent (allowlist v0 rather than blocklist legacy). The rename to `SupportedTransactionMessage` reads well too. LGTM.

## Things to watch out for

- **Breaking-ness**: this is a type-only tightening. Any caller currently passing a v1 message will start seeing a type error. Given v1 messages can't participate in ALT compression anyway, those callsites were already broken at runtime (or were dead code), so `patch` bump feels right and matches the changeset. Worth a quick sanity check that no internal callers in the repo were relying on the wider `NotLegacy` type.
- **Type test coverage**: the split into separate `legacy` and `v1` `@ts-expect-error` blocks is a nice improvement — it now proves both rejections independently rather than relying on the union.

- PR author confirms they audited the rest of `transaction-messages` and higher-level APIs for the same oversight and found nothing else. Worth a spot-check on anything else that constructs or consumes ALT-related shapes (e.g. compilation paths, signers) to make sure v1 isn't silently accepted there either, but I didn't find anything obvious.
- No runtime code changes — the function body still works identically for v0 inputs.

lorisleiva (approved):

PR #1875, opened 2026-07-31
Refactor the example app to use the system progam plugin

mcintyre94:
Small refactor of the example app to use the `systemProgram()` plugin and its instruction builders, instead of importing instruction builder functions.

As the app is using the react hooks I'm just using the plugin to build the instructions, not its `.plan/sendTransaction` functionality - which would need to be wrapped in `useAction`.

PR #1876, opened 2026-07-31
Add react hooks for usePayer and useIdentity

mcintyre94:
This PR adds `usePayer` and `useIdentity` react hooks, to surface the `client.payer` and `client.identity` values

These hooks are reactive, if the client implements the `subscribeTo[Payer/Identity]` then the hooks subscribe to it (using `useSyncExternalStore`) and always provide the latest value.

Unlike `payer` and `identity` which are not defined as optional on the client (the wallet plugin throws when it owns them and doesn't have a connected wallet to provide), I've made the hooks return `TransactionSigner | undefined`. This is because `useSyncExternalStore` doesn't really work with exceptions, it seems to tear down the subscription if the value it's trying to read throws, so it doesn't recover properly from a wallet simply being disconnected and reconnected etc. even with an `ErrorBoundary` in place. I think in react it makes more sense to just make these values optional.

### DIFF
--- a/packages/react/src/useIdentity.ts
+++ b/packages/react/src/useIdentity.ts
@@ -1,0 +1,47 @@
+import type { ClientWithIdentity, ClientWithSubscribeToIdentity, TransactionSigner } from '@solana/kit';
+import { useCallback, useSyncExternalStore } from 'react';
+
+const NOOP_UNSUBSCRIBE = () => {};
+
+/**
+ * Reads `client.identity` and re-renders whenever it changes, returning `undefined` while no
+ * identity is available.
+ *
+ * The identity is the {@link TransactionSigner} representing the wallet whose on-chain assets the
+ * application is acting upon.
+ *
+ * When the client advertises {@link ClientWithSubscribeToIdentity},
+ * this hook subscribes via `client.subscribeToIdentity` so the returned value always reflects the
+ * latest identity. For a client whose identity is fixed for its lifetime, the hook falls back to a
+ * no-op subscription and simply reads the value once.
+ *
+ * @param client - A client with an identity plugin installed. If it also advertises
+ *   `subscribeToIdentity`, the hook tracks changes reactively.
+ * @returns The current `client.identity` signer, or `undefined` if no identity is currently
+ *   available.
+ *
+ * @example
+ * ```tsx
+ * const identity = useIdentity(client);
+ * return <span>{identity ? `Signed in as ${identity.address}` : 'Signed out'}</span>;
+ * ```
+ *
+ * @see {@link usePayer}
+ */
+export function useIdentity(
+    client: ClientWithIdentity & Partial<ClientWithSubscribeToIdentity>,
+): TransactionSigner | undefined {
+    const subscribe = useCallback(
+        (onStoreChange: () => void) =>
+            client.subscribeToIdentity ? client.subscribeToIdentity(onStoreChange) : NOOP_UNSUBSCRIBE,
+        [client],
+    );
+    const getSnapshot = useCallback((): TransactionSigner | undefined => {
+        try {
+            return client.identity;
+        } catch {
+            return undefined;
+        }
+    }, [client]);
+    return useSyncExternalStore(subscribe, getSnapshot, getSnapshot);
+}
--- a/packages/react/src/usePayer.ts
+++ b/packages/react/src/usePayer.ts
@@ -1,0 +1,44 @@
+import type { ClientWithPayer, ClientWithSubscribeToPayer, TransactionSigner } from '@solana/kit';
+import { useCallback, useSyncExternalStore } from 'react';
+
+const NOOP_UNSUBSCRIBE = () => {};
+
+/**
+ * Reads `client.payer` and re-renders whenever it changes, returning `undefined` while no payer is
+ * available.
+ *
+ * The payer is the {@link TransactionSigner} a client uses to sign and pay for transactions by
+ * default.
+ *
+ * When the client advertises {@link ClientWithSubscribeToPayer}, this hook subscribes via
+ * `client.subscribeToPayer` so the returned value always reflects the latest payer. For a client
+ * whose payer is fixed for its lifetime, the hook falls back to a no-op subscription and simply
+ * reads the value once.
+ *
+ * @param client - A client with a payer plugin installed. If it also advertises
+ *   `subscribeToPayer`, the hook tracks changes reactively.
+ * @returns The current `client.payer` signer, or `undefined` if no payer is currently available.
+ *
+ * @example
+ * ```tsx
+ * const payer = usePayer(client);
+ * return <span>{payer ? `Paying with ${payer.address}` : 'No payer'}</span>;
+ * ```
+ *
+ * @see {@link useIdentity}
+ */
+export function usePayer(client: ClientWithPayer & Partial<ClientWithSubscribeToPayer>): TransactionSigner | undefined {
+    const subscribe = useCallback(
+        (onStoreChange: () => void) =>
+            client.subscribeToPayer ? client.subscribeToPayer(onStoreChange) : NOOP_UNSUBSCRIBE,
+        [client],
+    );
+    const getSnapshot = useCallback((): TransactionSigner | undefined => {
+        try {
+            return client.payer;
+        } catch {
+            return undefined;
+        }
+    }, [client]);
+    return useSyncExternalStore(subscribe, getSnapshot, getSnapshot);
+}
--- a/.changeset/clear-teams-train.md
+++ b/.changeset/clear-teams-train.md
@@ -1,0 +1,7 @@
+---
+'@solana/react': minor
+---
+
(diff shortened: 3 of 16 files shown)


mcintyre94:
* **#1876** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1876?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1876?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds two new public hooks to `@solana/react`: `usePayer(client)` and `useIdentity(client)`. Both read the corresponding signer off the client, and — when the client also advertises `subscribeToPayer`/`subscribeToIdentity` per the `ClientWithSubscribeTo*` convention — subscribe reactively via `useSyncExternalStore` so the returned value always reflects the latest signer. Clients without the subscribe hook fall back to a no-op subscribe and read the value once. To make the hooks robust against wallet-plugin getters that throw when no wallet is connected, both return `TransactionSigner | undefined` and swallow any error thrown by the getter into `undefined` — a deliberate departure from the non-optional plugin type. Ships with test-utility `renderHook` browser tests, `__typetests__` for the public signature, and a `minor` changeset.

Also rolls in a batch of unrelated prettier reformatting across several existing test files and adds a blank line to `examples/react-app/src/routes/root.tsx` — presumably from `pnpm style:fix`. Harmless drive-by cleanup.

- **Swallowing all errors from the getter is a design choice, not just wallet-not-connected.** The `try { return client.<x>; } catch { return undefined; }` treats any exception — including bugs in a plugin's getter or a corrupted state — as "signer unavailable". The PR description justifies this via `useSyncExternalStore`'s incompatibility with throwing snapshots. It's the right call for the wallet-disconnected path, but it does mean genuine bugs will silently surface as an absent signer rather than an error boundary hit. Might be worth a note in the docblock and/or a `__DEV__` `console.warn` on caught errors so developers can spot unexpected throws during development.
- **Reactive contract requires a stable `client.<x>` reference for equal snapshots.** `useSyncExternalStore` uses `Object.is` between snapshots; a getter that returns a new signer object on every read would cause infinite re-renders. That's the plugin's contract to uphold (and it's implied by `subscribe-to.ts`'s "consumers that bail on reference-equal snapshots… will filter redundant notifications out for free"), but worth calling out for plugin authors.
- **README not updated.** The `@solana/react` README's "Kit client bindings" section documents `ClientProvider`, `useClient`, `useClientCapability`, etc. and the repo's TypeScript READMEs skill says new public APIs should be added there. These two hooks should get a short section — especially because the `undefined`-on-throw behaviour and the "mutation without `subscribeTo*` isn't observed until the next render" semantic (which the last test in each file exercises) aren't obvious from the type signature alone.

- The hooks intentionally take `client` as an argument rather than reading it from `ClientProvider` via `useClient`/`useClientCapability`. That's consistent with the newer hook style (`usePlanTransaction`, `useSendTransaction`, etc.) but does differ from the `useClient*`-based pattern documented in the README's "Kit client bindings" intro — worth checking whether this is the intended direction for capability hooks going forward, and whether the README's framing should be adjusted.
- The type tests use `ClientWithSubscribeToIdentity`/`ClientWithSubscribeToPayer` alone as the negative case (`@ts-expect-error` because the base capability is missing). That correctly verifies the base capability is required; there's no negative test for passing an object that has neither, which would also be worth locking in.
- `NOOP_UNSUBSCRIBE = () => {}` is duplicated across both files. Fine as-is (one line, module-local), but if a third `use<Capability>` hook lands with the same pattern it would be worth pulling into a shared internal helper.
- Changeset file has no trailing newline (`\ No newline at end of file` in the diff). Trivial — `style:fix` or the changeset CLI will normalize it.
- Prettier reformatting in `useAction-test.browser.tsx`, `useReactiveStoreLifecycle-test.browser.tsx`, `useSubscription-test.browser.tsx`, `useRequestQuery-test.browser.tsx`, `useRequestSWR-test.browser.tsx`, and `ClientProvider-test.browser.tsx` is behaviour-neutral; safe to skim past.

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Re-review

(Submitting as COMMENT since I don't have approve permissions on this repo — consider this an approval.)

Since my last pass, the main change is the new **"Payer & identity" README section** (`packages/react/README.md`), which addresses the primary point from my previous review. It documents both hooks, the `undefined`-while-unavailable return, and the reactive vs. fixed-client subscription behaviour — and the example is framed around `useClient<AppClient>()`, which also resolves my earlier note about aligning with the README's `useClient`-based intro. The changeset was also expanded to explicitly call out that a throwing plugin getter surfaces as `undefined`, which covers the discoverability concern about the swallow-all-errors behaviour at the release-notes level.

Everything else is unchanged from my prior review; I won't re-raise the minor points already noted there (`NOOP_UNSUBSCRIBE` duplication, missing trailing newline in the changeset, the optional `__DEV__` warn on caught getter errors, and the absent negative typetest for a client with neither capability). All were explicitly optional and none block merging.

One truly optional thought: the docblocks and README say the hooks return `undefined` "while no payer/identity is available", which is accurate for the intended wallet-disconnected case but doesn't spell out that *any* exception from the getter is coerced to `undefined`. The changeset covers it, so I'm fine shipping as-is — just flagging in case you'd like the docblock `@returns` to mention it too.

Implementation, tests, and typetests all look solid. LGTM. 🚀

packages/react/src/usePayer.ts
```diff
@@ -0,0 +1,44 @@
...
+ * const payer = usePayer(client);
+ * return <span>{payer ? `Paying with ${payer.address}` : 'No payer'}</span>;
+ * ```
+ *
+ * @see {@link useIdentity}
+ */
+export function usePayer(client: ClientWithPayer & Partial<ClientWithSubscribeToPayer>): TransactionSigner | undefined {
+    const subscribe = useCallback(
+        (onStoreChange: () => void) =>
+            client.subscribeToPayer ? client.subscribeToPayer(onStoreChange) : NOOP_UNSUBSCRIBE,
```
lorisleiva, line 33: nit: Shouldn't `NOOP_UNSUBSCRIBE` be called `NOOP_SUBSCRIBE`? I'd be also happy to dissolve it for `() => {}` since it's pretty self-explanatory.

packages/react/src/useIdentity.ts
```diff
@@ -0,0 +1,47 @@
...
+ * ```
+ *
+ * @see {@link usePayer}
+ */
+export function useIdentity(
+    client: ClientWithIdentity & Partial<ClientWithSubscribeToIdentity>,
+): TransactionSigner | undefined {
+    const subscribe = useCallback(
+        (onStoreChange: () => void) =>
+            client.subscribeToIdentity ? client.subscribeToIdentity(onStoreChange) : NOOP_UNSUBSCRIBE,
```
lorisleiva, line 36: Same here

lorisleiva (approved):

mcintyre94:
### Merge activity

* **Aug 4, 4:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1876).
* **Aug 4, 4:16 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1876) couldn't merge this pull request because a downstack PR #1841 failed to merge.
* **Aug 4, 5:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1876).
* **Aug 4, 5:12 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1876) couldn't merge this pull request because a downstack PR #1825 failed to merge.
* **Aug 5, 7:47 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1876).
* **Aug 5, 8:16 AM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1876) couldn't merge this pull request because a downstack PR #1869 failed to merge.
* **Aug 5, 4:53 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1876).
* **Aug 5, 5:23 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1876) couldn't merge this PR because it had merge conflicts.
* **Aug 5, 6:23 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1876).
* **Aug 5, 6:24 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1876).

PR #1883, opened 2026-08-04
Update `withCleanup` to use a minimal polyfill of `DisposableStack`

mcintyre94:
#### Problem

Safari does not ship the explicit resource management APIs, such as `DisposableStack`: https://caniuse.com/mdn-javascript_builtins_disposablestack 

We rely on this for our `withCleanup` implementation, which enables a `[Symbol.Dispose]` function to be attached to a client, allowing plugins to clean up resources when the client is disposed. The wallet plugin is one user of this functionality.

Because of Safari not supporting it, it would crash on an app that used a client including a plugin that uses `withCleanup`, such as the wallet plugin

#### Summary of Changes

We now handle a missing `DisposableStack` in the runtime, by implementing a minimal version that implements only the `defer` and `dispose` functions that we use. If `DisposableStack` is available then we use it, following the same pattern as with our other polyfills (eg. in WebCrypto). If it is not available then we use our minimal polyfill instead.

We add extra tests to ensure the polyfill has all the behaviours we expect, and we now run the tests both with the global `DisposableStack` (it is available in our test environment) and with it removed to force use of the polyfill. 

With this change our example app works correctly on Safari

### DIFF
--- a/packages/plugin-core/src/client.ts
+++ b/packages/plugin-core/src/client.ts
@@ -292,6 +292,11 @@ function toConfigurableDescriptors<T extends PropertyDescriptorMap>(descriptors:
  * If the client already implements `Symbol.dispose`, the existing dispose
  * logic is chained so that it runs after the new `cleanup` function.
  *
+ * Cleanups run in reverse order of registration, disposal is idempotent, and if more than one
+ * cleanup throws then the errors are aggregated into a `SuppressedError` chain. Runtimes that have
+ * not shipped explicit resource management are supported too, though a `using` declaration needs a
+ * `Symbol.dispose` polyfill there.
+ *
  * The return type is an {@link ExtendedClient}, which flattens the merged shape into a
  * single object literal so chained calls do not accumulate nested intersections in editor
  * tooltips and error messages.
@@ -314,13 +319,28 @@ function toConfigurableDescriptors<T extends PropertyDescriptorMap>(descriptors:
  *     };
  * }
  *
- * // Later, when the client is no longer needed:
- * using client = createClient().use(myPlugin();
+ * // Build the client in the scope that should own it:
+ * using client = createClient().use(myPlugin());
  * // `socket.close()` is called automatically when `client` goes out of scope.
+ * ```
+ *
+ * @example Disposing without a using declaration
+ * `using` requires explicit resource management, which Safari has not shipped as of Safari 27.
+ * Either dispose the client yourself, as below, or polyfill `Symbol.dispose` — installing the
+ * polyfill before any client is created, since a client registers its dispose method under
+ * whatever `Symbol.dispose` was at the time.
+ * ```ts
+ * const client = createClient().use(myPlugin());
+ *
+ * // Later, when the client is no longer needed:
+ * client[Symbol.dispose]();
+ * // `socket.close()` has now been called.
  * ```
  *
  * @see {@link extendClient}
  * @see {@link ExtendedClient}
+ * @remarks See https://caniuse.com/mdn-javascript_builtins_disposablestack for platform
+ * availability of `DisposableStack`.
  */
 export function withCleanup<TClient extends object>(
     client: TClient,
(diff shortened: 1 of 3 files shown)


mcintyre94:
@trevor-cortex

packages/plugin-core/src/client.ts
```diff
@@ -367,9 +385,80 @@ function addCleanupToClientWithoutExistingStack<TClient extends object>(
...
+            cleanups.length = 0;
+            if (hasError) {
+                throw error;
+            }
+        },
+    };
+}
+
+function createSuppressedError(error: unknown, suppressed: unknown): Error {
+    if (typeof SuppressedError === 'function') {
```
trevor-cortex: Tiny consistency nit: `createCleanupStack` detects the platform class via `globalThis.DisposableStack`, but this check uses a bare `SuppressedError` reference. Both work (`typeof` on an undeclared identifier doesn't throw), but using `globalThis.SuppressedError` here would make the two detection sites symmetrical — and it's also the exact global the test suite deletes, which makes the coupling explicit.

```suggestion
    if (typeof globalThis.SuppressedError === 'function') {
        return new globalThis.SuppressedError(error, suppressed);
```

trevor-cortex (commented):
## Summary

*(LGTM — submitting as a comment since I don't have approve permissions on this repo.)*

This PR fixes a Safari crash in `withCleanup` by introducing a minimal internal fallback for `DisposableStack`. The runtime's own `DisposableStack` is still preferred when present (detected per call via `typeof globalThis.DisposableStack === 'function'`), and the fallback implements only the narrow `CleanupStack` slice (`defer` + `dispose`) that `withCleanup` actually uses. A `createSuppressedError` helper likewise falls back to a shape-compatible plain `Error` when `SuppressedError` is missing. Tests now run the whole `withCleanup` suite twice via `describe.each` — once with the platform class, once with it deleted — plus dedicated suites for the no-`SuppressedError` path and platform detection.

I verified the fallback against native `DisposableStack` semantics and it matches on all the behaviours that matter here:

- LIFO cleanup order, including chaining of a pre-existing `Symbol.dispose`.
- Idempotent `dispose()` (second call is a no-op).
- `ReferenceError` from `defer()` after disposal.
- Error aggregation orientation: the loop builds `SuppressedError(e, error)` so the newer error suppresses the pending one, which matches the spec's `DisposeResources` behaviour — the test's `{ error: firstError, suppressed: secondError }` expectation is correct for LIFO execution.
- The fallback's `SuppressedError` message ("An error was suppressed during disposal") matches what native `DisposableStack` disposal produces.

The changeset is present with `patch` bumps and a clear description, consistent with a bug fix.

## Things to watch / notes for other reviewers

**1. `Symbol.dispose` is *also* undefined on Safari — the code self-consistently degrades, but there's an ordering hazard worth documenting.** On a runtime without explicit resource management, `Symbol.dispose` evaluates to `undefined`, so the computed key in `addCleanupToClientWithoutExistingStack`'s `additions` object (and the `Symbol.dispose in client` check, and the caller's `client[Symbol.dispose]()`) all coerce to the *string* key `"undefined"`. Because registration and lookup coerce identically, explicit disposal works — which is what the changeset promises. But two consequences:

- If a user installs a `Symbol.dispose` polyfill *after* a client has been created, `client[Symbol.dispose]` now resolves to the real symbol and returns `undefined`, stranding the cleanup under the `"undefined"` key. The docs suggest polyfilling `Symbol.dispose` for `using` support; it may be worth a sentence noting the polyfill must be loaded before clients are created.
- Clients on Safari will carry an own property literally named `"undefined"`, and `Symbol.dispose in client` will match any unrelated `"undefined"` key. Both are edge cases, but worth being aware of.

**2. The "Safari simulation" test suite can't fully simulate Safari.** Well-known symbols like `Symbol.dispose` are non-configurable, so the tests necessarily run with the real symbol present even when `DisposableStack`/`SuppressedError` are deleted. That means the `"undefined"`-key degradation path from note 1 is untested and untestable in Jest — the suites cover everything that *can* be covered, which is fine, just flagging the gap so nobody assumes the Safari environment is exercised end to end.

**3. Minor known divergence (fine to ship):** the fallback's `defer` accepts non-callables (native throws `TypeError` immediately), and the fallback `SuppressedError`'s `error`/`suppressed` properties are enumerable where the native ones are not. TypeScript's types guard the former, and nothing in Kit depends on the latter — mentioning only for completeness.

Nice work on the test rigour, particularly running the shared suite against both stacks so they can't drift.

mcintyre94:
> Symbol.dispose is also undefined on Safari — the code self-consistently degrades, but there's an ordering hazard worth documenting.

Added a note: polyfill (if you choose to) before creating a client

> The "Safari simulation" test suite can't fully simulate Safari. 

Claude came up with a hacky test that defines a version of `Symbol` that's missing `dispose`, which is added 

> the fallback's defer accepts non-callables

Fixed

lorisleiva (approved):
Thank you!

PR #1884, opened 2026-08-04
Fix array decoder empty check

### LINKED ISSUE #1877: "getArrayDecoder" emptiness check copies the whole remaining buffer, making large-account decoding quadratic
## Overview

While profiling a web application that froze for ~8 seconds decoding one large (2.58 MB) account with an SDK built on `@solana/kit`, a CPU profile pointed at `getArrayDecoder`'s `read()`. It tests "is the buffer empty?" by copying every remaining byte:

```js
// packages/codecs-data-structures/src/array.ts (line 181 on current main)
read: (bytes, offset) => {
    const array = [];
    if (typeof size === 'object' && bytes.slice(offset).length === 0) {   // <-- here
        return [array, offset];
    }
    ...
```

`bytes.slice(offset)` allocates and copies from `offset` to the end of the buffer purely to read `.length` off the result and compare it to `0`.

Two things make this expensive rather than merely wasteful:

1. **It runs on every array-like read with a size prefix — the default (`u32`).** Fixed-size and `'remainder'` arrays skip it, but account layouts overwhelmingly use the prefixed form. `getMapDecoder` and `getSetDecoder` are both thin wrappers over `getArrayDecoder`, so they inherit it.
2. **Each copy is proportional to the bytes remaining.** For an account containing K nested arrays/maps/sets, total copy volume is `O(K x bytes)` — quadratic in account size.

Still present on `main` (checked 2026-08-02) and in the current release (`7.0.0`).

## Steps to reproduce

Self-contained: only `@solana/kit` and the two `@solana/codecs-*` packages it re-exports, synthetic data, no RPC. The script decodes a struct of K `u64` arrays — payload grows linearly with K, so decode time should too — and A/Bs against a patched copy of the installed package in the same run.

1. Save the two files below into an empty directory.
2. `npm install`
3. `node repro.mjs`

Output (Node 20.20.2, Linux):

```
@solana/kit                 : 7.0.0
@solana/codecs-data-structures: 7.0.0
patched dist files          : 3

  arrays   payload      ORIGINAL ms  (growth)     PATCHED ms  (growth)
  ---------------------------------------------------------------------
    1000      20 KiB           5.9        -           2.8        -
    2000      39 KiB          13.7     2.3x           5.2     1.8x
    4000      78 KiB          40.4     3.0x           9.9     1.9x
    8000     156 KiB         120.9     3.0x          19.3     2.0x
   16000     313 KiB         413.0     3.4x          53.3     2.8x
   32000     625 KiB        2779.8     6.7x          96.3     1.8x

  end-to-end: payload x32  ->  ORIGINAL x473 time,  PATCHED x34 time
  (linear would be x32; quadratic would be x1024)
  at the largest size the fix is 28.9x faster
```

The patched column is the tell: **x32 payload produces ~x34 time — essentially linear.** All of the excess in the original column is the emptiness check.

<details>
<summary><code>package.json</code></summary>

```json
{
  "name": "kit-array-decoder-repro",
  "private": true,
  "type": "module",
  "dependencies": {
    "@solana/kit": "7.0.0",
    "@solana/codecs-data-structures": "7.0.0",
    "@solana/codecs-numbers": "7.0.0"
  }
}
```

</details>

<details>
<summary><code>repro.mjs</code></summary>

```js
/**
 * Self-contained reproduction for the getArrayDecoder O(n) emptiness check.
 *
 * Public-safe: uses only @solana/kit and synthetic data. No third-party account,
 * RPC endpoint, or application code required.
 *
 *   getArrayDecoder's read() tests "is the buffer empty?" with
 *       bytes.slice(offset).length === 0
 *   which allocates and copies every byte from `offset` to the end, just to read .length.
 *
 * Decoding a struct of K prefixed arrays does K array reads over a payload that is itself
 * O(K) bytes, so the copy volume is O(K x bytes) = O(K^2). Doubling K should double decode
 * time; instead it roughly quadruples.
 *
 * The script also A/Bs against a patched copy of the installed package, so the same run
 * shows the superlinearity disappearing.
 */
import { readFileSync, writeFileSync, cpSync, existsSync } from 'node:fs';

// Run from anywhere: resolve everything relative to this file.
process.chdir(import.meta.dirname);

const PKG = 'node_modules/@solana/codecs-data-structures';
const PATCHED = 'node_modules/codecs-patched';

/** Copy the installed package and apply the one-line fix to every dist bundle. */
function buildPatchedCopy() {
  cpSync(PKG, PATCHED, { recursive: true, force: true });
  const pj = JSON.parse(readFileSync(`${PATCHED}/package.json`, 'utf8'));
  pj.name = 'codecs-patched';
  writeFileSync(`${PATCHED}/package.json`, JSON.stringify(pj, null, 2));

  let patchedFiles = 0;
  for (const f of ['dist/index.node.mjs', 'dist/index.node.cjs', 'dist/index.browser.mjs']) {
    const path = `${PATCHED}/${f}`;
    if (!existsSync(path)) continue;
    const before = readFileSync(path, 'utf8');
    const after = before.replace(
      /(\w+)\.slice\((\w+)\)\.length === 0/g,
      (_m, bytes, off) => `${off} >= ${bytes}.length`
    );
    if (after !== before) { writeFileSync(path, after); patchedFiles++; }
  }
  if (patchedFiles === 0) throw new Error('bug pattern not found — already fixed upstream?');
  return patchedFiles;
}

console.log(`@solana/kit                 : ${JSON.parse(readFileSync('node_modules/@solana/kit/package.json', 'utf8')).version}`);
console.log(`@solana/codecs-data-structures: ${JSON.parse(readFileSync(`${PKG}/package.json`, 'utf8')).version}`);
console.log(`patched dist files          : ${buildPatchedCopy()}\n`);

const orig = await import('@solana/codecs-data-structures');
const fixed = await import('codecs-patched');
const { getU64Encoder, getU64Decoder } = await import('@solana/codecs-numbers');
const { getStructEncoder, getStructDecoder } = await import('@solana/codecs-data-structures');

const time = (fn) => { const t = performance.now(); fn(); return performance.now() - t; };
const best = (fn) => Math.min(...[0, 1, 2].map(() => time(fn)));

/** Struct of `fields` u64 arrays, 2 values each. Payload grows linearly with `fields`. */
const build = (mod, fields) => {
  const names = Array.from({ length: fields }, (_, i) => `f${i}`);
  const encoder = getStructEncoder(names.map((n) => [n, orig.getArrayEncoder(getU64Encoder())]));
  const decoder = getStructDecoder(names.map((n) => [n, mod.getArrayDecoder(getU64Decoder())]));
  const value = Object.fromEntries(names.map((n) => [n, [1n, 2n]]));
  return { bytes: encoder.encode(value), decoder };
};

console.log('  arrays   payload      ORIGINAL ms  (growth)     PATCHED ms  (growth)');
console.log('  ---------------------------------------------------------------------');

let prevO = null, prevP = null, firstO = null, firstP = null, firstBytes = null, lastBytes = null;
for (const fields of [1000, 2000, 4000, 8000, 16000, 32000]) {
  const o = build(orig, fields);
  const p = build(fixed, fields);
  o.decoder.decode(o.bytes); p.decoder.decode(p.bytes);        // warm

  const msO = best(() => o.decoder.decode(o.bytes));
  const msP = best(() => p.decoder.decode(p.bytes));
  const gO = prevO ? `${(msO / prevO).toFixed(1)}x` : '  -';
  const gP = prevP ? `${(msP / prevP).toFixed(1)}x` : '  -';

  console.log(`  ${String(fields).padStart(6)}   ${String(Math.round(o.bytes.length / 1024)).padStart(5)} KiB   ` +
              `${msO.toFixed(1).padStart(11)}   ${gO.padStart(6)}   ` +
              `${msP.toFixed(1).padStart(11)}   ${gP.padStart(6)}`);
  prevO = msO; prevP = msP;
  if (firstO === null) { firstO = msO; firstP = msP; firstBytes = o.bytes.length; }
  lastBytes = o.bytes.length;
}

// Per-step growth is noisy (GC timing). The end-to-end ratio is the robust statement.
const payloadX = lastBytes / firstBytes;
console.log(`
  end-to-end: payload x${payloadX.toFixed(0)}  ->  ORIGINAL x${(prevO / firstO).toFixed(0)} time` +
            `,  PATCHED x${(prevP / firstP).toFixed(0)} time`);
console.log(`  (linear would be x${payloadX.toFixed(0)}; quadratic would be x${(payloadX ** 2).toFixed(0)})`);
console.log(`  at the largest size the fix is ${(prevO / prevP).toFixed(1)}x faster`);

console.log(`
PATCHED tracks the payload almost exactly - that is the cost of the real decoding work.
ORIGINAL grows far faster, because total copy volume is O(arrays x bytes).
Per-step figures are noisy; the end-to-end ratio is the reliable one.

Fix (packages/codecs-data-structures/src/array.ts):
  - if (typeof size === 'object' && bytes.slice(offset).length === 0) {
  + if (typeof size === 'object' && offset >= bytes.length) {

For the non-negative offsets the codec framework guarantees, 'bytes.slice(offset).length === 0'
is true precisely when 'offset >= bytes.length', so this is semantics-preserving:
an O(n) allocate-and-copy becomes an O(1) compare.
getMapDecoder and getSetDecoder both delegate to getArrayDecoder and inherit the fix.`);
```

</details>

## Description of bug

I expected decode time to scale roughly linearly with account size. Instead it grows quadratically, because every prefixed array/map/set read copies all bytes remaining in the buffer just to test emptiness.

Real-world impact, measured on the shipped build of a client SDK decoding a single 2.58 MB account with deeply nested arrays/maps/sets:

- **Chromium 150: 8.4 s → 0.50 s (16.6x)** with the one-line fix below, byte-identical decoded output. The 8.4 s is one uninterrupted main-thread block — the browser offers to kill the tab.
- Node 20: 3.7–6.2 s → ~0.65 s (5–6x across runs). **Node understates it roughly 3x**, because the dominant cost is allocation, and GC behaviour differs between runtimes.
- Instrumented, that one decode performs **18,489 emptiness checks copying a combined 12.85 GiB** — 5,353x the size of the account, mean 729 KiB per check.
- `node --cpu-prof` puts **~65–70% of self time** in the array decoder's `read`, plus ~13% in the byte-slice helper it calls.

So this is better characterised as an **allocation** bug than a CPU bug: the wasted work is gigabytes of short-lived buffers, which also shows up as sustained GC pressure well beyond the decode itself.

### Suggested fix

```diff
- if (typeof size === 'object' && bytes.slice(offset).length === 0) {
+ if (typeof size === 'object' && offset >= bytes.length) {
```

Semantics-preserving: for the non-negative offsets the codec framework guarantees (`decode()` starts at 0, composite reads only ever advance the offset, and `offsetCodec` normalises and asserts its offsets into range before the inner codec sees them), `bytes.slice(offset).length === 0` is true precisely when `offset >= bytes.length`. It replaces an O(n) allocate-and-copy with an O(1) integer compare, and `getMapDecoder` / `getSetDecoder` inherit the fix for free.

Verified byte-identical output on a full 2.58 MB decoded account graph (8,347,212 characters serialised, normalising `BigInt` / `Map` / `Set`), in both Node and Chromium.

Happy to open a PR if useful.

Swift42:
#### Problem

"getArrayDecoder" emptiness check copies the whole remaining buffer, making large-account decoding quadratic

#### Summary of Changes

For the non-negative offsets the codec framework guarantees (decode() starts at 0, composite reads only ever advance the offset, and offsetCodec normalises and asserts its offsets into range before the inner codec sees them), bytes.slice(offset).length === 0 is true precisely when offset >= bytes.length. 

Fixes #1877

### DIFF
--- a/.changeset/quick-arrays-decode.md
+++ b/.changeset/quick-arrays-decode.md
@@ -1,0 +1,7 @@
+---
+'@solana/codecs-data-structures': patch
+---
+
+Avoid copying the remaining buffer in `getArrayDecoder`'s emptiness check
+
+`getArrayDecoder`'s `read()` tested for an empty byte array with `bytes.slice(offset).length === 0`, which allocates and copies every byte from `offset` to the end just to read `.length` off the result. On large accounts containing many prefixed arrays, maps, or sets this made decoding quadratic in account size. The check is now the equivalent O(1) comparison `offset >= bytes.length`. `getMapDecoder` and `getSetDecoder` delegate to `getArrayDecoder` and benefit as well.
--- a/packages/codecs-data-structures/src/array.ts
+++ b/packages/codecs-data-structures/src/array.ts
@@ -178,7 +178,7 @@ export function getArrayDecoder<TTo>(item: Decoder<TTo>, config: ArrayCodecConfi
         ...(fixedSize !== null ? { fixedSize } : { maxSize }),
         read: (bytes: ReadonlyUint8Array | Uint8Array, offset) => {
             const array: TTo[] = [];
-            if (typeof size === 'object' && bytes.slice(offset).length === 0) {
+            if (typeof size === 'object' && offset >= bytes.length) {
                 return [array, offset];
             }
 


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
LGTM ✅ (submitting as a comment — I don't have approve permissions on this repo)

Replaces `bytes.slice(offset).length === 0` with `offset >= bytes.length` in `getArrayDecoder`'s prefixed-size emptiness check, turning an O(remaining bytes) allocation+copy into an O(1) comparison. Since `getMapDecoder` and `getSetDecoder` delegate to `getArrayDecoder`, they pick up the fix for free. Fixes the quadratic decoding behaviour from #1877.

I verified the equivalence claim:

- For `offset` in `[0, bytes.length)`, `slice(offset)` is non-empty and `offset >= bytes.length` is false — same result.
- For `offset >= bytes.length`, `slice(offset)` is empty and the comparison is true — same result.
- The only divergence is a negative `offset` against an *empty* buffer (old code returned `[[], offset]`, new code falls through to the prefix read and throws), which the framework rules out: `decode()` starts at 0 and reads only ever advance the offset, as the PR description lays out. Arguably the new behaviour (throwing on an out-of-range offset) is more correct there anyway.

The check still only applies when `typeof size === 'object'` (prefixed sizes), so `'remainder'` and fixed-number sizes are untouched.

Changeset is present with a `patch` bump for `@solana/codecs-data-structures`, which matches the perf-only, behaviour-preserving nature of the change.

Notes for other reviewers: nothing needs extra attention — it's a one-line hot-path change with a clear equivalence argument. If you want extra confidence, the existing array/map/set decoder tests covering empty-bytes decoding (`decode(new Uint8Array([]))` returning `[]`) exercise the changed branch directly.

mcintyre94 (approved):
Awesome, thank you for this!

PR #1888, opened 2026-08-05
Add an `assertOffchainMessageV1Equal` helper

mcintyre94:
#### Problem

We now have `signOffchainMessage` as a Solana wallet-standard feature: https://github.com/anza-xyz/wallet-standard/blob/487553dd7db02b0fd10b26ad59408237087b909a/packages/core/features/src/signOffchainMessage.ts

This feature has the app send the input message with shape `{messageVersion, account, message, requiredSigners}`, with the wallet responsible for constructing an `OffchainMessage` and returning it serialized, along with a signature.

For an app, verifying the signature is valid for the returned message bytes is necessary but not sufficient - the wallet could have returned a different signed message to the one requested. The app also needs to verify that the returned message matches what it asked the wallet to sign.

Currently Kit does not have a function to verify this.

#### Summary of Changes

Add a new function `assertOffchainMessageV1Equal(receivedMessage: OffchainMessageV1, expectedMessage: OffchainMessageV1)`

This function throws if the content or requiredSignatories does not match the expected message. Note that the `account` in the wallet-standard feature input is not part of the `OffchainMessageV1` type, that's only part of the wallet feature input.

The intended use, as documented, is:

```ts
const expectedMessage: OffchainMessageV1 = {
    content: 'Sign in to Example App',
    requiredSignatories: [{ address: signerAddress }],
    version: 1,
}

// output comes from the wallet-standard feature, shaped { signedOffchainMessage: Uint8Array, signature: Uint8Array }

// decode the message
const receivedMessage = getOffchainMessageDecoder().decode(output.signedOffchainMessage);

// check the wallet signed the message we expected
if (receivedMessage.version !== 1) {
  throw new Error(`Expected the wallet to sign a version 1 message; got version ${receivedMessage.version}`);
}
assertOffchainMessageV1Equal(receivedMessage, expectedMessage);

// unchanged, verify the signatures
await verifyOffchainMessageEnvelope({
  content: output.signedOffchainMessage,
  signatures: { [signerAddress]: output.signature },
});
```

It's expected that most apps will use this through a wallet library, for example this verification would be performed by the wallet plugin, wallet-adapter, etc.

Note that requiring the app to check the version first is intentional, the app controls the `messageVersion` sent to the feature and should check it's the expected version. This approach makes it easy to add support if/when we add a v2.

See also #1725 which explored a similar approach


mcintyre94:
@trevor-cortex

packages/offchain-messages/src/message-v1.ts
```diff
@@ -8,3 +15,91 @@ export type OffchainMessageV1 = BaseOffchainMessageV1 &
     Readonly<{
         content: string;
     }>;
+
+function getSortedSignatoryAddresses(message: OffchainMessageV1): readonly Address[] {
+    return message.requiredSignatories.map(({ address }) => address).toSorted();
```
trevor-cortex, line 21: `Array.prototype.toSorted` is ES2023 (Node 20+, Chrome 110+, Safari 16.3+, Firefox 115+). The package's `engines` field covers Node, but the browserslist baseline (`supports bigint and not dead`) includes browsers that predate it, and I don't believe the build injects polyfills. If the repo hasn't already established an ES2023 floor elsewhere, plain `.sort()` is equivalent here since `.map(…)` already returns a fresh array:

```ts
return message.requiredSignatories.map(({ address }) => address).sort();
```

Happy to be told there's precedent for `toSorted` in the codebase already — in that case ignore this.

packages/offchain-messages/src/message-v1.ts
```diff
@@ -8,3 +15,91 @@ export type OffchainMessageV1 = BaseOffchainMessageV1 &
...
+ */
+export function assertOffchainMessageV1Equal(
+    receivedMessage: OffchainMessageV1,
+    expectedMessage: OffchainMessageV1,
+): void {
+    if (receivedMessage.content !== expectedMessage.content) {
+        throw new SolanaError(SOLANA_ERROR__OFFCHAIN_MESSAGE__CONTENT_DOES_NOT_MATCH_EXPECTED, {
+            actualLength: receivedMessage.content.length,
+            expectedLength: expectedMessage.content.length,
+        });
```
trevor-cortex, line 94: `String.prototype.length` counts UTF-16 code units, while other length-oriented errors in this domain (e.g. `…MAXIMUM_LENGTH_EXCEEDED`) report byte counts. For non-ASCII content the two diverge (an emoji counts as 2 here, 4 in UTF-8 bytes), which could momentarily confuse someone eyeballing the error. Not blocking — the lengths are diagnostic hints, not load-bearing — but consider either measuring with `getUtf8Encoder().getSizeFromValue(…)` for consistency, or noting "in UTF-16 code units" in the message template.

trevor-cortex (commented):
## Summary

Adds `assertOffchainMessageV1Equal(receivedMessage, expectedMessage)` to `@solana/offchain-messages`, which throws when a decoded v1 offchain message's `content` or `requiredSignatories` differ from the message the app asked an untrusted signer (e.g. a wallet) to sign. Two new error codes back it (`…CONTENT_DOES_NOT_MATCH_EXPECTED` at 5607018, `…REQUIRED_SIGNATORIES_DO_NOT_MATCH_EXPECTED` at 5607019), and the README gains a thorough guide covering single- and multi-signer verification flows.

This looks good to me overall — the design decisions are sound and well-documented:

- Content is compared first, so a doubly-mismatched message reports the content error — the higher-signal one. Documented and tested.
- Signatories are compared order-insensitively (decoded messages are in spec-mandated lexicographic order; the caller's may not be) but *not* as sets — duplicates are a mismatch. The docstring correctly notes a decoded message can never contain duplicates since the codec rejects them, so this only catches caller mistakes. Good call surfacing rather than hiding that.
- Message content is deliberately excluded from the error context to avoid leaking sensitive content to logs/error reporters — lengths only. Nice touch.
- Scoping to `OffchainMessageV1` (rather than accepting the union and throwing on other versions) keeps version policy in the caller's hands, and the typetest pins the requirement to narrow first.

The error-code plumbing (codes/context/messages) is inserted in the right alphabetical/numeric positions, and the changeset covers both published packages at `minor`, which fits a new feature.

## Things to watch

- **Canonical-encoding assumption in the multi-signer README example.** The example verifies each wallet's signature against locally compiled bytes (`compileOffchainMessageV1Envelope(expectedMessage).content`) rather than each wallet's own `signedOffchainMessage` bytes. This is only sound if v1 encoding is canonical — i.e. two byte strings that decode to equal messages are byte-identical. That holds today (fixed signing domain, mandated signatory ordering enforced by the decoder, deterministic length prefixes), and if it ever didn't, verification would fail closed rather than open, so I think this is fine — but it's worth a second reviewer confirming the v1 decoder rejects all non-canonical encodings.
- Two small inline comments: `toSorted()` runtime availability vs the package's browser support baseline, and the units reported by the content-length error context.

- The naming (`assert…Equal` rather than `assertIs…`) diverges from the package's type-assertion helpers, but that's appropriate — this asserts a relation between two values and performs no type narrowing.
- The README and changeset are unusually complete; both accurately reflect the implemented semantics (I cross-checked each documented behaviour against the tests).

packages/offchain-messages/src/message-v1.ts
mcintyre94: We have precedent for this in this package

packages/offchain-messages/src/message-v1.ts
mcintyre94: Agreed, v1 messages are utf-8, so changed this to use `utf8Encoder.getSizeFromValue`. The error code `SOLANA_ERROR__OFFCHAIN_MESSAGE__CONTENT_DOES_NOT_MATCH_EXPECTED` is version agnostic, so changed it to `actual|expectedBytes` to make how we define length across versions clearer.

lorisleiva (approved):
Thank you!

PR #1890, opened 2026-08-06
CI: skip preview docs on non-docs PRs, add retry

mcintyre94:
#### Problem

We have fairly regular failures on the deploy docs CI job, where the build is exhausting the runner's memory. For now this is intermittent and resolved by retrying.

#### Summary of Changes

- Only run the deploy job if a PR actually changes docs/*. This means that we can no longer preview the auto-generated pages for new/changed APIs. But should eliminate the most common CI issue on most PRs. Note that I've removed this job from the ruleset, ie PRs no longer require it to pass. Note that it also runs if we change the CI job (to verify that change), and therefore does run on this PR. 

- Adds automatic `retry` in the docs jobs, so that if they fail we try again twice before failing the job. They'll obviously take longer if they fail and then pass, but quicker and less annoying than re-running them ourselves. In most cases I think trying again will succeed, but it'll depend how well memory is cleaned up after the error. If we see this not helping we can drop it and manually retry, but worth trying I think.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -53,8 +53,30 @@ jobs:
           VERCEL_PROJECT_ID: ${{ secrets.VERCEL_DOCS_PROJECT_ID }}
           VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
         run: |
-          vercel pull --token="$VERCEL_TOKEN" --yes --environment=production
-          vercel build --token="$VERCEL_TOKEN" --prod
+          # `vercel build` fails intermittently — it can be killed when `next build` exhausts the
+          # runner's memory
+          retry() {
+            # Label attempts with the command and subcommand (eg. `vercel build`) but not the
+            # remaining arguments, which include the Vercel token
+            local label="$1 $2"
+            local attempt
+            for attempt in 1 2 3; do
+              if "$@"; then
+                return 0
+              fi
+              if [ "$attempt" -lt 3 ]; then
+                echo "::warning::'$label' failed (attempt $attempt of 3); retrying in $((attempt * 15))s"
+                sleep "$((attempt * 15))"
+              fi
+            done
+            echo "::error::'$label' failed after 3 attempts"
+            return 1
+          }
+          retry vercel pull --token="$VERCEL_TOKEN" --yes --environment=production
+          retry vercel build --token="$VERCEL_TOKEN" --prod
+          # Deliberately not retried: this uploads the prebuilt output, so it's not subject to the
+          # memory issue above, and a failure *after* the deployment was created would result in a
+          # second deployment on retry
           vercel deploy --token="$VERCEL_TOKEN" --archive=tgz --prebuilt --prod
 
       - name: Synchronize InKeep Search Index
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -2,7 +2,9 @@ name: Preview Documentation
 
 on:
   pull_request:
-  merge_group:
+    paths:
+      - 'docs/**'
+      - '.github/workflows/preview-docs.yml'
 
 permissions:
   contents: read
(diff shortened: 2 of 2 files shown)


mcintyre94:
@trevor-cortex

.github/workflows/preview-docs.yml
```diff
@@ -2,7 +2,9 @@ name: Preview Documentation
 
 on:
   pull_request:
-  merge_group:
```
trevor-cortex, line 5: With `merge_group:` removed, this workflow will never run (or report a check) for merge queue events. You mentioned removing the job from the ruleset — worth double-checking the merge queue's required-checks configuration was updated as well, since a required check that never reports will leave queued PRs hanging indefinitely.

.github/workflows/preview-docs.yml
```diff
@@ -52,17 +54,37 @@ jobs:
...
+          # These steps fail intermittently — `vercel build` can be killed when `next build`
+          # exhausts the runner's memory
+          retry() {
+            local attempt
+            for attempt in 1 2 3; do
+              if "$@"; then
+                return 0
+              fi
+              if [ "$attempt" -lt 3 ]; then
+                echo "::warning::'$1' failed (attempt $attempt of 3); retrying in $((attempt * 15))s"
```
trevor-cortex: Nit: `$1` is just the program name, so all three invocations here log `'vercel' failed` — you can't tell from the annotation whether it was `pull`, `build`, or `deploy` that flaked. Consider `'$1 $2'` or `$*` (same for the `::error::` line below, and the copy in `deploy-docs.yml`). Non-blocking.

.github/workflows/deploy-docs.yml
```diff
@@ -53,9 +53,25 @@ jobs:
...
+                echo "::warning::'$1' failed (attempt $attempt of 3); retrying in $((attempt * 15))s"
+                sleep "$((attempt * 15))"
+              fi
+            done
+            echo "::error::'$1' failed after 3 attempts"
+            return 1
+          }
+          retry vercel pull --token="$VERCEL_TOKEN" --yes --environment=production
+          retry vercel build --token="$VERCEL_TOKEN" --prod
+          retry vercel deploy --token="$VERCEL_TOKEN" --archive=tgz --prebuilt --prod
```
trevor-cortex: FYI / non-blocking: if `vercel deploy` fails *after* the deployment was actually created (e.g. a network error while reporting back), the retry will create a second production deployment. With Vercel that's benign — the latest successful deploy wins — but if you'd rather be conservative, you could scope `retry` to just `pull`/`build` here, since `deploy` is the least likely step to hit the memory issue (it uploads prebuilt output).

trevor-cortex (commented):
This PR reduces docs CI flakiness in two ways: (1) the preview-docs workflow now only runs when `docs/**` (or the workflow itself) changes, and (2) both docs workflows wrap the Vercel CLI steps in a `retry` helper that attempts each command up to 3 times with a growing backoff.

The bash is careful and correct: `"$@"` inside the `if` condition properly suppresses errexit during attempts while still failing the step after exhaustion, and `deploy_preview` being a function invoked in the current shell means `DEPLOY_OUTPUT` survives for the URL extraction. Quoting `"$GITHUB_OUTPUT"` is a nice drive-by fix.

Things to watch / verify:

- **`merge_group` trigger removal** — with the trigger gone, this workflow will never report a check on merge queue runs. You mentioned removing the job from the ruleset; just make sure any merge queue required-checks config was updated too, otherwise queued PRs will stall waiting for a check that never runs (inline comment).
- **Trade-off acknowledged in the description**: PRs that change packages (and thus the auto-generated API docs) will no longer get previews. Seems like the right call given the flake rate, but reviewers of API-heavy PRs lose that verification.
- **Duplicated `retry` helper** — the function is copy-pasted between the two workflows. Fine at two copies for 15 lines of bash; if a third consumer shows up, consider extracting it to a composite action.
- Minor: the retry log messages only print `$1`, so all three Vercel steps report as `'vercel' failed` (inline comment).

Nothing blocking — LGTM. (Submitting as a comment since I don't have approve permissions on this repo.)

.github/workflows/preview-docs.yml
mcintyre94: Removing `merge_group` is intentional, I don't think it works properly with the file filter, and we don't need to re-run this job in the merge queue. AFAICT in settings the merge queue doesn't have its own required status checks, it uses the same ones as the PR. If I'm wrong and merge PRs hang, I'll dig more through settings and find what I missed.

.github/workflows/preview-docs.yml
mcintyre94: Agree, replaced with `$1 $2` to avoid writing all the args to logs

.github/workflows/deploy-docs.yml
mcintyre94: Agree, removed the retry from deploy

lorisleiva (approved):
Thanks! Let's hope that makes our CI more enjoyable to work with. ☺️

PR #1893, opened 2026-08-06
Allow transactions without a fee payer signature in transaction plan results

mcintyre94:
#### Problem

Currently the success branch of `TransactionPlanResult` is typed such that `result.context.signature` is defined, ie the successful transaction is guaranteed to have a signature.

We would like to be able to define transaction plan executors that do not have this behaviour, for example one that partially signs the transaction(s) and may not necessarily be the fee payer, ie may not provide a fee payer signature.

In addition to the type, the current implementation of `createTransactionPlanExecutor` also stops this, as it expects a function that returns a `Signature | Transaction` , and in the case of a `Transaction` it reads its signature and throws if it is not present.

#### Summary of Changes

This PR widens the type of `TransactionPlanResult` such that a successful leaf has an optional signature. The existing default type means that existing code which does not specify the `TSingle` type param is unchanged, and continues to receive the tighter signature-required type.

This widening applies to all the helper functions, such that apps receiving a `TransactionPlanResultWithOptionalSignature` can do everything that doesn't require a signature (like flattening it, asserting it's successful, etc).

`createTransactionPlanExecutor` gets a new optional `allowMissingFeePayerSignature` config field. If this is set to true, and the `executeTransactionMessage` returns a `Transaction`, then a new function `successfulSingleTransactionPlanResultWithOptionalSignature` is used instead of `successfulSingleTransactionPlanResultFromTransaction`. This function does not throw if the fee payer signature is missing, it just doesn't add it to the context. The type of the result when `allowMissingFeePayerSignature: true` is used, is `TransactionPlanResultWithOptionalSignature`.

A sort-of related pre-existing bug fix is fixed as part of this: `traverseSingle` has a catch block which calls `getSignatureFromTransaction`.  So if an erroring executor wrote a transaction to context which was missing a fee payer signature, and then threw some error, we'd lose its actual error trying to read the signature. This now uses the same defensive/best-effort signature in the error context.

The result of all of this is:

- Existing code is unchanged, and an existing call to `createTransactionPlanExecutor` continues to return a result that guarantees the presence of `result.context.signature` at its leaves
- We can define an executor that returns a result where there may not be a signature

This will allow a future PR to define a new plugin interface for `client.signTransaction[s]`. The return type will be `TransactionPlanResultWithOptionalSignature` so that the interface can support partial signing by a signer that is not the fee payer.


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1893?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1893** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1893?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1893?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1894** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1894?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -215,17 +270,36 @@ async function traverseSingle<TContext extends TransactionPlanResultContext>(
...
+ */
+function withDerivedSignature<TContext extends TransactionPlanResultContext>(
+    context: BaseTransactionPlanResultContext & TContext,
+): BaseTransactionPlanResultContext & TContext {
+    if (context.signature != null || context.transaction == null) {
+        return context;
+    }
+    const signature = getFeePayerSignatureIfPresent(context.transaction);
+    return signature == null ? context : { ...context, signature };
+}
```
trevor-cortex, line 356: This fixes the main case from the PR description, but there's a residual variant of the same bug: `getFeePayerSignatureIfPresent` only swallows `SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING` and re-throws anything else. Since `withDerivedSignature` is called from inside `traverseSingle`'s `catch` block, any other error thrown while reading the signature (e.g. a malformed `context.transaction` written by a buggy executor) will still escape and replace the original execution error — the exact failure mode this PR set out to fix, just for a narrower class of errors.

Given the doc here describes this as best-effort, consider making the derivation fully non-throwing in this path, e.g.:

```ts
let signature: Signature | undefined;
try {
    signature = getFeePayerSignatureIfPresent(context.transaction);
} catch {
    return context;
}
```

If the re-throw behaviour is intentional (surfacing a corrupted transaction loudly rather than silently dropping it), a short note in the doc comment explaining that trade-off would help the next reader. Non-blocking either way.

packages/instruction-plans/src/transaction-signature.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Returns the fee payer signature of a transaction, or `undefined` when the fee payer has not
+ * signed it yet.
+ *
+ * This is a non-throwing counterpart to `getSignatureFromTransaction`. Any error other than
+ * a missing fee payer signature is re-thrown unchanged.
+ *
+ * @param transaction - The transaction to read the fee payer signature from.
+ * @return The transaction signature, or `undefined` if the fee payer slot is empty.
+ */
+export function getFeePayerSignatureIfPresent(transaction: Transaction): Signature | undefined {
```
trevor-cortex: Minor observation: this helper isn't exported from the package index (the test imports it from the source file directly), so it's internal-only for now. That seems intentional, but it reads like a general-purpose companion to `getSignatureFromTransaction` — if the upcoming `client.signTransaction[s]` work or downstream consumers end up wanting it, it may fit better in `@solana/transactions` itself. No action needed for this PR.

trevor-cortex (commented):
**Verdict: LGTM** (submitting as a comment since I don't have approve permissions on this repo).

This PR widens `TransactionPlanResult` so that successful leaves may omit `context.signature`, enabling executors that partially sign transactions without providing a fee payer signature. It introduces `*WithOptionalSignature` counterparts for the successful single result, the single result union, and the result tree; rebinds `TSingle` across all tree types, guards, and helpers to the looser bound (keeping the strict type as the default); adds an `allowMissingFeePayerSignature` config flag to `createTransactionPlanExecutor` that conditionally types the executor's return; and fixes a real bug where the `traverseSingle` catch path could replace the original execution error with `SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING`.

The design is careful and I verified the key invariants:

- **Backward compatibility**: every widened generic keeps `SingleTransactionPlanResult` as the default for `TSingle`, so existing annotations resolve to the strict types unchanged. The strict overloads on `parallelTransactionPlanResult`, `sequentialTransactionPlanResult`, `nonDivisibleSequentialTransactionPlanResult`, and `passthroughFailedTransactionPlanExecution` are listed first, so strict inputs keep resolving to strict outputs. The typetests pin all of this down, including the assignability direction (strict → loose but not loose → strict, since an optional `signature` isn't assignable to a required one).
- **Fail-closed conditional**: `[TAllowMissingFeePayerSignature] extends [false]` means an omitted or literal-`false` flag yields the strict executor, while a non-literal `boolean` yields the loose one — the safe direction. Nice that there's a typetest asserting this explicitly.
- **Runtime semantics**: the loose constructor omits the `signature` key entirely (rather than setting `undefined`) so `'signature' in context` stays meaningful, and it still populates the signature when the fee payer slot happens to be filled.

- `packages/instruction-plans/src/transaction-signature.ts` is a new internal module — `getFeePayerSignatureIfPresent` is not re-exported from the package index (its test imports it from the source file directly). That seems deliberate, but it's worth a quick thought on whether this helper belongs in `@solana/transactions` next to `getSignatureFromTransaction`, where it would be generally useful beyond instruction plans. Fine to defer.
- The changeset is a `minor` bump for a purely additive feature plus a bug fix, which matches the package's precedent.
- The one remaining sharp edge is in the error-path signature derivation — see the inline comment on `withDerivedSignature`. It's an exotic case and non-blocking, but worth a decision before merge.

packages/instruction-plans/src/transaction-plan-executor.ts
mcintyre94: Agree, updated to use the catch-all. Better to surface the original error in all cases, and context will include the malformed transaction.

packages/instruction-plans/src/transaction-signature.ts
mcintyre94: Agree, moved this to `@solana/transactions` in a (stacked before this one) new PR: #1894. This is useful to apps, and implementing it in `transactions` reduces repetition.

mcintyre94:
I've added a second commit, as a fixup to make it reversible. This adjusts the type of `createTransactionPlanExecutor` so that when the `allowMissingFeePayerSignature` is true, the return from `executeTransactionMessage` must be `Transaction`. Without this flag it can still be `Signature | Transaction`.

I think this narrowing makes sense, it just means that an executor that sometimes returns transactions with missing signatures can no longer sometimes return a plain signature. It must always return a transaction (with or without fee payer signature).

This then allows the type to constrain the context on such executors to guarantee that `context.transaction` is defined. In the same way that we guarantee `context.signature` when `allowMissingFeePayerSignature` is false (default).

This addresses a review comment in the upstack PR, that we wouldn't be able to use `createTransactionPlanExecutor` to return an executor that guarantees the presence of a transaction. See https://github.com/anza-xyz/kit/pull/1899#discussion_r3735940331 

I've also added a comment that suggests a future breaking change should remove the required signature from `SuccessfulTransactionPlanResult` and not make any guarantees about context on that type. This would allow us to deprecate and later delete the `WithOptionalSignature` types added here. This would be a breaking change for uses of `TransactionPlanResult`. But importantly, for results coming from an executor defined by `createTransactionPlanExecutor`, they'd still have `result.context.signature` since it types the context.

I think this would be the ideal end state - `TransactionPlanResult` has no guaranteed context, our executors that return a signature/transaction are typed to return the correct context, and other uses of `TransactionPlanResult` are responsible for typing the context as they require it. But I think this PR is the best stopgap until we want to make that breaking change.

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
**Verdict: LGTM** (submitting as a comment since I don't have approve permissions on this repo).

Re-reviewed the full diff after the latest changes. Both points from my previous review are resolved, and the new material since then holds up well.

**Resolved from last pass:**

- **Error-path signature derivation is now fully best-effort.** `withDerivedSignature` wraps the derivation in a `try/catch` that returns the untouched context on *any* throw, so a malformed `context.transaction` written by a buggy executor can no longer displace the execution error being reported. The new "does not mask the failure cause when the transaction is malformed" test pins this down with a transaction whose `signatures` is `undefined` — exactly the shape that would have thrown before.
- **Helper placement.** The internal `transaction-signature.ts` module is gone; the executor now imports `getSignatureFromTransactionIfPresent` from `@solana/transactions`, where it sits next to `getSignatureFromTransaction` with cross-referencing docs. That's the home I'd hoped for — presumably promoted in an earlier PR of the stack.

**New since last pass, verified:**

- **The `context.transaction` guarantee on `allowMissingFeePayerSignature: true`.** `ExecutorSingleResult`'s `true` branch now types successful leaves as `SuccessfulSingleTransactionPlanResultWithOptionalSignature<TContext & { transaction: Transaction }>`. I checked that this is structurally sound rather than asserted: setting the flag narrows `executeTransactionMessage` to return a `Transaction`, so `traverseSingle` can only reach the branch that passes the transaction into `successfulSingleTransactionPlanResultWithOptionalSignature`, which unconditionally stores it. Failed/canceled variants correctly keep plain `TContext`, since they can be produced before the callback returns. The `typeof result === 'string'` branch is unreachable under the narrowed callback type absent an unsound cast by the caller, which is the right trade.
- **`passthroughFailedTransactionPlanExecution`'s generic overload.** Loose results skip the monomorphic strict overloads (an optional signature isn't assignable to a required one) and land on the generic overload, which preserves `TContext` and the transaction guarantee while adding only the failed/canceled variants the failure path can introduce. The overload ordering comment plus the TRIPWIRE typetest — which spells out that the fix for a future structural collapse is deleting the monomorphic overloads, not relaxing the assertion — is exactly the kind of guard rail overload stacks need. Same for the typetest pinning the pre-existing `TContext`-discarding behaviour of the strict overloads.
- **Typetest coverage.** The "call-site shapes found in the wild" block (the `limitFunction`-wrapped callback with `satisfies TransactionPlanExecutorConfig`, matching `rpcTransactionPlanExecutor` in kit-plugins) is a smart regression guard against inference breakage in downstream consumers. The fail-closed non-literal-`boolean` cases now also assert that *neither* guarantee is established, not just that the strict one isn't.
- **Docs + changeset.** The new "Producing transactions without a fee payer signature" section explains the guarantee trade-off clearly, and the expanded changeset is accurate to the implementation (including the fail-closed behaviour and the passthrough overload). Still `minor`, which matches precedent for additive feature + bug fix.

**Minor, non-blocking notes:**

- In `docs/content/docs/advanced-guides/instruction-plans.mdx` around line 770, the `estimateAndSetResourceLimitsFactory` import reformat is unrelated to this PR — presumably a formatter artifact. Harmless, just noting it for anyone diffing the docs.
- The remaining `as` casts in `createTransactionPlanExecutor` (the `executeTransactionMessage` narrowing and the final result cast) are both necessary while `TAllowMissingFeePayerSignature` is unresolved, and both carry comments explaining why. No concerns.

Nothing further from me — nice work closing the loop on the guarantees.

mcintyre94:
Closing - superseded by #1910

PR #1894, opened 2026-08-07
Add getSignatureFromTransactionIfPresent

mcintyre94:
#### Summary of Changes

This PR adds a new function `getSignatureFromTransactionIfPresent`, a non-throwing variant of `getSignatureFromTransaction`

If the transaction has a fee-payer signature then it is returned, otherwise `undefined`

This allows apps to avoid dealing with the details of a transaction signature, while handling the case where the transaction may not be signed yet.

This is primarily added for use in #1893 , but exported from `@solana/transactions` because it's useful functionality for apps.

### DIFF
--- a/.changeset/common-cats-roll.md
+++ b/.changeset/common-cats-roll.md
@@ -1,0 +1,5 @@
+---
+'@solana/transactions': minor
+---
+
+Added `getSignatureFromTransactionIfPresent`, which returns the `Signature` that uniquely identifies a transaction, or `undefined` when its fee payer has not signed it yet. Use it in preference to `getSignatureFromTransaction` when a transaction is legitimately allowed to be unsigned by its fee payer — for instance one that has been partially signed by an authority and is destined for a relayer that will pay for it. `getSignatureFromTransaction` is now implemented in terms of it and continues to throw `SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING` in that case.
--- a/packages/transactions/README.md
+++ b/packages/transactions/README.md
@@ -65,6 +65,19 @@ const signature = getSignatureFromTransaction(tx);
 console.debug(`Inspect this transaction at https://explorer.solana.com/tx/${signature}`);
 ```
 
+#### `getSignatureFromTransactionIfPresent()`
+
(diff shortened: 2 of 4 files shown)


mcintyre94:
* **#1893** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1893?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1894** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

This PR adds `getSignatureFromTransactionIfPresent`, a non-throwing variant of `getSignatureFromTransaction` that returns `undefined` when the fee payer hasn't signed the transaction. The existing throwing function is refactored to delegate to the new one, so there's a single source of truth for the "first signature is the fee payer" extraction logic.

The change is clean and well-scoped:

- **Refactor is sound** — `getSignatureFromTransaction` keeps its exact behavior (same `SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING` error) and the existing tests for it are untouched, which confirms no behavioral drift.
- **Tests cover the right cases** — signed fee payer, no signatures at all, and the important edge case where a non-fee-payer has signed but the fee payer slot is still `null`. That last one verifies the function keys off the *first* entry in the ordered `SignaturesMap` rather than "any signature present".
- **Docs are thorough** — README, TSDoc with cross-`@see` links in both directions, and a changeset. The `minor` bump for a new exported function is correct.
- Since `signatures.ts` is re-exported via `export * from './signatures'`, the new function flows through to `@solana/kit` automatically.

- The `!transactionSignature` falsy check in the throwing wrapper is safe: the base58 decoding of 64 signature bytes can never produce an empty string, so the only falsy value is `undefined`.
- Nothing else in the monorepo needed updating — existing callers of `getSignatureFromTransaction` are unaffected.

LGTM. (Submitting as a comment since I don't have approve permissions on this repo.)

mcintyre94:
Falsy check is changed to `transactionSignature === undefined`. This avoids a regression in existing tests that use an empty string signature.

mcintyre94:
Closing - we closed #1893 and don't need this. We'll add it if there's demand or we find a use case

PR #1897, opened 2026-08-07
Add a ClientWithFetchAccounts interface

lorisleiva:
This adds a new `ClientWithFetchAccounts` plugin interface to `@solana/plugin-interfaces`, representing a client that can fetch the encoded content of accounts from their addresses via a `fetchAccounts(addresses, config?)` method. Like the other capability interfaces, it lets plugins provide or require account fetching without coupling to a concrete RPC. The returned array matches the provided addresses in length and order, using `MaybeEncodedAccount` to represent accounts that may not exist. The interface intentionally takes an array and returns an array (no single/array overload) to keep it easy to implement for custom fetch strategies.

### DIFF
--- a/packages/plugin-interfaces/src/fetch-accounts.ts
+++ b/packages/plugin-interfaces/src/fetch-accounts.ts
@@ -1,0 +1,40 @@
+import { FetchAccountsConfig, MaybeEncodedAccount } from '@solana/accounts';
+import { Address } from '@solana/addresses';
+
+/**
+ * Represents a client that can fetch the content of accounts from their addresses.
+ *
+ * Different implementations may fetch accounts differently — for example, by calling the
+ * `getAccountInfo` and/or `getMultipleAccounts` RPC methods, by reading from a local validator,
+ * or by using a locally cached value.
+ *
+ * Note that this interface only fetches encoded accounts. Callers that need decoded accounts should
+ * decode the returned {@link MaybeEncodedAccount | MaybeEncodedAccounts} themselves using the codec
+ * of their choice.
+ *
+ * If you have a raw `Rpc` object instead of a client, you can construct a client implementing this
+ * interface using the {@link createClientWithFetchAccountsFromRpc} helper from `@solana/kit`.
+ *
+ * @example
+ * ```ts
+ * async function fetchProgramAddresses(client: ClientWithFetchAccounts, addresses: Address[]) {
+ *     const accounts = await client.fetchAccounts(addresses);
+ *     return accounts.filter(account => account.exists).map(account => account.programAddress);
+ * }
+ * ```
+ */
+export type ClientWithFetchAccounts = {
+    /**
+     * Fetches the encoded content of the accounts at the provided addresses.
+     *
+     * The returned array matches the provided addresses in both length and order. Each item is a
+     * {@link MaybeEncodedAccount} so that missing accounts can be represented whilst keeping track
+     * of their address.
+     *
+     * @param addresses - The addresses of the accounts to fetch.
+     * @param config - Optional configuration for the fetch.
+     * @returns A promise resolving to an array of {@link MaybeEncodedAccount | MaybeEncodedAccounts}
+     *          in the same order as the provided addresses.
+     */
+    fetchAccounts: (addresses: Address[], config?: FetchAccountsConfig) => Promise<MaybeEncodedAccount[]>;
+};
--- a/.changeset/ten-sloths-call.md
+++ b/.changeset/ten-sloths-call.md
@@ -1,0 +1,7 @@
+---
+'@solana/plugin-interfaces': minor
+---
+
+Add a `ClientWithFetchAccounts` interface
+
+This new plugin interface represents a client that can fetch the encoded content of accounts from their addresses via a `fetchAccounts(addresses, config?)` method. Like the other `@solana/plugin-interfaces` capabilities, it lets plugins provide or require account-fetching without coupling to a concrete RPC. The returned array matches the provided addresses in length and order, using `MaybeEncodedAccount` to represent accounts that may not exist.
--- a/packages/plugin-interfaces/README.md
+++ b/packages/plugin-interfaces/README.md
@@ -138,6 +138,25 @@ function accountCreationPlugin() {
 }
 ```
 
+### `ClientWithFetchAccounts`
+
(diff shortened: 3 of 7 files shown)


lorisleiva:
* **#1898** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1898?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1897** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1897?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1897?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Summary

This PR adds a `ClientWithFetchAccounts` capability interface to `@solana/plugin-interfaces`, exposing a `fetchAccounts(addresses, config?)` method that resolves to `MaybeEncodedAccount[]` in the same length/order as the input addresses. It follows the established pattern of the other capability interfaces (`ClientWithAirdrop`, `ClientWithGetMinimumBalance`, etc.): a single-method object type, thorough docblocks, a README section with a plugin example, a typetest file, and a `minor` changeset.

The implementation is clean and consistent with the package's conventions — this would be an approval if I had the repository access to submit one:

- ✅ Changeset present with a `minor` bump — correct for a new feature per the repo's policy.
- ✅ README section added, matching the structure and style of the existing interface sections.
- ✅ Typetest follows the `// [DESCRIBE]` + nested-block structure from the testing conventions.
- ✅ Docblocks are complete with `@param`, `@returns`, and an `@example`.
- ✅ New `@solana/accounts` workspace dependency is correctly reflected in `package.json` and `pnpm-lock.yaml`.
- The array-only signature (no single/array overload) is a reasonable simplification for implementers, as called out in the PR description. Not preserving per-address literal types (unlike `fetchEncodedAccounts`'s mapped tuples) is consistent with that same "easy to implement" goal.
- The `@example` in the docblock relies on TS 5.5+ inferred type predicates for `.filter(account => account.exists)` to narrow to `EncodedAccount` — that works fine with the repo's TS version.

## Things to watch out for

1. **Forward reference to `createClientWithFetchAccountsFromRpc`** — the docblock links to a `@solana/kit` helper that doesn't exist in the repo yet (it's not exported from `packages/kit/src/index.ts`). If this is the next PR in a stack, fine — but worth confirming the helper lands before this doc ships, otherwise the `{@link}` will be broken in the generated API docs. (Inline comment below.)

2. **Config type reuse** — the interface reuses `FetchAccountsConfig` from `@solana/accounts`, which carries RPC-flavored options (`commitment`, `minContextSlot`) that non-RPC implementations (caches, LiteSVM-style clients) can't meaningfully honor. This is a deliberate trade-off worth a quick sanity check. (Inline comment below.)

- Nothing runtime here — the package exports types only (`export type * from '@solana/plugin-interfaces'` in kit), so there's no bundle-size or behavior risk.
- The `kit-plugins` repo is the main consumer of these interfaces; the follow-up implementing this capability there (or in kit) is where the interesting review will be.

packages/plugin-interfaces/src/fetch-accounts.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Different implementations may fetch accounts differently — for example, by calling the
+ * `getAccountInfo` and/or `getMultipleAccounts` RPC methods, by reading from a local validator,
+ * or by using a locally cached value.
+ *
+ * Note that this interface only fetches encoded accounts. Callers that need decoded accounts should
+ * decode the returned {@link MaybeEncodedAccount | MaybeEncodedAccounts} themselves using the codec
+ * of their choice.
+ *
+ * If you have a raw `Rpc` object instead of a client, you can construct a client implementing this
+ * interface using the {@link createClientWithFetchAccountsFromRpc} helper from `@solana/kit`.
```
trevor-cortex, line 16: `createClientWithFetchAccountsFromRpc` doesn't exist in `@solana/kit` yet (it's not in `packages/kit/src/index.ts`). If it's coming in the next PR of the stack, no problem — just flagging that if this merges and releases on its own, the docblock references a helper that isn't there and the `{@link}` will render broken in the API docs. An alternative is to add this paragraph in the PR that introduces the helper.

packages/plugin-interfaces/src/fetch-accounts.ts
```diff
@@ -0,0 +1,40 @@
...
+     * The returned array matches the provided addresses in both length and order. Each item is a
+     * {@link MaybeEncodedAccount} so that missing accounts can be represented whilst keeping track
+     * of their address.
+     *
+     * @param addresses - The addresses of the accounts to fetch.
+     * @param config - Optional configuration for the fetch.
+     * @returns A promise resolving to an array of {@link MaybeEncodedAccount | MaybeEncodedAccounts}
+     *          in the same order as the provided addresses.
+     */
+    fetchAccounts: (addresses: Address[], config?: FetchAccountsConfig) => Promise<MaybeEncodedAccount[]>;
```
trevor-cortex, line 39: Design question: reusing `FetchAccountsConfig` from `@solana/accounts` makes RPC-backed implementations trivially compatible with `fetchEncodedAccounts`, but it also bakes RPC-flavored options (`commitment`, `minContextSlot`) into the capability contract — options that cache-based or LiteSVM-style implementations can't honor. The other interfaces in this package define their own config types (e.g. `GetMinimumBalanceConfig`). If the RPC coupling is intentional, all good — but if you want the interface to stay implementation-agnostic, a package-local config (even one that starts as an alias or a subset like `{ abortSignal?: AbortSignal }`) would give you room to diverge later without a breaking change.

lorisleiva: I agree these are technically RPC options _but_ they are also valid as interface options IMO as these are essentially account concepts.

packages/plugin-interfaces/src/fetch-accounts.ts
lorisleiva: Yeah next PR in the stack.

mcintyre94 (approved):
Nice, LGTM!

PR #1898, opened 2026-08-07
Add helpers to create client interfaces from a raw Rpc

lorisleiva:
Add `createClientWithGetMinimumBalanceFromRpc`, `createClientWithFetchAccountsFromRpc` and `createClientWithInterfacesFromRpc` to `@solana/kit`. These convenience helpers let consumers that only have a raw `Rpc` object construct the corresponding client interfaces (`ClientWithGetMinimumBalance` and `ClientWithFetchAccounts`) without assembling a full Kit client. `createClientWithInterfacesFromRpc` fills in whichever interfaces the RPC supports and narrows its return type accordingly.

### DIFF
--- a/packages/kit/src/create-client-with-interfaces-from-rpc.ts
+++ b/packages/kit/src/create-client-with-interfaces-from-rpc.ts
@@ -1,0 +1,137 @@
+import { BASE_ACCOUNT_SIZE, fetchEncodedAccount, fetchEncodedAccounts } from '@solana/accounts';
+import type { ClientWithFetchAccounts, ClientWithGetMinimumBalance } from '@solana/plugin-interfaces';
+import type { GetAccountInfoApi, GetMinimumBalanceForRentExemptionApi, GetMultipleAccountsApi, Rpc } from '@solana/rpc';
+import { lamports } from '@solana/rpc-types';
+
+/**
+ * Creates a {@link ClientWithGetMinimumBalance} from a raw `Rpc` object.
+ *
+ * The returned client computes the minimum balance for rent exemption using the
+ * {@link GetMinimumBalanceForRentExemptionApi.getMinimumBalanceForRentExemption | getMinimumBalanceForRentExemption}
+ * RPC method. By default, the 128-byte account header is included on top of the provided `space`;
+ * pass `{ withoutHeader: true }` to compute the minimum balance for the data portion only.
+ *
+ * This is a convenience helper for consumers that only have a raw `Rpc` object rather than a Kit
+ * client. If you are building a full client, prefer composing it with a plugin such as `solanaRpc`
+ * (i.e. `createClient().use(solanaRpc(...))`), which provides `getMinimumBalance` amongst other
+ * capabilities.
+ *
+ * @param rpc - An object that supports the {@link GetMinimumBalanceForRentExemptionApi} of the
+ *              Solana RPC API.
+ *
+ * @example
+ * ```ts
+ * const client = createClientWithGetMinimumBalanceFromRpc(rpc);
+ * const rentExemptBalance = await client.getMinimumBalance(100);
+ * ```
+ */
+export function createClientWithGetMinimumBalanceFromRpc(
+    rpc: Rpc<GetMinimumBalanceForRentExemptionApi>,
+): ClientWithGetMinimumBalance {
+    return {
+        async getMinimumBalance(space, config) {
+            if (config?.withoutHeader) {
+                // The runtime computes rent as `rate * (BASE_ACCOUNT_SIZE + space)`, where `rate`
+                // folds in the per-byte cost and the exemption threshold (see Agave's
+                // `Rent::minimum_balance`). Querying `space = 0` therefore returns `rate * 128`,
+                // which divides evenly by `BASE_ACCOUNT_SIZE` to recover `rate` exactly. There is
+                // no truncation here: `rate * 128 / 128 === rate`.
+                const headerBalance = await rpc.getMinimumBalanceForRentExemption(0n).send();
+                const lamportsPerByte = headerBalance / BigInt(BASE_ACCOUNT_SIZE);
+                return lamports(lamportsPerByte * BigInt(space));
+            }
+            return await rpc.getMinimumBalanceForRentExemption(BigInt(space)).send();
+        },
+    };
+}
+
+/**
+ * Creates a {@link ClientWithFetchAccounts} from a raw `Rpc` object.
+ *
+ * The returned client fetches the encoded content of accounts from their addresses, dispatching on
+ * the number of requested addresses: a single account is fetched via the
+ * {@link GetAccountInfoApi.getAccountInfo | getAccountInfo} RPC method, whilst multiple accounts are
+ * fetched in a single round-trip via the
+ * {@link GetMultipleAccountsApi.getMultipleAccounts | getMultipleAccounts} RPC method. Fetching an
+ * empty list short-circuits to an empty array without issuing any RPC call.
+ *
+ * The dispatch is based purely on the number of addresses because a raw `Rpc` object's capabilities
+ * cannot be detected at runtime. For this reason, the `Rpc` is required to support both methods.
+ *
+ * This is a convenience helper for consumers that only have a raw `Rpc` object rather than a Kit
+ * client. If you are building a full client, prefer composing it with a plugin such as `solanaRpc`
+ * (i.e. `createClient().use(solanaRpc(...))`), which provides account fetching amongst other
+ * capabilities.
+ *
+ * @param rpc - An object that supports both the {@link GetAccountInfoApi} and the
+ *              {@link GetMultipleAccountsApi} of the Solana RPC API.
+ *
+ * @example
+ * ```ts
+ * const client = createClientWithFetchAccountsFromRpc(rpc);
+ * const accounts = await client.fetchAccounts([addressA, addressB]);
+ * ```
+ */
+export function createClientWithFetchAccountsFromRpc(
+    rpc: Rpc<GetAccountInfoApi & GetMultipleAccountsApi>,
+): ClientWithFetchAccounts {
+    return {
+        async fetchAccounts(addresses, config) {
+            if (addresses.length === 0) {
+                return [];
+            }
+            if (addresses.length === 1) {
+                return [await fetchEncodedAccount(rpc, addresses[0], config)];
+            }
+            return await fetchEncodedAccounts(rpc, addresses, config);
+        },
+    };
+}
+
+type ClientInterfacesFromRpc<TRpc> = (TRpc extends Rpc<GetAccountInfoApi & GetMultipleAccountsApi>
+    ? ClientWithFetchAccounts
+    : object) &
+    (TRpc extends Rpc<GetMinimumBalanceForRentExemptionApi> ? ClientWithGetMinimumBalance : object);
+
+/**
+ * Creates a client from a raw `Rpc` object, filling in whichever client interfaces the RPC supports.
+ *
+ * The returned object's type implements a {@link ClientWithGetMinimumBalance} when the RPC supports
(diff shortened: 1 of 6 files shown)


lorisleiva:
* **#1898** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1898?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1898?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1897** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1897?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/kit/src/create-client-with-interfaces-from-rpc.ts
```diff
@@ -0,0 +1,128 @@
...
+ * const rentExemptBalance = await client.getMinimumBalance(100);
+ * const accounts = await client.fetchAccounts([addressA, addressB]);
+ * ```
+ */
+export function createClientWithInterfacesFromRpc<
+    TRpc extends Rpc<GetAccountInfoApi> | Rpc<GetMinimumBalanceForRentExemptionApi> | Rpc<GetMultipleAccountsApi>,
+>(rpc: TRpc): ClientInterfacesFromRpc<TRpc> {
+    return {
+        ...('getMinimumBalanceForRentExemption' in rpc ? createClientWithGetMinimumBalanceFromRpc(rpc) : {}),
+        ...('getAccountInfo' in rpc || 'getMultipleAccounts' in rpc ? createClientWithFetchAccountsFromRpc(rpc) : {}),
```
trevor-cortex: **Critical:** these `in` checks are always `false` for any `Rpc` created via `createRpc` / `createSolanaRpc`. Those objects are Proxies (`makeProxy` in `@solana/rpc-spec`) with no `has` trap, wrapping the `createJsonRpcApi` proxy whose own target is `{}` — so `'getMinimumBalanceForRentExemption' in rpc` resolves to `Reflect.has({}, ...)` and returns `false`.

As a result, `createClientWithInterfacesFromRpc(createSolanaRpc(url))` returns `{}` at runtime while its type claims both interfaces. The tests pass only because they use plain-object mocks.

Note `typeof rpc.getAccountInfo === 'function'` won't help either — the proxy's `get` trap returns a function for *every* name. Since `ClientInterfacesFromRpc<TRpc>` already narrows statically, one option is to unconditionally build both interfaces and rely on the return type to hide the unsupported ones; alternatively, take an explicit config of which interfaces to construct. See the review body for more discussion.

packages/kit/src/create-client-with-interfaces-from-rpc.ts
```diff
@@ -0,0 +1,128 @@
...
+    rpc: Rpc<GetAccountInfoApi> | Rpc<GetMultipleAccountsApi>,
+): ClientWithFetchAccounts {
+    return {
+        async fetchAccounts(addresses, config) {
+            // Prefer `getAccountInfo` for a single account when it is available.
+            if (addresses.length === 1 && 'getAccountInfo' in rpc) {
+                return [await fetchEncodedAccount(rpc, addresses[0], config)];
+            }
+            if ('getMultipleAccounts' in rpc) {
+                return await fetchEncodedAccounts(rpc, addresses, config);
```
trevor-cortex: Same root cause as the spread checks below: for a real proxy-backed `Rpc`, both `'getAccountInfo' in rpc` and `'getMultipleAccounts' in rpc` are `false`, so this always falls through to the `Promise.all` branch — N parallel `getAccountInfo` calls even when the RPC supports `getMultipleAccounts`. It still *works* (the proxy accepts any method name), but it contradicts the documented "uses `getMultipleAccounts` when available" behaviour and multiplies request volume for the most common case.

Also worth deciding explicitly: `fetchAccounts([])` with `getMultipleAccounts` available will currently issue an RPC call with an empty address list — an early `return []` might be worth adding.

packages/kit/src/create-client-with-interfaces-from-rpc.ts
```diff
@@ -0,0 +1,128 @@
...
+ */
+export function createClientWithGetMinimumBalanceFromRpc(
+    rpc: Rpc<GetMinimumBalanceForRentExemptionApi>,
+): ClientWithGetMinimumBalance {
+    return {
+        async getMinimumBalance(space, config) {
+            if (config?.withoutHeader) {
+                const headerBalance = await rpc.getMinimumBalanceForRentExemption(0n).send();
+                const lamportsPerByte = headerBalance / BigInt(BASE_ACCOUNT_SIZE);
+                return lamports(lamportsPerByte * BigInt(space));
```
trevor-cortex, line 41: The derived-rate approach truncates: `headerBalance / BigInt(BASE_ACCOUNT_SIZE)` floors, so whenever the cluster's rent parameters don't divide evenly by 128, this *underestimates* the minimum balance — and an underfunded account fails rent exemption at runtime, which is the worst failure direction for this helper.

An exact alternative is subtraction, since `getMin(space) − getMin(0)` = `rate × (128 + space) − rate × 128` = `rate × space`:

```ts
const [fullBalance, headerBalance] = await Promise.all([
    rpc.getMinimumBalanceForRentExemption(BigInt(space)).send(),
    rpc.getMinimumBalanceForRentExemption(0n).send(),
]);
return lamports(fullBalance - headerBalance);
```

It costs a second (parallel) RPC call, but it's exact for any rent configuration. If the single-call version is preferred, at minimum round up rather than down.

trevor-cortex (commented):
## Summary

This PR adds three convenience helpers to `@solana/kit` — `createClientWithGetMinimumBalanceFromRpc`, `createClientWithFetchAccountsFromRpc`, and `createClientWithInterfacesFromRpc` — that wrap a raw `Rpc` object into the `ClientWithGetMinimumBalance` and/or `ClientWithFetchAccounts` plugin interfaces without assembling a full Kit client. It ships with unit tests, typetests, README docs, and a `minor` changeset (appropriate for new features).

The API shape, type-level narrowing in `ClientInterfacesFromRpc<TRpc>`, typetests, and docs are all well done. Unfortunately there's a fundamental runtime problem that the unit tests don't catch. *(I would have marked this "request changes", but I don't have that permission on this repo — treat this as such.)*

## Critical: `in rpc` capability checks never pass for real Kit `Rpc` objects

Every `Rpc` produced by `createRpc` / `createSolanaRpc` / `createSolanaRpcFromTransport` is a `Proxy` (see `makeProxy` in `packages/rpc-spec/src/rpc.ts`). That proxy defines only `get`, `defineProperty`, and `deleteProperty` traps, and its target is the `createJsonRpcApi` proxy — which itself has no `has` trap and wraps a plain `{}` target. The `in` operator therefore falls through both proxies to `Reflect.has({}, ...)` and **always returns `false`** for every method name.

Consequences:

- `createClientWithInterfacesFromRpc(createSolanaRpc(url))` returns an **empty object at runtime**, while its type claims both interfaces — `client.getMinimumBalance is not a function` for the primary real-world use case.
- `createClientWithFetchAccountsFromRpc` on a real RPC never takes the `getMultipleAccounts` path and silently degrades to N `getAccountInfo` calls, contradicting the documented behaviour.

The unit tests pass only because they use plain-object mocks, where `in` works. Note that `typeof rpc.getAccountInfo === 'function'` is no better as a detection mechanism — the `get` trap returns a function for *any* property name — so runtime feature detection on these proxies is fundamentally unreliable. Some options to consider:

1. Since `ClientInterfacesFromRpc<TRpc>` already narrows the return type statically, always build both interfaces at runtime and let the cast hide the untyped ones (the current `not.toHaveProperty` test expectations would need to change).
2. Accept an explicit config telling the helper which interfaces/methods to back.
3. Add a `has` trap to the RPC proxies in `@solana/rpc-spec` — though that's a larger change and `createJsonRpcApi` is intentionally a catch-all, so there may be no method list to consult.

## Other issues

- The `withoutHeader` computation derives a per-byte rate via integer division, which can silently underestimate rent (see inline comment). An exact subtraction-based approach is available.
- **Tests**: please add at least one test that exercises these helpers against a proxy-backed RPC (e.g. built with `createRpc` + `createJsonRpcApi` and a mock transport, or `createSolanaRpcFromTransport`). That test would have caught the `in` issue immediately, and it guards against future regressions in how the proxies interact with these helpers.

- Verify the intended semantics of `withoutHeader` against the runtime's rent formula (`(BASE_ACCOUNT_SIZE + data_len) × rate`) — the current implementation and the interface docs in `@solana/plugin-interfaces` agree on intent, but the arithmetic deserves a close look.
- The changeset (`minor`) and README additions look consistent with repo conventions; no concerns there.
- The typetests are thorough and the `ClientInterfacesFromRpc<TRpc>` conditional type handles the `Rpc<SolanaRpcApi>` superset case correctly.

packages/kit/src/create-client-with-interfaces-from-rpc.ts
lorisleiva: This is not applicable here since the minimum balance formula includes a `lamports_per_bytes` variable. Since `rpc.getMinimumBalanceForRentExemption(0n)` returns the price for 128 bytes, we can always divide it by 128 to access that value.

packages/kit/src/create-client-with-interfaces-from-rpc.ts
lorisleiva: Oof that's a good catch which sadly changes quite a lot of the architecture but I suppose that's the tradeoff of having a nice type-only RPC Proxy. I've made the changes.

packages/kit/src/create-client-with-interfaces-from-rpc.ts
lorisleiva: Related to the previous comment. Addressed, thank you!

trevor-cortex (commented):
## Re-review

**Verdict: approve.** *(I don't have permission to submit a formal `APPROVE` review on this repo, so this is a comment review — please treat it as an approval.)*

All points from my previous review have been addressed — nicely done:

- **`in rpc` capability checks (critical)**: Resolved by building both interfaces unconditionally and letting the `ClientInterfacesFromRpc<TRpc>` return type do the narrowing (option 1 from my review). The docs, README, and JSDoc now clearly state that a raw `Rpc`'s capabilities cannot be detected at runtime and that calling an unsupported method fails at the underlying RPC call. That's an honest, well-documented contract.
- **Proxy-backed RPC test**: The new `with a proxy-backed RPC` describe block exercises the helpers against a real `createRpc` + `createJsonRpcApi` proxy with a mock transport, covering `getMinimumBalance`, the single-account `getAccountInfo` path, and the multi-account `getMultipleAccounts` path. This is exactly the regression guard that would have caught the original issue.
- **`withoutHeader` arithmetic**: The per-byte rate is now derived from `getMinimumBalanceForRentExemption(0n)` with a documented argument for why the division by `BASE_ACCOUNT_SIZE` is exact (`rate * 128 / 128 === rate` under the runtime's linear rent formula). Single round-trip, exact under the current formula — good resolution.
- **Empty address list**: `fetchAccounts([])` now short-circuits to `[]` without issuing an RPC call, with test coverage.

- The exactness of the `withoutHeader` derivation relies on the header-only balance being an exact integer multiple of `BASE_ACCOUNT_SIZE`, which holds under the current rent parameters (`lamports_per_byte_year × exemption_threshold` is integral). If a cluster ever ran with rent parameters where that product is non-integral, a small truncation error could reappear — theoretical today, and the code comment documents the assumption, so nothing to change here.
- Typetests cover all the interesting `ClientInterfacesFromRpc<TRpc>` branches, including rejection of single-account-method RPCs and the `Rpc<SolanaRpcApi>` superset case.

Changeset (`minor`), README, and export wiring all look consistent with repo conventions. Ship it.

mcintyre94 (approved):
Nice!

PR #1899, opened 2026-08-07
Add ClientWithTransactionSigning interface

mcintyre94:
#### Summary of Changes

This PR adds the `ClientWithTransactionSigning` interface, which adds `signTransaction[s]` functions to the client.

These take the same inputs as `sendTransaction`. The implementation should finalise and sign the transaction, for example adding a lifetime that may not have been part of the planner.

This interface imposes no constraints on the returned context, taking advantage of our new flexible context. Implementations can return whatever context fields make sense, for example a `Transaction` may be common, while a `Signature` would depend on whether the returned transactions are fully signed (which depends on the implementation).

This PR also loosens the `ClientWithTransactionSending` type so that it similarly imposes no constraint on the context for the interface. For backward compatibility it defaults to `TransactionPlanResultContextWithSignature`, the previous type that required a `Signature`.

### DIFF
--- a/packages/plugin-interfaces/src/instruction-plans.ts
+++ b/packages/plugin-interfaces/src/instruction-plans.ts
@@ -5,6 +5,8 @@ import type {
     TransactionPlan,
     TransactionPlanInput,
     TransactionPlanResult,
+    TransactionPlanResultContext,
+    TransactionPlanResultContextWithSignature,
 } from '@solana/instruction-plans';
 
 type Config = { abortSignal?: AbortSignal };
@@ -65,6 +67,11 @@ export type ClientWithTransactionPlanning = {
  * transactions. It supports flexible input formats including instructions,
  * instruction plans, transaction messages or transaction plans.
  *
+ * @typeParam TContext - The context attached to the results. It defaults to
+ * {@link TransactionPlanResultContextWithSignature}, which guarantees a `context.signature` on
+ * every successful result. Supply a different context to change or drop that guarantee — for
+ * instance, a client whose executor records extra fields on the context.
+ *
  * @example
  * ```ts
  * async function executeTransfer(client: ClientWithTransactionSending) {
@@ -79,7 +86,9 @@ export type ClientWithTransactionPlanning = {
  * }
  * ```
  */
-export type ClientWithTransactionSending = {
+export type ClientWithTransactionSending<
+    TContext extends TransactionPlanResultContext = TransactionPlanResultContextWithSignature,
+> = {
     /**
      * Sends a single transaction to the network.
      *
@@ -96,7 +105,7 @@ export type ClientWithTransactionSending = {
     sendTransaction: (
         input: InstructionPlanInput | SingleTransactionPlan | SingleTransactionPlan['message'],
         config?: Config,
-    ) => Promise<SuccessfulSingleTransactionPlanResult>;
+    ) => Promise<SuccessfulSingleTransactionPlanResult<TContext>>;
 
     /**
      * Sends one or more transactions to the network.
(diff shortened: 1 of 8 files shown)


mcintyre94:
* **#1899** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1899?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1899?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/plugin-interfaces/src/instruction-plans.ts
```diff
@@ -116,3 +118,72 @@ export type ClientWithTransactionSending = {
...
+     * @returns A promise resolving to the successful transaction result, whose signature may be
+     * absent if the transaction is not yet signed by its fee payer.
+     *
+     * @see {@link InstructionPlanInput}
+     * @see {@link SingleTransactionPlan}
+     */
+    signTransaction: (
+        input: InstructionPlanInput | SingleTransactionPlan | SingleTransactionPlan['message'],
+        config?: Config,
+    ) => Promise<SuccessfulSingleTransactionPlanResultWithOptionalSignature>;
```
trevor-cortex: Design question, non-blocking: `SuccessfulSingleTransactionPlanResultWithOptionalSignature` types `context.transaction` as optional (it comes from `BaseTransactionPlanResultContext`), yet retrieving the signed transaction is the entire purpose of this interface. As typed, every caller must handle `result.context.transaction` being `undefined` — including your own README example, which passes `result.context.transaction` straight to `postToRelayer` and would fail a strict type-check.

In practice `successfulSingleTransactionPlanResultWithOptionalSignature` always populates `transaction`, so the type is looser than the reality. Would it be worth tightening the return type here to guarantee `transaction` is present (e.g. intersecting with `{ context: { transaction: Transaction } }`, or adding a dedicated result type in `@solana/instruction-plans`)? If the looseness is intentional — e.g. to allow executors that only report signatures — a sentence in the docblock explaining how callers should handle a missing `transaction` would help.

trevor-cortex (commented):
This PR adds the `ClientWithTransactionSigning` interface to `@solana/plugin-interfaces`, installing `signTransaction(s)` methods that mirror `sendTransaction(s)` — same flexible inputs, same config — but return the `…WithOptionalSignature` result types from #1893, since the fee payer signature may legitimately be absent. Docs pages, README, and typetests are all updated.

The implementation is solid:

- The interface mirrors `ClientWithTransactionSending` faithfully — input unions, `Config`, and docblock structure all match the house style.
- The typetests follow the repo's `// [DESCRIBE]` + nested-block convention, and the assertion that a `ClientWithTransactionSending`'s methods structurally satisfy the signing methods is a nice way to pin down the required-signature-assignable-to-optional relationship.
- README and both docs pages are updated consistently.

Two points before merge:

1. **Missing changeset.** This ships a new public API on the published `@solana/plugin-interfaces` package, so per the repo conventions it needs a changeset (`minor`, generated via `npx changeset add --empty`). I don't see one in the changed files — if it's intentionally deferred to another PR in the stack, ignore this, but as-is this PR wouldn't trigger a release.

2. **`context.transaction` is optional on the results** — see the inline comment. Not blocking, but worth a deliberate decision since handing back the transaction is the whole point of this interface.

For subsequent reviewers: the types this PR consumes are defined in #1893, so the main thing to verify here is the semantic contract (docblocks/README) rather than the type mechanics, which the typetests already cover.

packages/plugin-interfaces/src/instruction-plans.ts
mcintyre94: Agreed, changed this to narrow the context on this executor so that it has `transaction: Transaction` on the successful leaves.

mcintyre94:
@trevor-cortex

packages/plugin-interfaces/src/instruction-plans.ts
```diff
@@ -116,3 +122,107 @@ export type ClientWithTransactionSending = {
...
+ * `context.transaction` is guaranteed to be present: a signer that succeeded must have produced a
+ * transaction, even when that transaction is not yet signed by its fee payer and therefore has no
+ * signature to report.
+ *
+ * @see {@link ClientWithTransactionSigning}
+ * @see {@link SignedTransactionPlanResult}
+ */
+export type SignedSingleTransactionPlanResult = SuccessfulSingleTransactionPlanResultWithOptionalSignature<{
+    transaction: Transaction;
+}>;
```
trevor-cortex, line 139: Non-blocking: the `successfulSingleTransactionPlanResultWithOptionalSignature` factory in `@solana/instruction-plans` always populates `context.transaction`, but its declared return type leaves it optional — so an executor implementation built on that helper won't structurally satisfy `SignedSingleTransactionPlanResult` without a cast or manual narrowing. Consider tightening the factory's return type upstream (in a follow-up PR in the stack, since that package isn't touched here) so implementations of this interface type-check out of the box.

trevor-cortex (commented):
Thanks for addressing both points from the last pass — this now looks ready to go. (Would approve, but I don't have repository access to submit an approving review, so leaving this as a comment.)

**What changed since my last review:**

- A `minor` changeset for `@solana/plugin-interfaces` was added, resolving point 1.
- The optional-`context.transaction` concern (point 2) is resolved properly: the new `SignedSingleTransactionPlanResult` narrows `SuccessfulSingleTransactionPlanResultWithOptionalSignature` via the `TContext` slot so `context.transaction` is required, and `SignedTransactionPlanResult` threads that through the result tree for `signTransactions`. Neat use of the context generic — the intersection of `transaction?: Transaction` with `{ transaction: Transaction }` does exactly the right thing, and it stays within the `TSingle` bound (`SingleTransactionPlanResultWithOptionalSignature`) used by the tree types.
- The typetests now cover all the new guarantees: `transaction` always present, `signature` still optional, the loose result *not* assignable to the signing result, and — nicely — the flipped assertion that a `ClientWithTransactionSending`'s methods no longer satisfy the signing methods, which documents the deliberate divergence introduced by the tightening.
- README example and docblocks updated to match (`result.context.transaction` is now safe unguarded), and `package.json`/lockfile pick up the two new type-only workspace deps.

**One non-blocking note** (see inline): the `successfulSingleTransactionPlanResultWithOptionalSignature` factory in `@solana/instruction-plans` always populates `transaction` but its return type leaves it optional, so implementers building on that helper won't structurally satisfy `SignedSingleTransactionPlanResult` without a cast. Worth tightening upstream in a follow-up.

**For subsequent reviewers:** the type mechanics are fully pinned by the typetests; the remaining thing to sanity-check is the semantic contract in the docblocks/README (relayer use case, fee-payer-signature-optional semantics), which reads accurately to me.

packages/plugin-interfaces/src/instruction-plans.ts
mcintyre94: This should now be resolved downstack: https://github.com/anza-xyz/kit/pull/1893#issuecomment-5218675924

lorisleiva (approved):
I know it's still in draft, but looks good! 😊

mcintyre94:
### Merge activity

* **Aug 18, 3:02 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1899).
* **Aug 18, 3:09 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1899) rebased this pull request as part of a merge.
* **Aug 18, 3:13 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1899).

PR #1900, opened 2026-08-07
feat(react): add createProgramHooks factory for Codama program plugins

amilz:
A Codama-generated program plugin is already a fully typed runtime map — codecs with `fetch` helpers, instruction builders with `sendTransaction`, PDA finders. `createProgramHooks` projects that map into React hooks, so one factory covers every program with no per-program codegen.

```tsx
const programHooks = createProgramHooks<MyClient>();
const { useAccount, useSendInstruction } = programHooks('token');
const { usePda } = programHooks('associatedToken');

const ata = usePda('associatedToken', { mint, owner });
const balance = useAccount('token', ata.data?.[0] ?? null);
const mintTo = useSendInstruction('mintTo');
```

Point the type argument at the client type and an uninstalled plugin is a compile-time error. A `null` address, address list or seed object disables a hook rather than requiring a conditional call.

## Hooks

Each one maps a single plugin function onto an existing primitive in this package, and is named for the primitive it uses.

| Hook | Plugin fn | Primitive |
| --- | --- | --- |
| `useAccount` | `.fetch` | `useRequest` |
| `useMaybeAccount` | `.fetchMaybe` | `useRequest` |
| `useAllAccounts` | `.fetchAll` | `useRequest` |
| `useAllMaybeAccounts` | `.fetchAllMaybe` | `useRequest` |
| `usePda` | `find*Pda` | `useRequest` |
| `useTrackedAccount` | `.decode` | `useTrackedData` |
| `useSendInstruction` | builder + `.sendTransaction` | `useAction` |

`useTrackedAccount` is the only one that opens a subscription: initial `getAccountInfo` plus `accountNotifications`, slot-deduped, both decoded with the plugin's own codec. It's opt-in rather than the default for `useAccount`, since holding a subscription per address isn't the right cost for every app.

## Notes

- Type projections (`ProgramAccountKey`, `ProgramInstructionInput`, …) derive everything from the plugin type — no Codama naming convention is reimplemented.
- `useStableValue` holds inline array/object arguments stable by value, so `usePda('associatedToken', { mint, owner })` doesn't rebuild its store every render.
- The curried call recovers the capability literal from a multi-plugin client type. The direct form `createProgramHooks<{ token: TokenPlugin }>('token')` still works for a single-plugin slice; on a multi-plugin type it degrades to an instructive error literal rather than silently typing against the union. Happy to simplify to curried-only if you'd prefer one call shape.

## Test plan

- 16 unit tests (`createProgramHooks-test.browser.tsx`) and a type test with a self-contained plugin fixture.
- `packages/react`: prettier, lint, typecheck, browser unit (313), node unit (77), treeshakability — all green.

### DIFF
--- a/packages/react/src/programNamespace.ts
+++ b/packages/react/src/programNamespace.ts
@@ -1,0 +1,82 @@
+import {
+    type AccountNotificationsApi,
+    type Address,
+    type ClientWithRpc,
+    type ClientWithRpcSubscriptions,
+    type GetAccountInfoApi,
+    getBase64Encoder,
+    type GetMultipleAccountsApi,
+    type ReadonlyUint8Array,
+} from '@solana/kit';
+
+/** Config shared by every account fetch the program hooks perform. */
+type FetchConfig = { abortSignal?: AbortSignal };
+
+/** Runtime view of one entry in a program plugin's `accounts` map. */
+export type ProgramAccountEntry = {
+    decode: (bytes: ReadonlyUint8Array, offset?: number) => unknown;
+    fetch: (address: Address, config?: FetchConfig) => Promise<unknown>;
+    fetchAll: (addresses: Address[], config?: FetchConfig) => Promise<unknown>;
+    fetchAllMaybe: (addresses: Address[], config?: FetchConfig) => Promise<unknown>;
+    fetchMaybe: (address: Address, config?: FetchConfig) => Promise<unknown>;
+};
+
+/** Runtime view of one entry in a program plugin's `instructions` map. */
+export type ProgramInstructionEntry = (input: never) => {
+    sendTransaction: (config?: unknown) => Promise<unknown>;
+};
+
+/** Runtime view of a program plugin as installed on the client. */
+export type ProgramNamespace = {
+    accounts?: Record<string, ProgramAccountEntry>;
+    instructions?: Record<string, ProgramInstructionEntry>;
+    pdas?: Record<string, (seeds: never) => unknown>;
+};
+
+/** The client capabilities the program hooks read from. */
+export type ProgramHooksClient = ClientWithRpc<GetAccountInfoApi & GetMultipleAccountsApi> &
+    ClientWithRpcSubscriptions<AccountNotificationsApi>;
+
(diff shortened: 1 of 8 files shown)


packages/react/src/programPlugin.ts
```diff
@@ -0,0 +1,112 @@
+/**
+ * Type-level projections over a Codama-generated program plugin.
```
GuiBibeau, line 2: is this file automatically generated each build? Curious if we truly need it committed to the repo vs having it as part of the build artifacts and NPM will deploy it even then.

packages/react/src/useStableValue.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (Array.isArray(a) || Array.isArray(b)) {
+        if (!Array.isArray(a) || !Array.isArray(b) || a.length !== b.length) return false;
+        return a.every((item, index) => isEqual(item, b[index]));
+    }
+    const aRecord = a as Record<string, unknown>;
+    const bRecord = b as Record<string, unknown>;
+    const aKeys = Object.keys(aRecord);
+    if (aKeys.length !== Object.keys(bRecord).length) return false;
+    return aKeys.every(key => Object.hasOwn(bRecord, key) && isEqual(aRecord[key], bRecord[key]));
+}
```
GuiBibeau, line 19: This is extremely expensive to run on every render. I clanked a few options and that version seems needed but it does raise a few alarm bells for me to see deep equality checks in a hook.

packages/react/src/programNamespace.ts
```diff
@@ -0,0 +1,82 @@
...
+    return (client as Record<string, ProgramNamespace>)[capability];
+}
+
+export function getAccountEntry(client: object, capability: string, key: string): ProgramAccountEntry {
+    const entry = getNamespace(client, capability).accounts?.[key];
+    if (!entry) {
+        throw new Error(`Program plugin "${capability}" does not declare an account named "${key}".`);
+    }
+    return entry;
+}
```
GuiBibeau, line 53: I'd add tsdocs strings to each function with examples for documentation and sharing intent on usage

PR #1902, opened 2026-08-07
Add signing error codes and factories

mcintyre94:
#### Summary of Changes

This PR adds Kit error codes for a `signTransaction` plugin to use: `SOLANA_ERROR__FAILED_TO_SIGN_TRANSACTION[s]`

These have the same shaped context as sign, because while `signTransaction` is not expected to actually send a transaction to the network, it may for example simulate to estimate CUs. In this case `logs` and `preflightData` are useful context for a failed simulation during signing.

We also add helpers `createFailedToSignTransactionError[s]`, again mirroring how the equivalent send helpers construct these errors. The only difference is that we skip the call to `getFailedIndicator` which attaches a preflight indicator or a signature, neither of which are useful for the sign error message.

The logic to implement these is shared with the send helpers.


mcintyre94:
* **#1902** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1902?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1902?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds `SOLANA_ERROR__FAILED_TO_SIGN_TRANSACTION[S]` error codes (13/14) to `@solana/errors` and their `createFailedToSignTransaction[s]Error` factories in `@solana/instruction-plans`, mirroring the existing failed-to-send codes. The implementation is a clean refactor: the bodies of the two send factories are extracted into shared `getSingleFailureContext` / `getMultipleFailuresContext` helpers, parameterized by an `includeSubmissionIndicator` flag that the sign variants set to `false` so the message drops the `(preflight)` / signature prefix while keeping `logs` and `preflightData` intact.

## What I checked

- **Behavior preservation for send**: the extracted helpers are line-for-line equivalent to the previous inlined logic when `includeSubmissionIndicator` is `true` — no behavioral change to the send factories.
- **Registry consistency**: codes 13/14 are appended sequentially in the correct range, and the `context.ts` / `messages.ts` / `SolanaErrorCode` union entries are all present and alphabetically ordered. Context shapes exactly mirror the send counterparts, including the non-enumerable `transactionPlanResult`.
- **Message formatting**: with an empty indicator, the single message becomes `Failed to sign transaction: <cause>` and plural lines become `[Tx #1] <cause>` — both covered by tests, including the edge case where a stray signature exists in the result context and is correctly omitted.
- **Tests**: thorough coverage of preflight unwrapping, cause propagation (single vs. multiple failures), cancellation with/without abort reason, non-enumerable `transactionPlanResult`, and error codes — matching the structure of the existing send tests.
- **Changeset**: present, `minor` for both packages, which fits an additive error-code + factory feature.

## Notes for other reviewers

- The `true /* includeSubmissionIndicator */` / `false /* ... */` call-site comments make the boolean flag readable enough that I don't think an options object is warranted for an internal helper.
- `createFailedToSignTransactionsError` reuses `TransactionPlanResultWithOptionalSignature` as its parameter type even though sign flows never produce signatures. That's the pragmatic choice since the plumbing is shared with send; a narrower type would add surface area for no real safety gain, so I wouldn't push on it.
- The helpers return `Record<string, unknown>` and rely on the `SolanaError` constructor accepting it, same as the pre-existing inlined code — no change in type rigor there.

LGTM — would approve, but I don't have approval permissions on this repo, so submitting as a comment instead.

lorisleiva (approved):
Thanks!

mcintyre94:
### Merge activity

* **Aug 18, 3:02 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1902).
* **Aug 18, 3:15 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1902) rebased this pull request as part of a merge.
* **Aug 18, 3:19 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1902).

PR #1906, opened 2026-08-11
Recover failed transaction signatures on a best-effort basis

mcintyre94:
#### Problem

When a `TransactionPlanExecutor` throws, we catch that error and return it as `failedSingleTransactionPlanResult`. This allows a `TransactionPlanResult` to include all errors from all transactions that were attempted.

In this block though, we use `getSignatureFromTransaction` on whatever transaction made it to the context at the point the error was thrown. There is no guarantee that this transaction has a signature, or is valid. Therefore `getSignatureFromTransaction` may throw. In this case, currently we will throw eg a fee payer missing error from the catch block of `traverseSingle`, and  lose both the original caught error and the coalescing of errors across the plan.

#### Summary of Changes

This PR makes reading the signature best effort, by wrapping it with a try-catch and swallowing any error thrown. We assume that the error originally thrown by the executor and caught is the most useful error to return.

This intentionally covers all errors, not just a missing fee payer. In this branch something has gone wrong with the executor and we make no assumptions about the validity of the transaction it has written to context.

Notes:

- The `transaction` is still available on context regardless, if it was written there
- This does not use the `getSignatureFromTransactionIfPresent` helper in a draft PR, because we intentionally swallow all errors, not just the fee payer signature missing one.
- This is related to the planned work to enable executors to return partially signed transactions in general, but this PR specifically is just fixing an existing bug on main.

### DIFF
--- a/packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
+++ b/packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
@@ -1,5 +1,6 @@
 import '@solana/test-matchers/toBeFrozenObject';
 
+import { Address } from '@solana/addresses';
 import {
     SOLANA_ERROR__INSTRUCTION_ERROR__INVALID_ARGUMENT,
     SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN,
@@ -9,6 +10,7 @@ import {
 } from '@solana/errors';
 import { Signature } from '@solana/keys';
 import { TransactionMessage, TransactionMessageWithFeePayer } from '@solana/transaction-messages';
+import { Transaction } from '@solana/transactions';
 
 import {
     canceledSingleTransactionPlanResult,
@@ -25,7 +27,7 @@ import {
     successfulSingleTransactionPlanResultFromTransaction,
     TransactionPlanResult,
 } from '../index';
-import { createMessage, createTransaction, FOREVER_PROMISE } from './__setup__';
+import { createMessage, createPartiallySignedTransaction, createTransaction, FOREVER_PROMISE } from './__setup__';
 
 jest.useFakeTimers();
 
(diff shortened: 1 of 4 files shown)


mcintyre94:
* **#1906** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1906?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1906?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -218,11 +218,21 @@ async function traverseSingle<TContext extends TransactionPlanResultContext>(
             : successfulSingleTransactionPlanResultFromTransaction(transactionPlan.message, result, context);
     } catch (error) {
         traverseConfig.canceled = true;
-        const contextWithSignature =
-            'transaction' in context && typeof context.transaction === 'object' && context.signature == null
-                ? { ...context, signature: getSignatureFromTransaction(context.transaction) }
-                : context;
-        return failedSingleTransactionPlanResult(transactionPlan.message, error as Error, contextWithSignature);
+        // The callback may have stored a transaction on the context before throwing. If that
+        // transaction was signed by its fee payer, we can recover its signature. We may not be able to though, if it doesn't have a fee payer signature or is malformed in some way. In these cases we catch and ignore that error, maintaining the original error instead. We leave the `signature` property absent.
```
trevor-cortex: Nit: the second line of this comment is one very long line (~300 chars) while the first wraps at the usual width. Prettier preserves comment prose, so it won't get fixed automatically — consider reflowing it to match the surrounding comment style.

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -218,11 +218,21 @@ async function traverseSingle<TContext extends TransactionPlanResultContext>(
...
+            try {
+                signature = getSignatureFromTransaction(context.transaction);
+            } catch {
+                /* Swallowed on purpose. The error we are handling takes precedence. */
+            }
+        }
+        return failedSingleTransactionPlanResult(
+            transactionPlan.message,
+            error as Error,
+            signature ? { ...context, signature } : context,
```
trevor-cortex: Nit: `signature !== undefined` would express the intent slightly more precisely than truthiness. Behaviorally identical today (a base58-encoded 64-byte signature can never be the empty string), so feel free to ignore.

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
```diff
@@ -299,6 +301,54 @@ describe('createTransactionPlanExecutor', () => {
...
+        });
+
+        it('does not add a signature to a failed context when the stored transaction is malformed', async () => {
+            expect.assertions(1);
+            const messageA = createMessage('A');
+            // A transaction whose fee payer signature slot holds something that is not signature bytes.
+            const transactionA = Object.freeze({
+                id: 'A',
+                signatures: { ['' as Address]: 'not signature bytes' },
+            }) as unknown as Transaction;
```
trevor-cortex, line 335: This test's failure mode is implementation-coupled: it relies on the base58 decoder throwing when given a plain string (strings lack the array methods the decoder uses). That's fine as a representative "any malformed transaction" case, but a short note in the comment about *why* this throws would help whoever debugs it if the codecs internals ever change and this stops throwing.

trevor-cortex (commented):
This PR fixes a real bug in `traverseSingle`'s error path: when the `executeTransactionMessage` callback stored a transaction on the context and then threw, the executor called `getSignatureFromTransaction` on that transaction while building the failed result. If the fee payer hadn't signed (or the transaction was malformed), that call threw from inside the catch block — replacing the original error and destroying the partial `TransactionPlanResult`. The fix wraps the signature recovery in a try-catch and deliberately swallows any error, leaving `signature` absent from the failed context while the original error propagates.

The approach is right. Swallowing *all* errors (not just `SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING`) is the correct call here — as the PR description says, once the executor has thrown, no assumptions should be made about the validity of whatever it wrote to context, and the caught error is the one worth preserving. The `transaction` itself is still preserved on the context either way, so no information is lost.

Other notes:

- Both new tests cover the two distinct failure modes (fee payer signature slot empty vs. garbage in the slot), use `passthroughFailedTransactionPlanExecution` for clean assertions, and follow the repo's `expect.assertions(n)`-in-async-tests convention.
- The `createPartiallySignedTransaction` setup helper is well-documented and mirrors the existing `createTransaction` shape.
- Changeset is present, patch bump for `@solana/instruction-plans` is correct for a bug fix, and the CLI-style filename suggests it was generated properly.

For subsequent reviewers: worth being aware that the "malformed transaction" test depends on the base58 decoder throwing when handed a plain string instead of a `Uint8Array` — that's an implementation detail of the codecs internals, so if the decoder ever became more permissive this test could start failing for non-obvious reasons. It's a reasonable representative case for "any error is swallowed" today, just something to keep in mind if it ever breaks.

Only nits inline. LGTM — would approve, but I don't have approval permissions on this repo, so submitting as a comment.

packages/instruction-plans/src/transaction-plan-executor.ts
mcintyre94: Forgot Prettier doesn't fix these! Fixed

packages/instruction-plans/src/transaction-plan-executor.ts
mcintyre94: Agreed, updated

packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
mcintyre94: The test file doesn't currently have any mocks so I left it as-is for now but added a comment.

mcintyre94:
Closing - superseded by #1913

PR #1907, opened 2026-08-11
Stop inflating every release to a major via the `@solana/react` peer dependency

mcintyre94:
TLDR: I noticed the open changesets PR #1830 has marked everything as 8.0 despite only minor/patch changesets. Fix and description below is all Claude 

#### Problem

The pending release PR (#1830) calculates the next version as **8.0.0**, even though every changeset on `main` is `minor` or `patch`. The expected version is 7.1.0.

The cause is a rule in `@changesets/assemble-release-plan` ([`shouldBumpMajor`](https://github.com/changesets/changesets/blob/main/packages/assemble-release-plan/src/determine-dependents.ts)): when a package receives a `minor` or `major` bump, any dependent that lists it in `peerDependencies` is bumped `major`, on the theory that tightening a peer range is breaking for consumers. `@solana/react` declares `@solana/kit` as a peer, `@solana/kit` has minor changesets, so `@solana/react` is forced to `major` — and because every publishable package shares one `fixed` group, that major propagates to all 48 packages.

`changeset status` on `main` today reports all 48 packages as `major`, most with an empty `changesets: []`.

#### Summary of Changes

Two changes, both required — either one alone still produces 8.0.0:

1. **`.changeset/config.json`** opts into `___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH.onlyUpdatePeerDependentsWhenOutOfRange`, so a peer dependent is bumped `major` only when the new version actually leaves its declared range, rather than unconditionally.
2. **`packages/react/package.json`** changes the `@solana/kit` peer range from `workspace:*` to `workspace:^`. This is necessary because changesets resolves `workspace:*` to the *exact* current version, so any bump at all leaves the range even with the flag enabled.

Plus one piece of documentation:

3. **`CLAUDE.md`** documents that a `@changesets/cli` upgrade should be reviewed by re-running `changeset status`, since the flag is explicitly unstable — it lives under `___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH`, and the dependency is declared as `^2.31.1`, so a routine Dependabot bump could change or drop the behaviour without any signal. Catching it at review time on the upgrade PR is better than discovering it on a release PR.

The caret range is also the more correct thing to publish, independent of the version-inflation issue. `workspace:*` publishes as an exact pin — `@solana/react@7.0.0` on npm today declares `"@solana/kit": "7.0.0"` — which is a semver-breaking declaration on every release by construction, since no one on a different `@solana/kit` version can satisfy it. In practice it means a consumer who advances `@solana/kit` without advancing `@solana/react` in the same step gets an unsatisfiable peer range even though the two are compatible. The lockstep release guarantee comes from the `fixed` group in the changesets config, not from the peer range, so nothing about which versions ship together changes. It also brings `@solana/kit` in line with the package's other peers (`^5.0.0`, `>=18`, `^2.4.2`), none of which are pinned exact.

Verified locally:

| Configuration | Calculated release |
| --- | --- |
| `main` as-is | 8.0.0 (48 × `major`) |
| flag only | 8.0.0 |
| `workspace:^` only | 8.0.0 |
| **both (this PR)** | **7.1.0 (48 × `minor`, 0 majors)** |

Also confirmed by packing a throwaway pnpm workspace that `workspace:^` publishes as `^7.1.0` while `workspace:*` publishes as the exact `7.1.0`, and that `pnpm install --lockfile-only` produces no lockfile change (`@solana/kit` is already a devDependency of `@solana/react`, so the peer specifier is not separately recorded).

Once this lands on `main`, the changesets action will regenerate #1830 as 7.1.0.

### DIFF
--- a/.changeset/brown-deserts-shake.md
+++ b/.changeset/brown-deserts-shake.md
@@ -1,0 +1,5 @@
+---
+'@solana/react': patch
+---
+
+Widen the `@solana/kit` peer dependency of `@solana/react` from an exact version to a caret range. `@solana/react` previously declared `"@solana/kit": "workspace:*"`, which publishes as an exact pin (`"@solana/kit": "7.0.0"`), so a consumer who advanced `@solana/kit` without advancing `@solana/react` in the same step hit an unsatisfiable peer range even though the two are compatible. It now declares `workspace:^` and publishes as `^7.1.0`. The two packages continue to be released in lockstep at identical versions, so this does not loosen which combinations are actually shipped — it only stops describing a compatible pair as incompatible.
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -18,5 +18,8 @@
     "version": false,
     "tag": false
   },
-  "updateInternalDependencies": "patch"
+  "updateInternalDependencies": "patch",
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
+    "onlyUpdatePeerDependentsWhenOutOfRange": true
+  }
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -94,6 +94,7 @@ RPC method response types live in `packages/rpc-api/src/<methodName>.ts`; types 
 - `patch` for fixes, `minor` for features, `major` for breaking changes (while pre-1.0, `minor` may be treated as breaking).
 - No changeset needed for docs-only changes, CI config, dev dependency updates, or test-only changes.
 - All publishable packages are version-locked via the `fixed` config in `.changeset/config.json`.
+- **Re-check the calculated version after any `@changesets/cli` bump.** Keeping releases off a spurious major depends on `___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH.onlyUpdatePeerDependentsWhenOutOfRange` in `.changeset/config.json`, and that option is explicitly unstable — the name promises it may change in a patch. Since `@changesets/cli` is declared as `^2.31.1`, a routine Dependabot bump could alter or drop the behaviour. When reviewing a PR that upgrades `@changesets/cli`, run `npx changeset status` and confirm the calculated versions still match the highest bump among the pending changesets. Without the option, changesets bumps any package to `major` when a package it lists in `peerDependencies` gets a minor — `@solana/react` peer-depends on `@solana/kit`, so that alone would push all packages in the `fixed` group to a major. The peer range must also stay `workspace:^` rather than `workspace:*`, since changesets resolves `workspace:*` to an exact version that every bump leaves.
 
 ## CI
 
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -147,7 +147,7 @@
         "swr": "^2.4.2"
     },
     "peerDependencies": {
-        "@solana/kit": "workspace:*",
+        "@solana/kit": "workspace:^",
         "@tanstack/react-query": "^5.0.0",
         "react": ">=18",
         "swr": "^2.4.2"


mcintyre94:
@trevor-cortex

.changeset/config.json
```diff
@@ -18,5 +18,8 @@
     "version": false,
     "tag": false
   },
-  "updateInternalDependencies": "patch"
+  "updateInternalDependencies": "patch",
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
+    "onlyUpdatePeerDependentsWhenOutOfRange": true
+  }
```
trevor-cortex, line 24: Since this lives under `___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH` and the root `package.json` takes `@changesets/cli` as `^2.31.1`, a routine patch/minor bump of changesets could legitimately change this flag's semantics. Two cheap mitigations worth considering (not blocking): document the dependency between this flag and correct version calculation in `CLAUDE.md`'s Changesets section, and make it a habit to run `changeset status` after any `@changesets/*` upgrade before merging.

.changeset/brown-deserts-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solana/react': patch
+---
+
+Widen the `@solana/kit` peer dependency of `@solana/react` from an exact version to a caret range. `@solana/react` previously declared `"@solana/kit": "workspace:*"`, which publishes as an exact pin (`"@solana/kit": "7.0.0"`), so a consumer who advanced `@solana/kit` without advancing `@solana/react` in the same step hit an unsatisfiable peer range even though the two are compatible. It now declares `workspace:^` and publishes as `^7.1.0`. The two packages continue to be released in lockstep at identical versions, so this does not loosen which combinations are actually shipped — it only stops describing a compatible pair as incompatible.
```
trevor-cortex, line 5: Tiny nit: this hard-codes `^7.1.0` as the published range. If anything shifts the next version before release (e.g. a major landing first), the changelog entry would name the wrong version. Phrasing it version-agnostically ("publishes as a caret range on the release version") would keep it accurate regardless. Fine to leave as-is if 7.1.0 is imminent.

trevor-cortex (commented):
This PR fixes version inflation where every release was being computed as a major: `@solana/kit` receiving any `minor`/`major` bump forced its peer dependent `@solana/react` to `major` (changesets' `shouldBumpMajor` rule), which then propagated to all 48 packages via the shared `fixed` group. The fix is two-part — the `onlyUpdatePeerDependentsWhenOutOfRange` experimental flag plus widening the peer range from `workspace:*` to `workspace:^` — and the description correctly explains why both are required (with `workspace:*` resolving to an exact version, any bump leaves the range even with the flag on).

The analysis checks out — this would be an approval if I had the repository access to submit one:

- `workspace:^` publishes as `^<version>` (pnpm rewrites at pack time), so a `minor`/`patch` bump of `@solana/kit` stays inside the declared range and no longer triggers the dependent-major rule. A genuine `@solana/kit` major would still leave the range and correctly major-bump `@solana/react` — which the `fixed` group would do anyway.
- The caret range is the more honest published contract regardless: the lockstep guarantee comes from the `fixed` group in `.changeset/config.json`, not from an exact peer pin, and exact pins make any skewed-but-compatible install unsatisfiable. It also matches the style of the package's other peers.
- `bumpVersionsWithWorkspaceProtocolOnly: true` is unaffected since `workspace:^` is still the workspace protocol, and the `@solana/kit` devDependency staying `workspace:*` is fine because `prepublishOnly` strips devDependencies.
- Changeset is `patch` on `@solana/react` only, generated via the CLI naming scheme, and the config change itself correctly ships no changeset. All consistent with the repo's conventions.

The local verification matrix (both changes required, 7.1.0 with 0 majors, no lockfile churn) is exactly the right evidence for this kind of change.

**Things to watch / notes for future reviewers:**

1. **The flag is explicitly unstable.** It lives under `___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH`, and the root `package.json` declares `@changesets/cli` with a caret range (`^2.31.1`), so a routine dependency bump could silently change or remove this behavior. Worth keeping an eye on release PRs after any changesets upgrade — if #1830 ever snaps back to a surprise major, this flag is the first suspect. Consider documenting this in the "Changesets & Releases" section of `CLAUDE.md` so a future changesets bump doesn't get merged without re-checking `changeset status`.
2. **The fix is only as durable as the peer ranges.** If a future package adds a `workspace:*` (or otherwise exact) peer dep on a fixed-group sibling, the all-majors inflation quietly returns. A cheap guard would be a CI check (or lint rule) asserting that intra-workspace `peerDependencies` use `workspace:^`, or simply asserting `changeset status` contains no unexplained majors. Not blocking for this PR.
3. After merge, verify the regenerated #1830 actually lands on 7.1.0 as expected — that's the end-to-end confirmation of the local verification.

.changeset/config.json
mcintyre94: Added to CLAUDE.md. Unfortunately this is most likely to be updated by a changeset PR without a claude review, but still useful to check if it's updated anywhere else.

.changeset/brown-deserts-shake.md
mcintyre94: This is fine, it's just the changeset, and I'm confident we'll release a minor before v8.

lorisleiva (approved):

PR #1910, opened 2026-08-12
Let `TContext` decide what a result context contains

mcintyre94:
This PR takes a new approach to enabling the `signTransaction` executor, by removing the guarantee that `TContext` on a `SuccessfulSingleTransactionPlanResult` extends `SuccessfulBaseTransactionPlanResultContext`, which enforces `signature: Signature`. The goal is to be able to define an executor that does not return a signature, nor a transaction from which one can be obtained.

We deprecate `SuccessfulBaseTransactionPlanResultContext`, but duplicate it as a new type `TransactionPlanResultContextWithSignature`.

The `TContext` on `SuccessfulSingleTransactionPlanResult` is now simply `TransactionPlanResultContext` (a dictionary with no particular keys), but defaults to `TransactionPlanResultContextWithSignature` for backward compatibility. That is, uses that do not define their own `TContext` type can continue to rely on `result.context.signature` being defined.

However, uses that do define a context type, eg. `{startedAt: number}`, will now need to fully define the context, eg. by intersecting their type `TransactionPlanResultContextWithSignature & { startedAt: number }`, or just including what they want, `{ signature: Signature, startedAt: number }`.

Alongside this, the types are made more sound:

The executor callback now receives `Partial<TContext>`, ie the context starts empty and it's up to the callback to populate it. Previously an executor trying to read properties of its context that it hasn't written would typecheck.

Failed and Canceled results now type their context as `Readonly<Partial<TContext>>`. The executor can throw or be cancelled at any time, and there are no guarantees which parts of context will have been written. Previously reading context fields from a failed/canceled result would typecheck.

Additionally, a bunch of helpers over `TransactionPlanResult`, eg. `passthroughFailedTransactionPlanExecution`, gain a `TContext` parameter. Previously they didn't have this generic and so implicitly required the with-signature context. 

This PR **does not** include runtime changes to enable an executor to return a transaction with no signature, that is planned for a following PR. This focuses on the types. We still auto-populate `context.signature` on the success path in this PR.

This PR:

- Supersedes https://github.com/anza-xyz/kit/pull/1893

The follow-up PR would:

- Supersede #1894 and #1906
- Become the new parent of #1899 and #1902 

Updated: This PR no longer updates docs, this is moved to #1912 - we'll update docs properly using the new types after publishing.

### DIFF
--- a/packages/instruction-plans/src/transaction-plan-executor.ts
+++ b/packages/instruction-plans/src/transaction-plan-executor.ts
@@ -24,11 +24,11 @@ import {
     parallelTransactionPlanResult,
     sequentialTransactionPlanResult,
     SingleTransactionPlanResult,
-    type SuccessfulBaseTransactionPlanResultContext,
     successfulSingleTransactionPlanResult,
     successfulSingleTransactionPlanResultFromTransaction,
     type TransactionPlanResult,
     type TransactionPlanResultContext,
+    type TransactionPlanResultContextWithSignature,
 } from './transaction-plan-result';
 
 /**
@@ -36,6 +36,10 @@ import {
  *
  * This function traverses the transaction plan tree, executing each transaction
  * message and collecting results that mirror the structure of the original plan.
+ *
+ * The zero-argument spelling defaults `TContext` to {@link TransactionPlanResultContextWithSignature},
+ * so its results guarantee a `context.signature` on every successful single result — but a different
+ * `TContext` may drop that guarantee entirely.
  *
  * @typeParam TContext - The type of the context object that may be passed along with results.
  * @param transactionPlan - The transaction plan to execute.
@@ -46,19 +50,18 @@ import {
  * @see {@link TransactionPlanResult}
  * @see {@link createTransactionPlanExecutor}
  */
-export type TransactionPlanExecutor<TContext extends TransactionPlanResultContext = TransactionPlanResultContext> = (
+export type TransactionPlanExecutor<
+    TContext extends TransactionPlanResultContext = TransactionPlanResultContextWithSignature,
+> = (
     transactionPlan: TransactionPlan,
     config?: { abortSignal?: AbortSignal },
 ) => Promise<TransactionPlanResult<TContext>>;
 
-type SuccessfulTransactionPlanResultContext<TContext extends TransactionPlanResultContext> =
-    SuccessfulBaseTransactionPlanResultContext & TContext;
-
 type ExecuteTransactionMessage<
     TContext extends TransactionPlanResultContext,
-    TReturn = Signature | SuccessfulTransactionPlanResultContext<TContext> | Transaction,
+    TReturn = Signature | TContext | Transaction,
 > = (
-    context: BaseTransactionPlanResultContext & TContext,
+    context: Partial<TContext>,
     transactionMessage: TransactionMessage & TransactionMessageWithFeePayer,
     config?: { abortSignal?: AbortSignal },
 ) => Promise<TReturn>;
@@ -74,8 +77,8 @@ export type TransactionPlanExecutorConfig<
     /**
      * Called whenever a transaction message must be sent to the blockchain.
      *
-     * It should return the context that the successful result must carry — which, since a
-     * successful result always has a signature, must include one. Returning a {@link Signature} or
+     * It should return the context that the successful result must carry — every property
+     * `TContext` promises, which by default includes a `signature`. Returning a {@link Signature} or
      * a {@link Transaction} instead is deprecated.
      */
     executeTransactionMessage: ExecuteTransactionMessage<TContext>;
@@ -105,7 +108,7 @@ export type TransactionPlanExecutorConfig<
  * @see {@link TransactionPlanExecutorConfig}
  */
 export function createTransactionPlanExecutor<
-    TContext extends TransactionPlanResultContext = TransactionPlanResultContext,
+    TContext extends TransactionPlanResultContext = TransactionPlanResultContextWithSignature,
 >(config: {
     executeTransactionMessage: ExecuteTransactionMessage<TContext, Signature | Transaction>;
 }): TransactionPlanExecutor<TContext>;
@@ -124,9 +127,9 @@ export function createTransactionPlanExecutor<
  * saved to the context will still be available in the plan result, which can be useful
  * for debugging failures or building recovery plans.
  *
- * The callback should return the context that a successful result must carry. Since a successful
- * result always has a signature, that context must include one — the executor derives nothing on
- * the callback's behalf. On success the returned context is merged over the one the callback
+ * The callback should return the context that a successful result must carry — every property
+ * `TContext` promises, which by default includes a `signature`, since the executor derives nothing
+ * on the callback's behalf. On success the returned context is merged over the one the callback
  * mutated, with the returned value taking precedence, so a property stored on the context but left
  * out of the return value is still reported.
  *
@@ -141,12 +144,7 @@ export function createTransactionPlanExecutor<
  * ```
  *
  * Note that the callback cannot simply return the context it was given, since every property on it
- * is optional. Build the return value from the values you have instead. Custom context properties
- * are typed by supplying `TContext` explicitly:
- *
- * ```ts
- * createTransactionPlanExecutor<{ startedAt: number }>(config);
- * ```
+ * is optional. Build the return value from the values you have instead.
  *
  * Returning a {@link Signature} or a full {@link Transaction} object instead of a context is
  * deprecated. Those return values are still honoured — a returned signature is stored as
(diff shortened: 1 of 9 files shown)


mcintyre94:
* **#1910** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1910?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1910?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/content/docs/advanced-guides/instruction-plans.mdx
```diff
@@ -483,18 +483,45 @@ const transactionPlanExecutor = createTransactionPlanExecutor({
...
+        // ...
+        return {} as Signature;
+    },
+});
+```
+
 When accessing the context from a result, you can retrieve both the base properties and your custom properties:
 
-```ts twoslash
+```ts
```
mcintyre94: Need to remove `twoslash` temporarily because of the new type

Probably something we should do after each release is have Claude go and check if any of our `ts` blocks can have twoslash

mcintyre94:
@trevor-cortex

docs/content/docs/advanced-guides/instruction-plans.mdx
```diff
@@ -483,18 +483,45 @@ const transactionPlanExecutor = createTransactionPlanExecutor({
...
+it and every property on it — base or custom — is optional. That means your callback cannot annotate
+its `context` parameter with required properties. If you want a custom property to be guaranteed on
+the result, declare it as an explicit type argument to `createTransactionPlanExecutor` instead:
+
+```ts
+import {
+    createTransactionPlanExecutor,
+    Signature,
+    TransactionPlanResultContextWithSignature,
+} from '@solana/kit';
```
trevor-cortex: This block is plain ` ```ts ` rather than ` ```ts twoslash `, so the `// ---cut-before---` marker and the setup imports above it will render literally in the published docs instead of being stripped. The same applies to the block below (~L511), which was downgraded from `ts twoslash` to `ts` while keeping its cut marker at ~L521.

Was twoslash dropped because these snippets fail type-checking against the docs' resolved `@solana/kit` (the new `TransactionPlanResultContextWithSignature` export)? If so it'd be good to note why; otherwise restoring `twoslash` seems preferable — it keeps the snippets verified and the cut markers functional. If it has to stay plain `ts`, the markers and hidden setup imports should be removed so readers don't see them.

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -213,16 +243,33 @@ async function traverseSingle<TContext extends TransactionPlanResultContext>(
...
     } catch (error) {
         traverseConfig.canceled = true;
+        // `TContext` no longer promises that a stored transaction is a `Transaction`, so this
+        // reads it back through the base context's shape before narrowing it at runtime.
+        const storedTransaction = context.transaction as BaseTransactionPlanResultContext['transaction'];
         const contextWithSignature =
-            'transaction' in context && typeof context.transaction === 'object' && context.signature == null
-                ? { ...context, signature: getSignatureFromTransaction(context.transaction) }
+            'transaction' in context && typeof storedTransaction === 'object' && context.signature == null
+                ? { ...context, signature: getSignatureFromTransaction(storedTransaction) }
```
trevor-cortex, line 348: Now that `TContext` no longer guarantees that the `transaction` key holds a `Transaction`, this call can throw for an executor whose custom context stores something else under that key — `typeof storedTransaction === 'object'` passes for any object (and for `null`, though that part is pre-existing), then `getSignatureFromTransaction` throws on anything without a valid signatures map. Since we're already inside the `catch`, that thrown error would *replace* the original failure, which is exactly the situation the failed-result path exists to preserve.

`getSignatureFromContext` in `transaction-plan-errors.ts` takes the defensive route for the same reason ("this narrows at runtime rather than trusting the type"). Consider mirroring that here: either a stronger structural check before calling, or wrapping the recovery in a `try/catch` that falls back to the bare `context` so signature recovery can never mask the real error.

packages/instruction-plans/src/transaction-plan-result.ts
```diff
@@ -498,14 +526,14 @@ export function parallelTransactionPlanResult<
...
+    TContext extends TransactionPlanResultContext = TransactionPlanResultContextWithSignature,
     TTransactionMessage extends TransactionMessage & TransactionMessageWithFeePayer = TransactionMessage &
         TransactionMessageWithFeePayer,
 >(
     plannedMessage: TTransactionMessage,
     transaction: Transaction,
-    context?: Omit<BaseTransactionPlanResultContext, 'signature' | 'transaction'> & TContext,
-): SuccessfulSingleTransactionPlanResult<TContext, TTransactionMessage> {
+    context?: TContext,
+): SuccessfulSingleTransactionPlanResult<TContext & TransactionPlanResultContextWithSignature, TTransactionMessage> {
```
trevor-cortex: The return type intersects the full `TransactionPlanResultContextWithSignature`, in which `transaction` is *optional* — yet the runtime unconditionally stores the required `transaction` argument in the context. So the type claims less than the function delivers, and it's slightly at odds with both the `@param` docblock above ("plus the `signature` and `transaction` derived from the `transaction` argument") and the changeset, which says both fields are "intersected into the result's context type".

Consider `SuccessfulSingleTransactionPlanResult<TContext & { signature: Signature; transaction: Transaction }, TTransactionMessage>` instead. That makes `transaction` required to match runtime, and as a bonus avoids re-adding the index signature and optional `message` from the full `WithSignature` type — which the current intersection quietly reintroduces despite the docblock's "claims exactly what you pass" framing. (Callers needing assignability to default-context results still get it, since `signature` is the only required field there.)

trevor-cortex (commented):
This PR removes the hardcoded `SuccessfulBaseTransactionPlanResultContext & TContext` intersection from the result types and makes `TContext` the sole authority on what a context contains, with `TransactionPlanResultContextWithSignature` as the new default everywhere to preserve zero-type-argument behaviour. Alongside that it makes the types more honest: the executor callback receives `Partial<TContext>` (the context starts empty), failed/canceled results type their context as `Readonly<Partial<TContext>>`, and the error/passthrough helpers gain a `TContext` parameter instead of implicitly demanding the with-signature context.

The approach is sound and the execution is thorough. The typetests are excellent — the `Eq` mutual-assignability checks on the zero-argument result shapes are a particularly good guard against future regressions in the defaults, and the changeset does a great job walking through every observable typing change. The two `as unknown as TContext` casts on the happy path are clearly commented and genuinely unavoidable given the design.

**Things to look at:**

1. **`TransactionPlanExecutorConfig` default divergence** (`transaction-plan-executor.ts` ~L71–77, unchanged lines so noting it here): it still defaults `TContext` to `TransactionPlanResultContext`, while `TransactionPlanExecutor`, `createTransactionPlanExecutor`, and every result type now default to `TransactionPlanResultContextWithSignature`. A standalone `const config: TransactionPlanExecutorConfig = {...}` passed to `createTransactionPlanExecutor(config)` will infer a no-guarantee executor, silently diverging from the inline spelling of the same code. If deliberate (the config is an input type, so the loose default is arguably "correct"), a short comment would help; otherwise aligning it would avoid surprises.

2. **Docs blocks demoted from `ts twoslash` to `ts`** — this leaves `// ---cut-before---` markers and setup imports rendering literally in the published docs. Inline comment below.

3. **Runtime hazard in the executor's catch branch** now that `TContext` no longer constrains what lives under the `transaction` key. Inline comment below.

4. **`successfulSingleTransactionPlanResultFromTransaction` return type** claims less than the runtime delivers (`transaction` stays optional). Inline comment below.

**For subsequent reviewers:** worth verifying that nothing else in the monorepo (and in the external `anza-xyz/kit-plugins` repo, which builds on these packages) parameterises these result types with a bare custom context and relies on the previously-implicit `signature`/`transaction` fields — the migration burden described in the changeset applies to those consumers too. Also worth confirming the follow-up PR (runtime changes for signature-less executors) revisits the `context.signature` auto-population in `traverseSingle`, since this PR intentionally keeps it.

packages/instruction-plans/src/transaction-plan-executor.ts
mcintyre94: Intentionally deferring this to the follow-up PR. Any change here would be removed in that PR, which will stop adding signature at all.

packages/instruction-plans/src/transaction-plan-result.ts
mcintyre94: Agreed, return type now uses `TContext & { signature: Signature; transaction: Transaction }`

docs/content/docs/advanced-guides/instruction-plans.mdx
mcintyre94: I've removed the docs change from this PR, we'll do it after publishing now to avoid this issue.

packages/instruction-plans/src/transaction-plan-result.ts
```diff
@@ -116,13 +116,40 @@ export interface BaseTransactionPlanResultContext {
...
+ *         context.startedAt = Date.now();
+ *         // ...
+ *     },
+ * });
+ * ```
+ *
+ * @see {@link TransactionPlanResultContext}
+ */
+export type TransactionPlanResultContextWithSignature = BaseTransactionPlanResultContext &
+    TransactionPlanResultContext & { signature: Signature };
```
lorisleiva: I could be wrong but I don't think we need `TransactionPlanResultContext` to be explicitly intersected here since it's just defined as:

```ts
type TransactionPlanResultContext = { [key: number | string | symbol]: unknown }
```

So we only need it as some slightly more explicit `object` so we can use it as a type parameter constraint.

packages/instruction-plans/src/transaction-plan-result.ts
lorisleiva, line 122: Could we deprecate this type now that we have the default `TransactionPlanResultContextWithSignature`?

Then we can simplify `TransactionPlanResultContextWithSignature` to be:

```ts
export interface TransactionPlanResultContextWithSignature {
    signature: Signature;
    message?: TransactionMessage & TransactionMessageWithFeePayer;
    transaction?: Transaction;
}
```

That way we can just kill it in v8 (unless it's still useful somewhere else?).

packages/instruction-plans/src/transaction-plan-executor.ts
lorisleiva: Was about to make the same comment haha.

lorisleiva (approved):
Thanks, nice refactoring! Just a couple of comments/questions.

packages/instruction-plans/src/transaction-plan-result.ts
mcintyre94: Agreed - refactored to not use `BaseTransactionPlanResultContext` for `TransactionPlanResultContextWithSignature` here. I've removed it in the following PR just to avoid conflicts with the other code that one removes.

mcintyre94:
### Merge activity

* **Aug 18, 3:02 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1910).
* **Aug 18, 3:02 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1910).

PR #1911, opened 2026-08-12
Fix getBitArrayEncoder returning the wrong next offset

latent-9:
#### Problem

`getBitArrayEncoder`'s `write` returns `size` instead of `offset + size`. Every other codec's `write` returns `offset` plus the number of bytes written, and the bit-array decoder itself returns `offset + size`, so the encoder is the only outlier.

Because `getStructEncoder`/`getTupleEncoder` chain fields with `offset = codec.write(value, bytes, offset)`, a bit array that is not the last field makes the next field write at `size` instead of `offset + size`, overwriting the bit array.

Example: a struct `[['a', u8], ['b', bitArray(1)], ['c', u8]]` with `{ a: 0x11, b: 0xa0, c: 0x22 }` encodes to `0x112200` instead of `0x11a022`.

The existing bit-array tests only call `.encode()` at offset 0, where `size === offset + size`, so this is invisible in isolation and only shows up under composition.

#### Summary of Changes

`write` now returns `offset + size`, matching the decoder and the other codecs. Added a test that writes a bit array at a non-zero offset, plus a changeset.

### DIFF
--- a/.changeset/fix-bit-array-encoder-offset.md
+++ b/.changeset/fix-bit-array-encoder-offset.md
@@ -1,0 +1,5 @@
+---
+'@solana/codecs-data-structures': patch
+---
+
+Fix `getBitArrayEncoder` returning the wrong next offset. Its `write` returned `size` instead of `offset + size`, so a bit array placed before another field in a struct or tuple was overwritten by the following field. It now returns `offset + size`, matching the decoder and the other codecs.
--- a/packages/codecs-data-structures/src/__tests__/bit-array-test.ts
+++ b/packages/codecs-data-structures/src/__tests__/bit-array-test.ts
@@ -70,6 +70,16 @@ describe('getBitArrayCodec', () => {
         );
     });
 
+    it('writes at a non-zero offset and returns the next offset', () => {
+        const a = (bits: string) => [...bits].map(bit => bit === '1');
+        // Writing at a non-zero offset must place the bytes at `offset` and
+        // return `offset + size`, so a following field is written after it
+        // rather than overwriting the bit array.
+        const bytes = new Uint8Array(2);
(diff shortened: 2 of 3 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Fixes `getBitArrayEncoder`'s `write` returning `size` instead of `offset + size`. Since `getStructEncoder`/`getTupleEncoder` chain fields via `offset = codec.write(value, bytes, offset)`, any field following a bit array at a non-zero offset would rewind and clobber it. The fix brings the encoder in line with the decoder (which already returns `offset + size`) and every other codec in the family.

## Verification

- Confirmed the decoder's `read` returns `[booleans, offset + size]`, so this was genuinely an encoder-only inconsistency.
- The new test exercises exactly the failure mode: writing at offset 1 into a 2-byte buffer, asserting both the returned offset (`2`) and byte placement (`0x00a0`). The existing tests only went through `.encode()` at offset 0 where the bug is invisible, so this is a meaningful coverage addition rather than a redundant one.
- Changeset is present with a `patch` bump on `@solana/codecs-data-structures`, which is right for a behavioral bug fix.

## Notes for other reviewers

Nothing to flag — this is a minimal, well-scoped one-line fix with a targeted regression test. One thing worth knowing for downstream consumers: any code that was compensating for the old return value (e.g. manually tracking offsets after a bit-array `write`) would double-count after this fix, but the old behavior was unambiguously wrong per the codec contract.

*(Submitting as a comment rather than an approval — I don't have approval permissions on this repo, but this has my LGTM.)*

mcintyre94 (approved):
Thankyou!

PR #1913, opened 2026-08-12
Stop writing to the execution context in `createTransactionPlanExecutor`

mcintyre94:
This PR follows from #1910 

With this PR we no longer write anything to the context in `createTransactionPlanExecutor`. Just like the new underlying types, the caller is now in complete control of the context.

`executeTransactionMessage` is now `Promise<TContext>`. Previously it returned a `Transaction`, which we added (along with a derived context) to `signature`. As with the previous PR, now we allow the executor to completely control what `TContext` is and to return it. We no longer make any changes to what the executor returns.

Note that an alternative design considered was to return `Promise<void>`, but this loses type safety. Our success case is typed with `TContext`, but nothing in the type system forces an executor to write what it promises to, to context. The mutable `context` is only for failed/cancelled cases, which are typed `Partial<TContext>`. This would be a less dramatic breaking change/migration, but I think the type safety is more valuable. 

This refactor means that we no longer derive the signature from a returned transaction, and we also drop the PR that refactored that to improve safety. If a `signature` is not on `TContext` then it won't be available in context, and if it is on `TContext` then the executor must return it itself. Similarly, the catch handler no longer derives a signature from the transaction in context.

Supersedes: #1894 

Supersedes: [#1906](/github/pr-or-issue/anza-xyz/kit/1906) 

Note that if we decide this is the correct approach, I will open a PR that makes it *possible* to return `TContext` but still allows signature and transaction (ideally deprecated). This will make migration easier, and in particular I'll update our plugins to return context before we move to v8.

### DIFF
--- a/packages/instruction-plans/src/transaction-plan-executor.ts
+++ b/packages/instruction-plans/src/transaction-plan-executor.ts
@@ -5,10 +5,8 @@ import {
     SOLANA_ERROR__INVARIANT_VIOLATION__INVALID_TRANSACTION_PLAN_KIND,
     SolanaError,
 } from '@solana/errors';
-import type { Signature } from '@solana/keys';
 import { getAbortablePromise } from '@solana/promises';
 import type { TransactionMessage, TransactionMessageWithFeePayer } from '@solana/transaction-messages';
-import { getSignatureFromTransaction, type Transaction } from '@solana/transactions';
 
 import type {
     ParallelTransactionPlan,
@@ -18,14 +16,12 @@ import type {
 } from './transaction-plan';
 import { createFailedToExecuteTransactionPlanError } from './transaction-plan-errors';
 import {
-    BaseTransactionPlanResultContext,
     canceledSingleTransactionPlanResult,
     failedSingleTransactionPlanResult,
     parallelTransactionPlanResult,
     sequentialTransactionPlanResult,
     SingleTransactionPlanResult,
     successfulSingleTransactionPlanResult,
-    successfulSingleTransactionPlanResultFromTransaction,
     type TransactionPlanResult,
     type TransactionPlanResultContext,
     type TransactionPlanResultContextWithSignature,
@@ -57,14 +53,11 @@ export type TransactionPlanExecutor<
     config?: { abortSignal?: AbortSignal },
 ) => Promise<TransactionPlanResult<TContext>>;
 
-type ExecuteTransactionMessage<
-    TContext extends TransactionPlanResultContext,
-    TReturn = Signature | TContext | Transaction,
-> = (
+type ExecuteTransactionMessage<TContext extends TransactionPlanResultContext> = (
     context: Partial<TContext>,
     transactionMessage: TransactionMessage & TransactionMessageWithFeePayer,
     config?: { abortSignal?: AbortSignal },
-) => Promise<TReturn>;
+) => Promise<TContext>;
 
 /**
  * Configuration object for creating a new transaction plan executor.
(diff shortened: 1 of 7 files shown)


mcintyre94:
* **#1913** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1913?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1913?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
**Verdict: LGTM** *(submitting as a comment since I don't have approve permissions on this repo)*

## Summary

This PR completes the inversion started in #1910: `createTransactionPlanExecutor` no longer writes anything to the execution context. The `executeTransactionMessage` callback now returns a complete `TContext`, which becomes the successful result's context (merged over whatever was stored on the mutable context, with the returned value winning). The two signature derivations are gone — the success path no longer calls `getSignatureFromTransaction` on a returned transaction, and the catch handler no longer derives one from a stored transaction.

This fixes two real bugs for executors producing partially signed transactions (no fee payer signature):

1. They can now **succeed** — previously `successfulSingleTransactionPlanResultFromTransaction` would throw `SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING`.
2. A failure no longer has its **original error replaced** by the signature-derivation throw in the catch handler.

Both are covered by new tests using the new `createPartiallySignedTransaction` helper, and the removal of the `as unknown as TContext` assertion on the happy path is a genuine type-safety win — the callback now proves the context is populated instead of the executor taking it on faith.

## Things to watch

- **Return-type inference is parameter-sensitivity dependent.** A zero-parameter callback gets `TContext` inferred from its return value, while any callback that declares parameters falls back to the default `TransactionPlanResultContextWithSignature`. The typetests pin this behavior explicitly (nice), but it's a subtle inference quirk that depends on TypeScript's context-sensitivity rules — worth keeping an eye on across TS version bumps, since a change in inference order would silently alter which executors compile.
- **`Promise<void>` vs `Promise<TContext>` trade-off.** The PR description's reasoning holds up: `void` would be a gentler migration but loses the one guarantee that makes `TContext` on successful results honest. The typetest `// Mutating the context does not discharge the obligation to return it` captures exactly the failure mode the return type prevents.
- **Docs are consistent.** The TSDoc on `createTransactionPlanExecutor`, the `TransactionPlanResultContextWithSignature` example, the token-airdrop example, and the changeset all tell the same story, including the "return an object built from the values you have; the mutable context itself doesn't satisfy `TContext`" caveat. The changeset's migration diff matches the actual new API.
- **Changeset bump.** `major` for `@solana/instruction-plans` matches the scale of the break (callback return type change); consistent with the majors on the preceding PRs in this series.

- The merge in `traverseSingle` (`{ ...context, ...returnedContext }`) is the one spot with non-obvious runtime semantics — see the inline note about explicitly-`undefined` returned properties.
- `successfulSingleTransactionPlanResultFromTransaction` still derives `signature`/`transaction` and still throws on missing fee payer signatures — that's intentional and called out in the changeset as the opt-in helper for the fully-signed case.
- The `return context as Context; // Never reached` casts in the failure-path tests are unreachable by construction (the callback always throws before them) and only exist to satisfy the new return type; they don't weaken what the tests assert.

mcintyre94:
Note that if we decide this is the correct approach, I will open a PR that makes it *possible* to return `TContext` but still allows signature and transaction (ideally deprecated). This will make migration easier, and in particular I'll update our plugins to return context before we move to v8.

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -237,39 +261,22 @@ async function traverseSingle<TContext extends TransactionPlanResultContext>(
...
         );
-        // Only on the happy path do we claim the context is fully populated. We cannot verify
-        // that — the callback promised these properties by way of `TContext` and we take it at
-        // its word — so this is the one place an assertion is unavoidable.
-        return typeof result === 'string'
-            ? successfulSingleTransactionPlanResult<TContext>(transactionPlan.message, {
-                  ...context,
-                  signature: result,
-              } as unknown as TContext)
-            : successfulSingleTransactionPlanResultFromTransaction<TContext>(
```
lorisleiva: Since this is a major bump and we don't seem to need `successfulSingleTransactionPlanResultFromTransaction` anymore, should we kill it?

lorisleiva (approved):
Nice, thank you!

packages/instruction-plans/src/transaction-plan-executor.ts
mcintyre94: Makes sense! Only question is whether it's useful for apps, but it's probably not, people writing custom executors/contexts probably don't want any of our opinionated helpers.

mcintyre94: Now removed in this PR

mcintyre94:
### Merge activity

* **Aug 18, 3:02 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1913).
* **Aug 18, 3:04 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1913) rebased this pull request as part of a merge.
* **Aug 18, 3:07 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit/1913).

PR #1914, opened 2026-08-12
Add client capability guards

### LINKED ISSUE #1909: Add `isX` / `assertIsX` helpers for `Client` objects
## Motivation

See https://github.com/anza-xyz/kit-plugins/pull/361#discussion_r3764610661

Many plugins include code along the lines of:

```ts
if (!client.planTransaction || !client.planTransactions) {
  throw new Error('some error message')
}
```

Other objects in Kit, like `TransactionMessage`, have a rich API of `isX` and `assertIsX` helpers to check their runtime properties + type narrow.

We should add similar helpers for clients.

## Example use case

```ts
assertIsClientWithPlanningFunctions(client)
```

## Details

It would probably make sense to use the `ClientWithX` boundaries, so test for both `planTransaction` and `planTransactions` together for instance.

These would be added to the `plugin-interfaces` package alongside these types

Would also make sense to add to the wallet plugin (which doesn't have its Client type in Kit, so this is a change in kit-plugins) 

Would throw `SolanaError` for the `assertIsX` helpers, removing a non-typed error path from most plugins.

SharkHand3:
#### Problem

Plugins currently repeat ad hoc runtime checks for client capabilities and often throw untyped errors. `@solana/plugin-interfaces` only exposes compile-time `ClientWithX` types, so plugin authors cannot validate unknown client objects and narrow them consistently.

#### Summary of Changes

Add `isClientWithX` and `assertIsClientWithX` helpers for every existing client capability interface. Multi-method capabilities are checked together, assertion failures throw a typed `SolanaError` that reports only the missing capabilities, and property-based capabilities are detected without invoking reactive getters. Convert `@solana/plugin-interfaces` from a type-only package to a standard runtime package, and add runtime tests, type-narrowing tests, package metadata, tree-shakability coverage, and README documentation.

Fixes #1909


lorisleiva:
Hi there, I appreciate the fancy structure of this implementation but it is highly inconsistent with how the rest of the code is implemented in this library. The single error for multiple capabilities and using shared helpers are fine but the rest needs revisiting:
- Having constants just to feed through a complex internal helper is overkill, just have a boolean return statement for the `isX` helper and use it inside the `assertIsX` helper to throw the error with the shared error code. I appreciate you wanting to potentially merge multiple capabilities into the same error but you're introducing the complexity of this solution without the benefits because these are all internal helpers/constants.
- Make sure the tests are also consistent. The `it.each` is more confusing then helping, add a `describe` block for each `isX` and `assertIsX` and a couple of test under each of them. Same with the typetests. `// [DESCRIBE]` should be followed by the name of the type or function your type-testing.

PR #1915, opened 2026-08-12
Let transaction plan executors return a result context

mcintyre94:
This PR enables migration for #1913 

In that PR, requiring a major changeset, we will require transaction plan executors to return their `TContext` (which at that point is much more flexible than the current one), instead of returning a `Transaction` or a `Signature`.

To enable that migration, this PR allows executors to return a `TContext`, and returns it if they do so.

As of this PR, executors may still return a `Transaction` or a `Signature`, but will get a deprecation warning. The behaviour if they do so is unchanged, ie we derive a signature from the transaction.

Executors that migrate to returning `TContext` should not have any breaking changes from #1913 

If this is approved, I'll stack #1910  and #1913 on it, this will slightly simplify #1913 as there's some duplication.

### DIFF
--- a/packages/instruction-plans/src/transaction-plan-executor.ts
+++ b/packages/instruction-plans/src/transaction-plan-executor.ts
@@ -24,6 +24,7 @@ import {
     parallelTransactionPlanResult,
     sequentialTransactionPlanResult,
     SingleTransactionPlanResult,
+    type SuccessfulBaseTransactionPlanResultContext,
     successfulSingleTransactionPlanResult,
     successfulSingleTransactionPlanResultFromTransaction,
     type TransactionPlanResult,
@@ -50,11 +51,17 @@ export type TransactionPlanExecutor<TContext extends TransactionPlanResultContex
     config?: { abortSignal?: AbortSignal },
 ) => Promise<TransactionPlanResult<TContext>>;
 
-type ExecuteTransactionMessage<TContext extends TransactionPlanResultContext> = (
+type SuccessfulTransactionPlanResultContext<TContext extends TransactionPlanResultContext> =
+    SuccessfulBaseTransactionPlanResultContext & TContext;
+
+type ExecuteTransactionMessage<
+    TContext extends TransactionPlanResultContext,
+    TReturn = Signature | SuccessfulTransactionPlanResultContext<TContext> | Transaction,
+> = (
     context: BaseTransactionPlanResultContext & TContext,
     transactionMessage: TransactionMessage & TransactionMessageWithFeePayer,
     config?: { abortSignal?: AbortSignal },
-) => Promise<Signature | Transaction>;
+) => Promise<TReturn>;
 
 /**
  * Configuration object for creating a new transaction plan executor.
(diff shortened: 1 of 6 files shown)


mcintyre94:
* **#1915** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1915?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -226,6 +313,19 @@ async function traverseSingle<TContext extends TransactionPlanResultContext>(
...
+ * {@link Signature} has been ruled out: the context a successful result should carry, or the
+ * deprecated {@link Transaction}. Every returned context has a `signature` — that is what a
+ * successful result guarantees — and a `Transaction` never does, since it keeps its signatures in a
+ * `signatures` map, so that property is a reliable discriminator.
+ */
+function isSuccessfulContext<TContext extends TransactionPlanResultContext>(
+    returnValue: SuccessfulTransactionPlanResultContext<TContext> | Transaction,
+): returnValue is SuccessfulTransactionPlanResultContext<TContext> {
+    return 'signature' in returnValue;
+}
```
mcintyre94, line 327: Note: this is a bit hacky, but it's only temporary for this migration. It'll be removed when we drop the non-context returns.

mcintyre94:
@trevor-cortex

trevor-cortex:
```
AI provider [anthropic-oauth] is overloaded.
```

mcintyre94:
@trevor-cortex

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -64,10 +71,44 @@ type ExecuteTransactionMessage<TContext extends TransactionPlanResultContext> =
...
+ * Unlike this deprecated path, returning a context never derives a signature from a transaction, so
+ * it also works for transactions their fee payer has not signed.
+ *
+ * @see {@link TransactionPlanExecutorConfig}
+ */
+export function createTransactionPlanExecutor<
+    TContext extends TransactionPlanResultContext = TransactionPlanResultContext,
+>(config: {
+    executeTransactionMessage: ExecuteTransactionMessage<TContext, Signature | Transaction>;
+}): TransactionPlanExecutor<TContext>;
```
trevor-cortex, line 111: The position of this overload is load-bearing: because TypeScript resolves calls against the first matching overload, this narrower `Signature | Transaction` signature must stay *above* the general one for legacy callbacks to get the deprecation flag. If the two were ever swapped, legacy callbacks would match the general overload first and the deprecation would silently stop firing (the typetest's `// It can still return a signature or a full transaction, using the deprecated overload` comment would still pass, since it doesn't assert deprecation). A one-line code comment here noting that the ordering is intentional would cheaply protect against a future refactor reordering them.

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -81,14 +122,48 @@ export type TransactionPlanExecutorConfig<
...
+ *
+ * ```ts
+ * createTransactionPlanExecutor<{ startedAt: number }>(config);
+ * ```
+ *
+ * Returning a {@link Signature} or a full {@link Transaction} object instead of a context is
+ * deprecated. Those return values are still honoured — a returned signature is stored as
+ * `context.signature`, and a returned transaction is stored as `context.transaction` with its
+ * signature derived from it — but deriving that signature throws
+ * {@link SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING} when the fee payer has not signed,
```
trevor-cortex: Tiny nit: `SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING` isn't imported in this file (unlike the other error codes referenced via `{@link}` here, which are all imported for runtime use), so this link likely won't resolve in the generated API docs. Since importing it just for a docblock would trip unused-import lint, plain backticks may be the pragmatic choice — same for the copy of this reference in the changeset, which is fine as-is since changesets aren't TypeDoc-processed.

trevor-cortex (commented):
This PR adds a third return option to `executeTransactionMessage`: the context the successful result should carry. It's implemented as two overloads of `createTransactionPlanExecutor` — a `@deprecated` one that only matches callbacks returning the legacy `Signature | Transaction`, and the general one that also accepts a `SuccessfulBaseTransactionPlanResultContext & TContext`. At runtime, a returned context is told apart from a `Transaction` by the presence of `signature` (a `Transaction` keeps its signatures in a `signatures` map, so this is a sound discriminator within Kit's types). On success, the returned context is merged over whatever the callback stored on the mutable context, with the return value winning.

The design is clean and the migration story is well thought out — the type-level trick of requiring `signature` on the returned context both enforces the successful-result invariant at compile time and doubles as the runtime discriminator. Tests cover the merge semantics, precedence, custom contexts, and the new partially-signed-transaction capability; the typetests cover both the deprecated and new overloads plus the missing-signature and missing-custom-property failure cases. The changeset is `minor`, which matches the repo's policy for a feature.

This would be an approval, but I don't have approve permissions on this repo — treat it as one.

Things to watch / notes for subsequent reviewers:

1. **Overload ordering is load-bearing.** The deprecated `Signature | Transaction` overload must stay *before* the general one — TypeScript resolves calls against the first matching overload, so if these were ever reordered, legacy callbacks would silently match the general overload and lose their deprecation flag. I left an inline note suggesting a small code comment to protect against an innocent future refactor.

2. **Worth a quick IDE sanity check** that `@deprecated` on only one overload strikes through legacy call sites but not the function in completions/auto-import. TypeScript handles per-signature deprecation for call resolution, but editor behavior around partially-deprecated overload sets has had rough edges historically — a 30-second check in VS Code would confirm.

3. **Failure-path error masking (possible follow-up, not blocking).** In `traverseSingle`'s catch block (unchanged by this PR, around L307-310), `getSignatureFromTransaction(context.transaction)` can itself throw `SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING` when a fee-payer-unsigned transaction was stored on the context. Since that throw happens *inside* the catch, it escapes `traverseSingle` entirely — the executor rejects with the derivation error instead of producing a structured `FailedSingleTransactionPlanResult`, losing the original failure. The new docs handle this honestly by telling callers not to store such transactions, but now that this PR makes partially signed transactions a first-class success-path pattern, wrapping that derivation in a try/catch (keep the original error, skip the derived signature) seems like a natural hardening — perhaps in #1913, which is touching this area anyway.

4. The changeset and docblocks are unusually thorough for a feature of this size, in a good way — they document the merge precedence, the discriminator rationale, and the failure-path caveat accurately against the implementation. I verified each documented behavior against the code and tests.

mcintyre94:
> The changeset and docblocks are unusually thorough for a feature of this size, in a good way — they document the merge precedence, the discriminator rationale, and the failure-path caveat accurately against the implementation.

Claude giving Claude a medal jpeg

packages/instruction-plans/src/transaction-plan-executor.ts
mcintyre94: Updated

lorisleiva (approved):
Thank you!

PR #1916, opened 2026-08-12
Add a note in CLAUDE.md about docs using published packages

mcintyre94:
I've noticed that Claude is quite eager to update docs which is good, but doesn't recognise that they won't compile when they reference new types that aren't yet published.

This PR adds some notes on this to CLAUDE.md so that it'll hopefully keep updating docs when appropriate, but not try to add docs using new unreleased types.


lorisleiva (approved):
Good shout

PR #1917, opened 2026-08-12
fix: stop upcasting transaction `version` to `bigint`

amilz:
Kit's RPC response transformer upcasts every JSON integer to a `bigint` unless the field's keypath appears in a per-method allow-list (`getAllowedNumericKeypaths()` in `packages/rpc-api/src/index.ts`). `version` was missing from that list on `getTransaction`, `getBlock` transactions, and `getTransactionsForAddress`, so it arrives at runtime as `0n` while still typechecking as `TransactionVersion` (`'legacy' | 0 | 1`).

The result is a silent mismatch between the declared type and the runtime value:

```ts
if (transaction.version === 0) {
    // never runs: the value is `0n`, and `0n === 0` is false
}
```

There is no compiler error and no runtime error; the branch simply never runs. This was found in Explorer, where it silently disabled an entire feature.

## Changes

Allow-list `version` on all three methods. Three added keypaths.

## Testing

New unit tests in `packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts` push canned responses through `createSolanaRpcApi()` and assert `version` arrives as a number. Three of the four fail against the unpatched allow-list; the fourth covers the `'legacy'` string case, which was never affected and guards against over-correcting.

## Follow-ups, deliberately not in this PR

**Token balance numerics.** The same audit found that `uiTokenAmount.uiAmount` is missing from the allow-list on these same three methods, and that `simulateTransaction` has no token balance keypaths allow-listed at all. `uiAmount` is an `f64`, and the transformer's guard is `Number.isInteger`, so a whole-number balance was upcast while a fractional one was not. That fix is ready and will follow immediately after this merges. It is kept separate because it adds a new public export to `@solana/rpc-transformers` and so carries a `minor` bump, whereas this change is a pure `patch`.

**`@solana/rpc-subscriptions-api`.** Same class of bug, different root cause. Its allow-list is keyed by notification method name (`blockNotifications`), but the response transformer is invoked with the subscribe request, whose `methodName` is `blockSubscribe` (`rpc-subscriptions-pubsub-plan.ts:96`). The lookup never matches, so the entire subscriptions allow-list is dead code and every field it names is upcast. There is a second defect underneath: the publisher is memoized per `(channel, responseTransformer)` and its closure captures the first `subscribeRequest`, so the transform is resolved per channel where it needs to be per subscription. Fixing it flips affected fields from `bigint` to `number` at runtime, which is breaking for anyone who adapted, so it needs its own change and its own bump. Adding correct keypaths to a list that is never consulted would ship an inert change that reads as fixed.

### DIFF
--- a/.changeset/small-birds-cheer.md
+++ b/.changeset/small-birds-cheer.md
@@ -1,0 +1,9 @@
+---
+'@solana/rpc-api': patch
+---
+
+Stop upcasting transaction `version` to `bigint`
+
+The response transformer upcasts every JSON integer to a `bigint` unless its keypath appears in an allow-list. `version` was missing from that allow-list on `getTransaction`, `getBlock` transactions, and `getTransactionsForAddress`, so it arrived at runtime as `0n` while still typechecking as `TransactionVersion` (`'legacy' | 0 | 1`).
+
+A check like `if (transaction.version === 0)` therefore compiled cleanly and was always false, with no compiler error and no runtime error. The keypath is now allow-listed and `version` arrives as a number, matching its declared type.
--- a/packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts
+++ b/packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts
@@ -1,0 +1,65 @@
+import type { Address } from '@solana/addresses';
+import type { Signature } from '@solana/keys';
+import { createRpc, type Rpc } from '@solana/rpc-spec';
+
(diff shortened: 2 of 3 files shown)


mcintyre94 (approved):
Nice, thanks!

On your follow ups, `uiTokenAmount.uiAmount` is defined as `Option<f64>` on the RPC, so agreed that shouldn't be bigint either 

I'm unsure about the notifications one, happy to take a look at a PR but feel free to just open an issue on that if it's easier/needs more investigation.

PR #1919, opened 2026-08-13
fix: stop upcasting token balance `uiAmount` and related numerics to `bigint`

amilz:
Follow-up to #1917, which allow-listed `version`. This covers the token balance numerics found in the same audit.

Opened as a draft because it carries a `minor` bump for `@solana/rpc-transformers` (it adds a public export) and you may prefer a different shape — see "Alternative" below.

## The bug

The response transformer upcasts every JSON integer to a `bigint` unless the field's keypath appears in a per-method allow-list. The guard is `Number.isInteger`, so a float whose value happens to be whole is upcast while a fractional one passes through untouched.

`uiTokenAmount.uiAmount` is an `f64` on the server and was not allow-listed on `getTransaction`, `getBlock`, or `getTransactionsForAddress`. So its runtime type depended on the value, not the field. Reproduced against the published `7.0.0`, all four rows from a single `getTransaction` response:

```
where                         value   typeof   verdict
meta.preTokenBalances  bob    1.5     number   OK
meta.preTokenBalances  alice  1n      bigint   BUG
meta.postTokenBalances bob    2n      bigint   BUG
meta.postTokenBalances alice  0.5     number   OK
```

The same balance fetched through `getTokenAccountBalance` returns `2` rather than `2n`, because that method already allow-lists `uiAmount`. So the same field had two different runtime types depending on which method you called.

`simulateTransaction` had no token balance keypaths allow-listed at all despite returning `preTokenBalances` and `postTokenBalances`, so `accountIndex` and `uiTokenAmount.decimals` were affected there too.

## Changes

- Allow-list `uiTokenAmount.uiAmount` on `getTransaction`, `getBlock`, and `getTransactionsForAddress` token balances.
- Allow-list the full token balance set on `simulateTransaction`.
- Extract the token balance keypaths into a shared `tokenBalancesConfigs` fragment, alongside the existing `innerInstructionsConfigs` and `messageConfig`, since they are now used at four call sites. This is the new export from `@solana/rpc-transformers`, and the reason for the `minor`.

## Alternative, if you would rather not take a `minor`

The keypaths can be inlined at all four call sites instead, keeping this a pure `patch` to `@solana/rpc-api` with no change to `@solana/rpc-transformers`. That trades the deduplication for the smaller bump. Happy to switch — say the word.

## Testing

Extends the tests added in #1917. Eight cases total in `packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts`, pushing canned responses through `createSolanaRpcApi()` and asserting the values arrive as numbers. Verified the token balance cases fail against the unpatched allow-list.

## Still outstanding

`@solana/rpc-subscriptions-api` has the same symptom from a different root cause, and is not addressed here. Its allow-list is keyed by notification method name (`blockNotifications`), but the response transformer is invoked with the subscribe request, whose `methodName` is `blockSubscribe` (`rpc-subscriptions-pubsub-plan.ts:96`). The lookup never matches, so the entire subscriptions allow-list is dead code and every field it names is upcast. There is a second defect underneath: the publisher is memoized per `(channel, responseTransformer)` and its closure captures the first `subscribeRequest`, so the transform resolves per channel where it needs to be per subscription. Fixing it flips affected fields from `bigint` to `number` at runtime, which is breaking for anyone who adapted, so it wants its own change and its own bump. Happy to take it on if you want it.

### DIFF
--- a/packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts
+++ b/packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts
@@ -2,10 +2,27 @@ import type { Address } from '@solana/addresses';
 import type { Signature } from '@solana/keys';
 import { createRpc, type Rpc } from '@solana/rpc-spec';
 
-import { createSolanaRpcApi, GetBlockApi, GetTransactionApi, GetTransactionsForAddressApi } from '../index';
+import {
+    createSolanaRpcApi,
+    GetBlockApi,
+    GetTransactionApi,
+    GetTransactionsForAddressApi,
+    SimulateTransactionApi,
+} from '../index';
 
 const MOCK_SIGNATURE =
     '4nHvMbxHURt2AXd7yQpKSKM5XCVKQiNbfsFmvPtHNJnJPSJHFT6cGUUNQGYK3wcxDCTvBMTLpQFf6HGqhLTUsxwj' as Signature;
+
+const MOCK_TOKEN_BALANCE = {
+    accountIndex: 1,
+    mint: 'So11111111111111111111111111111111111111112',
+    uiTokenAmount: {
+        amount: '1000000000',
+        decimals: 9,
+        uiAmount: 1,
+        uiAmountString: '1',
+    },
+};
 
 function createMockRpc<TApi>(result: unknown): Rpc<TApi> {
     return createRpc({
@@ -32,6 +49,18 @@ describe('the default response transformer for the Solana RPC', () => {
                 .send();
             expect(result?.version).toBe('legacy');
         });
+        it('leaves token balance `decimals` and `uiAmount` as numbers', async () => {
+            expect.assertions(2);
+            const rpc = createMockRpc<GetTransactionApi>({
+                meta: { postTokenBalances: [MOCK_TOKEN_BALANCE], preTokenBalances: [MOCK_TOKEN_BALANCE] },
+                slot: 1,
+            });
+            const result = await rpc
+                .getTransaction(MOCK_SIGNATURE, { encoding: 'json', maxSupportedTransactionVersion: 0 })
+                .send();
+            expect(result?.meta?.preTokenBalances?.[0].uiTokenAmount).toMatchObject({ decimals: 9, uiAmount: 1 });
+            expect(result?.meta?.postTokenBalances?.[0].uiTokenAmount).toMatchObject({ decimals: 9, uiAmount: 1 });
+        });
     });
 
     describe('getBlock', () => {
@@ -43,6 +72,18 @@ describe('the default response transformer for the Solana RPC', () => {
             });
             const result = await rpc.getBlock(1n, { encoding: 'json', maxSupportedTransactionVersion: 0 }).send();
             expect(result?.transactions[0].version).toBe(0);
+        });
+        it('leaves token balance `decimals` and `uiAmount` as numbers', async () => {
+            expect.assertions(1);
+            const rpc = createMockRpc<GetBlockApi>({
+                blockhash: '4nHvMbxHURt2AXd7yQpKSKM5XCVKQiNbfsFmvPtHNJnJ',
+                transactions: [{ meta: { preTokenBalances: [MOCK_TOKEN_BALANCE] } }],
+            });
+            const result = await rpc.getBlock(1n, { encoding: 'json', maxSupportedTransactionVersion: 0 }).send();
+            expect(result?.transactions[0].meta?.preTokenBalances?.[0].uiTokenAmount).toMatchObject({
+                decimals: 9,
+                uiAmount: 1,
+            });
         });
     });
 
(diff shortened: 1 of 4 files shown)


mcintyre94 (dismissed):
LGTM, thank you!

mcintyre94:
@trevor-cortex

trevor-cortex:
```
AI provider [anthropic-oauth] is overloaded.
```

mcintyre94:
@trevor-cortex

packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts
```diff
@@ -62,4 +103,23 @@ describe('the default response transformer for the Solana RPC', () => {
             expect(result.data[0].version).toBe(0);
         });
     });
+
+    describe('simulateTransaction', () => {
```
trevor-cortex, line 128: While adding this block, consider also adding the equivalent token balance case to the `getTransactionsForAddress` describe above (anchoring here since that block is unchanged in this diff). It's the only one of the four allow-list call sites without token balance coverage, and its hand-written keypath prefix (`['data', KEYPATH_WILDCARD, 'meta', ...]`) is exactly the part the shared `tokenBalancesConfigs` fragment can't protect. Something like:

```ts
it('leaves token balance `decimals` and `uiAmount` as numbers', async () => {
    expect.assertions(1);
    const rpc = createMockRpc<GetTransactionsForAddressApi>({
        data: [{ meta: { preTokenBalances: [MOCK_TOKEN_BALANCE] } }],
    });
    const result = await rpc.getTransactionsForAddress(MOCK_ADDRESS).send();
    expect(result.data[0].meta?.preTokenBalances?.[0].uiTokenAmount).toMatchObject({ decimals: 9, uiAmount: 1 });
});
```

(Adjust the call args to match the existing `version` test in that block.) This would also square the test count with the "seven cases" mentioned in the PR description — I count six as-is.

trevor-cortex (commented):
## Summary

This PR fixes the value-dependent `bigint` upcasting of token balance numerics. Since the transformer only upcasts values passing `Number.isInteger`, `uiTokenAmount.uiAmount` (an `f64` server-side) arrived as `bigint` for whole values and `number` for fractional ones on `getTransaction`, `getBlock`, and `getTransactionsForAddress` — and `simulateTransaction` had no token balance keypaths allow-listed at all. The fix allow-lists `uiAmount` on the three transaction-shaped methods, adds the full token balance set to `simulateTransaction`, and extracts the shared keypaths into a new `tokenBalancesConfigs` export from `@solana/rpc-transformers` (hence the `minor` there, `patch` for `@solana/rpc-api`).

The change is correct and well-scoped. The refactor to `tokenBalancesConfigs.flatMap(...)` is behavior-preserving for the previously listed keypaths (`accountIndex`, `uiTokenAmount.decimals`) and adds exactly `uiTokenAmount.uiAmount` — I diffed the expansions at each call site against the removed literals and they match. The `as const` on the new export plus spreading into fresh arrays at the call sites keeps the `KeyPath[]` typing happy while making the fragment itself immutable, which is a nice touch even though the neighboring `innerInstructionsConfigs`/`messageConfig` aren't `as const`.

## Notes

- **Test coverage gap**: `getTransactionsForAddress` gets the same allow-list change as `getTransaction`/`getBlock` but is the only one of the four call sites without a token balance test — its describe block still only covers `version`. The shared `tokenBalancesConfigs` fragment makes a per-site regression unlikely, but the keypath *prefix* (`['data', KEYPATH_WILDCARD, 'meta', ...]`) is still hand-written per site and is exactly the part a test would catch. Relatedly, the PR description says seven cases; I count six tests in the file after this change. Worth a quick follow-up commit, though I don't consider it blocking.
- The `toMatchObject` assertions do the right thing here: Jest treats `1n` and `1` as unequal, so these tests fail against the unpatched allow-list as advertised.
- **Changeset shape**: `minor` for the new `tokenBalancesConfigs` export and `patch` for `@solana/rpc-api` matches the repo's changeset conventions. Whether to take the inlined-keypaths alternative to avoid the `minor` is a maintainer call — the dedup seems worth it to me given four call sites, and it keeps future token-balance-bearing methods honest.

## For subsequent reviewers

- The `rpc-subscriptions-api` issue described under "Still outstanding" (allow-list keyed by notification name vs. subscribe method name, plus the per-channel memoization capturing the first `subscribeRequest`) is real and worth tracking as its own issue/PR — fixing it flips runtime types from `bigint` to `number`, so it needs its own bump discussion and shouldn't ride along here.
- Verify the changeset was CLI-generated (it looks well-formed) and that CI runs the new tests against the built transformer, not just source.

*(Submitting as COMMENT rather than APPROVE — I don't have approval permissions on this repo — but this would otherwise be an approval.)*

amilz:
@trevor-cortex thanks — the coverage gap is a fair catch, and for the right reason. The `tokenBalancesConfigs` fragment is shared, but the keypath prefix that wraps it (`['data', KEYPATH_WILDCARD, 'meta', ...]`) is still written out per call site, so that's precisely the part a per-site test protects.

Fixed in acd2eb89, which adds the missing token balance case to the `getTransactionsForAddress` describe block. I confirmed it earns its place: with the `tokenBalancesConfigs.flatMap(...)` entry removed from `getTransactionsForAddress`, it is the only one of the eight tests that fails.

mcintyre94 (approved):
Thanks!

PR #1924, opened 2026-08-14
Deprecate transaction plan result APIs being removed in v8

mcintyre94:
We have some types that we plan to remove in #1915 which will ship as v8, this PR marks them as deprecated and updates docs to no longer use them. This can publish as a 7.x release to enable gradual migration before we publish v8 

- Docs updated to 7.1.0, instruction-plans docs updated to return context, and derive signature in the callback. 
- `successfulSingleTransactionPlanResultFromTransaction` is marked deprecated
- `BaseTransactionPlanResultContext` is marked deprecated. Note that `SuccessfulBaseTransactionPlanResultContext` extends it and remains non-deprecated. It is deprecated (not removed) in #1915, but we don't have an alternative to point to until that PR.

### DIFF
--- a/.changeset/sixty-streets-tap.md
+++ b/.changeset/sixty-streets-tap.md
@@ -1,0 +1,28 @@
+---
+'@solana/instruction-plans': patch
+---
+
+Deprecate transaction plan result APIs being removed in v8
+
+Both of the following are deprecated and will be removed in the next major version.
+
+`successfulSingleTransactionPlanResultFromTransaction` derives the transaction `signature` on your behalf by calling `getSignatureFromTransaction`, which throws when the transaction's fee payer has not signed it. Construct results with `successfulSingleTransactionPlanResult` instead, passing the context explicitly:
+
+```diff
+- successfulSingleTransactionPlanResultFromTransaction(message, transaction, context);
++ successfulSingleTransactionPlanResult(message, {
++   ...context,
++   signature: getSignatureFromTransaction(transaction),
++   transaction,
++ });
+```
+
+`BaseTransactionPlanResultContext` goes away together with the intersections that graft it onto every `SingleTransactionPlanResult`. The context of a result is becoming entirely caller-defined — it will be exactly the `TContext` you supply — so there will be no separate base shape to merge in. If you refer to this type, declare whichever of its fields you need on your own context type instead:
+
+```ts
+type MyContext = {
+    message?: TransactionMessage & TransactionMessageWithFeePayer;
+    signature?: Signature;
+    transaction?: Transaction;
+};
+```
--- a/docs/content/docs/advanced-guides/instruction-plans.mdx
+++ b/docs/content/docs/advanced-guides/instruction-plans.mdx
@@ -379,7 +379,7 @@ The `executeTransactionMessage` callback receives the following arguments:
 - **`message`**: The transaction message to execute.
 - **`config`**: An optional configuration object that may include an `abortSignal`.
 
-The callback must return either a `Signature` or a full `Transaction` object directly.
+The callback must return the context object that the successful result should carry — at minimum, an object containing the transaction `signature`. Anything stored on the mutable `context` argument but left out of the return value is preserved as well, with the returned value taking precedence.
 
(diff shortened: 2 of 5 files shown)


mcintyre94:
* **#1924** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1924?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/1924?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

This PR prepares for the v8 removals in #1915 by deprecating `successfulSingleTransactionPlanResultFromTransaction` and `BaseTransactionPlanResultContext` in `@solana/instruction-plans`, publishing as a 7.x patch. It also updates the instruction-plans docs so every executor example returns a context (`{ signature, transaction }`) from `executeTransactionMessage` instead of a `Transaction`, and bumps the docs' `@solana/kit` pin to 7.1.0 so the twoslash examples type-check against the release that added context-return support.

LGTM overall — no blocking issues.

## What I verified

- **Deprecation notices match actual behavior.** The `@deprecated` text on `successfulSingleTransactionPlanResultFromTransaction` correctly describes the throw path (`getSignatureFromTransaction` on an unsigned-fee-payer transaction), and the docs' claims about context merging ("returned value taking precedence, anything stored but not returned is preserved") match the `{ ...context, ...result }` merge in `traverseSingle`. The claim that failed results derive `signature` from any `transaction` on the context also matches the executor's catch block.
- **Docs examples are correct.** `getSignatureFromTransaction` is called right after `signTransactionMessageWithSigners`, so the fee payer signature is present and the derivation is safe. All four updated examples are consistent with each other.
- **Deliberate partial deprecation is fine.** `SuccessfulBaseTransactionPlanResultContext` extending the now-deprecated `BaseTransactionPlanResultContext`, and the non-deprecated `Failed`/`Canceled` result types still intersecting it, is intentional per the PR description — the replacement story lands with #1915. No objection.
- **Patch bump** looks right for a JSDoc-only deprecation with no runtime change; docs/CI/lockfile changes need no changeset.

## Notes for other reviewers

- The docs pin bump to `@solana/kit@7.1.0` is load-bearing: the examples now rely on the context-return overload of `createTransactionPlanExecutor` that shipped in 7.1.0, so the exact-pin bump and lockfile churn belong in this PR.
- One minor suggestion inline about the migration snippet not covering the deprecated helper's optional third `context` argument.
- The docs now say the callback "must return the context object" — strictly speaking `Signature`/`Transaction` returns still work (deprecated) in 7.x. I think presenting only the forward-looking API in docs is the right call, but flagging it in case you'd prefer a migration note for existing readers.

.changeset/sixty-streets-tap.md
```diff
@@ -0,0 +1,27 @@
...
+Both of the following are deprecated and will be removed in the next major version.
+
+`successfulSingleTransactionPlanResultFromTransaction` derives the transaction `signature` on your behalf by calling `getSignatureFromTransaction`, which throws when the transaction's fee payer has not signed it. Construct results with `successfulSingleTransactionPlanResult` instead, passing the context explicitly:
+
+```diff
+- successfulSingleTransactionPlanResultFromTransaction(message, transaction);
++ successfulSingleTransactionPlanResult(message, {
++   signature: getSignatureFromTransaction(transaction),
++   transaction,
++ });
```
trevor-cortex: Minor: the deprecated helper also accepts an optional third `context` argument. Callers using it would need to spread it into the new context, which this snippet doesn't show:

```diff
- successfulSingleTransactionPlanResultFromTransaction(message, transaction, context);
+ successfulSingleTransactionPlanResult(message, {
+   ...context,
+   signature: getSignatureFromTransaction(transaction),
+   transaction,
+ });
```

Might be worth covering (here and in the matching `@deprecated` JSDoc in `transaction-plan-result.ts`) so those callers aren't left guessing — though if you'd rather keep the common case minimal, that's reasonable too.

lorisleiva (approved):
Nice, thank you!

PR #1925, opened 2026-08-15
fix: consult the subscriptions numeric allow-list

### LINKED ISSUE #1921: Subscriptions numeric allow-list is never consulted, so all notification numerics are upcast to bigint
The numeric allow-list in `@solana/rpc-subscriptions-api` is never consulted, so every field it names arrives as a `bigint` while still typechecking as a `number`.

`getAllowedNumericKeypaths()` is keyed by notification method name (`blockNotifications`, `accountNotifications`, `programNotifications`), but the transformer is invoked with the subscribe request, whose name the plan executor rewrote to `blockSubscribe` (`rpc-subscriptions-pubsub-plan.ts:96`). The lookup misses and falls back to an empty keypath list.

```
subscribeRequest.methodName = "blockSubscribe"
via subscribeRequest   : {"value":{"block":{"transactions":[{"version":"0n"}]}}}
via *Notifications key : {"value":{"block":{"transactions":[{"version":0}]}}}
```

Two things make this more than a one-line fix:

1. **Passing the notification request through isn't sufficient.** The demultiplexed publisher is memoized per `(channel, responseTransformer)` and its closure captures the first `subscribeRequest` it sees, so subscriptions sharing a channel are all transformed under whichever one created the publisher. Closing over the request per execution doesn't work either, since the transformer's identity is the memoization key — that yields one demultiplexer per subscription instead of one per channel.
2. **`blockNotifications`' list has drifted** from its `getBlock` sibling. It's ~160 lines of hand-expanded keypaths and is missing transaction `version`, token balance `uiTokenAmount.uiAmount`, and `stackHeight` on both message and inner instructions. Deriving it from `innerInstructionsConfigs` / `messageConfig` / `tokenBalancesConfigs` fixes all three.

Auditing the rest found nothing further to allow-list — `slot`, `slotsUpdates`, `vote`, `root`, `logs` and `signature` notifications declare no `number`-typed fields.

Note this changes runtime values from `bigint` to `number` for the affected fields, which is breaking for anyone who adapted to the current behaviour.

Related, same class of bug on the request side: #1917 (merged, `version`) and #1919 (open, token balance `uiAmount`). This one is the subscriptions equivalent.

Happy to open a PR if that's useful — I have a fix and tests working locally — but it touches `rpc-subscriptions-spec`'s public config and needs a call on the bump, so flagging it as an issue first.

o-mid:
## Summary
- The numeric allow-list is keyed by API names (`blockNotifications`), but the plan executor invoked the transformer with the rewritten subscribe request (`blockSubscribe`). The lookup missed, so every notification numeric arrived as `bigint` while still typechecking as `number`.
- The plan executor now derives the method name from each notification payload (`blockNotification` → `blockNotifications`) so subscriptions that share a channel are not transformed under whichever request created the publisher. The subscriptions transformer also maps `*Subscribe` / `*Notification` as a fallback.
- `blockNotifications` is derived from the same `innerInstructionsConfigs` / `messageConfig` / `tokenBalancesConfigs` as `getBlock`, so transaction `version`, token balance `uiAmount`, and `stackHeight` stay numbers.

Fixes #1921

## Test plan
- [x] `rpc-subscriptions-spec` pubsub-plan tests (transformer gets `thingNotifications`, not `thingSubscribe`; shared-channel notifications keep their own method names)
- [x] `rpc-transformers` maps `blockSubscribe` / `blockNotification` onto the `blockNotifications` allow-list
- [x] `rpc-subscriptions-api` leaves `version`, `uiAmount`, and `stackHeight` as numbers on a mocked `blockNotification`


lorisleiva:
@trevor-cortex

packages/rpc-transformers/src/response-transformer.ts
```diff
@@ -62,11 +62,21 @@ export function getDefaultResponseTransformerForSolanaRpc<TApi>(
...
  */
+function toAllowedNumericMethodName(methodName: string): string {
+    if (methodName.endsWith('Subscribe')) {
+        return `${methodName.slice(0, -'Subscribe'.length)}Notifications`;
+    }
+    if (methodName.endsWith('Notification') && !methodName.endsWith('Notifications')) {
+        return `${methodName}s`;
+    }
+    return methodName;
+}
```
trevor-cortex: This helper landed between the TSDoc block ("Returns the default response transformer for the Solana RPC Subscriptions API…") and `getDefaultResponseTransformerForSolanaRpcSubscriptions`, so the doc comment now attaches to the private helper instead of the exported function — TypeDoc and editor hovers will both pick it up wrong. Consider moving `toAllowedNumericMethodName` above the doc comment (or below the exported function) so the docs stay attached to the public API.

packages/rpc-subscriptions-spec/src/rpc-subscriptions-pubsub-plan.ts
```diff
@@ -31,6 +31,22 @@ type RpcNotification<TNotification> = Readonly<{
...
+ * does not fall through to an empty keypath list.
+ */
+function getApiMethodNameFromNotificationMethod(notificationMethod: string): string {
+    if (notificationMethod.endsWith('Subscribe')) {
+        return `${notificationMethod.slice(0, -'Subscribe'.length)}Notifications`;
+    }
+    if (notificationMethod.endsWith('Notification') && !notificationMethod.endsWith('Notifications')) {
+        return `${notificationMethod}s`;
+    }
+    return notificationMethod;
```
trevor-cortex, line 48: Two questions on this helper:

1. The `*Subscribe` branch looks unreachable here — `message.method` comes from the server notification payload, which is always the singular `*Notification` wire name, never a `*Subscribe` name. Is it intentional belt-and-braces, or copied over from the transformer-side helper where the branch *is* reachable? If the former, a short note in the doc comment would prevent someone "cleaning it up" later; if not, it could be dropped to keep the mapping tight.

2. This duplicates `toAllowedNumericMethodName` in `@solana/rpc-transformers` almost verbatim. I assume that's deliberate to avoid giving `rpc-subscriptions-spec` a dependency on `rpc-transformers` — worth confirming, and maybe cross-referencing the two in comments so they don't drift apart.

packages/rpc-subscriptions-spec/src/rpc-subscriptions-pubsub-plan.ts
```diff
@@ -93,7 +109,10 @@ function getMemoizedDemultiplexedNotificationPublisherFromChannelAndResponseTran
                     return;
                 }
                 const transformedNotification = responseTransformer
-                    ? responseTransformer(message.params.result, subscribeRequest)
+                    ? responseTransformer(message.params.result, {
+                          methodName: getApiMethodNameFromNotificationMethod(message.method),
+                          params: subscribeRequest.params,
```
trevor-cortex, line 119: Non-blocking, mostly pre-existing: the publisher is memoized per `(channel, responseTransformer)`, so `subscribeRequest` here is whichever request *created* the publisher. Now that `methodName` is derived per-message, `params` is the one remaining field that can belong to a different subscription than the notification being transformed on a shared channel. The default transformer ignores `params` so nothing breaks today, but a custom transformer relying on it would get the wrong ones. A short code comment documenting this caveat would save the next person some head-scratching.

packages/rpc-subscriptions-api/src/__tests__/allowed-numeric-keypaths-test.ts
```diff
@@ -0,0 +1,136 @@
...
+    const notificationListener = jest.fn();
+    publisher.on('notification', notificationListener);
+    for (const listener of channel.listeners.message) {
+        listener({
+            jsonrpc: '2.0',
+            method: notificationMethod,
+            params: { result, subscription: subscriptionId },
+        });
+    }
+    return notificationListener.mock.calls[0][0];
```
trevor-cortex, line 71: Nit: if the notification never fires (e.g. the single `await Promise.resolve()` on L44 stops flushing enough microtasks after a refactor), this indexes into an empty `mock.calls` and the test dies with an opaque `TypeError: Cannot read properties of undefined`. An explicit guard, e.g. `if (!notificationListener.mock.calls.length) throw new Error('notification listener was never called');`, would make the failure mode self-explanatory without disturbing the `expect.assertions` counts.

trevor-cortex (commented):
## Summary

This fixes #1921 on two levels:

1. **The lookup miss**: the numeric allow-list is keyed by API method names (`blockNotifications`), but the plan executor called the response transformer with the rewritten subscribe request (`blockSubscribe`), so the lookup always fell through to an empty keypath list and every numeric got upcast to `bigint`. The plan executor now derives the method name from each notification payload's wire `method` (`blockNotification` → `blockNotifications`), which also fixes the shared-channel case where the memoized publisher previously transformed all notifications under whichever request created it. `getDefaultResponseTransformerForSolanaRpcSubscriptions` gains an equivalent `*Subscribe`/`*Notification` → `*Notifications` mapping as a fallback for custom plan executors.
2. **The allow-list itself**: `blockNotifications` is rebuilt from the shared `tokenBalancesConfigs` / `innerInstructionsConfigs` / `messageConfig`, aligning it with `getBlock`. This is a behavior addition beyond the lookup fix — `uiAmount`, `stackHeight`, and transaction `version` weren't in the old list at all.

I verified the composed keypaths against the old hand-written list: every previous path is covered (`accountIndex`, `decimals`, `commission`, `index`, `programIdIndex`, `accounts`, address-table indexes, header fields), and the only additions are the three above, matching `getBlock` in `rpc-api` exactly. Wire notification names (`accountNotification`, `slotsUpdatesNotification`, etc.) all map correctly to their API keys, including the tricky `slotsUpdates` plural.

Overall this looks correct and well-tested — I'd approve, but I don't have approve permissions on this repo, so leaving as a comment review.

## Things to watch out for

- **Custom response transformers now see a synthesized request.** `executeRpcPubSubSubscriptionPlan` previously passed the whole `subscribeRequest` to the transformer; it now passes `{ methodName: <derived>, params: subscribeRequest.params }`. The default transformer is unaffected, but any consumer passing a custom `responseTransformer` that inspected `request.methodName` will now see the API name instead of the subscribe name. Patch bump seems defensible since the old value was arguably the bug, but worth a maintainer sanity-check.
- **Stale `params` on shared channels** — see inline comment. The methodName is now derived per-message, but `params` still comes from whichever request created the memoized publisher. Pre-existing, non-blocking, but worth a code comment.
- **TSDoc attachment regression** in `response-transformer.ts` — see inline comment. Easy fix.

- The name-mapping helper is duplicated between `rpc-subscriptions-spec` and `rpc-transformers` (identical logic, different names). Presumably deliberate to keep `rpc-subscriptions-spec` free of a `rpc-transformers` dependency — I've asked inline to confirm.
- Changeset is present with patch bumps for all three touched packages, consistent with the repo's changeset conventions for a bug fix.
- Tests cover all three layers (plan executor, transformer fallback, end-to-end API allow-list) and match the stated test plan.

packages/rpc-transformers/src/response-transformer.ts
o-mid: Moved `toAllowedNumericMethodName` above the TSDoc so it stays attached to `getDefaultResponseTransformerForSolanaRpcSubscriptions`.

packages/rpc-subscriptions-spec/src/rpc-subscriptions-pubsub-plan.ts
o-mid: The `*Subscribe` branch was copied from the transformer helper and isn't reachable here (`message.method` is always the wire `*Notification` name), so I dropped it.

The duplication is deliberate — `rpc-subscriptions-spec` shouldn't depend on `rpc-transformers`. Both helpers now cross-reference each other; the transformer one still maps `*Subscribe` because custom plan executors can pass the request name.

packages/rpc-subscriptions-spec/src/rpc-subscriptions-pubsub-plan.ts
o-mid: Added a comment on the memoized publisher: `methodName` is per-notification, `params` still come from whichever request created it. The default transformer ignores `params`.

packages/rpc-subscriptions-api/src/__tests__/allowed-numeric-keypaths-test.ts
o-mid: Added that guard so a missed notification fails with `notification listener was never called` instead of indexing into an empty `mock.calls`.

mcintyre94 (dismissed):
LGTM, thank you!

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Re-reviewed the latest changes — all four points from my previous review have been addressed:

1. **TSDoc regression** — `toAllowedNumericMethodName` now carries its own doc block and no longer sits between `getDefaultResponseTransformerForSolanaRpcSubscriptions` and its documentation. ✅
2. **`*Subscribe` branch / helper duplication** — kept deliberately, and the new doc comments on both helpers explain why: the duplication avoids a package dependency from `rpc-subscriptions-spec` onto `rpc-transformers`, and the `*Subscribe` branch guards against custom plan executors that still pass the request name. The cross-references ("keep in sync" notes) in both directions make the intent clear. ✅
3. **Stale `params` on shared channels** — now explicitly documented at the call site in the plan executor, including the note that the default transformer ignores `params`. ✅
4. **Test guard** — `subscribeAndNotify` now throws a descriptive error if the notification listener never fires, instead of failing opaquely on `mock.calls[0][0]`. ✅

No new issues in this pass. This would be an approval, but as before I don't have approve permissions on this repo — treat this as a ✅ from me. Callum has already approved, so this looks good to land.

mcintyre94:
It looks like there's a few CI issues on the `rpc-subscriptions-api` package

The easiest way to check this locally is to do `pnpm turbo compile:js compile:typedefs` (since this PR spans multiple packages), and then make sure `pnpm dev` is passing

mcintyre94 (approved):

PR #1926, opened 2026-08-15
fix(codecs-strings): report buffer length as next offset when base-X decoder reads no bytes

latent-9:
`getBaseXDecoder` returns an offset of `0` when the slice at the requested offset is empty, instead of the buffer length.

Every other return path in `read()` reports `rawBytes.length` as the next offset, and the sibling `getBaseXResliceCodec` does the same in `baseX-reslice.ts` (with an explicit test for it). Returning `0` here rewinds the composed-decoder cursor, so any decoder placed after a base-X decoder that reads an empty remainder restarts from offset `0` and silently mis-decodes the following fields.

This is the decoder-side counterpart of the same class fixed in #1911 (`getBitArrayEncoder` returning the wrong next offset).

Fix: return `rawBytes.length` to match the other paths. Added a regression test in `base58-test.ts` asserting `read(bytes, offset)` reports the end of the buffer when nothing remains to decode.

### DIFF
--- a/.changeset/great-melons-camp.md
+++ b/.changeset/great-melons-camp.md
@@ -1,0 +1,5 @@
+---
+'@solana/codecs-strings': patch
+---
+
+Fix `getBaseXDecoder` returning an offset of `0` instead of the buffer length when there are no bytes left to decode, which corrupted the offset of any decoder composed after it.
--- a/packages/codecs-strings/src/__tests__/base58-test.ts
+++ b/packages/codecs-strings/src/__tests__/base58-test.ts
@@ -98,4 +98,13 @@ describe('getBase58Codec', () => {
         ]);
         expect(base58.read(bytes, -(bytes.byteLength + 1))).toStrictEqual(base58.read(bytes, 0));
     });
+
+    it('returns the end of the buffer as the next offset when no bytes remain to decode', () => {
+        const base58 = getBase58Codec();
+        // Reading at an offset that leaves nothing to consume must report the end of the
+        // buffer as the next offset, like every other read() path. Rewinding to 0 here
+        // corrupts the cursor of any decoder composed after this one.
+        const bytes = new Uint8Array([255]);
+        expect(base58.read(bytes, 1)).toStrictEqual(['', 1]);
+    });
 });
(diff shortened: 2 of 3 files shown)


lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Summary

One-character-class fix: `getBaseXDecoder.read()` returned `['', 0]` when the slice at the requested offset was empty, while every other return path in the same function reports `rawBytes.length` as the next offset. Since composed decoders (structs, tuples, etc.) thread the returned offset into the next field's `read()`, returning `0` rewound the cursor and silently mis-decoded everything after a base-X decoder that consumed an empty remainder.

Verified against the codebase:

- The sibling `getBaseXResliceDecoder` in `baseX-reslice.ts` handles the identical case with `if (bytes.length === 0) return ['', rawBytes.length];` — this PR brings `baseX.ts` in line with it.
- The other two return paths in `getBaseXDecoder.read()` (all-zero bytes and the general case) both return `rawBytes.length`, so this was clearly an oversight, not a deliberate choice.
- The regression test mirrors the existing reslice test's intent and correctly exercises the empty-remainder path (`read([255], 1)` → `['', 1]`).
- Changeset is present with a `patch` bump on `@solana/codecs-strings`, which matches the bug-fix nature of the change.

## Notes for reviewers

One pre-existing edge worth knowing about (not introduced here, and shared with `baseX-reslice.ts`, so I'm not requesting changes): if `read()` is called with an offset *beyond* the buffer end (e.g. offset `5` on a 1-byte buffer), the slice is also empty and the reported next offset (`rawBytes.length`) will be *less* than the offset passed in. Callers going through `decode()` are protected by the offset range assertions in `codecs-core`, so this only matters for direct `read()` calls with invalid offsets — fine to leave as-is, just flagging for awareness.

LGTM — would approve, but I don't have approve permissions on this repo, so submitting as a comment review instead.

lorisleiva (approved):

latent-9:
@lorisleiva @trevor-cortex thank you  am happy to make a small contribution to this ecosystem.

PR #1930, opened 2026-08-18
Remove the retry we added to docs CI

mcintyre94:
This doesn't work because the failure case is the runner running out of memory, and it just gets killed

Retrying is just wasting time when the build legitimately fails, for example with a compile error


lorisleiva (approved):
It was worth a shot, thank you for your work on this painful CI. 🙇

PR #1948, opened 2026-08-19
Remove previously deprecated APIs

mcintyre94:
This PR removes some of the APIs that we previously deprecated as part of the v8 release

- `estimateComputeUnitLimitFactory` is removed, was replaced by `estimateResourceLimitsFactory`
- Same with `estimateAndSetComputeUnitLimitFactory` and `fillTransactionMessageProvisoryComputeUnitLimit`
- `getBigIntDowncastRequestTransformer` was deprecated when we no longer needed it for Kit's default RPC transformer, now removed
- `TRANSACTION_SIZE_LIMIT` and other constants were deprecated because v1 transactions make transaction size non-constant, now removed
- `SuccessfulBaseTransactionPlanResultContext` was deprecated and is now removed, replaced by `TransactionPlanResultContextWithSignature` in v8


lorisleiva (approved):
Nice, I love seeing deprecated code being cleaned up. 🤤

PR #1949, opened 2026-08-19
docs(sysvars): drop unsupported Fees sysvar from docs

rajanpanth:
`packages/sysvars` documents a `Fees` sysvar in two places — the README's "Supported Sysvars" list and the `@packageDocumentation` block in `src/index.ts` — but it isn't supported:

- there is no `src/fees.ts` module;
- `src/index.ts` exports nothing for it (clock, epoch-rewards, epoch-schedule, last-restart-slot, recent-blockhashes, rent, slot-hashes, slot-history, stake-history, sysvar — no fees);
- the README's link to `packages/sysvars/src/fees.ts` returns 404.

`Fees` is deprecated on-chain, so the omission looks intentional and it's the docs that are stale. This removes both mentions. Every other sysvar in that list has a matching module and export, so the rest is accurate.

Happy to instead implement `fees.ts` if support is actually wanted — just say so.


mcintyre94 (approved):
Thanks! Agree on removing this. The linked file was removed in c88068718 which includes in its changeset:

> Removed the fees sysvar which has been disabled on the network for a year, and has now been removed from the test validator in Agave 2.0

PR #1950, opened 2026-08-19
Allow v1 transactions to typecheck

mcintyre94:
Kit has runtime support for v1 transactions, but we've been waiting until they're close to available on the network before enabling code using them to type check

This PR updates the types, removing all the `TransactionVersionWithoutV1` temporary types so that `TransactionVersion` including v1 flows through all the `TransactionMessage` types.

The plan is to include this in the kit v8 release

### DIFF
--- a/.changeset/empty-gifts-shave.md
+++ b/.changeset/empty-gifts-shave.md
@@ -1,0 +1,7 @@
+---
+'@solana/transaction-messages': minor
+---
+
+Add support for version 1 transaction messages to `createTransactionMessage`. You can now pass `{ version: 1 }` to create an empty v1 transaction message.
+
+This means that code using version 1 transaction messages will now type check.
--- a/packages/kit/src/__typetests__/scenarios/transaction-signers-typetest.ts
+++ b/packages/kit/src/__typetests__/scenarios/transaction-signers-typetest.ts
@@ -16,21 +16,17 @@ import {
     TransactionVersion,
 } from '@solana/transaction-messages';
 
-// Temporary, until we support v1 transactions in `createTransactionMessage`
-// When this is removed, use `TransactionVersion`
-type TransactionVersionWithoutV1 = Exclude<TransactionVersion, 1>;
-
 // [DESCRIBE] TransactionMessageWithSigners type
 {
     // It is satisfied by a transaction message from `createTransactionMessage`
     {
-        const result = createTransactionMessage({ version: null as unknown as TransactionVersionWithoutV1 });
+        const result = createTransactionMessage({ version: null as unknown as TransactionVersion });
         result satisfies TransactionMessage & TransactionMessageWithSigners;
     }
 
     // It is satisfied after adding an address fee payer
     {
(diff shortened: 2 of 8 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

This PR removes the temporary `TransactionVersionWithoutV1` / `SupportedTransactionVersion` type guards so that `TransactionVersion` (including `1`) flows through `createTransactionMessage` and the surrounding `TransactionMessage` types. Changes are almost entirely type-level: the `createTransactionMessage` generic constraint is widened, the corresponding `@ts-expect-error` suppressions are dropped from the v1 decompiler and size tests, and the typetests are updated to assert v1 creation works (with proper negative assertions against legacy/v0).

The change is additive from a consumer perspective — code that previously failed to typecheck now succeeds — so the `minor` changeset on `@solana/transaction-messages` looks right, and dependent packages (`kit`, `transactions`) will get their bumps via changesets' dependency handling.

Overall this looks good — no blocking issues, just one cleanup worth doing while we're here.

## Notes

- The typetest coverage is thorough: v1 now has the same positive/negative `satisfies` matrix as legacy and v0 in `create-transaction-message-typetest.ts`, including the `TransactionMessageWithinSizeLimit` check.
- **Leftover cast in `packages/transaction-messages/src/decompile/v1/message.ts` (~L43–44)**: the comment `// Won't need this cast after we support v1 on \`createTransactionMessage\`` and the `as unknown as TransactionMessage & { version: 1 }` cast are now stale — this PR is precisely the change that comment was waiting for. See inline comment.
- **Doc nit (not anchorable)**: the doc comment on `createTransactionMessage` in `create-transaction-message.ts` (~L18–24) still only shows `version: 0` in its example. Since v1 is the headline of this change, adding a `{ version: 1 }` example (or mentioning supported versions) would help discoverability.

## For subsequent reviewers

- Verify no other `Exclude<TransactionVersion, 1>` aliases remain elsewhere in the repo (e.g. in `kit` or `transactions` sources) — the typetest suite should catch stragglers, but a quick grep is cheap.
- Since this is slated for the kit v8 release, confirm the release coordination is tracked outside this PR (the changeset alone will publish on the next release cycle once merged).

mcintyre94:
> Leftover cast in packages/transaction-messages/src/decompile/v1/message.ts

Knew I'd miss something, fixed!

> Doc nit (not anchorable): the doc comment on createTransactionMessage in create-transaction-message.ts (~L18–24) still only shows version: 0 in its example.

Leaving this for now, probably makes sense to update whenever we make v1 the default across the ecosystem/docs though 

> Verify no other Exclude<TransactionVersion, 1> aliases remain elsewhere in the repo

Checked this. The only cases are `Exclude<TransactionVersion, 'legacy'> and are legit. Also grepped for "support v1" and no matches, so hopefully got everything now!

lorisleiva (approved):
It's happening! 🙌

PR #1951, opened 2026-08-19
feat: fill in transaction v1 (SIMD-0385) gaps

amilz:
## Summary

- Export the `v1-transaction-config` module from `@solana/transaction-messages`, which was built and published but missing from the package index, leaving `setTransactionMessageConfig` and `V1TransactionConfig` unreachable.
- Correct the `computeUnitLimit` and `loadedAccountsDataSizeLimit` docs: an unset value on a version 1 transaction is budgeted zero, not the legacy 200,000 CU per instruction.
- Add `transactionConfig` to the transaction message response types in `@solana/rpc-types` and `@solana/rpc-api` so a confirmed version 1 transaction's compute budget can be read without a cast.
- Allow-list the three `u32` config fields in the response transformer so they stay JS numbers, leaving `priorityFee` as the only `Lamports` bigint.
- Mirror the same shape in the `@solana/rpc-graphql` schema as `TransactionMessageConfig`.

Closes DEV-922

### DIFF
--- a/packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts
+++ b/packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts
@@ -13,6 +13,13 @@ import {
 const MOCK_SIGNATURE =
     '4nHvMbxHURt2AXd7yQpKSKM5XCVKQiNbfsFmvPtHNJnJPSJHFT6cGUUNQGYK3wcxDCTvBMTLpQFf6HGqhLTUsxwj' as Signature;
 
+const MOCK_TRANSACTION_CONFIG = {
+    computeUnitLimit: 20_000,
+    heapSize: 32_768,
+    loadedAccountsDataSizeLimit: 65_536,
+    priorityFee: 5_000,
+};
+
 const MOCK_TOKEN_BALANCE = {
     accountIndex: 1,
     mint: 'So11111111111111111111111111111111111111112',
@@ -49,6 +56,23 @@ describe('the default response transformer for the Solana RPC', () => {
                 .send();
             expect(result?.version).toBe('legacy');
         });
+        it('leaves the `u32` fields of `transactionConfig` as numbers but upcasts `priorityFee`', async () => {
+            expect.assertions(4);
+            const rpc = createMockRpc<GetTransactionApi>({
+                meta: null,
+                slot: 1,
+                transaction: { message: { transactionConfig: MOCK_TRANSACTION_CONFIG } },
+                version: 1,
+            });
+            const result = await rpc
+                .getTransaction(MOCK_SIGNATURE, { encoding: 'json', maxSupportedTransactionVersion: 0 })
+                .send();
+            const transactionConfig = result?.transaction.message.transactionConfig;
+            expect(transactionConfig?.computeUnitLimit).toBe(20_000);
+            expect(transactionConfig?.heapSize).toBe(32_768);
+            expect(transactionConfig?.loadedAccountsDataSizeLimit).toBe(65_536);
+            expect(transactionConfig?.priorityFee).toBe(5_000n);
+        });
         it('leaves token balance `decimals` and `uiAmount` as numbers', async () => {
             expect.assertions(2);
             const rpc = createMockRpc<GetTransactionApi>({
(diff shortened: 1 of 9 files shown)


mcintyre94 (dismissed):
Thankyou!

mcintyre94:
@trevor-cortex

packages/rpc-graphql/src/schema/type-defs/transaction.ts
```diff
@@ -46,6 +46,13 @@ export const transactionTypeDefs = /* GraphQL */ `
         writableIndexes: [Int]
     }
 
+    type TransactionMessageConfig {
+        computeUnitLimit: Int
+        heapSize: Int
+        loadedAccountsDataSizeLimit: Int
+        priorityFee: BigInt
```
trevor-cortex: Nit: `TransactionMeta` uses the `Lamports` scalar for fee fields (`fee: Lamports`, `preBalances: [Lamports]`), and `rpc-types` types this field as `Lamports | null`. Consider `priorityFee: Lamports` here for consistency with the rest of the schema.

packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts
```diff
@@ -49,6 +56,23 @@ describe('the default response transformer for the Solana RPC', () => {
...
+        it('leaves the `u32` fields of `transactionConfig` as numbers but upcasts `priorityFee`', async () => {
+            expect.assertions(4);
+            const rpc = createMockRpc<GetTransactionApi>({
+                meta: null,
+                slot: 1,
+                transaction: { message: { transactionConfig: MOCK_TRANSACTION_CONFIG } },
+                version: 1,
+            });
+            const result = await rpc
+                .getTransaction(MOCK_SIGNATURE, { encoding: 'json', maxSupportedTransactionVersion: 0 })
```
trevor-cortex: Nit: the mock returns `version: 1` but the request says `maxSupportedTransactionVersion: 0` (here and in the `getBlock` test below). It has no effect on the transformer under test, but using `1` would make the fixture self-consistent — a real server would reject a v1 transaction when the caller caps at 0.

trevor-cortex (commented):
This PR closes several gaps in transaction v1 (SIMD-0385) support:

1. **`@solana/transaction-messages`**: exports the previously-orphaned `v1-transaction-config` module (making `setTransactionMessageConfig`, `V1TransactionConfig`, and the bit-mask helpers reachable) and adds/corrects docs.
2. **`@solana/rpc-types` / `@solana/rpc-api`**: adds an optional `transactionConfig` to the transaction message response shapes, with a new exported `TransactionConfig` type and typetests.
3. **`@solana/rpc-transformers`**: allow-lists the three `u32` config fields in `messageConfig` so they stay JS numbers while `priorityFee` upcasts to `bigint`/`Lamports`.
4. **`@solana/rpc-graphql`**: mirrors the shape as `TransactionMessageConfig` on `TransactionMessage`.

**Verified:**

- The doc corrections match SIMD-0385: an unset compute-unit limit or loaded-accounts-data-size limit on v1 resolves to **zero** (unlike the legacy 200k-CU-per-instruction fallback), and an unset heap size defaults to 32KiB. The 200k claim being removed was indeed wrong for v1.
- The `messageConfig` additions in `rpc-transformers` automatically propagate to `getBlock`, `getTransaction`, and `getTransactionsForAddress` via the keypath mapping in `rpc-api/src/index.ts` — no per-method wiring was missed, and the new tests cover both `getTransaction` and `getBlock`.
- The `transactionConfig?: TransactionConfig` field is correctly optional (absent on legacy/v0), and the typetests pin the `number` vs `Lamports` split.
- Changeset covers all five packages; `minor` is appropriate for new API surface.

- The field-name split is intentional and correct: the client-side builder type (`V1TransactionConfig`) uses `priorityFeeLamports`, while the RPC response type (`TransactionConfig`) uses `priorityFee` to mirror the server's JSON. Don't flag it as an inconsistency.
- `rpc-graphql` gets the new type via default resolvers (plain field pass-through), which should just work since the GraphQL client reuses the transformed RPC response — but there's no rpc-graphql test exercising `transactionConfig`. If a follow-up is planned, a query-level test there would confirm the `BigInt`/number split survives the resolver layer.

Two minor inline nits below; neither blocks. LGTM.

*(Submitting as COMMENT — I don't have approve permissions on this repo.)*

packages/rpc-api/src/__tests__/allowed-numeric-keypaths-test.ts
amilz: 2bf1e2d

mcintyre94 (approved):

PR #1957, opened 2026-08-20
Fix toArrayBuffer returning the whole backing buffer for offset-zero views

mcintyre94:
`toArrayBuffer` is a function that converts a `Uint8Array` to a `ArrayBuffer`, and then returns an (optional) offset/length slice. If the offset/length encompasses the entire array then the buffer should correspond to the entire array. It does this by using `bytes.buffer` to get the underlying buffer.

It has a short-circuit path, when the requested slice covers all of the buffer. In this case it does not do `buffer.slice` and just returns buffer. But this check was comparing against the **bytes** length instead of the buffer, and then returning the buffer.

The problem is that if we do `arr.subarray(0,3)`, this returns a `Uint8Array` with length 3, backed by a buffer with the original length:

```ts
> const arr = new Uint8Array([1,2,3,4,5])
> arr.buffer
ArrayBuffer { [Uint8Contents]: <01 02 03 04 05>, [byteLength]: 5 }
> arr.subarray(0,3).buffer
ArrayBuffer { [Uint8Contents]: <01 02 03 04 05>, [byteLength]: 5 }
```

This means that if we passed something like `decodedTransaction.subarray(0, len - 64*numSigs)` and requested the whole array as a buffer, it would return `decodedtransaction.buffer`, ie the entire decoded transaction including the signatures.

The fast path now compares offsets and lengths against the **buffer** length, instead of the **input bytes**. 

This fixes a bug when decoding v1 transactions, where an app might use subarray to account for the signatures being at the end.

This is a pre-existing bug, but was not triggered by the same logic for legacy or v0 transactions because they have signatures first, so `bytesOffset` would be non-zero if you wanted to skip the signatures.

### DIFF
--- a/.changeset/long-words-obey.md
+++ b/.changeset/long-words-obey.md
@@ -1,0 +1,5 @@
+---
+'@solana/codecs-core': patch
+---
+
+Fix `toArrayBuffer` returning the entire backing buffer when given a `Uint8Array` view that starts at byte offset zero but is shorter than its underlying `ArrayBuffer`. This caused `signBytes` and `verifySignature` to operate on the wrong bytes — and `getBase64Decoder().decode()` to include trailing data — for such views, most notably the `messageBytes` of a decoded version 1 transaction, whose wire envelope places the message first and the signatures last
--- a/packages/codecs-core/src/__tests__/array-buffers-test.ts
+++ b/packages/codecs-core/src/__tests__/array-buffers-test.ts
@@ -21,6 +21,18 @@ describe('toArrayBuffer', () => {
             expect(toArrayBuffer(byteArray, offset, length)).toBe(byteArray.buffer);
         },
     );
+    it('returns a new buffer when the byte array is an offset-zero view over a larger buffer', () => {
+        const byteArray = new Uint8Array([1, 2, 3, 4, 5]).subarray(0, 3);
+        const arrayBuffer = toArrayBuffer(byteArray);
+        expect(arrayBuffer).not.toBe(byteArray.buffer);
+        expect(new Uint8Array(arrayBuffer)).toStrictEqual(new Uint8Array([1, 2, 3]));
+    });
+    it('returns a new buffer when the specified slice encompasses an entire offset-zero view over a larger buffer', () => {
+        const byteArray = new Uint8Array([1, 2, 3, 4, 5]).subarray(0, 3);
+        const arrayBuffer = toArrayBuffer(byteArray, 0, 3);
+        expect(arrayBuffer).not.toBe(byteArray.buffer);
+        expect(new Uint8Array(arrayBuffer)).toStrictEqual(new Uint8Array([1, 2, 3]));
+    });
     it.each([
         [-3, 0],
         [-3, 1],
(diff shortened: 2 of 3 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
This PR fixes a real correctness bug in `toArrayBuffer`: the fast path that returns the backing buffer unsliced was comparing the requested slice against the **view's** length/offset instead of the **buffer's**, so an offset-zero `subarray` view over a larger buffer (e.g. the `messageBytes` of a decoded v1 transaction, where signatures trail the message) would leak the entire backing buffer, including bytes outside the view.

The fix is minimal and correct. The fast path now fires only when the requested slice is exactly `[0, buffer.byteLength)`, which is precisely the case where `buffer.slice(bytesOffset, bytesOffset + bytesLength)` would reproduce the whole buffer anyway — so the identity-return optimization is now semantically equivalent to the slow path (modulo copy vs. alias, which the existing tests pin down deliberately). I traced the edge cases:

- Plain array, full-slice request (`[0, 3]` / `[-3, 3]` on a 3-byte array): unchanged, fast path still fires (`buffer.byteLength === bytes.byteLength` when the view covers the whole buffer).
- Offset-zero view shorter than buffer, with or without explicit offset/length args: now correctly takes the slow path and copies. Both new tests cover this, and they assert content, not just identity. 👍
- `SharedArrayBuffer` branch: there `buffer` is a fresh copy sized `bytes.length`, so `buffer.byteLength` equals the old `bytes.byteLength` comparison — no behavior change on that path.
- Changeset: patch bump on `@solana/codecs-core` with a clear description; matches the repo's semver policy for fixes.

Two pre-existing issues I noticed while verifying, both out of scope for this PR but possibly worth follow-ups:

1. **SAB views with non-zero `byteOffset` are broken.** In the `SharedArrayBuffer` branch (lines 16–18 of `array-buffers.ts`), `buffer` is a copy of the *view's* bytes only, but `bytesOffset` still includes `bytes.byteOffset`. So for a SAB-backed view at, say, offset 4, the code slices the already-offset-adjusted copy by 4 again, returning truncated/empty garbage. This predates this PR and is untouched by it.
2. **Negative-offset semantics are murky in the slow path.** `buffer.slice(bytesOffset, bytesOffset + bytesLength)` with a negative `bytesOffset` produces an end index that crosses zero (e.g. `slice(-1, 0)` → empty buffer), so most negative offset/length combinations return empty buffers rather than tail slices. The existing `it.each` tests only assert `not.toBe(buffer)`, so they don't catch this. Again pre-existing; this PR faithfully preserves the one tested negative case (`[-3, 3]` on a full-width array).

Nothing blocking — LGTM (would approve, but I don't have approval rights on this repo). Nice catch and a tidy fix.

mcintyre94:
> Two pre-existing issues I noticed while verifying, both out of scope for this PR but possibly worth follow-ups:

Will open issues for both of these, worth fixing but not v8 (or v1 transaction) blockers

lorisleiva (approved):
Well spotted, thank you!

PR #1958, opened 2026-08-20
Fix getTransactionSizeLimit misclassifying single-signer legacy transactions as v1

mcintyre94:
`getTransactionSizeLimit` identifies the transaction version by looking at the first byte:

- for legacy, it's the number of signatures, <128
- for v0 it's 128 (0 | 0x80)
- for v1 it's 129 (1 | 0x80)

The byte arithmetic used was `versionByte & 0b01111111` (ie 127)

But this doesn't work for a legacy transaction with a single signer, because their first byte is 1 for 1 signer:

```ts
> 1 & 0b01111111
1
```

This means that the assertions treated a single signer legacy transaction as a v1 transaction, and would allow a legacy transaction with one signature to have >1232 bytes. 

The check now mirrors the envelope detection in the transaction codec: a transaction is only treated as v1 when the version flag (high bit) of the first message byte is set and the version number is 1.

Note that `getTransactionMessageSizeLimit` (message-based) was unaffected — it switches on `transactionMessage.version` directly.

### DIFF
--- a/.changeset/clear-rings-taste.md
+++ b/.changeset/clear-rings-taste.md
@@ -1,0 +1,5 @@
+---
+'@solana/transactions': patch
+---
+
+Fix `getTransactionSizeLimit` misclassifying single-signer legacy transactions as version 1. A legacy message has no version byte, so its first byte is the number of required signatures — `1` for every single-signer transaction — which was mistaken for a v1 version byte. As a result, `isTransactionWithinSizeLimit`, `assertIsTransactionWithinSizeLimit`, and `assertIsSendableTransaction` accepted legacy transactions of up to 4096 bytes that the network rejects at 1232 bytes. The version flag (high bit) of the first message byte is now required to be set before treating a transaction as version 1
--- a/packages/transactions/src/__tests__/transaction-size-test.ts
+++ b/packages/transactions/src/__tests__/transaction-size-test.ts
@@ -13,6 +13,7 @@ import { compileTransaction } from '../compile-transaction';
 import {
     assertIsTransactionWithinSizeLimit,
     getTransactionSize,
+    getTransactionSizeLimit,
     isTransactionWithinSizeLimit,
 } from '../transaction-size';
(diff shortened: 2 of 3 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
LGTM — clear, well-scoped bug fix. (Submitting as a comment since I don't have approve permissions on this repo.)

The old check `(firstByte & 0b01111111) === 1` matched any single-signer legacy transaction (whose first byte is `num_required_signatures = 1`), granting it the 4096-byte v1 limit instead of 1232. The new check requires the version flag (high bit) to be set before comparing the version number, which exactly mirrors the envelope detection in `getTransactionVersionDecoder` (`packages/transaction-messages/src/codecs/transaction-version.ts`) — the source of truth for the wire format.

Verified:

- The logic matches the codec: legacy messages never have the high bit set (signature counts are single-byte shortvec values < 128), so `firstByte & 0x80` is a reliable discriminator.
- Test coverage is thorough: the previously-broken case (single-signer legacy over the legacy limit) is covered across `getTransactionSizeLimit`, `isTransactionWithinSizeLimit`, and `assertIsTransactionWithinSizeLimit`, plus a new direct `getTransactionSizeLimit` describe block covering legacy/v0/v1.
- The expected `transactionSize: 1403` in the new assert test is consistent with the v0 equivalent's 1405 (legacy drops the version byte and the empty address-table-lookup count byte).
- Changeset is present with a `patch` bump, which fits a bug fix, and the description accurately explains the user-facing impact (including `assertIsSendableTransaction`).

Notes for other reviewers:

- Versioned-but-unsupported first bytes (e.g. `0x82` for v2) still fall through to the legacy limit rather than throwing, same as before this PR. That's fine here — such bytes can't be produced by `compileTransaction`, and `getTransactionSizeLimit` isn't a validity check — but worth knowing the function is intentionally permissive rather than delegating to `getTransactionVersionDecoder`, which would throw on unsupported versions.
- `getTransactionMessageSizeLimit` is confirmed unaffected: it switches on `transactionMessage.version` directly and never inspects bytes.

lorisleiva (approved):
Oof good catch, thank you!

PR #1968, opened 2026-08-21
Revert `fumadocs-core` bump to 16.14.4 in /docs

mcintyre94:
Reverts #1965, which broke the docs build on `main` ([failing run](https://github.com/anza-xyz/kit/actions/runs/32461788761/job/96710009248)).

As an aside, this is an issue because we made the preview docs job optional, ie it only runs on PRs when docs/ changes now. And so it's not required to pass to merge a PR. It failed, but the PR merged. See https://github.com/anza-xyz/kit/actions/runs/32461788761/job/96710009248

AFAIK Github doesn't have a way to say "it only runs if X, but if it does run then it must pass", which is what we really want. 

Claude explanation below:

--

The fumadocs packages can only be upgraded together: `fumadocs-ui@15.2.10` pins `fumadocs-core@15.2.10` exactly, `fumadocs-mdx@12` only supports core `^14 || ^15`, and fumadocs v16 removed the `fumadocs-core/server` entrypoint that the docs pages import `getPageTreePeers` from. Bumping core alone therefore cannot work. Moving to fumadocs v16 is a coordinated migration (`fumadocs-ui` 15→16, `fumadocs-mdx` 12→15, plus code changes across the docs layouts) that should be done as its own PR — the docs already run Next 16 and React 19.2, so the peers are satisfiable when we get to it.

A second commit updates the dependabot config so this cannot recur: fumadocs packages are grouped into a single PR, and major fumadocs updates in /docs are ignored until the v16 migration happens.

Verified locally that `pnpm install --frozen-lockfile` and a full docs build pass after the revert.

No changeset: docs dependency + CI config changes only.


mcintyre94:
@trevor-cortex

docs/pnpm-lock.yaml
```diff
@@ -23,17 +23,30 @@ importers:
         specifier: ^0.7.1
         version: 0.7.1
       fumadocs-core:
+<<<<<<< HEAD
         specifier: 16.14.4
         version: 16.14.4(@mdx-js/mdx@3.1.1)(@types/estree-jsx@1.0.5)(@types/hast@3.0.5)(@types/mdast@4.0.4)(@types/react@19.2.18)(lucide-react@1.34.0(react@19.2.8))(next@16.3.1(@babel/core@7.28.5)(@types/node@25.4.0)(react-dom@19.2.8(react@19.2.8))(react@19.2.8))(react-dom@19.2.8(react@19.2.8))(react@19.2.8)(zod@4.2.1)
+=======
+        specifier: 15.2.10
+        version: 15.2.10(@types/react@19.2.18)(next@16.3.1(@babel/core@7.28.5)(@types/node@25.4.0)(react-dom@19.2.8(react@19.2.8))(react@19.2.8))(react-dom@19.2.8(react@19.2.8))(react@19.2.8)
+>>>>>>> dd702c907 (Revert "Bump fumadocs-core from 15.2.10 to 16.14.4 in /docs (#1965)")
```
trevor-cortex: Unresolved merge conflict markers — this makes the lockfile invalid YAML and `pnpm install --frozen-lockfile` will fail. There are at least five more conflicted blocks further down in this file (`fumadocs-docgen`, `npm-to-yarn`, `picomatch`, `remark-rehype`, `unist-util-visit`).

Rather than resolving these by hand, I'd regenerate the lockfile: check out the reverted `docs/package.json`, run `pnpm install` in `docs/`, and commit the result. Hand-merging a pnpm lockfile is very likely to leave the resolution graph inconsistent (see my comment on `package.json` — the `fumadocs-docgen` importer entry is already stale).

trevor-cortex (commented):
This PR reverts the standalone `fumadocs-core` 15.2.10 → 16.14.4 bump from #1965 (which broke the docs build because the fumadocs packages must move in lockstep, and v16 removed the `fumadocs-core/server` entrypoint the docs import from), and hardens the dependabot config so it can't recur: fumadocs packages are grouped into a single PR, and major fumadocs updates are ignored until a coordinated v16 migration.

The revert direction and the dependabot changes are exactly right. However, **the lockfile as pushed is broken and this cannot merge as-is** (I'd request changes, but I don't have that permission on this repo):

1. **`docs/pnpm-lock.yaml` contains unresolved merge conflict markers** (`<<<<<<< HEAD` / `=======` / `>>>>>>> dd702c907`) in at least six places — the importers block, `fumadocs-core@16.14.4`/`fumadocs-docgen@2.1.0`, `npm-to-yarn`, `picomatch`, `remark-rehype`, and `unist-util-visit`. The file is not valid YAML, so `pnpm install --frozen-lockfile` will fail in CI. The PR description says a frozen-lockfile install and full docs build were verified locally, so this looks like a rebase/force-push that went wrong *after* that verification — the working copy that was tested is not what got pushed.

2. **Even the non-conflicted parts of the lockfile are stale.** The importers block still records `fumadocs-docgen` with `specifier: ^3.1.0` resolving to `3.1.0(...fumadocs-core@16.14.4...)`, while `docs/package.json` now says `^2.1.0`. Similarly `fumadocs-twoslash` still resolves against `fumadocs-core@16.14.4` in the retained (HEAD) side. Hand-resolving these conflicts is not the right fix — please check out the reverted `package.json`, regenerate the lockfile via `pnpm install` in `docs/`, push the result, then confirm `pnpm install --frozen-lockfile` passes on the pushed SHA.

- The `.github/dependabot.yml` and `docs/package.json` changes themselves look correct and match the PR description. The trailing-newline fix on `package.json` is a nice touch.
- No changeset is correctly omitted here — docs dependency + CI config only, nothing published.
- Once the lockfile is repushed, it's worth re-checking that the CI docs build actually runs on this PR (not just install) before merging, since that's the failure #1965 slipped past.

docs/package.json
```diff
@@ -14,9 +14,9 @@
         "@inkeep/cxkit-react": "^0.5.119",
         "@radix-ui/react-popover": "^1.1.23",
         "class-variance-authority": "^0.7.1",
-        "fumadocs-core": "16.14.4",
-        "fumadocs-docgen": "^3.1.0",
-        "fumadocs-mdx": "15.3.1",
+        "fumadocs-core": "15.2.10",
+        "fumadocs-docgen": "^2.1.0",
```
trevor-cortex, line 18: This correctly reverts `fumadocs-docgen` to `^2.1.0`, but the lockfile's importers block (outside the conflicted regions) still records `specifier: ^3.1.0` resolving to `fumadocs-docgen@3.1.0` against `fumadocs-core@16.14.4`. Another sign the pushed lockfile isn't the one that was verified locally — regenerating it should fix both.

.github/dependabot.yml
```diff
@@ -40,7 +40,21 @@ updates:
...
+      # The fumadocs packages must be upgraded to v16 together (fumadocs-ui
+      # pins an exact fumadocs-core version) and the upgrade requires code
+      # changes to the docs app. Remove this once the docs are migrated to
+      # fumadocs v16.
+      - dependency-name: 'fumadocs-core'
+        update-types: ['version-update:semver-major']
+      - dependency-name: 'fumadocs-ui'
+        update-types: ['version-update:semver-major']
+      - dependency-name: 'fumadocs-mdx'
+        update-types: ['version-update:semver-major']
```
trevor-cortex, line 53: Consider adding `fumadocs-docgen` to this ignore list too. It was also reverted a major (`^3.1.0` → `^2.1.0`) because docgen 3.x resolves against `fumadocs-core@16`, so a solo major docgen bump from dependabot could reproduce the same class of breakage the comment above describes. The `fumadocs*` group helps but grouping alone doesn't prevent a major bump PR from being opened.

(Minor: worth double-checking whether `fumadocs-twoslash` majors have the same coupling — it's currently on 3.x which supports both, so probably fine to leave out.)

mcintyre94:
Oops messed up rebasing

@trevor-cortex

trevor-cortex:
No worries — the repush looks good. I re-checked the branch:

- **No conflict markers** in `docs/pnpm-lock.yaml` anymore, and the lockfile is structurally valid again.
- **Importers block is now consistent with `package.json`**: `fumadocs-core` → 15.2.10, `fumadocs-docgen` → `^2.1.0` resolved to 2.1.0, `fumadocs-mdx` → 12.0.3, `fumadocs-ui` → 15.2.10, and `fumadocs-twoslash` now resolves against core 15 instead of 16. The stale `fumadocs-core@16.14.4` resolution graph from the previous push is gone.
- The larger diff footprint (+360/−833 vs the original +118/−137) is what you'd expect from a proper `pnpm install` regeneration rather than a hand-merge — the extra churn is just transitive deps (esbuild, emnapi, etc.) settling back to their pre-#1965 resolutions.
- The dependabot config also now includes the `fumadocs-docgen` major-update ignore I mentioned last time. 👍

Only remaining thing is to let CI confirm `pnpm install --frozen-lockfile` + the docs build on this exact push — that's the definitive check the previous push failed. Assuming that's green, this looks ready to merge.

.github/dependabot.yml
```diff
@@ -40,7 +40,23 @@ updates:
...
+      - dependency-name: 'fumadocs-docgen'
+        update-types: ['version-update:semver-major']
```
lorisleiva, line 55: Do we need this now that we have the fumadocs group? We can just ignore the fumadocs major bump PR from dependabot and make the changes inside that PR in the future when we want to tackle it?

lorisleiva (approved):

.github/dependabot.yml
mcintyre94: The problem is that if dependabot opens that PR then it will pass CI (docs build will fail but not block CI because it's optional), and then get auto merged.

I think the better fix would be to make preview-docs run unconditionally, and do the "did docs change?" check ourselves within that job, with a no-op success if nothing changed. Then we can make it block PRs again. It feels really dumb that Github can't do "this job sometimes runs, when it does it must pass" but it really doesn't seem to be possible. https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks just says "Avoid requiring workflows that can be skipped."

I can look into that, but I think it's probably worth landing this PR for now to get green CI on main, we need the revert from here either way.

Issue #1960, opened 2026-08-20
toArrayBuffer: Negative-offset semantics are confusing/probably wrong


mcintyre94:
Issue flagged by Trevor here: https://github.com/anza-xyz/kit/pull/1957#pullrequestreview-4982438072

> Negative-offset semantics are murky in the slow path. buffer.slice(bytesOffset, bytesOffset + bytesLength) with a negative bytesOffset produces an end index that crosses zero (e.g. slice(-1, 0) → empty buffer), so most negative offset/length combinations return empty buffers rather than tail slices. The existing it.each tests only assert not.toBe(buffer), so they don't catch this.

It seems like this is probably not behaving as we'd expect. Should probably try to make this behave exactly the same as native `slice` in this case.

edycutjong:
I'd like to take this one.

The failure is specifically when `bytesOffset` is negative but `bytesOffset + bytesLength` reaches zero or beyond, so the end index crosses zero: `toArrayBuffer(new Uint8Array([1,2,3,4,5]), -1, 1)` becomes `slice(-1, 0)` and returns an empty buffer instead of `[5]`. Offsets whose end index stays negative, like `(-3, 1)`, happen to return the right bytes, which is probably why this went unnoticed.

My approach is to resolve the offset against the end of the buffer before deriving the end index, the way `Array.prototype.slice` does, which also subsumes the existing `bytesOffset === -buffer.byteLength` special case since that normalises to `0`.

On the tests: the current `it.each` block does already cover `(-1, 1)` and `(-2, 1)`, but only asserts `.not.toBe(byteArray.buffer)`, so it passes against the broken behaviour. I've added cases that assert the returned contents and check parity with native `slice`. They fail on `main` and pass with the fix.

One coordination note — #1970 is open against the same function for #1959. The two changes are independent (it touches the `SharedArrayBuffer` branch, this touches the return), but they overlap textually, so I'm happy to rebase on top of it once it lands rather than have it rebase on me.

PR #1970, opened 2026-08-23
Fix SharedArrayBuffer view slicing

### LINKED ISSUE #1959: toArrayBuffer: SharedArrayBuffer with non-zero bytesOffset is broken
Pre-existing issue flagged by Trevor here: https://github.com/anza-xyz/kit/pull/1957#pullrequestreview-4982438072

> SAB views with non-zero byteOffset are broken. In the SharedArrayBuffer branch (lines 16–18 of array-buffers.ts), buffer is a copy of the view's bytes only, but bytesOffset still includes bytes.byteOffset. So for a SAB-backed view at, say, offset 4, the code slices the already-offset-adjusted copy by 4 again, returning truncated/empty garbage. This predates this PR and is untouched by it.

Relevant code:

https://github.com/anza-xyz/kit/blob/213bcdbf467556bcc6367231d274099b3f860fcd/packages/codecs-core/src/array-buffers.ts#L16-L18

Here it's only writing `bytes` (which would be the subarray) into the buffer, and then later we do `buffer.slice(bytesOffset, ...)` which will re-apply the offset

koriyoshi2041:
#### Problem

`toArrayBuffer` copies a SharedArrayBuffer-backed view into a new zero-offset ArrayBuffer, but then slices that copy using the original view's `byteOffset`. A view beginning after byte zero therefore has its offset applied twice and returns truncated or empty data.

#### Summary of Changes

- Treat the copied SharedArrayBuffer view as a zero-offset buffer before applying the optional caller-provided slice.
- Add regressions for both a whole non-zero-offset view and an explicit slice relative to that view.
- Add the patch changeset for `@solana/codecs-core`.

Local verification:

```
pnpm --filter @solana/codecs-core test:unit:node -- --runInBand src/__tests__/array-buffers-test.ts
pnpm --filter @solana/codecs-core test:lint
pnpm --filter @solana/codecs-core test:prettier
pnpm turbo compile:typedefs --filter=@solana/codecs-core...
pnpm --filter @solana/codecs-core test:typecheck
pnpm changeset status
git diff --check
```

Fixes #1959

### DIFF
--- a/packages/codecs-core/src/array-buffers.ts
+++ b/packages/codecs-core/src/array-buffers.ts
@@ -8,7 +8,7 @@ import { ReadonlyUint8Array } from './readonly-uint8array';
  * Source: https://stackoverflow.com/questions/37228285/uint8array-to-arraybuffer
  */
 export function toArrayBuffer(bytes: ReadonlyUint8Array | Uint8Array, offset?: number, length?: number): ArrayBuffer {
-    const bytesOffset = bytes.byteOffset + (offset ?? 0);
+    let bytesOffset = bytes.byteOffset + (offset ?? 0);
     const bytesLength = length ?? bytes.byteLength;
     let buffer: ArrayBuffer;
     if (typeof SharedArrayBuffer === 'undefined') {
@@ -16,6 +16,7 @@ export function toArrayBuffer(bytes: ReadonlyUint8Array | Uint8Array, offset?: n
     } else if (bytes.buffer instanceof SharedArrayBuffer) {
         buffer = new ArrayBuffer(bytes.length);
         new Uint8Array(buffer).set(new Uint8Array(bytes));
+        bytesOffset = offset ?? 0;
     } else {
         buffer = bytes.buffer;
     }
--- a/.changeset/calm-buffers-slice.md
+++ b/.changeset/calm-buffers-slice.md
@@ -1,0 +1,5 @@
+---
+'@solana/codecs-core': patch
+---
+
+Fix `toArrayBuffer` slicing SharedArrayBuffer-backed views from a non-zero byte offset.
--- a/packages/codecs-core/src/__tests__/array-buffers-test.ts
+++ b/packages/codecs-core/src/__tests__/array-buffers-test.ts
@@ -7,6 +7,24 @@ describe('toArrayBuffer', () => {
         const arrayBuffer = toArrayBuffer(byteArray);
         expect(arrayBuffer).not.toBeInstanceOf(SharedArrayBuffer);
     });
+    it('copies a non-zero-offset shared array buffer view without applying its offset twice', () => {
+        const sharedArrayBuffer = new SharedArrayBuffer(5);
+        const byteArray = new Uint8Array(sharedArrayBuffer);
+        byteArray.set([1, 2, 3, 4, 5]);
+
+        const arrayBuffer = toArrayBuffer(byteArray.subarray(2, 5));
+
+        expect(new Uint8Array(arrayBuffer)).toStrictEqual(new Uint8Array([3, 4, 5]));
+    });
+    it('applies an explicit slice relative to a non-zero-offset shared array buffer view', () => {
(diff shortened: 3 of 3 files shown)


mcintyre94:
@trevor-cortex

packages/codecs-core/src/array-buffers.ts
```diff
@@ -8,14 +8,15 @@ import { ReadonlyUint8Array } from './readonly-uint8array';
...
-    const bytesOffset = bytes.byteOffset + (offset ?? 0);
+    let bytesOffset = bytes.byteOffset + (offset ?? 0);
     const bytesLength = length ?? bytes.byteLength;
     let buffer: ArrayBuffer;
     if (typeof SharedArrayBuffer === 'undefined') {
         buffer = bytes.buffer as ArrayBuffer;
     } else if (bytes.buffer instanceof SharedArrayBuffer) {
         buffer = new ArrayBuffer(bytes.length);
         new Uint8Array(buffer).set(new Uint8Array(bytes));
+        bytesOffset = offset ?? 0;
```
trevor-cortex, line 19: Correct fix — the copy above only contains the view's own bytes, so it's inherently zero-offset and re-adding `bytes.byteOffset` was the bug.

One non-blocking note: negative `offset` values now behave differently between the SAB branch and the non-shared branches. For a regular view, `bytes.byteOffset + offset` lets a negative offset reach *before* the view into the underlying buffer; for a SAB-backed view those preceding bytes were never copied, so a negative offset falls into `ArrayBuffer.prototype.slice`'s end-relative semantics instead (e.g. `slice(-1, 0)` → empty). This asymmetry pre-dates this PR and is an inherent consequence of the copy-only-the-view approach, so I don't think it needs addressing here — just flagging it in case negative offsets on views ever become a supported path worth documenting or asserting on.

trevor-cortex (commented):
## Summary

Fixes a double-offset bug in `toArrayBuffer` for `SharedArrayBuffer`-backed views. The SAB branch copies only the view's own bytes into a fresh zero-offset `ArrayBuffer`, but the subsequent slice still used `bytes.byteOffset + (offset ?? 0)` — applying the view's offset a second time and truncating (or emptying) the result. The fix resets `bytesOffset` to `offset ?? 0` in the copy branch, adds two regression tests (whole non-zero-offset view, and an explicit slice relative to such a view), and includes a `patch` changeset for `@solana/codecs-core`.

**Verdict: LGTM** (submitting as a comment since I don't have approval permissions on this repo).

## Verification

I traced both regressions against the fixed logic and they check out:

- `subarray(2, 5)` with no slice args → copy is `[3, 4, 5]`, `bytesOffset = 0`, `bytesLength = 3 = buffer.byteLength`, so the fast path returns the copy directly. ✅
- `subarray(1, 5)` with `(1, 2)` → copy is `[2, 3, 4, 5]`, `slice(1, 3)` = `[3, 4]`. ✅

The changeset follows the repo's conventions (patch bump for a fix, generated-style filename), and test-file placement/naming matches the existing suite. The non-SAB branches are untouched, so there's no behavior change for regular `ArrayBuffer`-backed views.

- The fix is intentionally minimal (a `let` + one reassignment in the SAB branch). A slightly more declarative alternative would be to compute the view-base offset once, e.g. `const bytesOffset = (isSharedArrayBufferBacked ? 0 : bytes.byteOffset) + (offset ?? 0);` — but that's purely stylistic and not worth blocking on.
- There's a pre-existing asymmetry around negative `offset` values that this PR doesn't (and needn't) resolve — see the inline note. The existing `bytesOffset === -buffer.byteLength` fast path still covers the "whole view via negative offset" case correctly after this change.

packages/codecs-core/src/array-buffers.ts
mcintyre94: Agreed, negative offset tracked separately: https://github.com/anza-xyz/kit/issues/1960

mcintyre94 (approved):
Thanks for this!

PR #1971, opened 2026-08-23
Preserve fixedSize for single-field struct codecs

### LINKED ISSUE #1738: getStructEncoder/Decoder/Codec do not infer a literal fixedSize from their fields

shin4141:
#### Problem

Single fixed-size field structs currently widen their literal `fixedSize` to `number`.

#### Context

A broader type-arithmetic solution was explored in #1761. Maintainer feedback identified exactly-one-fixed-size-field support as a simpler middle ground.

#### Summary of Changes

Add narrow overloads that forward the only field's existing literal `fixedSize` for `getStructEncoder`, `getStructDecoder`, and `getStructCodec`. Add focused typetests and a patch changeset for `@solana/codecs-data-structures`.

Multi-field structs intentionally retain their existing `number` behavior. This introduces no type arithmetic and no runtime behavior changes.

#### Tests

- `pnpm --filter @solana/codecs-data-structures test:typecheck`
- `pnpm turbo compile:typedefs --filter=@solana/codecs-data-structures...`
- `pnpm --filter @solana/codecs-data-structures test:lint`
- `pnpm --filter @solana/codecs-data-structures test:prettier`
- `pnpm --filter @solana/codecs-data-structures test:unit:node`
- `pnpm --filter @solana/codecs-data-structures test:unit:browser`
- `pnpm --filter @solana/codecs-data-structures compile:js`, with Node, browser, and native runtime bundles byte-identical to the pristine build
- Generated declaration consumer typecheck
- `pnpm changeset status`
- `git diff --check`

#### AI assistance

AI assistance was used during implementation and verification. Shin reviewed the final change and owns the submission.

Fixes #1738

### DIFF
--- a/packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
+++ b/packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
@@ -11,6 +11,8 @@ import { getUtf8Codec, getUtf8Decoder, getUtf8Encoder } from '@solana/codecs-str
 
 import { getStructCodec, getStructDecoder, getStructEncoder } from '../struct';
 
+type IsWidenedNumber<T extends number> = number extends T ? true : false;
+
 {
     // [getStructEncoder]: It knows if the encoder is fixed size or variable size.
     getStructEncoder([
@@ -22,6 +24,7 @@ import { getStructCodec, getStructDecoder, getStructEncoder } from '../struct';
         ['age', {} as FixedSizeEncoder<number>],
     ]) satisfies VariableSizeEncoder<{ age: number; name: string }>;
     getStructEncoder([['age', getU32Encoder()]]) satisfies FixedSizeEncoder<{ age: number }>;
+    getStructEncoder([['age', getU32Encoder()]]) satisfies FixedSizeEncoder<{ age: bigint | number }, 4>;
     getStructEncoder([['name', getUtf8Encoder()]]) satisfies VariableSizeEncoder<{ name: string }>;
 }
 
@@ -36,6 +39,7 @@ import { getStructCodec, getStructDecoder, getStructEncoder } from '../struct';
         ['age', {} as FixedSizeDecoder<number>],
     ]) satisfies VariableSizeDecoder<{ age: number; name: string }>;
     getStructDecoder([['age', getU32Decoder()]]) satisfies FixedSizeDecoder<{ age: number }>;
+    getStructDecoder([['age', getU32Decoder()]]) satisfies FixedSizeDecoder<{ age: number }, 4>;
     getStructDecoder([['name', getUtf8Decoder()]]) satisfies VariableSizeDecoder<{ name: string }>;
 }
 
@@ -50,7 +54,73 @@ import { getStructCodec, getStructDecoder, getStructEncoder } from '../struct';
         ['age', {} as FixedSizeCodec<number>],
     ]) satisfies VariableSizeCodec<{ age: number; name: string }>;
     getStructCodec([['age', getU32Codec()]]) satisfies FixedSizeCodec<{ age: number }>;
+    getStructCodec([['age', getU32Codec()]]) satisfies FixedSizeCodec<{ age: bigint | number }, { age: number }, 4>;
     getStructCodec([['name', getUtf8Codec()]]) satisfies VariableSizeCodec<{ name: string }>;
+}
+
+{
+    // [getStructEncoder/Decoder/Codec]: It only preserves literal sizes for exactly one fixed-size field.
+    {
+        // It forwards a second literal size without hard-coding the U32 size.
+        getStructEncoder([['value', {} as FixedSizeEncoder<number, 8>]]) satisfies FixedSizeEncoder<
+            { value: number },
+            8
+        >;
+        getStructDecoder([['value', {} as FixedSizeDecoder<number, 8>]]) satisfies FixedSizeDecoder<
+            { value: number },
+            8
+        >;
+        getStructCodec([['value', {} as FixedSizeCodec<number, number, 8>]]) satisfies FixedSizeCodec<
+            { value: number },
+            { value: number },
+            8
+        >;
+    }
+
+    {
+        // Two fixed-size fields stay widened; their sizes are not added by the type system.
+        const encoder = getStructEncoder([
+            ['a', {} as FixedSizeEncoder<number, 4>],
+            ['b', {} as FixedSizeEncoder<number, 8>],
+        ]);
+        const decoder = getStructDecoder([
+            ['a', {} as FixedSizeDecoder<number, 4>],
+            ['b', {} as FixedSizeDecoder<number, 8>],
+        ]);
+        const codec = getStructCodec([
+            ['a', {} as FixedSizeCodec<number, number, 4>],
+            ['b', {} as FixedSizeCodec<number, number, 8>],
+        ]);
+        [encoder.fixedSize, decoder.fixedSize, codec.fixedSize] satisfies number[];
+        true satisfies IsWidenedNumber<typeof encoder.fixedSize>;
+        true satisfies IsWidenedNumber<typeof decoder.fixedSize>;
+        true satisfies IsWidenedNumber<typeof codec.fixedSize>;
+    }
+
+    {
+        // Non-tuple field arrays stay widened even when their items have a literal size.
+        const encoderFields: readonly (readonly [string, FixedSizeEncoder<number, 4>])[] = [];
+        const decoderFields: readonly (readonly [string, FixedSizeDecoder<number, 4>])[] = [];
+        const codecFields: readonly (readonly [string, FixedSizeCodec<number, number, 4>])[] = [];
+        const encoder = getStructEncoder(encoderFields);
+        const decoder = getStructDecoder(decoderFields);
+        const codec = getStructCodec(codecFields);
+        [encoder.fixedSize, decoder.fixedSize, codec.fixedSize] satisfies number[];
+        true satisfies IsWidenedNumber<typeof encoder.fixedSize>;
+        true satisfies IsWidenedNumber<typeof decoder.fixedSize>;
+        true satisfies IsWidenedNumber<typeof codec.fixedSize>;
+    }
+
+    {
+        // Empty structs preserve their existing widened fixed-size classification.
+        const encoder = getStructEncoder([]);
+        const decoder = getStructDecoder([]);
+        const codec = getStructCodec([]);
+        [encoder.fixedSize, decoder.fixedSize, codec.fixedSize] satisfies number[];
+        true satisfies IsWidenedNumber<typeof encoder.fixedSize>;
+        true satisfies IsWidenedNumber<typeof decoder.fixedSize>;
+        true satisfies IsWidenedNumber<typeof codec.fixedSize>;
+    }
 }
 
 {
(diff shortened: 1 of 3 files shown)


packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
```diff
@@ -11,6 +11,8 @@ import { getUtf8Codec, getUtf8Decoder, getUtf8Encoder } from '@solana/codecs-str
 
 import { getStructCodec, getStructDecoder, getStructEncoder } from '../struct';
 
+type IsWidenedNumber<T extends number> = number extends T ? true : false;
```
mcintyre94: I appreciate what you're doing here, we want to check that the type is exactly `Number` and not a specific fixed size literal.

We have precedent for this with `Equal`: https://github.com/anza-xyz/kit/blob/be30e328304ed055817572fb67712d79454ca12e/packages/plugin-core/src/__typetests__/client-typetest.ts#L37

Please could you copy over that and its long explanatory comment? It covers some edge cases that this one doesn't, and has a very detailed explanation

packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
```diff
@@ -50,9 +54,75 @@ import { getStructCodec, getStructDecoder, getStructEncoder } from '../struct';
         ['age', {} as FixedSizeCodec<number>],
     ]) satisfies VariableSizeCodec<{ age: number; name: string }>;
     getStructCodec([['age', getU32Codec()]]) satisfies FixedSizeCodec<{ age: number }>;
+    getStructCodec([['age', getU32Codec()]]) satisfies FixedSizeCodec<{ age: bigint | number }, { age: number }, 4>;
     getStructCodec([['name', getUtf8Codec()]]) satisfies VariableSizeCodec<{ name: string }>;
 }
 
+{
+    // [getStructEncoder/Decoder/Codec]: It only preserves literal sizes for exactly one fixed-size field.
```
mcintyre94: Please only test one of these at a time, so three different blocks for these tests

packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
```diff
@@ -50,9 +54,75 @@ import { getStructCodec, getStructDecoder, getStructEncoder } from '../struct';
...
+            ['b', {} as FixedSizeDecoder<number, 8>],
+        ]);
+        const codec = getStructCodec([
+            ['a', {} as FixedSizeCodec<number, number, 4>],
+            ['b', {} as FixedSizeCodec<number, number, 8>],
+        ]);
+        [encoder.fixedSize, decoder.fixedSize, codec.fixedSize] satisfies number[];
+        true satisfies IsWidenedNumber<typeof encoder.fixedSize>;
+        true satisfies IsWidenedNumber<typeof decoder.fixedSize>;
+        true satisfies IsWidenedNumber<typeof codec.fixedSize>;
```
mcintyre94: This should be 3 blocks too, but to demonstrate how we'd use `Equal` here, this is how I'd write the encoder test:

```ts
{
    // [getStructEncoder]: Two fixed-size fields keep a widened size; it is not narrowed to a literal.
    const encoder = getStructEncoder([
        ['a', {} as FixedSizeEncoder<number, 4>],
        ['b', {} as FixedSizeEncoder<number, 8>],
    ]);
    encoder satisfies FixedSizeEncoder<{ a: number; b: number }>;
    true satisfies Equal<typeof encoder.fixedSize, number>;
}}
```

mcintyre94 (changes requested):
Thanks for this! We need to tighten these typetests up a bit but this looks like a great solution to me!

shin4141:
Thanks! Addressed in 18d0aebe. I copied the existing `Equal` helper and explanatory comment, split the encoder, decoder, and codec cases into separate blocks, and now assert exact `number` equality for the two-field, non-tuple, and empty widened-size cases. The follow-up commit only changes `struct-typetest.ts`; production and the changeset are untouched.

Re-ran package typecheck, `compile:typedefs` (6/6), lint (65/65), Prettier (70/70), Node and browser tests (303/303 each), the five runtime bundle hash checks (unchanged), and `git diff --check`.

mcintyre94:
@trevor-cortex

packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
```diff
@@ -11,6 +11,30 @@ import { getUtf8Codec, getUtf8Decoder, getUtf8Encoder } from '@solana/codecs-str
...
+ *    to mutable). `Equal` distinguishes `{ x: T }` from `{ x?: T }` and from
+ *    `{ readonly x: T }` because the inferred-position generic comparison
+ *    uses identity rather than assignability for the type parameters.
+ *
+ * Use `Equal` when the exact shape (including modifiers) matters. Use
+ * `satisfies` when one-way assignability is the actual requirement (e.g.
+ * "this value is usable where `Disposable & X` is expected").
+ */
+type Equal<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ? 1 : 2 ? true : false;
+
```
trevor-cortex, line 37: Nit: this doc comment's rationale doesn't match how `Equal` is actually used in this file. There are no `Omit<...>` wrappers in these codec types to regress into, and none of the assertions below compare object shapes where `?`/`readonly` modifiers matter — every usage is `Equal<typeof x.fixedSize, number>`.

The real reason `Equal` is needed here is simpler and worth stating directly: `x.fixedSize satisfies number` passes even when `fixedSize` is a literal like `4` (literals are assignable to `number`), so it can't assert that widening was *preserved*. `Equal` catches accidental narrowing.

Suggest trimming this to a two-or-three-line comment saying exactly that — the current text reads like it was carried over from a different context and will confuse future readers of this file.

packages/codecs-data-structures/src/struct.ts
```diff
@@ -82,6 +82,9 @@ type GetDecoderTypeFromFields<TFields extends Fields<Decoder<any>>> = DrainOuter
  *
  * @see {@link getStructCodec}
  */
+export function getStructEncoder<const TFields extends readonly [readonly [string, FixedSizeEncoder<any>]]>(
+    fields: TFields,
+): FixedSizeEncoder<GetEncoderTypeFromFields<TFields>, TFields[0][1]['fixedSize']>;
```
trevor-cortex, line 87: Minor: inserting this overload between the JSDoc block and the previous first overload means the doc comment now attaches to this narrow single-field signature. Calls that resolve to the multi-field fixed-size overload (arguably the most common case) will resolve to a signature with no JSDoc, which may drop hover documentation in editors depending on TS server fallback behavior.

Worth a quick hover check in an editor after building typedefs. If docs do disappear, the fix is cheap: keep the JSDoc on the broad fixed-size overload and give this one its own short `/** ... */`, or duplicate the block. Same applies to the decoder and codec overloads below.

trevor-cortex (commented):
This PR adds a narrow overload to `getStructEncoder`, `getStructDecoder`, and `getStructCodec` that forwards the field's literal `fixedSize` when the struct has exactly one fixed-size field, plus focused typetests and a patch changeset. No runtime changes — overloads only.

The solution looks solid and the typetests are now nicely tightened per the prior review — this would be an approval, but I don't have approve permissions on this repo:

- **Overload resolution is correct.** The single-element-tuple constraint (`readonly [readonly [string, FixedSizeEncoder<any>]]`) only matches literal one-field tuples. Multi-field tuples, non-tuple `Fields<...>` arrays, and `[]` all fall through to the existing widened overloads — and the new `Equal`-based typetests pin each of those cases down exactly, which `satisfies number` alone couldn't do (a literal `4` would still satisfy `number`).
- **No generic-call-site regressions.** A call with a generic `TFields extends Fields<FixedSizeEncoder<any>>` doesn't match the tuple overload and resolves to the widened one, same as before.
- **Changeset level looks right.** Narrowing `number` to a literal is non-breaking for consumers (`FixedSizeEncoder<T, 4>` remains assignable to `FixedSizeEncoder<T, number>`), so patch fits.

Two minor nits inline, neither blocking: the `Equal` helper's doc comment describes rationale that doesn't match its actual use in this file, and the JSDoc attachment in `struct.ts` shifts to the new overload (worth a quick hover check in an editor).

For subsequent reviewers: the main thing to verify is that the generated `.d.ts` output keeps the overload order intact — the narrow overload must stay first for resolution to work in consuming packages. The PR author reports a declaration-consumer typecheck was run, which covers this.

packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
mcintyre94: This is fine, we're just copying it from the other use

packages/codecs-data-structures/src/struct.ts
mcintyre94: I don't think that's how Typescript works, we always use this pattern of putting the docblock before the first overload. Overload order is correct too

mcintyre94 (approved):
Awesome, thanks for this! 🚀

PR #1972, opened 2026-08-23
validate compute unit limit and heap size in transaction messages

### LINKED ISSUE #1961: Transaction messages: Validate heap size config
The heap size requested for a transaction must be a multiple of 1 KiB between 32 KiB - 256 KiB. This is a consensus/transaction sanitisation check: https://github.com/anza-xyz/agave-sdk/blob/711c4dfce6866b3db56f33b745b82b401234cbd8/transaction-view/src/sanitize.rs#L83 

We could validate this in Kit and throw a `SolanaError` if the value is invalid

This is slightly risky because the network could change this value and we'd need to adjust our constraint, but in practice the constraint is the same for v1 transaction config as it was for the compute budget instruction.

### LINKED ISSUE #1962: transaction messages: validate compute unit limit config
The max CUs that can be requested is 1.4M, but we don't validate the value set using Kit. We could do so, and throw a `SolanaError` before it becomes a failed transaction when simulated/sent.

Related to #1961

skyyycodes:
#### Problem

So the problem was kit accepts any numbers you hand to 'setTransactionMessageComputeUnitLimit' or
`setTransactionMessageHeapSize` but the runtime won't. A transaction can request at most 1.4 million compute units and a requested heap frame has to be a whole number of kib between 32 KiB and 256 kib, go outside either of those and nothing complains until you simulate or send at which point the failure comes back from the network and doesn't point anywhere near the line where the
bad value was set

#### Summary of Changes

added `resource-limit-validation.ts` with two assert helpers and called them from `setTransactionMessageComputeUnitLimit` `setTransactionMessageHeapSize` and `setTransactionMessageConfig`. Both of them throw a `SolanaError` with the value you passed and the expected bounds so the error message is actually useful. The module isn't exported from the package index same as `MAX_COMPUTE_UNIT_LIMIT` today so the public surface only grows by the two new error codes. The numbers are coming from the runtime. `MAX_COMPUTE_UNIT_LIMIT` is 1.4m `MAX_HEAP_FRAME_BYTES` is `256 * 1024` and `MIN_HEAP_FRAME_BYTES` is `HEAP_LENGTH` which is 32 KiB. The multiple of 1024 rule is already in `sanitize_config`. One thing from the risk you raised in #1961 is the sanitiser doesn't directly compare against these constants. It reads the min and max from a `SanitizeConfig` so technically the bounds are
configurable they match the constants today but if that ever changes we'll have to follow that.

The checks sit above the version switch so they apply to legacy and v0 as well not just the v1 config. Both issues are worded around config but the same limits apply when the value goes out as a compute budget instruction so validating one and not the other felt odd to me. Can narrow it to v1 only if you prefer.

Separately `decompileTransactionMessage` was building the message through `setTransactionMessageConfig`. Because of that adding validation there meant Kit would refuse to decode a transaction just because the values inside it were something the runtime would reject. That felt wrong to me. If some weird transaction reaches a wallet you should still be able to decode and inspect it. So now decode attaches the config directly instead and I added a test which decodes a message with a 1400001 compute unit limit without throwing, I also added `Number.isInteger` to the checks. Without that `NaN` was able to get through the compute unit validation since both `NaN < 0` and `NaN > 1_400_000` are false. It would then get serialized as `null`. Heap was catching it by accident because `NaN % 1024 !== 0` happens to be true. A few test fixtures were using `HEAP_SIZE_A = 30_000` and `HEAP_SIZE_B = 50_000`. Neither is a valid heap size with this rule so I changed them to 32 KiB and 64 KiB

two things I left for now,`estimateResourceLimits` in `@solana/kit` caps the simulated compute units at u32 max instead of `MAX_COMPUTE_UNIT_LIMIT` and then passes that into the setter. I don't think it's really reachable in practice since the runtime stops execution at 1.4M but capping it at the actual limit would be more correct. Also `loadedAccountsDataSizeLimit` has a max too but since neither issue mentioned it so i left it out. Happy to pick either of those up if you say so.

#### Testing

ran everything on node and browser for both packages and it's all green

- `@solana/transaction-messages` 1348 unit tests on node and 1348 on browser
- `@solana/errors` 263 on node and 263 on browser
- typecheck lint and prettier clean on both
- treeshakability passing on node browser and native
- typedoc builds with 0 errors
- 163 turbo tasks green in total across `errors` `transaction-messages` `transactions` `signers`
  `instruction-plans` `transaction-introspection` `rpc` and `kit`

65 of those tests are new. they cover over the max and under the min, heap sizes that aren't a multiple of 1024, the boundaries themselves so 0 and 1400000 for compute units and 32 KiB and 256 KiB for heap, passing undefined to clear the value, and `NaN ` , `Infinity` and nonintegers. All of them run against legacy v0 and v1.

#### Fixes 

closes #1962
closes #1961


.changeset/social-areas-bathe.md
```diff
@@ -0,0 +1,10 @@
+---
+'@solana/transaction-messages': minor
+'@solana/errors': minor
+---
+
+Validate compute unit limit and heap size when they are set on a transaction message
+
+`setTransactionMessageComputeUnitLimit`, `setTransactionMessageHeapSize` and `setTransactionMessageConfig` now reject values the runtime would refuse, throwing `SOLANA_ERROR__TRANSACTION__COMPUTE_UNIT_LIMIT_OUT_OF_RANGE` or `SOLANA_ERROR__TRANSACTION__INVALID_HEAP_SIZE` rather than letting the transaction fail at simulation or send time. A compute unit limit must be an integer in the range [0, 1,400,000]; a heap size must be an integer multiple of 1 KiB between 32 KiB and 256 KiB inclusive.
+
+Decoding is unaffected: `decompileTransactionMessage` still returns messages carrying out-of-range values so that callers can inspect any transaction that reaches them.
```
mcintyre94, line 10: This is the right behaviour, thank you!

mcintyre94:
@trevor-cortex

packages/transaction-messages/src/decompile/v1/message.ts
```diff
@@ -40,7 +40,11 @@ export function decompileTransactionMessage(
 
     return pipe(
         createTransactionMessage({ version: 1 }),
-        m => setTransactionMessageConfig(transactionConfig, m),
+        // Attach the decoded config directly instead of going through `setTransactionMessageConfig`,
+        // which validates resource limits. Decoding must not reject a transaction merely because it
+        // carries values the runtime would refuse; callers need to be able to inspect such messages.
+        m =>
+            isV1ConfigEmpty(transactionConfig) ? m : Object.freeze({ ...m, config: Object.freeze(transactionConfig) }),
```
trevor-cortex, line 47: The reasoning here is sound and the behavior is equivalent today (fresh message, no prior config, `isV1ConfigEmpty` handles the empty case). One maintainability thought: this hand-rolls a subset of `setTransactionMessageConfig`'s semantics (freeze, empty-config normalization), so if that function's merge/normalization behavior ever changes, this path silently diverges. Consider extracting an internal unchecked helper — e.g. `setTransactionMessageConfigUnchecked` that both this and the validating public function call — so there's a single source of truth for how a config gets attached. Non-blocking.

packages/transaction-messages/src/v1-transaction-config.ts
```diff
@@ -119,6 +120,12 @@ type SupportedTransactionVersions = Extract<TransactionVersion, 1>;
 export function setTransactionMessageConfig<
     TTransactionMessage extends TransactionMessage & { version: SupportedTransactionVersions },
 >(config: V1TransactionConfig, transactionMessage: TTransactionMessage): TTransactionMessage {
+    if (config.computeUnitLimit !== undefined) {
+        assertIsValidComputeUnitLimit(config.computeUnitLimit);
+    }
+    if (config.heapSize !== undefined) {
+        assertIsValidHeapSize(config.heapSize);
+    }
```
trevor-cortex, line 128: Only the incoming partial `config` is validated, not `mergedConfig`. That means a message carrying an invalid value — e.g. one obtained from `decompileTransactionMessage` — keeps that invalid value through any subsequent `setTransactionMessageConfig` call that touches other fields (say, setting `priorityFeeLamports`), and nothing throws until the network rejects it.

I suspect this is the right trade-off: validating `mergedConfig` would prevent incrementally *fixing* an invalid decoded message field by field. But it's worth being deliberate about — maybe a short comment here documenting that pre-existing config values are intentionally not re-validated?

.changeset/social-areas-bathe.md
```diff
@@ -0,0 +1,10 @@
+---
+'@solana/transaction-messages': minor
+'@solana/errors': minor
+---
+
+Validate compute unit limit and heap size when they are set on a transaction message
+
+`setTransactionMessageComputeUnitLimit`, `setTransactionMessageHeapSize` and `setTransactionMessageConfig` now reject values the runtime would refuse, throwing `SOLANA_ERROR__TRANSACTION__COMPUTE_UNIT_LIMIT_OUT_OF_RANGE` or `SOLANA_ERROR__TRANSACTION__INVALID_HEAP_SIZE` rather than letting the transaction fail at simulation or send time. A compute unit limit must be an integer in the range [0, 1,400,000]; a heap size must be an integer multiple of 1 KiB between 32 KiB and 256 KiB inclusive.
```
trevor-cortex: Worth adding a sentence about the most likely real-world break: code that buffers an estimated compute unit limit (e.g. `setTransactionMessageComputeUnitLimit(unitsConsumed * 1.1, tx)`) will now throw when the result is a non-integer or exceeds 1.4M, where it previously truncated silently at encode time. A pointer to wrap in `Math.min(Math.ceil(...), 1_400_000)` would save consumers a debugging session.

trevor-cortex (commented):
This PR adds client-side validation of compute unit limits and heap frame sizes, mirroring the runtime's sanitization rules: CU limit must be an integer in `[0, 1_400_000]`, heap size an integer multiple of 1 KiB in `[32 KiB, 256 KiB]`. Validation fires in `setTransactionMessageComputeUnitLimit`, `setTransactionMessageHeapSize`, and `setTransactionMessageConfig` across all message versions, with two new `SolanaError` codes carrying the offending value and bounds. `decompileTransactionMessage` deliberately bypasses validation so out-of-range transactions remain inspectable.

LGTM overall (submitting as a comment since I can't formally approve). The implementation is clean and the test coverage is genuinely thorough — boundaries, `NaN`/`Infinity`/non-integers, `undefined` clearing, all three versions, and the decode-stays-permissive case. The `Number.isInteger` catch for `NaN` (which sails past both range comparisons) is a nice find. I verified the decompile change preserves prior behavior: `decompileTransactionConfig` only sets present keys, and since the config is attached to a freshly created message, the hand-rolled freeze is equivalent to what `setTransactionMessageConfig` would have produced (including the empty-config same-reference case via `isV1ConfigEmpty`).

Things to watch / notes for maintainers:

- **`estimateResourceLimits` interaction** (`packages/kit/src/resource-limit-estimation.ts`, ~L143): the downcast caps `unitsConsumed` at u32 max, and `estimateAndSetResourceLimitsFactory` feeds that straight into the now-throwing setter. It's unreachable through Kit's own factory today — the simulation requests exactly 1.4M CUs, so `unitsConsumed` can't exceed it — but the docs explicitly invite custom estimator wrappers that "apply a buffer to the returned values". A wrapper doing `computeUnitLimit * 1.1` near the max (or producing a non-integer) now throws where it previously truncated silently. I'd support the author's suggested follow-up to cap at `MAX_COMPUTE_UNIT_LIMIT` instead of u32 max, and ideally clamp rather than throw inside that specific path.
- **Sanitizer bounds are technically configurable**: as the author notes, Agave reads these limits from `SanitizeConfig` rather than comparing against the constants directly. They match today; if that ever diverges, these hardcoded bounds need to follow. Worth keeping in mind but not actionable now.
- **Bump level**: `minor` seems defensible to me since every newly-rejected value was already guaranteed to fail at the network — this converts a remote failure into a local, actionable one. Flagging only so a maintainer confirms it matches the repo's policy on setters that newly throw.
- Error codes 5663039/5663040 are correctly appended sequentially in the transaction range, and the union/import orderings in `codes.ts`/`context.ts`/`messages.ts` are alphabetized correctly.
- Applying the checks above the version switch (so legacy/v0 get them too) is the right call — the limits are runtime-enforced regardless of whether they travel as a compute budget instruction or a v1 config.

packages/transaction-messages/src/resource-limit-validation.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/** A requested heap frame size must be a whole number of KiB. */
+export const HEAP_SIZE_MULTIPLE_OF = 1024;
+
+/**
+ * Throws if the given compute unit limit is outside the range the runtime accepts.
+ *
+ * A transaction may request at most {@link MAX_COMPUTE_UNIT_LIMIT} compute units. Requesting more
+ * is rejected when the transaction is sanitized, so we fail here rather than let it become a failed
+ * transaction at simulation or send time.
```
mcintyre94: Sorry but this is an error I made when I opened this issue, requesting too many CUs does not fail a transaction. It gets clamped to the 1.4M max on the server. See https://github.com/anza-xyz/agave/blob/1f6e0fbf3d8ae4a37364a1d9eb31c6b5bcac8869/compute-budget-instruction/src/compute_budget_instruction_details.rs#L122-L128

I still think that having validation against the max CUs in Kit makes sense, but this docblock is a bit misleading as-is.

mcintyre94 (changes requested):
Thankyou, this looks really good! Just one nitpick on a docblock, otherwise this is good to go.

You're right about loaded accounts data size having a limit too. It doesn't need to be part of this PR, but I'll open an issue for that. Thankyou!

skyyycodes:
> Thankyou, this looks really good! Just one nitpick on a docblock, otherwise this is good to go.
> 
> You're right about loaded accounts data size having a limit too. It doesn't need to be part of this PR, but I'll open an issue for that. Thankyou!

Looking into it

skyyycodes:
> You're right about loaded accounts data size having a limit too. It doesn't need to be part of this PR, but I'll open an issue for that. Thankyou!

sorry, my bad . fixed it.
went and read the sanitizer properly and yeah the two are different
 heap actually errors out but CU just gets .min(MAX_COMPUTE_UNIT_LIMIT). so reworded the CU one to say it clamps and left the heap docblock as is since that one's fine. fixed the same wording in the changeset too.
and sure, happy to pick up the loaded accounts data size one whenever you open it

packages/transaction-messages/src/v1-transaction-config.ts
mcintyre94: Agree that this is the right tradeoff

mcintyre94 (approved):
Looks great now, thankyou!

PR #1979, opened 2026-08-24
fix: stop `getBase64Decoder` crashing on payloads larger than ~65KB

amilz:
In browser builds, `getBase64Decoder().read` converted bytes to a string via
`String.fromCharCode(...slice)`. The spread passes one argument per byte, so any
input past V8's argument limit (~65K) threw `RangeError: Maximum call stack size
exceeded` instead of returning a string.

Found this in doing some work on explorer for larger accounts.

The conversion now runs in 32KB chunks, keeping every `String.fromCharCode` call
well under the limit regardless of input size. Also swapped `bytes.slice(offset)`
for `bytes.subarray(offset)` to drop a full copy that nothing mutates.

Added a round-trip test over a 200KB array, which reproduces the `RangeError` on
`main` and passes here.

### DIFF
--- a/.changeset/every-dots-ring.md
+++ b/.changeset/every-dots-ring.md
@@ -1,0 +1,5 @@
+---
+'@solana/codecs-strings': patch
+---
+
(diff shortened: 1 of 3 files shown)


lorisleiva (approved):
I had no idea! Thank you for the fix! 🙏

PR #1986, opened 2026-08-26
fix(codecs-numbers): reject truncated and overlong shortU16 encodings

latent-9:
## Problem

The `shortU16` decoder loop had no upper bound, violating the codec's own documented contract of a **1 to 3 byte** encoding (`maxSize: 3`). Two concrete failure modes:

1. **Overlong chains accepted**: an input like `80 80 80 80 ...` with continuation bits keeps decoding, producing values far outside the `0..65535` domain the encoder asserts via `assertNumberIsBetweenForCodec`.
2. **Truncated input silently misread**: once the chain ran past the end of the buffer, `bytes[offset + byteIndex]` returned `undefined`; `undefined & 0b10000000 === 0` terminated the loop, so truncated input decoded to garbage and returned an offset beyond `bytes.length` instead of failing. For example, `decode(new Uint8Array([0x80]))` currently returns `[0, 2]`.

Every other number decoder guards its reads through `assertByteArrayHasEnoughBytesForCodec`; `shortU16` was the odd one out.

## Change

- Bound the read loop at three bytes and assert per-byte availability with `assertByteArrayHasEnoughBytesForCodec('shortU16', ...)`, so truncation throws `INVALID_BYTE_LENGTH` exactly like the fixed-size number decoders.
- If a continuation bit is still set after three bytes, throw `NUMBER_OUT_OF_RANGE`, mirroring the encoder's own range assertion for the same domain violation.

## Test plan

Two regression tests fail on current main and pass with this change:

- `decode([0x80])` and `decode([0x80, 0x80])` (truncated chains) now throw `SOLANA_ERROR__CODECS__INVALID_BYTE_LENGTH`.
- `decode([0xff, 0xff, 0xff, 0x00])` (four-byte chain) now throws `SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE`.

Full `codecs-numbers` suite passes (28/28), including the existing round-trip sweep over all `\[0, 0xffff\]` values; package typecheck is clean.

### DIFF
--- a/.changeset/quiet-shorts-refuse.md
+++ b/.changeset/quiet-shorts-refuse.md
@@ -1,0 +1,5 @@
+---
+'@solana/codecs-numbers': patch
+---
+
+Fixed `getShortU16Decoder()` silently accepting malformed input: a truncated buffer decoded to garbage with an offset past the end of the byte array, and continuation chains longer than the documented three-byte encoding produced values outside the u16 domain. Truncated input now throws `SOLANA_ERROR__CODECS__INVALID_BYTE_LENGTH` and overlong chains throw `SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE`, matching the guards every other number decoder already uses.
--- a/packages/codecs-numbers/src/__tests__/short-u16-test.ts
+++ b/packages/codecs-numbers/src/__tests__/short-u16-test.ts
@@ -1,3 +1,9 @@
+import {
+    SOLANA_ERROR__CODECS__INVALID_BYTE_LENGTH,
+    SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE,
+    SolanaError,
+} from '@solana/errors';
+
 import { getShortU16Codec } from '../short-u16';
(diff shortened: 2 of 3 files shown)


latent-9:
Reopening — this was closed by the stale bot, not reviewed. I re-verified every claim in the PR body by executing the actual pre/post code with the real `@solana/codecs-core` + `@solana/errors` packages (10/10 checks):

**Old code reproduces exactly as described:**
- `read([0x80])` → `[0, 2]` (garbage value, offset past end of buffer)
- `read([0x80, 0x80])` → silently accepted
- `read([0xff, 0xff, 0xff, 0x00])` → accepted with value 2097151 (far outside u16)

**Fixed code:**
- `[0x80]` → `INVALID_BYTE_LENGTH` `{bytesLength: 1, expected: 2}`
- `[0x80, 0x80]` → `INVALID_BYTE_LENGTH` `{bytesLength: 2, expected: 3}`
- overlong chain → `NUMBER_OUT_OF_RANGE` `{min: 0, max: 65535, value: 2097151}`
- Valid 1–3 byte decodes unchanged; full 0..65535 encode→decode sweep intact.

Happy to address any review feedback. 🙏

PR #1987, opened 2026-08-26
fix(codecs-data-structures): preserve struct fixed-size literals

atharva-patill:
## What changed

Struct codecs were losing the precise fixed size of their fields and widening it to `number`.

This preserves the combined fixed-size literal for struct encoders, decoders, and codecs when all fields have known fixed sizes.

For example, a struct containing a `u32` and a `u8` now infers a fixed size of `5` instead of `number`.

## Tests

Added type tests for:

- Single-field fixed sizes
- Summed sizes such as `u32 + u8 = 5`
- Larger fixed-size combinations
- Widening when the size cannot be represented precisely
- Variable-size fields remaining variable-size

Verified with:

- `@solana/codecs-data-structures` typecheck
- `@solana/transaction-messages` typecheck
- ESLint

No runtime behavior was changed.

### DIFF
--- a/packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
+++ b/packages/codecs-data-structures/src/__typetests__/struct-typetest.ts
@@ -6,7 +6,15 @@ import {
     VariableSizeDecoder,
     VariableSizeEncoder,
 } from '@solana/codecs-core';
-import { getU32Codec, getU32Decoder, getU32Encoder } from '@solana/codecs-numbers';
+import {
+    getU8Codec,
+    getU8Decoder,
+    getU8Encoder,
+    getU32Codec,
+    getU32Decoder,
+    getU32Encoder,
+    getU64Encoder,
+} from '@solana/codecs-numbers';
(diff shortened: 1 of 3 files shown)


mcintyre94:
Hi, thanks for this but we're tracking this issue at https://github.com/anza-xyz/kit/issues/1738. A previous PR https://github.com/anza-xyz/kit/pull/1761 was closed because we decided against summing the fields like this.

We decided to only infer a fixed size for single-field structs when closing that PR, and another open PR is already looking at that: https://github.com/anza-xyz/kit/pull/1971. So I'll close this for now.

atharva-patill:
> Hi, thanks for this but we're tracking this issue at #1738. A previous PR #1761 was closed because we decided against summing the fields like this.
> 
> We decided to only infer a fixed size for single-field structs when closing that PR, and another open PR is already looking at that: #1971. So I'll close this for now.

Thanks for the clarification! I wasn't aware that #1971 was already covering this direction. I'll leave this one here.

PR #1993, opened 2026-08-26
Add heap size validation to transaction messages

atharva-patill:
## What changed

Transaction heap sizes were previously accepted without checking the protocol limits, so invalid values could make it through to serialization and only fail later when submitted to the network.

This adds validation when setting the heap size:

- must be between 32 KiB and 256 KiB
- must be a multiple of 1 KiB
- must be a finite integer
- `undefined` remains valid for removing the setting

The same validation is applied to the V1 transaction config path while keeping transaction decompilation permissive for existing on-chain data.

## Tests

Added coverage for the boundary, alignment, and invalid-value cases across legacy, V0, and V1 transactions, along with config validation and permissive decoding.

Also verified that failed validation does not mutate the transaction message.

All relevant tests, typechecks, lint, and formatting checks pass.

No runtime behavior outside heap-size validation was changed.


packages/errors/src/context.ts
```diff
@@ -960,6 +933,12 @@ export type SolanaErrorContext = ReadonlyContextValue<
                 configName: string;
                 expectedKind: 'u32' | 'u64';
             };
+            [SOLANA_ERROR__TRANSACTION__INVALID_HEAP_SIZE]: {
+                heapSize: number;
+                maxHeapSize: number;
+                minHeapSize: number;
+                multipleOf: number;
```
mcintyre94, line 940: My initial instinct is that I'm not sure it makes sense for this to be configurable, but it's probably worth having in case these values change in future. We have similar with things like max accounts in an instruction too.

packages/transaction-messages/src/__tests__/heap-size-test.ts
```diff
@@ -174,6 +182,112 @@ describe('setTransactionMessageHeapSize', () => {
     );
 });
 
+describe('setTransactionMessageHeapSize validation', () => {
+    describe.each([{ version: 'legacy' as const }, { version: 0 as const }, { version: 1 as const }])(
+        'given a $version transaction',
+        ({ version }) => {
+            const baseTx = { instructions: [], version } as TransactionMessage & { version: 1 };
```
mcintyre94, line 189: Should this be:

```suggestion
            const baseTx = { instructions: [], version } as TransactionMessage & { version };
```

packages/transaction-messages/src/compute-budget-instruction.ts
```diff
@@ -13,6 +13,15 @@ export const COMPUTE_BUDGET_PROGRAM_ADDRESS =
...
 export const MAX_COMPUTE_UNIT_LIMIT = 1_400_000;
 
+/** The minimum heap frame size that can be requested for a transaction (32 KiB). */
+export const MIN_HEAP_FRAME_SIZE = 32 * 1024;
+
+/** The maximum heap frame size that can be requested for a transaction (256 KiB). */
+export const MAX_HEAP_FRAME_SIZE = 256 * 1024;
+
+/** The requested heap frame size must be a multiple of this value (1 KiB). */
+export const HEAP_FRAME_SIZE_MULTIPLE = 1024;
```
mcintyre94, line 23: I'm hesitant about exporting these, but given the MAX_COMPUTE_UNIT_LIMIT precedent I think it makes sense

packages/transaction-messages/src/decompile/v1/message.ts
```diff
@@ -39,8 +39,20 @@ export function decompileTransactionMessage(
...
 
     return pipe(
+        // @ts-expect-error We don't expose v1 on `createTransactionMessage` yet
         createTransactionMessage({ version: 1 }),
-        m => setTransactionMessageConfig(transactionConfig, m),
+        // Decoding must remain permissive, so the decompiled config is assigned directly instead of
+        // passing through the validating `setTransactionMessageConfig` setter.
+        // Won't need this cast after we support v1 on `createTransactionMessage`
+        m => {
+            const mWithVersion = m as unknown as TransactionMessage & { version: 1 };
```
mcintyre94, line 48: This line is just casting, but we're not passing it to anything, and we're casting again after. I don't think we need this, can just use m directly.

packages/transaction-messages/src/decompile/v1/message.ts
```diff
@@ -39,8 +39,20 @@ export function decompileTransactionMessage(
     );
 
     return pipe(
+        // @ts-expect-error We don't expose v1 on `createTransactionMessage` yet
         createTransactionMessage({ version: 1 }),
-        m => setTransactionMessageConfig(transactionConfig, m),
+        // Decoding must remain permissive, so the decompiled config is assigned directly instead of
+        // passing through the validating `setTransactionMessageConfig` setter.
+        // Won't need this cast after we support v1 on `createTransactionMessage`
```
mcintyre94, line 46: This comment line ("won't need this cast") can be removed, it doesn't apply to your cast (and is outdated on main)

packages/transaction-messages/src/decompile/v1/message.ts
```diff
@@ -39,8 +39,20 @@ export function decompileTransactionMessage(
     );
 
     return pipe(
+        // @ts-expect-error We don't expose v1 on `createTransactionMessage` yet
         createTransactionMessage({ version: 1 }),
-        m => setTransactionMessageConfig(transactionConfig, m),
+        // Decoding must remain permissive, so the decompiled config is assigned directly instead of
```
mcintyre94, line 44: Agree with keeping decoding permissive, good spot!

mcintyre94 (commented):
@trevor-cortex

mcintyre94:
Thanks for this! https://github.com/anza-xyz/kit/pull/1972 includes heap size validation, so I'm going to close this in favour of that PR.

PR #1994, opened 2026-08-26
fix: read abort reasons from captured signals

matusbalascak:
#### Problem

In React Native 0.86, the abort event passed to the RPC coalescer can have `target === null`. Reading `event.target.reason` then throws: `TypeError: Cannot read property 'reason' of null`

I ran into this while testing Solana transaction confirmation in Trezor Suite: https://github.com/trezor/trezor-suite/issues/30691

The timeout confirmation code used the same pattern.

#### Summary of Changes

Use the abort signal already in scope instead of reading it from the event. Added regression tests for both places.

I ran the node and browser tests, typecheck, lint, and Prettier for both affected packages.

### DIFF
--- a/.changeset/chilly-ads-retire.md
+++ b/.changeset/chilly-ads-retire.md
@@ -1,0 +1,6 @@
+---
+'@solana/rpc': patch
+'@solana/transaction-confirmation': patch
+---
+
+Fix transaction confirmation and coalesced RPC request cancellation in environments where abort events have a null target.
--- a/packages/rpc/src/__tests__/rpc-request-coalescer-test.ts
+++ b/packages/rpc/src/__tests__/rpc-request-coalescer-test.ts
@@ -96,6 +96,25 @@ describe('RPC request coalescer', () => {
             const transportAbortSignal = mockTransport.mock.lastCall![0].signal!;
             expect(transportAbortSignal.aborted).toBe(false);
         });
+        it('rejects with the signal reason when the abort event target is null', async () => {
+            expect.assertions(2);
+            const abortController = new AbortController();
+            const addEventListenerSpy = jest
+                .spyOn(abortController.signal, 'addEventListener')
+                .mockImplementation(() => {});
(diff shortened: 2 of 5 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
### Summary

Works around a React Native 0.86 quirk where the `abort` event delivered to a listener can have `target === null`, which made `(e.target as AbortSignal).reason` throw. Both call sites now read `.reason` from the signal already captured in scope (`signal` in the RPC request coalescer, `callerAbortSignal` in the timeout confirmation strategy), and regression tests simulate the null-target event in both packages.

### Why this is correct

In both cases the listener is registered directly on the same signal object that's now being read. Per spec, `e.target` for an `abort` event *is* that signal, so `signal.reason` is semantically identical to the old expression in conforming environments — and it keeps working where the event implementation fails to set `target`. Dropping the now-unused `e` parameter is the right cleanup, and the `as AbortSignal` casts go away as a bonus.

The tests take the only realistic route to reproduce this in Node: spy out `addEventListener`, capture the handler, and invoke it manually with a synthetic `Event('abort')` whose `target` is asserted to be null before the handler runs. Aborting the controller first ensures `signal.reason` is populated, mirroring the real-world sequence. Slightly implementation-coupled, but appropriate here.

Changeset with `patch` bumps for `@solana/rpc` and `@solana/transaction-confirmation` matches the scope of the fix.

### Notes for other reviewers

- It may be worth a quick repo-wide grep for other `abort` listeners reading `e.target` (e.g. in the subscriptions or WebSocket packages) so the same RN issue doesn't resurface elsewhere. This PR fixes the two sites named in the linked issue.
- No behavioral change for platforms with conforming `AbortSignal` implementations — this is purely a defensive source change.

LGTM — would approve, but I don't have approval rights on this repo.

mcintyre94 (approved):
Thanks for this!

PR #1995, opened 2026-08-26
Add concurrent transaction plan leaf executor

mcintyre94:
This PR adds a new helper `createTransactionPlanExecutorWithConcurrentLeaves` as a sibling to `createTransactionPlanExecutor`.

This creates an intentionally much simpler executor, that ignores dependencies between transactions and concurrently executes the callback (slightly simplified) `SingleTransactionPlan -> Promise<SingleTransactionPlanResult<TContext>>` for every leaf. It returns a `TransactionPlanResult` with the same shape as the input `TransactionPlan`. It is intended for cases where dependencies are irrelevant, such as signing the transactions without executing them.

See also https://github.com/anza-xyz/kit-plugins/pull/395 which first added a similar more specific version of this function. That PR will be refactored to use this function.

Failures are aggregated, similar to `createTransactionPlanExecutor`: if a callback throws for any leaf then this is returned as a `FailedSingleTransactionPlanResult` at that leaf, independently of other callbacks. Unlike `createTransactionPlanExecutor`, this does not cancel dependent leafs, as they are treated independently. If there are any failures then we throw `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN` with the plan.

### DIFF
--- a/.changeset/polite-clubs-notice.md
+++ b/.changeset/polite-clubs-notice.md
@@ -1,0 +1,5 @@
+---
+'@solana/instruction-plans': minor
+---
+
+Add `createTransactionPlanExecutorWithConcurrentLeaves` to concurrently transform every transaction plan leaf while preserving the plan's result shape.
--- a/packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
+++ b/packages/instruction-plans/src/__tests__/transaction-plan-executor-test.ts
@@ -1,9 +1,11 @@
 import '@solana/test-matchers/toBeFrozenObject';
 
 import {
+    isSolanaError,
     SOLANA_ERROR__INSTRUCTION_ERROR__INVALID_ARGUMENT,
     SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN,
     SOLANA_ERROR__INSTRUCTION_PLANS__NON_DIVISIBLE_TRANSACTION_PLANS_NOT_SUPPORTED,
+    SOLANA_ERROR__INVARIANT_VIOLATION__INVALID_TRANSACTION_PLAN_KIND,
     SOLANA_ERROR__TRANSACTION_ERROR__INSUFFICIENT_FUNDS_FOR_FEE,
     SolanaError,
 } from '@solana/errors';
@@ -14,15 +16,19 @@ import { Transaction } from '@solana/transactions';
 import {
     canceledSingleTransactionPlanResult,
     createTransactionPlanExecutor,
+    createTransactionPlanExecutorWithConcurrentLeaves,
     failedSingleTransactionPlanResult,
     nonDivisibleSequentialTransactionPlan,
+    nonDivisibleSequentialTransactionPlanResult,
     parallelTransactionPlan,
     parallelTransactionPlanResult,
     passthroughFailedTransactionPlanExecution,
     sequentialTransactionPlan,
     sequentialTransactionPlanResult,
+    SingleTransactionPlan,
     singleTransactionPlan,
     successfulSingleTransactionPlanResult,
+    TransactionPlan,
     TransactionPlanResult,
     TransactionPlanResultContext,
     TransactionPlanResultContextWithSignature,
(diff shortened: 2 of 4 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
This PR adds `createTransactionPlanExecutorWithConcurrentLeaves`, a deliberately simple sibling to `createTransactionPlanExecutor` that maps a `SingleTransactionPlan -> Promise<SingleTransactionPlanResult<TContext>>` callback over every leaf concurrently, ignoring sequential dependencies, and rebuilds a `TransactionPlanResult` with the same nesting, order, and divisibility as the input plan. Intended for dependency-irrelevant operations like signing without executing (extracted from anza-xyz/kit-plugins#395).

The implementation is tidy and well-scoped — this would be an approval if I had the repo access to file one:

- The recursive traversal handles all three plan kinds with an exhaustive `kind satisfies never` fallback throwing `SOLANA_ERROR__INVARIANT_VIOLATION__INVALID_TRANSACTION_PLAN_KIND`, consistent with the existing `traverse` function in the same file.
- Unlike `createTransactionPlanExecutor`, this one supports non-divisible sequential plans and reconstructs them via `nonDivisibleSequentialTransactionPlanResult` — that's correct given dependencies are irrelevant here, and the first unit test covers it explicitly.
- The abort signal is forwarded rather than enforced (no `getAbortablePromise`, no pre-flight `aborted` check). The docs are explicit that cancellation handling is the callback's responsibility, so I read this as intentional.
- TSDoc is thorough and matches the actual behavior, including the `@throws` documentation for callback errors.
- Tests cover shape preservation, concurrent start of sequential leaves, abort-signal forwarding, and the unknown-kind invariant. Type tests cover explicit and inferred `TContext` plus the callback signature and context obligation.
- Changeset is present with a `minor` bump, which matches precedent for new public API in this repo.

Two things flagged inline:

1. `ConcurrentLeafTransactionPlanExecutorConfig` is imported in the typetest but never used — likely a lint failure waiting to happen.
2. A question on error semantics: a throwing leaf callback propagates the raw error via `Promise.all`, with no partial result shape (unlike the `FAILED_TO_EXECUTE_TRANSACTION_PLAN` error the sibling executor throws, which `passthroughFailedTransactionPlanExecution` understands). This looks intentional given callbacks are expected to return failed/canceled results rather than throw, but worth confirming.

For subsequent reviewers: the main thing to sanity-check is the error-propagation contract in point 2 above, especially how the kit-plugins#395 refactor will consume this — if that call site relies on `passthroughFailedTransactionPlanExecution`, callbacks there must never throw.

mcintyre94:
I've aligned the failure aggregation behaviour, and the abort signal handling, with `createTransactionPlanExecutor`, and added symmetric tests.

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Re-reviewed after the latest changes — both points from my previous review are resolved, and the error/abort semantics have been substantially strengthened. This would be an approval if I could file one.

**What changed since last review:**

1. **Unused import — resolved.** `ConcurrentLeafTransactionPlanExecutorConfig` is now exercised by the new typetest block asserting that the config type has no default `TContext` (via `@ts-expect-error`).

2. **Error semantics — resolved, and better than what I asked for.** Instead of propagating raw callback errors through `Promise.all`, thrown errors are now caught per-leaf and converted to `FailedSingleTransactionPlanResult`s without affecting sibling leaves. After all leaves settle, any failed/canceled result in the tree produces a `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN` via `createFailedToExecuteTransactionPlanError`, carrying the complete result tree. This means `passthroughFailedTransactionPlanExecution` now works uniformly across both executors — which directly resolves my concern about the kit-plugins#395 refactor inheriting surprising failure behavior.

3. **Abort signal is now enforced, not just forwarded.** Pre-aborted signals cancel leaves without invoking callbacks; mid-flight aborts race callbacks via `getAbortablePromise` and record failed results carrying the abort reason, while preserving already-completed results. The `abortReason` lands in the error context. TSDoc was updated to match.

**Verification notes:**

- The failed-vs-canceled asymmetry (pre-abort → `canceled`, mid-flight abort → `failed` with the abort reason as the error) mirrors `traverseSingle` in the sibling executor exactly, is documented in the TSDoc, and is asserted by the abort tests — I read it as intentional and consistent.
- The `error as Error` cast in the leaf catch matches the identical pre-existing pattern in `traverseSingle`; no new concern.
- Test coverage is thorough: shape preservation, concurrent starts of sequential leaves, per-leaf failure aggregation with `error.cause` pointing at the first tree failure, waiting for all siblings to settle before throwing (the fake-timers assertion at 99ms/100ms is a nice touch), failed and canceled callback results, both abort paths, and the invariant-violation fallback.

**One tiny observation, no action needed:** there's a theoretical window where the signal aborts *after* every leaf has settled successfully but *before* the final `aborted` check, producing a `FAILED_TO_EXECUTE` error wrapping a fully successful result tree. The sibling executor has an equivalent window with its `canceled` flag, so behavior is consistent across the two — just noting it for completeness.

For subsequent reviewers: nothing outstanding from my side. The main thing left to verify is the kit-plugins#395 refactor consuming this cleanly, which the new aggregation contract should make straightforward.

lorisleiva (approved):
Perfect, thank you!

PR #2002, opened 2026-08-27
Add a note about the `Equal` helper in CLAUDE.md

mcintyre94:
This is an occasionally useful typetest helper, and we don't want to have weaker versions of it used in other tests.

Added a note to CLAUDE.md to just copy it with its large explanatory docblock when it's needed, instead of reinventing something like it.


lorisleiva (approved):
Nice!

PR #2003, opened 2026-08-27
Update docs CI so that it can be required

mcintyre94:
#### Problem

Currently the docs CI job is not required to merge a PR. Most PRs don't update docs, and the job is flaky because of running out of memory on the runners. So we skip it if there are no docs changes. This means that it is no longer a required job on the CI.

This means that if there *are* docs changes, the preview-docs job will not block merging if it fails. In the case of Dependabot, passing CI means that it merges automatically.

Previously Dependabot bumped Fumadocs packages and broke the CI. We reverted this and told it to ignore major updates. This didn't work, because it merged a minor... and that also broke. See https://github.com/anza-xyz/kit/pull/1996

#### Summary of Changes

Annoyingly Github doesn't have a simple way to specify a job that only runs sometimes, but must pass if it runs. 

See https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

The top row of that table specifies that if we skip a job because of path filtering (ie only if docs/ changes, what we currently use), then it stays pending, and blocks merging if it's marked required. 

So instead we rely on the second row, which says that if a job is skipped because of a conditional, then it is marked success.

We now split `preview-docs` into 3 jobs:

- `detect-docs-changes` runs unconditionally. It uses the [dorny/paths-filter](https://github.com/dorny/paths-filter) action to check for changes to docs/ (or the docs workflow). It outputs `true` if there are changes. This runs on all PRs but only takes a few seconds. 
- `preview-docs` uses the **condition** `if: needs.detect-docs-changes.outputs.docs == 'true'`. If there are no changes then it does not run but GitHub treats it as successful. If there are changes, then it does our whole Vercel build/preview deploy 
- `preview-docs-gate` is the job that will become required to pass. It just sanity checks the outputs of the two jobs, failing if `detect-docs-changes` failed, and failing if `detect-docs-changes` returned true and `preview-docs` was unsuccessful. This also runs on all PRs and takes a few seconds. 

Note that this would work with just the first two jobs, with the second one as a required check. But this would behave wrong if `detect-docs-changes` returned something invalid or failed. The third job also makes our conditions clearer I think, with less implicit reliance on Github's behaviour. 

Together this enables the behaviour that we want: most PRs just run a quick check, no docs changes, success. PRs that do change docs, fail the CI and block merging (when we make `preview-docs-gate` required) if they fail. When Dependabot tries to break our docs, it won't be able to merge its PRs.

Another alternative would be to just run `preview-docs` unconditionally again, but I think it is better to avoid running it on most PRs.


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

This restructures the docs preview workflow so it can be made a required check. The workflow now runs on every PR: `detect-docs-changes` uses `dorny/paths-filter` to detect docs changes, `preview-docs` runs conditionally on that output (skipped-via-`if` jobs report success, unlike skipped-via-path-filter jobs which stay pending), and `preview-docs-gate` verifies both results and is the job intended to become required.

The mechanics are correct. I traced the gate logic through the relevant states — detect fails → gate fails; docs changed + preview failed/cancelled → gate fails; docs unchanged → gate passes; detect output invalid/empty → gate fails via the third check. `if: always()` is also the right choice here (see inline comment), and the paths-filter action is SHA-pinned with a version comment. The explicit `contents: read` added to `preview-docs` is a good catch too, since job-level `permissions` replace the workflow-level block entirely.

## Key concern

**The gate doesn't actually block the Dependabot scenario that motivated this PR.** The existing `Deploy to Vercel` step — which contains the *entire* build (`vercel pull` / `vercel build` / `vercel deploy`) — is skipped when `github.actor == 'dependabot[bot]'` or on fork PRs. So when Dependabot bumps Fumadocs in `docs/package.json`, `preview-docs` will run checkout + installs, skip the build entirely, and succeed — and the gate will pass. See the inline comment for a suggested fix (run a secrets-free `pnpm build` for those PRs).

## Notes for other reviewers

- The required check in branch protection should target the job name `Documentation Preview Gate` once this lands.
- Cost note (acknowledged in the description): every PR now spins up two small jobs (`detect-docs-changes` + gate) instead of zero. That's the unavoidable price of the "skipped counts as success" pattern.
- No changeset needed — CI-only change, consistent with the repo's changeset policy.

.github/workflows/preview-docs.yml
```diff
@@ -23,8 +20,29 @@ env:
...
+        id: filter
+        with:
+          filters: |
+            docs:
+              - 'docs/**'
+              - '.github/workflows/preview-docs.yml'
+
   preview-docs:
+    needs: detect-docs-changes
+    if: needs.detect-docs-changes.outputs.docs == 'true'
```
trevor-cortex, line 43: This wiring is correct, but there's a gap versus the PR's stated goal: the existing `Deploy to Vercel` step below (~L64) is guarded by `github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]'`, and that step contains the entire build (`vercel pull`/`vercel build`/`vercel deploy`). So on a Dependabot PR that bumps Fumadocs in `docs/package.json` — the exact #1996 scenario — this job runs checkout + installs, skips the build entirely, and reports success, and the gate passes. Dependabot could still auto-merge a broken docs bump. (The `pnpm install --ignore-workspace` step does run `fumadocs-mdx` via `postinstall`, so *some* breakage would surface there, but `next build` never runs.)

A fix that keeps secrets out of Dependabot/fork runs: add a build-only fallback step with the inverse condition that runs `pnpm run build` in `./docs/` (it's a plain `next build`, no Vercel secrets needed). That way every docs-touching PR validates the build, and only trusted same-repo PRs deploy a preview. Happy to leave it for a follow-up PR if you prefer, but as-is the "block Dependabot" goal isn't met.

.github/workflows/preview-docs.yml
```diff
@@ -72,3 +90,33 @@ jobs:
           PREVIEW_URL: '${{ steps.vercel_deploy.outputs.preview_url }}'
         run: |
           gh pr comment $PR_NUMBER --body "Documentation Preview: $PREVIEW_URL" --create-if-none --edit-last
+
+  preview-docs-gate:
+    needs:
+      - detect-docs-changes
+      - preview-docs
+    if: always()
```
trevor-cortex, line 100: `always()` is the right call here, and worth preserving deliberately: the common advice to prefer `!cancelled()` would be a security hole for a gate job, because a job skipped via `if` counts as **success** for required checks — cancelling the run would let the PR merge without the gate ever evaluating. Might be worth a short comment in the YAML so a future cleanup doesn't "fix" this to `!cancelled()`.

.github/workflows/preview-docs.yml
mcintyre94: This is not an issue because on these PRs we have a GitHub action that updates the docs lockfile, and then the CI runs including deploy to Vercel.

See https://github.com/anza-xyz/kit/pull/1996/commits for example

- First commit failed at install isolated docs dependencies, because of the outdated lockfile (dependabot won't update it)
- Second commit failed at deploy to Vercel with a compile error

.github/workflows/preview-docs.yml
mcintyre94: Added a comment

mcintyre94:
This PR now also reverts #1996, because otherwise it (correctly!) fails CI

mcintyre94:
This PR now also makes a small change to `instruction-plans.mdx` to make it compile for v8. Another dependabot PR https://github.com/anza-xyz/kit/pull/1998 bumped docs to Kit v8, which I don't want to revert.

lorisleiva (approved):
Nice, thank you!

PR #2006, opened 2026-08-27
fix: accept class-based signers in `@solana/signers` type guards

amilz:
The signer type guards declared their parameter as `{ [key: string]: unknown; address: Address<TAddress> }`. TypeScript only grants implicit index signatures to object-literal and mapped types, so class instances implementing a signer interface failed to compile when passed to a guard even though the check passes at runtime. This forced web3.js v3's `Keypair` to add a `readonly [key: string]: unknown` index signature as a workaround, weakening typo detection for its consumers.

```ts
export class Keypair implements KeyPairSigner<KitAddress> {
   readonly [key: string]: unknown; // <-- not awesome, let me delete it plz!
   // ...
```
[ref](https://github.com/solana-foundation/solana-web3.js/blob/6cc68377587a2a158260d82967b1b1dc8686d626/src/keypair.ts#L26-L35)

The guards are now generic over `TValue extends { address: Address<TAddress> }` and narrow to `TValue & <Signer>`, so any addressable value — plain object or class instance — is accepted, and narrowing preserves the input's other properties. Guard bodies are unchanged since `in` narrowing works without index signatures. A typetest covers a class-based `MessagePartialSigner` passing both guard forms without casts.

### DIFF
--- a/packages/signers/src/__typetests__/message-partial-signer-typetest.ts
+++ b/packages/signers/src/__typetests__/message-partial-signer-typetest.ts
@@ -18,3 +18,19 @@ const signMessages = () => {};
     assertIsMessagePartialSigner(potentialSigner);
     potentialSigner satisfies MessagePartialSigner<'1'>;
 }
+
+{
+    // [isMessagePartialSigner]: It accepts class instances, which have no implicit index signature.
+    class MyMessagePartialSigner implements MessagePartialSigner<'1'> {
+        readonly address = address('1');
+        signMessages() {
+            return Promise.resolve([]);
+        }
+    }
+    const potentialSigner: MyMessagePartialSigner = new MyMessagePartialSigner();
+    if (isMessagePartialSigner(potentialSigner)) {
+        potentialSigner satisfies MessagePartialSigner<'1'>;
+    }
+    assertIsMessagePartialSigner(potentialSigner);
+    potentialSigner satisfies MessagePartialSigner<'1'>;
+}
--- a/packages/signers/src/__typetests__/message-partial-signer-typetest.ts
+++ b/packages/signers/src/__typetests__/message-partial-signer-typetest.ts
@@ -20,17 +20,28 @@ const signMessages = () => {};
 }
 
 {
-    // [isMessagePartialSigner]: It accepts class instances, which have no implicit index signature.
-    class MyMessagePartialSigner implements MessagePartialSigner<'1'> {
+    // [isMessagePartialSigner]: It accepts a class instance, which has no implicit index signature.
+    class MyClassBasedSigner implements MessagePartialSigner<'1'> {
         readonly address = address('1');
         signMessages() {
             return Promise.resolve([]);
         }
     }
-    const potentialSigner: MyMessagePartialSigner = new MyMessagePartialSigner();
+    const potentialSigner = null as unknown as MyClassBasedSigner;
     if (isMessagePartialSigner(potentialSigner)) {
         potentialSigner satisfies MessagePartialSigner<'1'>;
     }
+}
+
+{
+    // [assertIsMessagePartialSigner]: It accepts a class instance, which has no implicit index signature.
+    class MyClassBasedSigner implements MessagePartialSigner<'1'> {
+        readonly address = address('1');
+        signMessages() {
+            return Promise.resolve([]);
+        }
+    }
+    const potentialSigner = null as unknown as MyClassBasedSigner;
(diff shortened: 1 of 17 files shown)


packages/signers/src/__typetests__/message-partial-signer-typetest.ts
```diff
@@ -18,3 +18,19 @@ const signMessages = () => {};
...
+
+{
+    // [isMessagePartialSigner]: It accepts class instances, which have no implicit index signature.
+    class MyMessagePartialSigner implements MessagePartialSigner<'1'> {
+        readonly address = address('1');
+        signMessages() {
+            return Promise.resolve([]);
+        }
+    }
+    const potentialSigner: MyMessagePartialSigner = new MyMessagePartialSigner();
```
mcintyre94: Nit: we usually do `null as unknown as MyMessagePartialSigner` in tests, I think that'll work for a class too

packages/signers/src/__typetests__/message-partial-signer-typetest.ts
```diff
@@ -18,3 +18,19 @@ const signMessages = () => {};
...
+    if (isMessagePartialSigner(potentialSigner)) {
+        potentialSigner satisfies MessagePartialSigner<'1'>;
+    }
+    assertIsMessagePartialSigner(potentialSigner);
+    potentialSigner satisfies MessagePartialSigner<'1'>;
```
mcintyre94, line 46: I'd split this into two tests, one for `isMessagePartialSigner` and one for `assertIsMessagePartialSigner`

packages/signers/src/__typetests__/message-partial-signer-typetest.ts
```diff
@@ -18,3 +18,19 @@ const signMessages = () => {};
     assertIsMessagePartialSigner(potentialSigner);
     potentialSigner satisfies MessagePartialSigner<'1'>;
 }
+
+{
+    // [isMessagePartialSigner]: It accepts class instances, which have no implicit index signature.
```
mcintyre94: I think it'd be worth adding tests like this to the other signer typetests too

mcintyre94 (changes requested):
A few nits about the tests but this is a good idea, thank you!

amilz:
Thanks @mcintyre94 -- addressed in 3435c91.

mcintyre94 (approved):
LGTM, thank you!

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Rewrites all eight signer type guards (and their `assertIs*` variants) from taking `{ [key: string]: unknown; address: Address<TAddress> }` to being generic over `TValue extends { address: Address<TAddress> }`, narrowing to `Signer<TAddress> & TValue`. This fixes the long-standing papercut where class instances implementing a signer interface couldn't be passed to the guards without casts, since TypeScript only grants implicit index signatures to object-literal and mapped types.

The approach is sound — this would be an approval if I had the repo access to submit one:

- `TAddress` is correctly inferred from the constraint on `TValue`, and the existing typetests confirm the address type parameter is still tracked (e.g. narrowing to `KeyPairSigner<'1'>`).
- Narrowing to `TValue & Signer` is a strict improvement over the old behavior — it preserves the input's other properties instead of erasing them to the bare signer type.
- Guard bodies are untouched, which works because `in` narrowing applies to generic type parameters without needing an index signature. The composite guards (`isMessageSigner`, `isTransactionSigner`) pass `TValue` through to the sub-guards whose constraints it already satisfies.
- Typetest coverage is thorough: each of the eight guards gets both an `is*` and `assertIs*` class-instance case.

## One minor note (non-blocking)

There's a subtle edge for callers that explicitly instantiate the type argument, e.g. `isKeyPairSigner<'1234..5678'>(value)`. That compiled before with one type parameter, but now fails with "Expected 2 type arguments" since `TValue` has no default and TS doesn't do partial inference. This is a rare pattern for type guards (inference handles the common case), so a patch bump still seems reasonable, but it's worth being aware of in case an issue gets filed. A default like `TValue extends { address: Address<TAddress> } = { address: Address<TAddress> }` would restore single-type-arg call sites if it ever comes up, at the cost of a slightly noisier signature.

## For subsequent reviewers

Nothing needs extra attention — the change is mechanical and consistent across all eight signer modules, runtime behavior is unchanged, and the changeset accurately describes the relaxation. This also unblocks deleting the `readonly [key: string]: unknown` workaround in web3.js v3's `Keypair`.

PR #2007, opened 2026-08-28
Feat: add resilient subscription helpers with reconnection and gap recovery

### LINKED ISSUE #1990: feat: Add subscription reconnection helpers with automatic recovery
## Description
Create reusable utilities for managing RPC subscription lifecycle with automatic reconnection, exponential backoff, and coordinated message gap recovery.

## Goals
- Implement automatic reconnection with exponential backoff strategy
- Provide utilities to detect and recover from message gaps in subscriptions
- Support subscription state persistence across reconnections
- Create helper to coordinate recovery across multiple subscriptions
- Minimize data loss and ensure consistency

## Expected Behavior
Developers should be able to create resilient subscriptions:

import { createResilientSubscription } from '@solana/rpc-subscriptions';

const subscription = createResilientSubscription(rpcSubscriptions, {
  onSubscribe: async () => rpcSubscriptions.accountNotifications(address),
  onGap: async () => {
    // Catch up on missed updates
    const { value } = await rpc.getBalance(address).send();
    return { value };
  },
  backoffConfig: { maxAttempts: 10, initialDelayMs: 100 },
});

for await (const notification of subscription) {
  console.log('Received:', notification);
}

MayurK-cmd:
Add createResilientSubscription and coordinateSubscriptionRecovery utilities to help applications build robust WebSocket subscriptions that automatically reconnect on failure and recover from message gaps.

createResilientSubscription wraps a subscription factory with:
- Automatic reconnection on failure
- Exponential backoff with jitter (configurable)
- Optional message gap recovery via onGap callback
- AbortSignal support for graceful shutdown
- Async generator interface for streaming results

coordinateSubscriptionRecovery runs multiple recovery functions in parallel and collects any errors, useful for synchronizing catch-up logic across multiple subscriptions after network failures.

Both utilities include comprehensive JSDoc with usage examples and follow Kit conventions (functional composition, TypeScript strict mode, proper error handling).

Closes #1990 

Testing: Manual integration tests conducted - all 4 test scenarios pass:
- Test 1: Basic reconnection with gap recovery ✓
- Test 2: Parallel recovery coordination ✓
- Test 3: AbortSignal handling ✓
- Test 4: Edge case (empty recovery functions) ✓

### DIFF
--- a/packages/rpc-subscriptions/src/index.ts
+++ b/packages/rpc-subscriptions/src/index.ts
@@ -17,5 +17,6 @@ export * from './rpc-subscriptions-clusters';
 export * from './rpc-subscriptions-coalescer';
 export * from './rpc-subscriptions-json-bigint';
 export * from './rpc-subscriptions-json';
+export * from './rpc-subscriptions-reconnector';
 export * from './rpc-subscriptions-transport';
 export * from './rpc-subscriptions';
--- a/packages/rpc-subscriptions/src/rpc-subscriptions-reconnector.ts
+++ b/packages/rpc-subscriptions/src/rpc-subscriptions-reconnector.ts
@@ -1,0 +1,264 @@
+/**
+ * Utilities for creating resilient subscriptions with automatic reconnection and message gap recovery.
+ *
+ * @packageDocumentation
+ */
+
+import type { RpcSubscriptionsApiMethods, RpcSubscriptionsPlan } from '@solana/rpc-subscriptions-spec';
+
+/**
+ * Configuration for backoff strategy when reconnecting.
+ */
+export type BackoffConfig = Readonly<{
+    /**
+     * The initial delay in milliseconds before the first retry.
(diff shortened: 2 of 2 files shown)


mcintyre94:
Thanks for this, but my instinct is that this it too opinionated for Kit. It's definitely too opinionated for `packages/rpc-subscriptions`. If we did add something like this then we'd probably want it in `packages/kit`, which is where the higher-level more opinionated functionality lives. But then we'd be positioning this as the default, and that needs some careful design thought from us.

I think this may be better off as an independent package outside of Kit, perhaps published as a [Kit Plugin](https://github.com/anza-xyz/kit-plugins/).

MayurK-cmd:
Thanks for the feedback! Makes sense — I'll publish this as a community plugin instead and open a PR to kit-plugins to get it listed in the Community Plugins table. Appreciate the pointer to that repo.

PR #2008, opened 2026-08-28
Also run preview-docs in the merge queue

mcintyre94:
In order to require this job, we need to run it on PRs in the merge queue also

This requires adding a checkout step before `dorny/paths-filter` in the merge queue.

We run `pnpm build` for docs in the merge queue, but do not include the Vercel deploy

### DIFF
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -2,6 +2,7 @@ name: Preview Documentation
 
 on:
   pull_request:
+  merge_group:
 
 permissions:
   contents: read
@@ -29,6 +30,10 @@ jobs:
     outputs:
       docs: ${{ steps.filter.outputs.docs }}
     steps:
+      - name: Checkout Merge Group
+        if: github.event_name == 'merge_group'
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
       - name: Detect Documentation Changes
         uses: dorny/paths-filter@ceb8a2b8f2d89434be7ff52d3de7ec3738c5cc9d # v4.0.3
         id: filter
@@ -58,11 +63,18 @@ jobs:
         shell: bash
         run: pnpm install --ignore-workspace
 
+      - name: Build Documentation for Merge Queue
+        if: github.event_name == 'merge_group'
+        working-directory: ./docs/
+        shell: bash
+        run: pnpm build
+
       - name: Install Vercel CLI
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]'
         run: pnpm install -g vercel
 
       - name: Deploy to Vercel
-        if: github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]'
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]'
         shell: bash
         id: vercel_deploy
         env:


mcintyre94:
@trevor-cortex

.github/workflows/preview-docs.yml
```diff
@@ -58,11 +63,18 @@ jobs:
         shell: bash
         run: pnpm install --ignore-workspace
 
+      - name: Build Documentation for Merge Queue
+        if: github.event_name == 'merge_group'
+        working-directory: ./docs/
+        shell: bash
+        run: pnpm build
```
trevor-cortex, line 70: Non-blocking thought: with this gated on `merge_group` only, PRs from forks and dependabot still build nothing at all — the Vercel steps are skipped for them and this step doesn't run, so the soon-to-be-required gate passes on those PRs without ever compiling the docs, and a broken docs change would only surface when it bounces out of the merge queue. If you want earlier signal, you could widen this step to also run whenever the Vercel deploy conditions aren't met, e.g.:

```yaml
if: github.event_name == 'merge_group' || github.event.pull_request.head.repo.full_name != github.repository || github.actor == 'dependabot[bot]'
```

The merge queue already backstops this, so totally fine to leave as-is.

trevor-cortex (commented):
This PR extends the `preview-docs` workflow to run in the merge queue so that the `preview-docs-gate` job can be made a required check. It adds the `merge_group` trigger, a conditional checkout before `dorny/paths-filter` (required because paths-filter can only use the PR API on `pull_request` events — for other events it needs a local git checkout and diffs against the default branch), a `pnpm build` step for merge-queue runs, and explicit `github.event_name == 'pull_request'` guards on the Vercel steps.

The logic looks sound — LGTM:

- The Vercel deploy would already have been skipped on `merge_group` (the old `if` would evaluate falsy since `github.event.pull_request` is empty), but making it explicit — and also gating the previously-unconditional `Install Vercel CLI` step — is clearly better.
- The `Comment on PR with Preview URL` step checks `steps.vercel_deploy.outcome == 'success'`, which correctly evaluates to `'skipped'` in the merge queue, so no changes needed there.
- The gate job runs for both event types and requires no changes. No changeset needed (CI-only change), per repo convention.

Two things worth knowing / verifying, neither blocking:

1. **paths-filter semantics differ between the two events.** On `merge_group`, paths-filter diffs the merge group ref against the default branch, so the combined changes of *all* PRs in the group are considered — a docs change in any queued PR triggers the build for the whole group. That's the correct behavior for a merge queue (the group is tested as a unit), just be aware the docs job may run on groups where "your" PR didn't touch docs. The shallow default checkout is fine; paths-filter deepens the fetch itself.
2. **`pnpm build` isn't identical to what Vercel runs.** `vercel build` pulls project settings and preview environment variables from Vercel before building. If the docs build depends on any env vars configured in the Vercel project, the plain `pnpm build` in the merge queue could fail (or silently build differently). Worth confirming a merge-queue run passes before flipping the check to required.

For subsequent reviewers: the main thing to double-check is the required-check configuration itself — the branch protection / merge queue settings should require `preview-docs-gate` (which always runs), not `preview-docs` (which is skipped when no docs changed).

.github/workflows/preview-docs.yml
```diff
@@ -29,6 +30,10 @@ jobs:
     outputs:
       docs: ${{ steps.filter.outputs.docs }}
     steps:
+      - name: Checkout Merge Group
+        if: github.event_name == 'merge_group'
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
trevor-cortex, line 35: Anchor note for future readers: this checkout is what switches `dorny/paths-filter` from API mode (used on `pull_request`) to local-git mode. Without a `base:` input it diffs against the default branch, which for a merge group ref means the combined changes of all queued PRs — the right semantics here, since the queue tests the group as a unit.

.github/workflows/preview-docs.yml
mcintyre94: Intentional, `pnpm install --ignore-workspace` fails on dependabot PRs (until our follow up action fixes the lock file)

lorisleiva (approved):

PR #2013, opened 2026-08-28
Unify the concurrent leaf executor’s callback contract with `createTransactionPlanExecutor`

mcintyre94:
I think this API as published was a mistake. Its callback has each leaf return a `SingleTransactionPlanResult`, which worked as a specialised helper in the kit-plugins PRs, but diverges from `createTransactionPlanExecutor`.

This PR changes it to use the same config (so, the same callback shape) as `createTransactionPlanExecutor`: it returns whatever it specified as `TContext`. The executor handles wrapping this in a successful response. It receives the mutable `Partial<TContext>` and is able to set values on it during execution. If it fails/is cancelled, the partial context is made available on the result.

This is a breaking change to the new API, but I've made it a minor because it's brand new and unused. I think it's worth aligning this now, making its consumers simpler and making it more straightforward for anyone else using it.

### DIFF
--- a/packages/instruction-plans/src/transaction-plan-executor.ts
+++ b/packages/instruction-plans/src/transaction-plan-executor.ts
@@ -65,37 +65,18 @@ type ExecuteTransactionMessage<TContext extends TransactionPlanResultContext> = 
  * Configuration object for creating a new transaction plan executor.
  *
  * @see {@link createTransactionPlanExecutor}
+ * @see {@link createTransactionPlanExecutorWithConcurrentLeaves}
  */
 export type TransactionPlanExecutorConfig<
     TContext extends TransactionPlanResultContext = TransactionPlanResultContext,
 > = {
     /**
-     * Called whenever a transaction message must be sent to the blockchain.
+     * Called whenever a transaction message must be executed.
      *
      * It should return the context that the successful result must carry — every property
      * `TContext` promises, which by default includes a `signature`.
      */
     executeTransactionMessage: ExecuteTransactionMessage<TContext>;
-};
-
-/**
- * Configuration for an executor that processes every transaction plan leaf concurrently.
- *
- * @typeParam TContext - The context carried by each single transaction plan result.
- *
- * @see {@link createTransactionPlanExecutorWithConcurrentLeaves}
- */
-export type ConcurrentLeafTransactionPlanExecutorConfig<TContext extends TransactionPlanResultContext> = {
-    /**
-     * Executes a single transaction plan and returns its result.
-     *
-     * The executor calls this function concurrently for every leaf in the plan and forwards
-     * the optional abort signal.
-     */
-    executeSingleTransactionPlan: (
-        transactionPlan: SingleTransactionPlan,
-        config?: { abortSignal?: AbortSignal },
-    ) => Promise<SingleTransactionPlanResult<TContext>>;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


mcintyre94:
@trevor-cortex

packages/instruction-plans/src/transaction-plan-executor.ts
```diff
@@ -228,51 +209,58 @@ export function createTransactionPlanExecutor<
...
  *
- * @typeParam TContext - The caller-defined context carried by each single transaction plan result.
- * @param config - Configuration containing the function used to execute each leaf.
+ * @typeParam TContext - The context carried by each single transaction plan result.
+ * @param config - Configuration object containing the transaction message executor function.
  * @return A {@link TransactionPlanExecutor} that processes all transaction plan leaves concurrently.
  * @throws {@link SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN}
- *   if any leaf callback throws or returns a failed or canceled result. The error context contains
- *   a `transactionPlanResult` property with the complete result tree.
+ *   if any leaf callback throws or is canceled. The error context contains a
```
trevor-cortex: Tiny wording nit: it's the *leaf* that gets canceled, not the callback (canceled leaves never invoke their callbacks, per the paragraph above). Also, a plan can throw this error with an all-successful tree when the abort signal fires after every leaf has settled. Something like "if any leaf callback throws, any leaf is canceled, or the abort signal fires" would be more precise.

trevor-cortex (commented):
**Verdict: LGTM** (submitting as a comment — I don't have approve permissions on this repo).

This unifies `createTransactionPlanExecutorWithConcurrentLeaves` with `createTransactionPlanExecutor`: both now take the same `TransactionPlanExecutorConfig`, so the concurrent executor's callback receives a mutable per-leaf `Partial<TContext>` plus the message and returns the complete `TContext`, instead of building a whole `SingleTransactionPlanResult` itself. `ConcurrentLeafTransactionPlanExecutorConfig` is removed, and tests/typetests are updated to the new contract.

**What I verified:**

- The new leaf path in `traverseTransactionPlanLeavesConcurrently` mirrors `traverseSingle` faithfully: fresh empty context per leaf, `{ ...context, ...returnedContext }` merge on success (returned value wins, stored-but-unreturned properties kept), and context preservation on throw/abort. The semantics of the two executors now genuinely match, not just their types.
- The doc claim that "a value the callback resolves with afterwards is discarded" on abort holds: `failedSingleTransactionPlanResult` snapshots the context via `Object.freeze({ ...context })` at creation time, so a still-running callback can't retroactively mutate the failed result even though it holds a reference to the same mutable context object.
- The typetest documenting the inference quirk (a callback that declares parameters becomes context-sensitive, so `TContext` falls back to the signature-requiring default) is accurate and inherited from `createTransactionPlanExecutor` — consistent behavior is exactly the point of this PR.
- The removed tests (callback returning failed/canceled results) are correctly obsolete: the callback can no longer express those outcomes, and throw/abort paths remain covered.

**Notes for other reviewers:**

- The `minor` bump for a breaking change is justified in the PR description (API shipped in 8.1.0 and unused). I find the reasoning sound, but it's a release-policy judgment call worth a second pair of eyes.
- `ConcurrentLeafTransactionPlanExecutorConfig` and `executeSingleTransactionPlan` disappear from the public surface — worth a quick grep of the kit-plugins branches that motivated the original API so their consumer updates land in tandem.

lorisleiva (approved):
Nice, yeah the fact that they now share the same config feels like a better design. Thank you!

PR #2016, opened 2026-08-29
Add the `getAgGenesisCert` RPC method

mcintyre94:
This is a new RPC method giving information on Alpenglow for the cluster

Also adds `getTransactionsForAddress` to the `isSolanaRequest` check, which was previously missed

Supersedes #1703

### DIFF
--- a/packages/rpc-api/src/getAgGenesisCert.ts
+++ b/packages/rpc-api/src/getAgGenesisCert.ts
@@ -1,0 +1,34 @@
+import type { Slot } from '@solana/rpc-types';
+
+/**
+ * The Alpenglow genesis certificate, as gossiped between validators.
+ *
+ * This mirrors the `WireBlockCertMessage` type in the Agave validator.
+ */
+type GetAgGenesisCertApiResponse = Readonly<{
+    /** The block that the certificate is certifying */
+    block: Readonly<{
+        /** The block id of the certified block, as a 32-byte array */
+        blockId: readonly number[];
+        /** The slot of the certified block */
+        slot: Slot;
+    }>;
+    /** The signature of the certificate message */
+    signature: Readonly<{
+        /** Bitmap of the ranks of the validators included in the aggregate signature */
+        bitmap: readonly number[];
+        /** The 192-byte aggregate BLS signature */
+        signature: readonly number[];
+    }>;
+}>;
+
+export type GetAgGenesisCertApi = {
+    /**
+     * Returns the Alpenglow genesis certificate, or `null` if the node does not have one.
+     *
+     * @returns The certificate over the block at which the Alpenglow consensus protocol was
+     * activated.
+     * @see https://solana.com/docs/rpc/http/getaggenesiscert
+     */
+    getAgGenesisCert(): GetAgGenesisCertApiResponse | null;
+};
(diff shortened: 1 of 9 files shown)


mcintyre94:
@trevor-cortex

packages/rpc-api/src/getAgGenesisCert.ts
```diff
@@ -0,0 +1,33 @@
...
+        slot: Slot;
+    }>;
+    /** The signature of the certificate message */
+    signature: Readonly<{
+        /** Bitmap of the ranks of the validators included in the aggregate signature */
+        bitmap: number[];
+        /** The 192-byte aggregate BLS signature */
+        signature: number[];
+    }>;
+}>;
```
trevor-cortex, line 23: Convention nit: response types elsewhere in `rpc-api` use `readonly` arrays inside `Readonly<{...}>` shapes (e.g. `readonly Lamports[]` in `getBlockCommitment.ts`). Suggest `readonly number[]` for `blockId`, `bitmap`, and `signature` — and the corresponding update in the typetest.

packages/rpc-api/src/getAgGenesisCert.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export type GetAgGenesisCertApi = {
+    /**
+     * Returns the Alpenglow genesis certificate, or `null` if the node does not have one.
+     *
+     * @returns The certificate over the block at which the Alpenglow consensus protocol was
+     * activated.
+     */
```
trevor-cortex, line 32: Other methods in this package include an `@see https://solana.com/docs/rpc/http/...` link in the method doc. If a docs page exists (or is planned) for `getAgGenesisCert`, worth adding here; if not yet, fine to leave as is.

trevor-cortex (commented):
This PR adds the `getAgGenesisCert` RPC method to `@solana/rpc-api`, returning the Alpenglow genesis certificate (or `null` when the node doesn't have one). It wires the method into `SolanaRpcApiForAllClusters`, registers allowed numeric keypaths so the byte arrays (`blockId`, `bitmap`, `signature`) stay as plain numbers while `slot` is upcast to `bigint`, and adds the method to the `isSolanaRequest` allowlist — along with `getTransactionsForAddress`, which was previously missing from that list (a real fix, correctly shipped as a separate `patch` changeset).

Overall this is a clean, well-tested addition. The keypath wildcards are correct, test coverage spans the transformer behavior, a live-validator integration test, and a typetest, and both changesets look appropriately scoped and leveled.

A couple of minor points, anchored inline:

- The response type uses mutable `number[]` where the rest of `rpc-api` uses `readonly T[]` inside `Readonly<{...}>` response shapes.
- The method doc is missing the customary `@see` link — possibly deliberate if no docs page exists yet for this method.

- Placement in `SolanaRpcApiForAllClusters` seems right: the method exists on any node running a recent Agave build and returns `null` when there's no certificate, so it isn't cluster-gated the way `requestAirdrop` is. Worth a second opinion if Alpenglow-era methods are meant to be grouped differently.
- The `it.todo` for the positive-path integration test (validator *with* a genesis certificate) is a reasonable deferral since the shared test validator can't produce one today, but it's worth tracking so it doesn't linger.

packages/rpc-api/src/getAgGenesisCert.ts
mcintyre94: Added this for now, it's an Agave method that we can assume will be documented. We can drop this later if we find otherwise

lorisleiva (approved):

PR #2025, opened 2026-09-03
Add `HasAddress` and account meta overrides to program client inputs

lorisleiva:
This PR makes the instruction builders of generated program clients accept a much wider range of account inputs. The original motivation is interoperability: for instance, we want web3.js v3 to be able to pass its `PublicKey` class straight into Kit-generated helpers. Rather than special-casing that class, this PR introduces a structural `HasAddress` interface — any object exposing an `address: Address` property — so web3.js, and any other framework with its own address wrapper, can plug into generated clients by simply conforming to it.

## What this means for Kit

**New types.** `@solana/addresses` gains `HasAddress` (a shape that `TransactionSigner`, `AccountMeta` and transaction message fee payers already satisfy), and `@solana/instructions` gains `AccountNonSignerMeta` — the counterpart of `AccountSignerMeta` for the two non-signer roles.

**Wider instruction account inputs.** `@solana/program-client-core` defines the two unions that generated input types will use, giving callers all of the following options:

```ts
// Non-signer accounts — InstructionAccountInput<T>:
getTransferSolInstruction({ destination: address('9r5R…kFPR'), … });          // an Address, as before
getTransferSolInstruction({ destination: myPublicKey, … });                    // any { address } object, e.g. a web3.js PublicKey
getTransferSolInstruction({ destination: await findVaultPda({ owner }), … }); // a PDA, no more [address] unwrapping
getTransferSolInstruction({ destination: { address, role: AccountRole.READONLY }, … }); // an explicit role override

// Signer accounts — InstructionSignerInput<T>:
getTransferSolInstruction({ source: mySigner, … });                            // a TransactionSigner, as before
getTransferSolInstruction({ source: { address, role: AccountRole.READONLY_SIGNER, signer }, … }); // a role override
```

**Principled signer handling.** `ResolvedInstructionAccount` gains an optional `isSigner` flag describing the IDL's requirement. Implicit values now follow the IDL: a `TransactionSigner` passed to a non-signer account acts as a plain address carrier instead of being silently upgraded to a signer (an unintended authority grant under Solana semantics), and a missing signer for a required signer account throws a new error pointing at `createNoopSigner`. Omitting the flag is equivalent to `'either'`, which preserves the exact behaviour of previously generated clients. Explicit account metas always take precedence, serving as an escape hatch when the IDL is wrong.

**Type-level mirrors.** New `ResolvedInstructionAccountMeta` and `InstructionAccountInputAddress` helpers reproduce the runtime role logic and address extraction at the type level, so generated builders can accurately type the account metas they return — including inferring address brands out of `HasAddress` wrappers and PDA tuples. Their typetests replicate the full generated-client machinery (instruction type, input type and builder signature) and pin down the exact signature shapes the renderer must emit, including a canary guarding the TypeScript inference behaviour they rely on.

## What this means for the Codama JS renderer

A follow-up PR in `codama-idl/renderers-js` will adopt these types once published: the five-way account input branch collapses to the two unions above, `originalAccounts` gains the `isSigner` flag, and the instruction builders capture the caller's input in a single `TInput` type parameter resolved through the new helpers. This was validated end-to-end against a patched local renderer: all e2e programs regenerate, typecheck and pass their tests, and existing generated clients are unaffected since the new runtime behaviour is gated behind the `isSigner` flag that only newly generated code emits.

## Appendix: the full resolution matrix

For the curious, this is how `getAccountMetaFactory` resolves every combination of IDL `isSigner` setting and provided input. Legend: **`W`** = the IDL's writable flag as a role (`isWritable ? WRITABLE : READONLY`), **`W+S`** = `upgradeRoleToSigner(W)`, **`R`** = the explicit role carried by the input. ✅ = admitted by the generated input types, ⚠️ = honoured at runtime but not advertised in the types (deliberate escape hatch), ❌ = type error (runtime behaviour shown for JS callers).

| Provided input ↓ / IDL `isSigner` → | `false` | `'either'` (or omitted) | `true` |
| --- | --- | --- | --- |
| `Address` | ✅ `{ addr, W }` | ✅ `{ addr, W }` | ❌ → throws (points at `createNoopSigner`) |
| `HasAddress` — `{ address }` | ✅ `{ addr, W }` | ✅ `{ addr, W }` | ❌ → throws |
| `ProgramDerivedAddress` — `[addr, bump]` | ✅ `{ addr, W }` | ✅ `{ addr, W }` | ❌ → throws |
| `AccountNonSignerMeta` — `{ address, role: R }` | ✅ `{ addr, R }` | ✅ `{ addr, R }` | ⚠️ `{ addr, R }` (explicit demotion) |
| `TransactionSigner` | ✅ `{ addr, W }` — address carrier, no upgrade | ✅ `{ addr, W+S, signer }` | ✅ `{ addr, W+S, signer }` |
| `AccountSignerMeta` — `{ address, role: R, signer }` | ⚠️ `{ addr, R, signer }` (explicit promotion) | ✅ `{ addr, R, signer }` | ✅ `{ addr, R, signer }` |
| `null` / absent (optional account) | *same in every column:* strategy `'omitted'` → dropped; `'programId'` → `{ programAddress, READONLY }` | | |

*(Clients generated before the `isSigner` flag existed omit it entirely, so every account behaves per the `'either'` column — i.e. exactly as before this PR.)*


lorisleiva:
* **#2025** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2025?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2025?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/program-client-core/src/instruction-input-resolution.ts
```diff
@@ -129,22 +170,29 @@ export function getResolvedInstructionAccountAsTransactionSigner<T extends strin
     inputName: string,
     value: ResolvedInstructionAccount<T>['value'] | undefined,
 ): TransactionSigner<T> {
-    if (!isResolvedInstructionAccountSigner(value)) {
+    const signer = value && typeof value === 'object' && 'signer' in value ? value.signer : value;
```
trevor-cortex: This unwraps `.signer` whenever the value has a `signer` key, whereas `getAccountMetaFactory` treats a value as an explicit meta only when it has a numeric `role` (`hasExplicitRole`). The two paths can disagree for a `TransactionSigner` implementation that happens to carry a `signer` property (e.g. a wrapper signer delegating to an inner one): the factory would attach the outer signer, this function would return the inner one (or throw). Contrived, but cheap to keep consistent:

```ts
const signer = value && hasExplicitRole(value) && 'signer' in value ? value.signer : value;
```

`hasExplicitRole` is a function declaration so hoisting is fine even though it's defined further down.

packages/program-client-core/src/instruction-input-resolution.ts
```diff
@@ -153,35 +201,164 @@ export function getResolvedInstructionAccountAsTransactionSigner<T extends strin
...
+ * @see {@link InstructionAccountInputAddress}
+ */
+export type ResolvedInstructionAccountMeta<TInput, TAddress extends string, TSignerMeta = TAddress> = [TInput] extends [
+    { role: AccountRole },
+]
+    ? AccountMeta<TAddress> &
+          ([TInput] extends [{ signer: TransactionSigner<TAddress> }] ? AccountSignerMeta<TAddress> : unknown)
+    : [TInput] extends [TransactionSigner<TAddress>]
+      ? TSignerMeta
+      : TAddress;
```
trevor-cortex, line 335: Two non-blocking observations on the explicit-role branch:

1. Since `AccountSignerMeta<T>` already `extends AccountMeta<T>` with the role narrowed to the two signer roles, `AccountMeta<T> & AccountSignerMeta<T>` is just `AccountSignerMeta<T>`. The intersection works, but `? AccountSignerMeta<TAddress> : AccountMeta<TAddress>` reads more directly (the `Equal` typetests would need to compare against `AccountSignerMeta<'8888'>` etc.).

2. The docblock says the meta "widens to `AccountMeta` since the role is only known at runtime", but when the input is an `AccountNonSignerMeta` the type system does know the role is non-signer, and when the caller writes `{ address, role: AccountRole.READONLY }` inline, `TInput['role']` is inferred as the `AccountRole.READONLY` literal. Something like `AccountMeta<TAddress> & { readonly role: TInput['role'] }` (or `Pick<TInput, 'role'>`) would let an explicit `READONLY` override type as `ReadonlyAccount<TAddress>` instead of the full `AccountMeta` union, which is a nice precision win for the override use case. Totally fine to leave for a follow-up if you'd rather keep this PR focused.

packages/program-client-core/src/__tests__/instruction-input-resolution-test.ts
```diff
@@ -178,6 +212,109 @@ describe('getAccountMetaFactory', () => {
...
+            isSigner: false,
+            isWritable: false,
+            value: { address: mockAddress, role: AccountRole.READONLY_SIGNER, signer: mockSigner },
+        });
+        expect(promotedMeta).toEqual({ address: mockAddress, role: AccountRole.READONLY_SIGNER, signer: mockSigner });
+    });
+
+    it('ignores non-numeric role properties on address-bearing objects', () => {
+        const toAccountMeta = getAccountMetaFactory(mockAddress, 'programId');
+        const value = { address: mockAddress, role: 'admin' } as unknown as Address;
```
trevor-cortex: Nit: `as unknown as Address` is a slightly misleading cast for an object value. `as HasAddress` (imported from `@solana/addresses`) is sufficient here since the literal is directly assignable, and it better describes what the test is exercising.

trevor-cortex (commented):
## Summary

This PR widens what generated instruction builders accept as account inputs and makes signer handling follow the IDL rather than the input:

- `@solana/addresses` gains `HasAddress<T>` (structurally `Readonly<{ address: Address<T> }>`); `TransactionMessageWithFeePayer.feePayer` is retyped to it (no-op structurally).
- `@solana/instructions` gains `AccountNonSignerMeta<T>` (`ReadonlyAccount | WritableAccount`).
- `@solana/program-client-core` gains `InstructionAccountInput` / `InstructionSignerInput` unions, an optional `isSigner?: boolean | 'either'` on `ResolvedInstructionAccount`, and two type-level mirrors (`ResolvedInstructionAccountMeta`, `InstructionAccountInputAddress`). `getAccountMetaFactory` now honours explicit `role` overrides first, only upgrades `TransactionSigner`s when `isSigner !== false`, and throws the new `RESOLVED_INSTRUCTION_INPUT_MUST_BE_SIGNER` error when `isSigner === true` and no signer was supplied.
- Runtime tests cover the full resolution matrix (including a mock of the generated builder body), and the typetests replicate the generated-client machinery end to end, including a canary for TS's constraint-only inference behaviour.

I walked through the resolution matrix against the code and every cell checks out. The `[TInput] extends [...]` non-distributive checks, the `?? 'either'` fallback for pre-flag clients, and the `typeof role === 'number'` guard are all sensible choices. Docblocks and READMEs are in place, changesets are present with reasonable bump levels, and error code/context/message were added following the `@solana/errors` conventions. **This would be an approve from me** — I don't have repository access to formally approve, so submitting as a comment. Nothing below is blocking.

## Things to watch out for

- **Detection asymmetry between the factory and `getResolvedInstructionAccountAsTransactionSigner`.** The factory decides "this is an explicit meta" via `hasExplicitRole` (numeric `role`), while the signer extractor decides via `'signer' in value`. A `TransactionSigner` implementation with an incidental `signer` property would be treated differently by the two paths. Cheap to align — see inline.
- **`HasAddress` absorbs more than the type names suggest.** `AccountSignerMeta`, `AccountLookupMeta`, and any object with an `address` string are all admitted by `InstructionAccountInput`. The PR is explicit about the `AccountSignerMeta` case being a deliberate escape hatch (⚠️ cells in the matrix), and the typetest documents it honestly. Just noting it for reviewers: with a lookup meta the factory would return `{ address, role }` and silently drop `lookupTableAddress`/`addressIndex`. Not something anyone should be doing, so I don't think it needs handling.
- **Explicit type arguments lose signer precision.** With `getFoo<'8888', '9999', '7777'>({ owner: signer9 })`, `TInput` defaults to the declared union, `ResolvedInstructionAccountMeta` falls back to `TAddress`, and the `'either'` account types as `ReadonlyAccount` even though the runtime produces a signer meta. The typetest pins this down and it's a conscious trade-off (the current renderer's `(typeof input)['owner'] extends TransactionSigner ? … : …` distributes into a union here, so neither is precise). Worth a heads-up in the renderer PR description since it's user-visible.

- **Optional accounts aren't covered in the integration typetests.** The runtime test mocks `mint?: InstructionAccountInput | null`, but the `ResolvedInstructionAccountMeta (generated instruction integration)` describe block only exercises required accounts. When a caller omits an optional account, `TInput['mint']` resolves to `unknown` (indexed access on an instantiated type without the key, no access node), and when they pass `null` it's `null` — both should fall through to `TAddress`, but that `unknown` path is non-obvious and worth pinning with `Equal<…, WritableAccount<string>>`-style assertions for the omitted / `null` / provided cases. Suggest a follow-up or a small addition here.
- The docblock on `ResolvedInstructionAccountMeta` and the `IMPORTANT` typetest comment both stress that the renderer must emit `Input<…> & TInput` as the parameter type, not just `TInput extends Input<…>`. That's the key contract for the codama-idl/renderers-js follow-up — reviewers of that PR should check the emitted signature matches the shape in the `getMockInstructionBuilder` typetest exactly.
- `@solana/transaction-messages` gets a `minor` for a structurally identical retype. In lockstep versioning it makes no practical difference, so I'm not asking for a change — just flagging in case someone prefers `patch` for pure type-alias swaps.

.changeset/jolly-cups-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solana/instructions': minor
+---
+
+Add a new `AccountNonSignerMeta` type representing an `AccountMeta` whose role is guaranteed not to be a signer role — i.e. `ReadonlyAccount | WritableAccount`. It is the counterpart of the `AccountSignerMeta` type from `@solana/signers`. Additionally, the `role` member of `WritableAccount`, `ReadonlySignerAccount` and `WritableSignerAccount` is now marked `readonly`, consistently with `ReadonlyAccount` and `AccountMeta` (a structural no-op since `AccountMeta.role` was already `readonly`).
```
trevor-cortex: The "structural no-op" claim isn't accurate. In an intersection, TS only makes a property `readonly` when every constituent does, so `AccountMeta & { role: AccountRole.WRITABLE }` previously exposed a *mutable* `role` — that's exactly why `durable-nonce-test.ts` had to be rewritten in this PR to stop assigning `instruction.accounts[2].role`. Suggest rewording to something like: "…is now marked `readonly`, consistently with `ReadonlyAccount` and `AccountMeta`. Code that mutated `role` on these types will now fail to compile." Bump level is a maintainer call; I'd personally keep `minor` since mutation was already contrary to `AccountMeta`'s contract, but the note should be honest about it.

trevor-cortex (commented):
## Re-review

Re-fetched everything since my last pass. All the points from my previous review have been addressed:

- `getResolvedInstructionAccountAsTransactionSigner` now gates the `signer` unwrap on `hasExplicitRole(value)`, matching the factory, with a test for the incidental-`signer`-property case.
- `ResolvedInstructionAccountMeta` now preserves `TInput['role']` (`& { readonly role: TRole }`) and picks `AccountSignerMeta` vs `AccountMeta` based on the presence of `signer`, so inline `role: AccountRole.READONLY` overrides resolve to `ReadonlyAccount<…>` exactly. The redundant `AccountMeta & AccountSignerMeta` intersection is gone.
- Optional-account typetests were added for both builder shapes (omitted / `null` / provided), pinning the `unknown` → `string` fallback path.
- `as HasAddress` in the test.

The runtime matrix still checks out against `getAccountMetaFactory`, and the type helper mirrors it — including the `infer TRole extends AccountRole` guard, which rejects a non-numeric `role` at the type level the same way `typeof role === 'number'` does at runtime.

**Still an approve in substance** (submitting as a comment since I don't have repo access to formally approve).

## One new thing to watch

The role-type preservation forced `readonly role` onto `WritableAccount`, `ReadonlySignerAccount` and `WritableSignerAccount` in `@solana/instructions`. That's the right shape, but it is **not** a structural no-op as the `jolly-cups-call` changeset claims: in an intersection, TS only treats a property as `readonly` when *every* constituent marks it so, which is why `AccountMeta & { role: X }` used to expose a mutable `role` — and why `durable-nonce-test.ts` had to stop doing `instruction.accounts[2].role = …` in this very PR. Any downstream code that mutates `.role` on one of these aliases will break at compile time. I don't think that warrants a `major` on its own (mutating a meta's role was already contrary to `AccountMeta`'s contract), but the changeset wording should say so plainly rather than call it a no-op — see inline.

- The renderer contract is unchanged from my previous review: emit `Input<…> & TInput` as the parameter type with `TInput` defaulted to the concrete input type. The `getMockInstructionBuilder` typetest remains the executable spec.
- Edge case not worth blocking on: `HasAddress` still admits `{ address, role: AccountRole.READONLY, signer }` for a non-signer account. Runtime returns `{ address, role: READONLY, signer }`; the type helper yields `AccountSignerMeta & { role: READONLY }` (i.e. `role: never`). Nobody should be constructing that, so I'd leave it.

trevor-cortex (commented):
## Re-review (3rd pass)

Re-fetched all 17 files. The only change since my previous pass is the `jolly-cups-call` changeset, which now states plainly that `role` on `WritableAccount` / `ReadonlySignerAccount` / `WritableSignerAccount` is `readonly` and that code mutating it will stop compiling — with the justification that such mutation was already contrary to `AccountMeta`'s contract and would throw on frozen metas at runtime. That resolves the one point I raised last time; the `minor` bump is consistent with that framing.

Source, tests and typetests are byte-for-byte what I reviewed previously, so the earlier verdict stands: the runtime resolution matrix matches `getAccountMetaFactory`, the type helper mirrors it (including the `infer TRole extends AccountRole` guard paired with `typeof role === 'number'`), the signer-unwrap path is aligned via `hasExplicitRole`, and optional accounts are pinned in the integration typetests.

**Approve in substance** — submitting as a comment since I don't have repository access to approve formally. Nothing outstanding from my side.

- Renderer contract for the codama-idl/renderers-js follow-up is unchanged: emit `Input<…> & TInput` as the parameter type, default `TInput` to the concrete input type, and use `ResolvedInstructionAccountMeta` / `InstructionAccountInputAddress` per the `getMockInstructionBuilder` typetest.
- Previously noted, non-blocking: explicit type arguments (`getFoo<'8888', …>({ owner: signer })`) lose signer precision for `'either'` accounts; `HasAddress` structurally admits `AccountLookupMeta` and `{ address, role: READONLY, signer }` (the latter typing as `role: never`). All deliberate or not worth handling.

dhl:
Fantastic change to allow program clients to accept richer representation of account types. Thanks @lorisleiva!

PR #2029, opened 2026-09-04
Add u256/i256 number codecs and extract shared word-combining helpers

lorisleiva:
This PR introduces `u256` and `i256` number codecs to `@solana/codecs-numbers`, extending the large-integer codecs one width beyond the existing `u128`/`i128`. Each exposes the usual `getX256Encoder`, `getX256Decoder`, and `getX256Codec` trio, serializes to 32 bytes, supports little- and big-endian ordering via the `endian` option, and always decodes to a `bigint`.

Because `DataView` has no native 128-bit accessor, multi-word integers are serialized as a sequence of 64-bit words. Rather than duplicate that split-and-reassemble logic across four files, this PR extracts two shared helpers, `setBigIntWords` and `getBigIntWords`, into `utils.ts` and refactors all four codecs (`u128`, `i128`, `u256`, `i256`) onto them. The helpers handle endianness, the signed most-significant word, and two's-complement reassembly via `BigInt.asIntN`. The existing `u128`/`i128` test suites pass unchanged, confirming the refactor is behaviour-equivalent.

Documentation for the new codecs is intentionally deferred to a follow-up, since the docs site builds against the published `@solana/kit` and cannot reference the new symbols until a release ships.

This will enable Anchor v2 to use Kit codecs under the hood without having to re-implement their own `u256`/`i256` codecs.


lorisleiva:
* **#2029** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2029?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2029?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds `u256`/`i256` codecs (encoder/decoder/codec trio, 32 bytes, LE/BE, always decodes to `bigint`) and extracts the multi-word split/reassemble logic into `setBigIntWords` / `getBigIntWords` in `utils.ts`, then moves `u128`/`i128` onto those helpers. Also adds typetests, a `minor` changeset, and cross-links the 128-bit docblocks to the new 256-bit codecs.

I walked through the helpers for both endiannesses and both signednesses:

- **Offsets**: `i === 0` is the MSW; in LE it lands at `(wordCount - 1) * 8`, in BE at `0`. Correct for both.
- **Encoding negatives**: lower words use `(v >> shift) & U64_MASK`, which on a negative `bigint` yields the correct two's-complement low 64 bits (BigInt bitwise ops behave as infinite-width two's complement). MSW at `-2^255` shifts to exactly `-2^63`, which `setBigInt64` accepts.
- **Decoding**: unsigned reassembly then `BigInt.asIntN(wordCount * 64, …)` is equivalent to the old `(signedLeft << 64n) + right` for i128, so the refactor is behaviour-preserving as claimed.
- Test vectors for `MIN`, `MIN + 1`, `MAX`, `MAX - 1`, `-1`, `-42` in both endiannesses check out by hand.

`@solana/codecs` re-exports `codecs-numbers` via `export *`, so nothing else needs wiring up. Code-wise this looks good to merge (would approve, but the bot doesn't have approve rights on this repo).

## Things to address

**Package README is now stale.** The PR description defers docs to a follow-up because the docs *site* builds against the published Kit — that's correct for `solanakit.com`, but `packages/codecs-numbers/README.md` lives in this repo and isn't subject to that constraint. It currently says "ten codecs of five different byte sizes" / "Five of them store unsigned integers and the other five store signed integers" and "All integer codecs are of type `Codec<number>` except for the `u64`, `u128`, `i64` and `i128` codecs". Those sentences, plus the two example blocks (LE and BE), should pick up `u256`/`i256` in this PR per the repo's "add or update the package's README when adding a new public API" convention.

## Nits (inline, both optional)

- The `signed` branch in `setBigIntWords` is byte-identical to the unsigned branch, so the parameter could be dropped from the setter.
- `& U64_MASK` after `getBigUint64` is a no-op.

## For subsequent reviewers

Nothing risky here beyond the helper refactor, which the unchanged `u128`/`i128` suites cover. Worth a glance: the i256 range is written as `-BigInt('0x8000…')` whereas i128 uses `-BigInt('0x7fff…') - 1n` — same value, just a style difference; I don't think it matters.

packages/codecs-numbers/src/utils.ts
```diff
@@ -31,6 +31,78 @@ function isLittleEndian(config?: NumberCodecConfig): boolean {
...
+    for (let i = 0; i < wordCount; i++) {
+        // `i === 0` is the most significant word.
+        const shift = BigInt((wordCount - 1 - i) * 64);
+        // In little-endian, the least significant word is stored first.
+        const offset = (littleEndian ? wordCount - 1 - i : i) * 8;
+        if (signed && i === 0) {
+            view.setBigInt64(offset, bigintValue >> shift, littleEndian);
+        } else {
+            view.setBigUint64(offset, (bigintValue >> shift) & U64_MASK, littleEndian);
+        }
```
trevor-cortex, line 73: Optional simplification: this branch is byte-identical to the `else` branch. Per spec, `DataView.prototype.setBigInt64` runs the value through `ToBigInt64` (mod 2^64, then reinterpreted as signed) and `setBigUint64` through `ToBigUint64` (mod 2^64), so for any value `v`, `setBigInt64(off, v)` and `setBigUint64(off, v & U64_MASK)` write exactly the same 8 bytes. E.g. `-42n >> 192n === -1n`; `setBigInt64(-1n)` → `ff…ff`, and `setBigUint64(-1n & U64_MASK)` = `setBigUint64(0xffffffffffffffffn)` → `ff…ff`.

That means `signed` could be dropped from the setter entirely and the loop body reduced to the single `setBigUint64` call. Totally fine to keep it for symmetry with `getBigIntWords` (where `signed` *does* matter) — just flagging that the distinction is cosmetic on the write side.

packages/codecs-numbers/src/utils.ts
```diff
@@ -31,6 +31,78 @@ function isLittleEndian(config?: NumberCodecConfig): boolean {
...
+    view: DataView,
+    littleEndian: boolean | undefined,
+    wordCount: number,
+    signed: boolean,
+): bigint {
+    let value = 0n;
+    for (let i = 0; i < wordCount; i++) {
+        // In little-endian, the least significant word is stored first.
+        const offset = (littleEndian ? wordCount - 1 - i : i) * 8;
+        value = (value << 64n) | (view.getBigUint64(offset, littleEndian) & U64_MASK);
```
trevor-cortex: Nit: `getBigUint64` already returns a value in `[0, 2^64)`, so `& U64_MASK` is a no-op here. Harmless, but it reads as if the mask is doing something.

packages/codecs-numbers/src/utils.ts
lorisleiva: Keeping for symmetry.

PR #2030, opened 2026-09-04
Add tap codec helpers for observing values and bytes

lorisleiva:
This PR adds six `tap` helpers to `@solana/codecs-core` for observing the values and bytes flowing through a codec without modifying them. Any tap may throw to abort the operation, making them ideal for adding validation guards to existing codecs without resorting to an identity `transformEncoder`.

The **value family** observes the input value before encoding and the decoded value after decoding:

```ts
// Reject out-of-range input before it is encoded.
const boundedU8 = tapEncoder(getU8Encoder(), value => {
  if (value > 100) throw new Error('Value must not exceed 100');
});

// Validate the decoded result.
const nonZeroU8 = tapDecoder(getU8Decoder(), value => {
  if (value === 0n) throw new Error('Value must not be zero');
});
```

The **bytes family** observes the raw bytes after encoding and before decoding. Bytes are exposed as a `ReadonlyUint8Array`, so taps are pure observers:

```ts
// Guard malformed boolean bytes before decoding.
const safeBool = tapDecoderBytes(getBooleanDecoder(), (bytes, offset) => {
  if (bytes[offset] > 1) throw new Error('Expected a 0 or a 1 for booleans');
});
```

The `tapCodec` and `tapCodecBytes` variants mirror `transformCodec`'s `(codec, encodeTap, decodeTap?)` shape, and mixed guards compose cleanly:

```ts
// Guard the input value on encode and the raw bytes on decode.
const codec = combineCodec(
  tapEncoder(getBooleanEncoder(), value => { /* ... */ }),
  tapDecoderBytes(getBooleanDecoder(), (bytes, offset) => {
    if (bytes[offset] > 1) throw new Error('Expected a 0 or a 1 for booleans');
  }),
);
```

Each helper uses a single passthrough generic so the exact input type, including any extra properties, is preserved.

Documentation is intentionally deferred to a follow-up, since the docs site builds against the published `@solana/kit` and cannot reference the new symbols until a release ships.

### DIFF
--- a/.changeset/easy-results-sink.md
+++ b/.changeset/easy-results-sink.md
@@ -1,0 +1,5 @@
+---
+'@solana/codecs-core': minor
+---
+
+Add `tap` codec helpers for observing values and bytes without modifying them. The value family (`tapEncoder`/`tapDecoder`/`tapCodec`) observes the input value before encoding and the decoded value after decoding, whilst the bytes family (`tapEncoderBytes`/`tapDecoderBytes`/`tapCodecBytes`) observes the raw bytes after encoding and before decoding. Any tap may throw to abort the operation, making these helpers ideal for adding validation guards to existing codecs without an identity `transformEncoder`. For example, `tapDecoderBytes(getBooleanDecoder(), (bytes, offset) => { if (bytes[offset] > 1) throw new Error('Expected a 0 or a 1 for booleans'); })`.
--- a/packages/codecs-core/src/tap-codec.ts
+++ b/packages/codecs-core/src/tap-codec.ts
@@ -1,0 +1,173 @@
+import { Codec, createCodec, createDecoder, createEncoder, Decoder, Encoder } from './codec';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type AnyEncoder = Encoder<any>;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type AnyDecoder = Decoder<any>;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type AnyCodec = Codec<any>;
+
+/**
+ * Observes the input value of an encoder before it is encoded, without modifying it.
+ *
+ * This function takes an existing encoder and returns a new encoder of the exact same type that
+ * calls the provided `tap` function with each value before delegating to the original encoder.
+ * The value is passed through unchanged, making this useful for adding validation guards, logging,
+ * or other read-only side effects on the input value.
+ *
+ * If the `tap` function throws, the encoding is aborted and the error propagates to the caller.
+ *
+ * @typeParam TEncoder - The type of the encoder being tapped. The returned encoder has the same type.
+ *
+ * @param encoder - The encoder whose input value should be observed.
+ * @param tap - A function called with each value before encoding. It may throw to abort encoding.
+ * @returns A new encoder of the same type that observes its input value before encoding.
+ *
+ * @throws Whatever error the `tap` function throws, aborting the encoding.
+ *
+ * @example
+ * Guarding the input value of a `u8` encoder.
+ * ```ts
+ * const encoder = tapEncoder(getU8Encoder(), value => {
+ *     if (value > 100) throw new Error('Value must not exceed 100');
+ * });
+ * encoder.encode(42); // 0x2a
+ * encoder.encode(200); // Throws 'Value must not exceed 100'
+ * ```
+ *
+ * @remarks
+ * The tap fires from within the encoder's `write` step. For variable-size encoders, this means the
+ * encoded size is computed from the unvalidated value (via `getSizeFromValue`) *before* the tap runs,
+ * so a custom `getSizeFromValue` that rejects the value will throw before the tap has a chance to.
+ *
+ * To observe the encoded bytes instead of the input value, use {@link tapEncoderBytes}.
+ *
+ * @see {@link tapCodec}
+ * @see {@link tapDecoder}
+ * @see {@link tapEncoderBytes}
+ */
+export function tapEncoder<TEncoder extends AnyEncoder>(
+    encoder: TEncoder,
+    tap: (value: TEncoder extends Encoder<infer TFrom> ? TFrom : never) => void,
+): TEncoder {
+    return createEncoder({
+        ...encoder,
+        write: (value, bytes, offset) => {
+            tap(value);
+            return encoder.write(value, bytes, offset);
+        },
+    }) as TEncoder;
+}
+
+/**
+ * Observes the decoded value of a decoder after it is decoded, without modifying it.
+ *
+ * This function takes an existing decoder and returns a new decoder of the exact same type that
+ * calls the provided `tap` function with each decoded value before returning it. The value is
+ * passed through unchanged, making this useful for adding validation guards, logging, or other
+ * read-only side effects on the decoded value.
+ *
+ * If the `tap` function throws, the decoding is aborted and the error propagates to the caller.
+ *
+ * @typeParam TDecoder - The type of the decoder being tapped. The returned decoder has the same type.
+ *
+ * @param decoder - The decoder whose decoded value should be observed.
+ * @param tap - A function called with each decoded value. It may throw to abort decoding.
+ * @returns A new decoder of the same type that observes its decoded value.
+ *
+ * @throws Whatever error the `tap` function throws, aborting the decoding.
+ *
+ * @example
+ * Guarding the decoded value of a `u8` decoder.
+ * ```ts
+ * const decoder = tapDecoder(getU8Decoder(), value => {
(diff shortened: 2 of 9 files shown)


lorisleiva:
* **#2030** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2030?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2030?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/codecs-core/src/tap-codec.ts
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * @throws Whatever error the `tap` function throws, aborting the decoding.
+ *
+ * @example
+ * Guarding the decoded value of a `u8` decoder.
+ * ```ts
+ * const decoder = tapDecoder(getU8Decoder(), value => {
+ *     if (value === 0n) throw new Error('Value must not be zero');
+ * });
+ * decoder.decode(new Uint8Array([42])); // 42n
```
trevor-cortex: `getU8Decoder()` is a `FixedSizeDecoder<number, 1>`, so `value` is `number` here — `value === 0n` is a TS error ("This comparison appears to be unintentional because the types 'number' and 'bigint' have no overlap") and would never be true at runtime either. Should be `value === 0` and the result comment `// 42`.

packages/codecs-core/src/tap-codec.ts
```diff
@@ -0,0 +1,169 @@
...
+ * @example
+ * Guarding both the encoded and decoded values of a `u8` codec.
+ * ```ts
+ * const codec = tapCodec(
+ *     getU8Codec(),
+ *     value => {
+ *         if (value > 100) throw new Error('Value must not exceed 100');
+ *     },
+ *     value => {
+ *         if (value === 0n) throw new Error('Value must not be zero');
```
trevor-cortex: Same issue as above: `getU8Codec()` decodes to `number`, so this should be `value === 0`.

packages/codecs-core/src/tap-codec.ts
```diff
@@ -0,0 +1,169 @@
...
+    encoder: TEncoder,
+    tap: (value: TEncoder extends Encoder<infer TFrom> ? TFrom : never) => void,
+): TEncoder {
+    return createEncoder({
+        ...encoder,
+        write: (value, bytes, offset) => {
+            tap(value);
+            return encoder.write(value, bytes, offset);
+        },
+    }) as TEncoder;
```
trevor-cortex, line 59: Worth a sentence in the `@remarks`: the tap fires inside `write`, so for variable-size encoders `encode()` calls `getSizeFromValue(value)` on the unvalidated value *before* the tap runs (and the same happens when the encoder is nested in a struct — the parent sizes every field before writing any). Tapping `getSizeFromValue` too would double-fire the tap, so the current design is the right call; it's just a surprise waiting for anyone using this as a guard on an encoder whose `getSizeFromValue` throws on bad input.

.changeset/easy-results-sink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solana/codecs-core': minor
+---
+
+Add `tap` codec helpers for observing values and bytes without modifying them. The value family (`tapEncoder`/`tapDecoder`/`tapCodec`) observes the input value before encoding and the decoded value after decoding, whilst the bytes family (`tapEncoderBytes`/`tapDecoderBytes`/`tapCodecBytes`) observes the raw bytes after encoding and before decoding. Any tap may throw to abort the operation, making these helpers ideal for adding validation guards to existing codecs without an identity `transformEncoder`. For example, `tapDecoderBytes(getBooleanCodec(), (bytes, offset) => { if (bytes[offset] > 1) throw new Error('Expected a 0 or a 1 for booleans'); })`.
```
trevor-cortex: Nit: `tapDecoderBytes(getBooleanCodec(), …)` type-checks (a codec is a decoder) but reads as if the wrong helper is being used. `getBooleanDecoder()` matches the docblock example and the PR description.

trevor-cortex (commented):
## Summary

Adds six passthrough observers to `@solana/codecs-core`: `tapEncoder`/`tapDecoder`/`tapCodec` observe the value on either side, and `tapEncoderBytes`/`tapDecoderBytes`/`tapCodecBytes` observe the raw buffer (after `write`, before `read`). Each uses a single generic `TEncoder extends Encoder<any>` etc. so the input type — including fixed/variable-size discriminant and any extra props — round-trips exactly, which the typetests lock down with `Equal`. Runtime is a thin `createEncoder({ ...x, write })` / `createDecoder({ ...x, read })` wrap, matching the existing `reverseEncoder`/`transformDecoder` pattern. Changeset is `minor`, which is right for a new feature.

Implementation looks correct and the tests cover the important paths (offsets, non-zero read offsets, abort on throw, optional decode tap, size preservation). LGTM — the inline comments are docblock fixes and a couple of nits. (Submitting as a comment rather than an approval since I don't have approve rights on this repo.)

## Things to watch

- **Docblock examples don't type-check.** `getU8Decoder()`/`getU8Codec()` decode to `number`, not `bigint`, so `value === 0n` is a TS "no overlap" error and would never be true at runtime. Two spots, flagged inline.
- **`tapEncoder` runs inside `write`, not `getSizeFromValue`.** For variable-size encoders, `encode()` (and any parent struct's `getSizeFromValue`) evaluates the size on the unvalidated value *before* the tap fires. Not a bug, but for a helper pitched as a validation guard it's worth a sentence in the docblock so nobody is surprised when a custom `getSizeFromValue` blows up first. Flagged inline.
- **README.** `CLAUDE.md` asks for a README update when adding public API, and `packages/codecs-core/README.md` already has one section per helper (`transformCodec`, `resizeCodec`, `offsetCodec`, …) plus a "you may also be interested in `transformCodec`, `fixCodecSize` or `reverseCodec`" pointer near the top. Unlike the docs site, the README doesn't build against the published `@solana/kit`, so the deferral reason doesn't apply here. A short "Tapping codecs" section would fit naturally — fine as a follow-up if you'd rather keep this PR tight, but I'd not skip it.
- **Test asymmetry (nit).** `tap-codec-test.ts` has an "aborts encoding when the encode tap throws" case for `tapCodec`, but `tap-codec-bytes-test.ts` only covers the decode-abort path for `tapCodecBytes`. Cheap to add the mirror.

- The `as TEncoder` casts are unavoidable: `createEncoder` returns `Encoder<TFrom>`, not the caller's exact type. The typetests with `& { custom: 42 }` are what keep those casts honest.
- The `Encoder<any>`-bounded generic with an `infer` in the tap parameter is deliberately different from `transformCodec`'s overload ladder — it preserves the exact input type rather than reconstructing it, and it's why the tap's `value` type is inferred correctly for `Codec<number | string, string>` (typetest covers it).
- `tapDecoderBytes`/`tapEncoderBytes` type `bytes` as `ReadonlyUint8Array`, but the encoder tap is handed the real mutable buffer at runtime. That's a type-level contract only, consistent with how `ReadonlyUint8Array` is used elsewhere in the package.

.changeset/easy-results-sink.md
trevor-cortex: Nit (carried over): `tapDecoderBytes(getBooleanCodec(), …)` type-checks but reads oddly since the helper is decoder-side and every other example uses `getBooleanDecoder()`. Worth aligning since this text ends up in the changelog.

trevor-cortex (commented):
## Re-review

All points from the previous pass are addressed:

- `0n` → `0` in the `tapDecoder` and `tapCodec` docblock examples.
- `@remarks` on `tapEncoder` (and `tapEncoderBytes`) now spell out that `getSizeFromValue` runs on the unvalidated value before the tap fires.
- New "Tapping codecs" README section, slotted between "Transforming codecs" and "Fixing the size of codecs", plus `tapCodec` added to the helper list near the top. Examples type-check and read well.
- `tapCodecBytes` now has the mirror "aborts encoding when the encode tap throws" test.

Source, index exports, and typetests are otherwise unchanged from the last pass, so the implementation verdict stands: LGTM. (Commenting rather than approving since I don't have approve rights on this repo.)

One leftover nit inline: the changeset example still passes `getBooleanCodec()` to `tapDecoderBytes`. It type-checks (a `Codec` is a `Decoder`), but since it ships in the release notes it'd be cleaner as `getBooleanDecoder()` to match the README and docblocks. Not blocking.

Tiny optional wording note on the README, not worth a separate anchor: the paragraph before the last example says you can "compose separate encoder and decoder taps using `combineCodec`", but the example only taps the decoder side. Either tap the encoder too (e.g. a `tapEncoder` value guard, as in the PR description) or soften the sentence — purely cosmetic.

PR #2031, opened 2026-09-04
Add createLazyKeyPairSignerFromBytes to @solana/signers

lorisleiva:
This PR adds `createLazyKeyPairSignerFromBytes`, a synchronous counterpart to `createKeyPairSignerFromBytes`. It derives the address from the public key half of the 64-byte secret key up front, then defers the async `CryptoKey` import until the first signature is requested (memoised so it runs once).

```ts
// Returns synchronously — no `await`.
const signer = createLazyKeyPairSignerFromBytes(keypairBytes);
signer.address; // Available immediately.

// The CryptoKey import happens here, on first use, and is reused thereafter.
const [signatures] = await signer.signTransactions([transaction]);
```

This is useful when a signer must be created in a synchronous context (e.g. registering signers up front, as in the Anchor v2 integration) whilst signing itself stays async.

The returned signer implements both `MessagePartialSigner` and `TransactionPartialSigner`, but — because the `CryptoKeyPair` is imported lazily — it does not expose a `keyPair` property and is therefore not a full `KeyPairSigner`. Only the 64-byte length is validated synchronously; the cryptographic public/private-key match (which the eager variant checks up front) surfaces on the first signing attempt instead.

### DIFF
--- a/packages/signers/src/__tests__/lazy-keypair-signer-test.ts
+++ b/packages/signers/src/__tests__/lazy-keypair-signer-test.ts
@@ -1,0 +1,181 @@
+import '@solana/test-matchers/toBeFrozenObject';
+
+import { getAddressDecoder, getAddressFromPublicKey } from '@solana/addresses';
+import { SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH, SolanaError } from '@solana/errors';
+import { createKeyPairFromBytes, SignatureBytes, signBytes } from '@solana/keys';
+import {
+    partiallySignTransaction,
+    Transaction,
+    TransactionWithinSizeLimit,
+    TransactionWithLifetime,
+} from '@solana/transactions';
+
+import { createLazyKeyPairSignerFromBytes } from '../lazy-keypair-signer';
+import { createSignableMessage } from '../signable-message';
+
+// Partial mocks of the async crypto boundary that the lazy signer defers to.
+jest.mock('@solana/addresses', () => ({
+    ...jest.requireActual('@solana/addresses'),
+    getAddressFromPublicKey: jest.fn(),
+}));
+jest.mock('@solana/keys', () => ({
+    ...jest.requireActual('@solana/keys'),
+    createKeyPairFromBytes: jest.fn(),
+    signBytes: jest.fn(),
+}));
+jest.mock('@solana/transactions', () => ({
+    ...jest.requireActual('@solana/transactions'),
+    partiallySignTransaction: jest.fn(),
+}));
+
+// A valid 64-byte secret key (private key + public key).
+const MOCK_KEY_BYTES = new Uint8Array([
+    0xeb, 0xfa, 0x65, 0xeb, 0x93, 0xdc, 0x79, 0x15, 0x7a, 0xba, 0xde, 0xa2, 0xf7, 0x94, 0x37, 0x9d, 0xfc, 0x07, 0x1d,
+    0x68, 0x86, 0x87, 0x37, 0x6d, 0xc5, 0xd5, 0xa0, 0x54, 0x12, 0x1d, 0x34, 0x4a, 0x1d, 0x0e, 0x93, 0x86, 0x4d, 0xcc,
+    0x81, 0x5f, 0xc3, 0xf2, 0x86, 0x18, 0x09, 0x11, 0xd0, 0x0a, 0x3f, 0xd2, 0x06, 0xde, 0x31, 0xa1, 0xc9, 0x42, 0x87,
+    0xcb, 0x43, 0xf0, 0x5f, 0xc9, 0xf2, 0xb5,
+]);
+
+// The address is the base58 encoding of the last 32 bytes (the public key half).
+const EXPECTED_ADDRESS = getAddressDecoder().decode(MOCK_KEY_BYTES.slice(32));
+
+const getMockCryptoKeyPair = () => ({ privateKey: {}, publicKey: {} }) as CryptoKeyPair;
+
+describe('createLazyKeyPairSignerFromBytes', () => {
+    beforeEach(() => {
+        // By default, the deferred crypto boundary resolves to a mock key pair and address.
+        jest.mocked(createKeyPairFromBytes).mockResolvedValue(getMockCryptoKeyPair());
+        jest.mocked(getAddressFromPublicKey).mockResolvedValue(EXPECTED_ADDRESS);
+    });
+
+    it('derives the address synchronously from the public key half of the secret key', () => {
+        const signer = createLazyKeyPairSignerFromBytes(MOCK_KEY_BYTES);
+        expect(signer.address).toBe(EXPECTED_ADDRESS);
+    });
+
+    it('returns a frozen signer implementing both partial signer interfaces', () => {
+        const signer = createLazyKeyPairSignerFromBytes(MOCK_KEY_BYTES);
+        expect(signer).toBeFrozenObject();
+        expect(typeof signer.signMessages).toBe('function');
+        expect(typeof signer.signTransactions).toBe('function');
+    });
+
+    it('does not expose a keyPair property', () => {
+        const signer = createLazyKeyPairSignerFromBytes(MOCK_KEY_BYTES);
+        expect(signer).not.toHaveProperty('keyPair');
+    });
+
+    it('throws synchronously when the secret key is not 64 bytes long', () => {
+        expect(() => createLazyKeyPairSignerFromBytes(MOCK_KEY_BYTES.slice(0, 63))).toThrow(
+            new SolanaError(SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH, { byteLength: 63 }),
+        );
+    });
+
+    it('defers the key pair import until the first signature is requested', async () => {
+        expect.assertions(2);
+        const signer = createLazyKeyPairSignerFromBytes(MOCK_KEY_BYTES);
+
+        // The import has not happened yet.
+        expect(jest.mocked(createKeyPairFromBytes)).not.toHaveBeenCalled();
+
+        await signer.signTransactions([]);
+        expect(jest.mocked(createKeyPairFromBytes)).toHaveBeenCalledTimes(1);
+    });
+
+    it('imports the key pair only once across multiple sequential signing calls', async () => {
+        expect.assertions(1);
+        const signer = createLazyKeyPairSignerFromBytes(MOCK_KEY_BYTES);
+
+        await signer.signTransactions([]);
+        await signer.signTransactions([]);
+        await signer.signMessages([]);
+
+        expect(jest.mocked(createKeyPairFromBytes)).toHaveBeenCalledTimes(1);
+    });
+
+    it('imports the key pair only once across concurrent signing calls', async () => {
+        expect.assertions(1);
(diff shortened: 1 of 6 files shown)


lorisleiva:
* **#2031** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2031?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2031?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/signers/src/keypair-signer.ts
```diff
@@ -201,3 +205,80 @@ export async function createKeyPairSignerFromPrivateKeyBytes(
...
+ * This is useful when a signer must be created in a synchronous context — for instance, when
+ * registering signers up front — whilst the actual signing can remain asynchronous.
+ *
+ * @typeParam TAddress - Supply a string literal to define a signer having a particular address.
+ *
+ * @param bytes - A 64-bytes secret key, the first 32 of which represent the private key and the last
+ * 32 of which represent its associated public key.
+ * @param extractable - Setting this to `true` makes it possible to extract the bytes of the private
+ * key using the [`crypto.subtle.exportKey()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/exportKey)
+ * API. Defaults to `false`, which prevents the bytes of the private key from being visible to JS.
```
trevor-cortex: This text is inherited from the eager variants, but it doesn't hold here: regardless of `extractable`, the private key bytes are retained in plain JS memory by the `getKeyPairSigner` closure (`frozenBytes`) for as long as the signer is alive. Suggest rewording to something like:

> Defaults to `false`. Note that, because the import is deferred, this helper keeps a copy of the secret key bytes in memory until the first signing call completes, regardless of this setting.

(And ideally pair it with actually zeroing the copy after import — see below.)

packages/signers/src/keypair-signer.ts
```diff
@@ -201,3 +205,80 @@ export async function createKeyPairSignerFromPrivateKeyBytes(
...
+ */
+export function createLazyKeyPairSignerFromBytes<TAddress extends string = string>(
+    bytes: ReadonlyUint8Array,
+    extractable?: boolean,
+): MessagePartialSigner<TAddress> & TransactionPartialSigner<TAddress> {
+    if (bytes.byteLength !== 64) {
+        throw new SolanaError(SOLANA_ERROR__KEYS__INVALID_KEY_PAIR_BYTE_LENGTH, { byteLength: bytes.byteLength });
+    }
+
+    const frozenBytes = Uint8Array.from(bytes);
```
trevor-cortex: Nit: `frozenBytes` isn't frozen — it's a defensive copy. (Typed arrays can't be frozen anyway; `Object.freeze` throws on a non-empty `Uint8Array`.) `secretKeyBytes` or `bytesCopy` would read more accurately.

packages/signers/src/keypair-signer.ts
```diff
@@ -201,3 +205,80 @@ export async function createKeyPairSignerFromPrivateKeyBytes(
...
+    const address = getAddressDecoder().decode(frozenBytes.slice(32)) as Address<TAddress>;
```
trevor-cortex: Nit: `getAddressDecoder().decode(frozenBytes, 32)` reads the same 32 bytes without the intermediate `slice` allocation.

packages/signers/src/keypair-signer.ts
```diff
@@ -201,3 +205,80 @@ export async function createKeyPairSignerFromPrivateKeyBytes(
...
+
+    let keyPairSignerPromise: Promise<KeyPairSigner> | undefined;
+    const getKeyPairSigner = () => (keyPairSignerPromise ??= createKeyPairSignerFromBytes(frozenBytes, extractable));
```
trevor-cortex: Two related points on the memoisation:

**Release the key bytes after a successful import.** Since the promise is never reset, `frozenBytes` is dead once the import resolves — but the closure keeps it alive for the signer's lifetime, so a plaintext private key sits in memory even with `extractable: false`. Something like:

```ts
const getKeyPairSigner = () =>
    (keyPairSignerPromise ??= createKeyPairSignerFromBytes(secretKeyBytes, extractable).then(signer => {
        secretKeyBytes.fill(0);
        return signer;
    }));
```

**Rejections are cached forever.** If `createKeyPairFromBytes` rejects (mismatched halves, or something environmental like `assertPRNGIsAvailable`), every later `signMessages`/`signTransactions` call rejects with the same cached error. That's probably fine for a corrupt key (deterministic), but it's worth deciding explicitly. If you'd rather allow retries:

```ts
.catch(error => {
    keyPairSignerPromise = undefined;
    throw error;
})
```

Either way, a one-liner in `@remarks` describing the behaviour would help.

packages/signers/src/__tests__/lazy-keypair-signer-test.ts
```diff
@@ -0,0 +1,146 @@
...
+        expect(jest.mocked(createKeyPairFromBytes)).toHaveBeenCalledTimes(1);
+    });
+
+    it('imports the key pair only once across multiple signing calls', async () => {
+        expect.assertions(1);
+        const signer = createLazyKeyPairSignerFromBytes(MOCK_KEY_BYTES);
+
+        await signer.signTransactions([]);
+        await signer.signTransactions([]);
+        await signer.signMessages([]);
```
trevor-cortex, line 91: These calls are sequential, so this test would also pass if only the *resolved signer* were memoised rather than the *promise*. Consider adding a concurrent case to pin down the behaviour that actually matters for the "register signers up front, sign later" use case:

```ts
await Promise.all([signer.signTransactions([]), signer.signMessages([]), signer.signTransactions([])]);
expect(jest.mocked(createKeyPairFromBytes)).toHaveBeenCalledTimes(1);
```

trevor-cortex (commented):
## Summary

Adds `createLazyKeyPairSignerFromBytes`, a synchronous factory that derives the address from the public-key half of a 64-byte secret key up front and defers the `CryptoKey` import (via the existing `createKeyPairSignerFromBytes`) until the first `signMessages`/`signTransactions` call, memoising the resulting promise. Returns `MessagePartialSigner & TransactionPartialSigner` (no `keyPair`), so it deliberately isn't a `KeyPairSigner`. Comes with unit tests, type tests and a `minor` changeset for `@solana/signers`.

The core logic is sound: the sync length check mirrors `createKeyPairFromBytes`, the `??=` memoisation is clean, and delegating to the eager path means the public/private match check still runs (just later). Docblock is thorough and the trade-offs are honestly called out in `@remarks`.

## Things I'd address before merging

**1. Private key material lives in JS memory for the signer's lifetime, and the `extractable` docs say otherwise.** `frozenBytes` is a plaintext copy of the full secret key captured by the `getKeyPairSigner` closure, and it's never released — not even after the `CryptoKey` import succeeds and the bytes are no longer needed. The `@param extractable` text (copied from the eager variant) says `false` "prevents the bytes of the private key from being visible to JS", which isn't true for this helper: the bytes are sitting in the closure regardless of `extractable`. I'd (a) zero the copy once the import resolves — since the promise is never reset, the bytes are dead after that point — and (b) reword the docblock so the trade-off is explicit. See inline.

**2. README.** `CLAUDE.md` asks for a README update for any new public API, and `packages/signers/README.md` → "Creating and generating KeyPair signers" documents every sibling (`createKeyPairSignerFromBytes`, `createKeyPairSignerFromPrivateKeyBytes`, `grindKeyPairSigner`, `writeKeyPairSigner`…). A short `createLazyKeyPairSignerFromBytes()` subsection there, noting the sync return, lazy import, no `keyPair`, and deferred validation, would round this off.

**3. Memoised rejection — intentional?** If the deferred import rejects, the rejected promise is cached and every subsequent call fails with the same error. For a mismatched key that's deterministic and arguably desirable, but any transient failure (e.g. `assertPRNGIsAvailable` inside `createKeyPairFromBytes`) is also permanent. Either is a reasonable design; I'd just make it a conscious choice and document it, or reset `keyPairSignerPromise` on rejection so a retry is possible. Inline.

## Nits / optional

- `frozenBytes` isn't frozen (and typed arrays can't be — `Object.freeze` throws on a non-empty `Uint8Array`). `secretKeyBytes` or `bytesCopy` would be more honest.
- `getAddressDecoder().decode(frozenBytes, 32)` avoids the extra `slice` allocation.
- Tests: the "only once" test is sequential, so it would also pass if you memoised the *result* rather than the *promise*. A `Promise.all([...])` variant would pin down the concurrent case, which is the one that matters for the "registering signers up front" use case.
- Minor consistency: the function lives in `keypair-signer.ts` but the tests are in `lazy-keypair-signer-test.ts`. Fine if it's to keep the `createKeyPairFromBytes` mock isolated from the existing file, just noting the mismatch.

## For subsequent reviewers

- Verify the zeroing (if adopted) doesn't interact badly with the `forwards the extractable option` test — it uses `expect.any(Uint8Array)` so it should be fine, but `toHaveBeenCalledWith` on the actual bytes would break since Jest stores a reference.
- The changeset is `minor` for a new export, consistent with the repo's semver policy.
- No `index.ts` change is needed — `keypair-signer.ts` is already `export *`'d.

_(Submitting as a comment rather than "request changes" since I don't have write access on this repo — treat points 1–3 as the blocking ones.)_

trevor-cortex (commented):
## Re-review

All three blocking points from my first pass have been addressed, and the function has moved into its own `lazy-keypair-signer.ts` module (with matching test/typetest files and an `index.ts` export), which resolves the filename mismatch nit too.

**What changed since last time:**

- **Secret key retention** — `secretKeyBytes` is now `fill(0)`'d in the success branch of the `.then`, and the `@param extractable` / `@remarks` docs are explicit that the plaintext copy lives until the first signing call completes. Since `createKeyPairFromBytes` slices the input before handing it to `importKey` (and WebCrypto copies `BufferSource` inputs anyway), zeroing once the outer promise resolves is safe. The new test grabs the exact reference Jest recorded and asserts it's all zeros, which is a nice way to pin this without exposing internals.
- **Failure caching** — rejections now reset `keyPairSignerPromise` so a retry re-imports. Importantly the bytes are *not* zeroed on the failure path, so the retry actually has something to work with. Documented in `@remarks` and covered by the `mockRejectedValueOnce` → success test.
- **README** — new `createLazyKeyPairSignerFromBytes()` subsection under the KeyPair signers section, consistent in tone and wording with its siblings.
- **Nits** — `frozenBytes` → `secretKeyBytes`, `decode(secretKeyBytes, 32)` replaces the `slice`, and there's a `Promise.all` test that proves promise-level (not result-level) memoisation.

**Sanity checks I did on the new code:**

- Concurrency of the reset: `??=` assigns the chained promise synchronously, and the rejection handler only clears the variable — a new import can't be created until it's `undefined`, so there's no window where two imports run or where the reset clobbers a newer promise.
- The address derived synchronously from bytes 32–63 can't diverge from the one `createKeyPairSignerFromBytes` computes internally, because `createKeyPairFromBytes` verifies the public/private match before resolving.
- Changeset is still `minor` and now also describes the zeroing + retry behaviour.

Nothing further from me — LGTM. _(Posting as a comment since this account can't formally approve on this repo.)_

PR #2038, opened 2026-09-08
fix: preserve per-account types in assertAccountsDecoded

### LINKED ISSUE #49: `TAddress` on `assertAccountsDecoded` must be plural
The `assertAccountsDecoded` of `@solana/addresses` has a type parameter `TAddress`. This param is not plural, which implies that although `assertAccountsDecoded` operates over an array of addresses, the type parameter would enforce that every account has the same address.

Both `TAddress` and `TData` should probably be plural, and the return type should map over them to convert from something like `Account<Uint8Array | TData[P], TAddress[P]>` to `Account<TData[P], TAddress[P]>`

o-mid:
#### Problem

`assertAccountsDecoded` treated every account in the array as sharing one address type and one data type. A tuple of accounts with different addresses therefore narrowed to a union, and a tuple with different data types failed to type-check.

#### Summary of Changes

The helper now maps over the input tuple or array and strips encoded `Uint8Array` data from each element, so each index keeps its original address and decoded data type. Homogeneous arrays still narrow as they did before. Runtime behavior is unchanged.

Fixes #49

### DIFF
--- a/packages/accounts/README.md
+++ b/packages/accounts/README.md
@@ -297,14 +297,18 @@ account satisfies MaybeAccount<MockData>;
 
 This function asserts that all input accounts store decoded data, ie not a Uint8Array. As with `assertAccountDecoded` it does not check the shape of the data matches the decoded type, only that it is not a Uint8Array.
 
-```ts
-type MyAccountData = { name: string; age: number };
+When the input is a tuple, each element's address and decoded data type is preserved.
 
-const myAccounts: Account<MyAccountData | Uint8Array, Address>[];
+```ts
+type TokenData = { mint: Address };
+type MintData = { supply: bigint };
+
+const myAccounts: [Account<TokenData | Uint8Array, '1111..'>, Account<MintData | Uint8Array, '2222..'>] = [
+    {} as Account<TokenData | Uint8Array, '1111..'>,
+    {} as Account<MintData | Uint8Array, '2222..'>,
+];
 assertAccountsDecoded(myAccounts);
 
-// now the account data can be used as MyAccountData
-for (const a of account) {
-    account.data satisfies MyAccountData;
-}
+myAccounts[0] satisfies Account<TokenData, '1111..'>;
+myAccounts[1] satisfies Account<MintData, '2222..'>;
 ```
--- a/packages/accounts/src/__typetests__/decode-account-typetest.ts
+++ b/packages/accounts/src/__typetests__/decode-account-typetest.ts
@@ -5,7 +5,32 @@ import { Account, EncodedAccount } from '../account';
 import { assertAccountDecoded, assertAccountsDecoded, decodeAccount } from '../decode-account';
 import { MaybeAccount, MaybeEncodedAccount } from '../maybe-account';
 
+/**
+ * Strict type-equality helper used by typetests below. Resolves to `true` only
+ * if `A` and `B` are mutually assignable AND share the same modifier set (`?`,
+ * `readonly`); otherwise resolves to `false`.
+ *
+ * This is stricter than `satisfies` for two reasons:
+ *
+ * 1. **Bidirectionality.** `A satisfies B` only checks that `A` is assignable
+ *    to `B`. A test using `satisfies` passes if the actual type has extra
+ *    members beyond what we asserted — which would silently mask a regression
+ *    that re-introduced a nested `Omit<...>` wrapper, since `Omit<X, K> & A`
+ *    is still structurally assignable to a flat literal.
+ * 2. **Modifier strictness.** `A satisfies B` tolerates losing `?` (required
+ *    is assignable to optional) and losing `readonly` (readonly is assignable
+ *    to mutable). `Equal` distinguishes `{ x: T }` from `{ x?: T }` and from
+ *    `{ readonly x: T }` because the inferred-position generic comparison
+ *    uses identity rather than assignability for the type parameters.
+ *
+ * Use `Equal` when the exact shape (including modifiers) matters. Use
+ * `satisfies` when one-way assignability is the actual requirement (e.g.
+ * "this value is usable where `Disposable & X` is expected").
+ */
+type Equal<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ? 1 : 2 ? true : false;
+
 type MockData = { foo: 42 };
+type OtherMockData = { bar: 24 };
 type MockDataDecoder = Decoder<MockData>;
 
 {
@@ -41,5 +66,68 @@ type MockDataDecoder = Decoder<MockData>;
     accounts satisfies Account<MockData, Address>[];
     for (const a of accounts) {
         a.data satisfies MockData;
+    }
+}
+
+{
+    // It narrows a homogeneous array without requiring a tuple annotation.
+    const accounts = [] as Account<MockData | ReadonlyUint8Array, Address>[];
+    assertAccountsDecoded(accounts);
+    accounts satisfies Account<MockData, Address>[];
+}
+
+// [DESCRIBE] assertAccountsDecoded per-element types
+{
+    // It preserves a distinct address on each tuple element after narrowing.
+    {
+        const accounts: [
+            Account<MockData | ReadonlyUint8Array, '1111'>,
+            Account<MockData | ReadonlyUint8Array, '2222'>,
+            Account<MockData | ReadonlyUint8Array, '3333'>,
+        ] = [
+            {} as Account<MockData | ReadonlyUint8Array, '1111'>,
+            {} as Account<MockData | ReadonlyUint8Array, '2222'>,
+            {} as Account<MockData | ReadonlyUint8Array, '3333'>,
+        ];
+        assertAccountsDecoded(accounts);
+        true satisfies Equal<
+            typeof accounts,
+            [Account<MockData, '1111'>, Account<MockData, '2222'>, Account<MockData, '3333'>]
+        >;
+        // @ts-expect-error The first account must not be typed with the second address.
+        true satisfies Equal<
+            typeof accounts,
+            [Account<MockData, '2222'>, Account<MockData, '2222'>, Account<MockData, '3333'>]
+        >;
+    }
+
+    // It preserves a distinct data type on each tuple element after narrowing.
+    {
+        const accounts: [
+            Account<MockData | ReadonlyUint8Array, '1111'>,
+            Account<OtherMockData | ReadonlyUint8Array, '2222'>,
+        ] = [
+            {} as Account<MockData | ReadonlyUint8Array, '1111'>,
+            {} as Account<OtherMockData | ReadonlyUint8Array, '2222'>,
+        ];
+        assertAccountsDecoded(accounts);
+        true satisfies Equal<typeof accounts, [Account<MockData, '1111'>, Account<OtherMockData, '2222'>]>;
+        accounts[0].data satisfies MockData;
+        accounts[1].data satisfies OtherMockData;
+        // @ts-expect-error The first account must not be typed with the second data type.
+        accounts[0].data satisfies OtherMockData;
+    }
+
+    // It preserves per-element addresses on a MaybeAccount tuple.
+    {
+        const accounts: [
+            MaybeAccount<MockData | ReadonlyUint8Array, '1111'>,
+            MaybeAccount<OtherMockData | ReadonlyUint8Array, '2222'>,
+        ] = [
+            {} as MaybeAccount<MockData | ReadonlyUint8Array, '1111'>,
+            {} as MaybeAccount<OtherMockData | ReadonlyUint8Array, '2222'>,
+        ];
+        assertAccountsDecoded(accounts);
+        true satisfies Equal<typeof accounts, [MaybeAccount<MockData, '1111'>, MaybeAccount<OtherMockData, '2222'>]>;
     }
 }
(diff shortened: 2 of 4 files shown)


lorisleiva:
@trevor-cortex

packages/accounts/src/decode-account.ts
```diff
@@ -108,34 +108,61 @@ export function assertAccountDecoded<TData extends object, TAddress extends stri
...
+    ? [TAccount] extends [Account<ReadonlyUint8Array | object, string>]
+        ? Account<DecodedAccountData<TData>, TAddress>
+        : MaybeAccount<DecodedAccountData<TData>, TAddress>
+    : [TAccount] extends [Account<infer TData, infer TAddress>]
+      ? Account<DecodedAccountData<TData>, TAddress>
+      : TAccount;
+
+type AssertedDecodedAccounts<TAccounts extends readonly unknown[]> = {
+    -readonly [P in keyof TAccounts]: AssertedDecodedAccount<TAccounts[P]>;
+};
```
trevor-cortex, line 128: Is the `-readonly` needed for something to compile? Homomorphic mapped types already preserve/omit `readonly` based on the input, so stripping it here means a `readonly [Account<…>, Account<…>]` input narrows to a *mutable* tuple after the assertion (mutable is assignable to readonly, so the narrowing takes the candidate). The old overloads didn't accept readonly arrays at all, so this is a new, small soundness hole. If nothing depends on it I'd drop the modifier and add a typetest:

```ts
// It preserves readonly-ness of the input tuple.
{
    const accounts = {} as readonly [Account<MockData | ReadonlyUint8Array, '1111'>, Account<OtherMockData | ReadonlyUint8Array, '2222'>];
    assertAccountsDecoded(accounts);
    true satisfies Equal<typeof accounts, readonly [Account<MockData, '1111'>, Account<OtherMockData, '2222'>]>;
}
```

packages/accounts/src/decode-account.ts
```diff
@@ -108,34 +108,61 @@ export function assertAccountDecoded<TData extends object, TAddress extends stri
     }
 }
 
+type DecodedAccountData<TData> =
+    Exclude<TData, ReadonlyUint8Array | Uint8Array> extends infer TDecoded
+        ? TDecoded extends object
+            ? TDecoded
+            : never
+        : never;
```
trevor-cortex, line 116: Nit: this is equivalent to `Extract<Exclude<TData, ReadonlyUint8Array>, object>` — `Extract` is exactly `T extends U ? T : never` and, being a conditional whose true branch is the check type, TS gives it the `TData & object` constraint you need to satisfy `Account`'s `TData extends Uint8Array | object`. (`Uint8Array` is also redundant in the `Exclude` since it's a subtype of `ReadonlyUint8Array`, though keeping it for readability is fine.) Take it or leave it — only worth changing if tsc is happy with the shorter form.

packages/accounts/src/decode-account.ts
```diff
@@ -108,34 +108,61 @@ export function assertAccountDecoded<TData extends object, TAddress extends stri
...
+
+type AssertedDecodedAccount<TAccount> = [TAccount] extends [MaybeAccount<infer TData, infer TAddress>]
+    ? [TAccount] extends [Account<ReadonlyUint8Array | object, string>]
+        ? Account<DecodedAccountData<TData>, TAddress>
+        : MaybeAccount<DecodedAccountData<TData>, TAddress>
+    : [TAccount] extends [Account<infer TData, infer TAddress>]
+      ? Account<DecodedAccountData<TData>, TAddress>
+      : TAccount;
```
trevor-cortex, line 124: Small precision loss worth knowing about (and matching old behaviour, so not a blocker): an element typed `Account<X, A> & { exists: true }` — what you get after `assertAccountsExist` — matches the outer `MaybeAccount` test *and* the inner `Account` test, so it comes out as `Account<Decoded, A>` and drops the `exists: true` marker. If you want to keep it, one more branch does it:

```ts
? [TAccount] extends [{ readonly exists: true }]
    ? Account<DecodedAccountData<TData>, TAddress> & { readonly exists: true }
    : Account<DecodedAccountData<TData>, TAddress>
```

Also worth a one-line comment on why the checks are wrapped in `[…]` (to keep `MaybeAccount`'s own union from distributing), since that's the non-obvious part of this type.

packages/accounts/src/decode-account.ts
```diff
@@ -108,34 +108,61 @@ export function assertAccountDecoded<TData extends object, TAddress extends stri
...
-): asserts accounts is MaybeAccount<TData, TAddress>[];
-export function assertAccountsDecoded<TData extends object, TAddress extends string = string>(
-    accounts: (Account<ReadonlyUint8Array | TData, TAddress> | MaybeAccount<ReadonlyUint8Array | TData, TAddress>)[],
-): asserts accounts is (Account<TData, TAddress> | MaybeAccount<TData, TAddress>)[] {
+export function assertAccountsDecoded<
+    TAccounts extends readonly (
+        | Account<ReadonlyUint8Array | object, string>
+        | MaybeAccount<ReadonlyUint8Array | object, string>
+    )[],
+>(accounts: AssertedDecodedAccounts<TAccounts> | TAccounts): asserts accounts is AssertedDecodedAccounts<TAccounts> {
```
trevor-cortex, line 165: Suggest a short comment on the `AssertedDecodedAccounts<TAccounts> | TAccounts` parameter type — it's a well-known trick but it reads like a typo to anyone who hasn't hit "A type predicate's type must be assignable to its parameter's type" before. Something like `// The union lets TS accept the mapped type as an assertion target; inference always binds TAccounts from the naked branch.`

packages/accounts/README.md
```diff
@@ -297,14 +297,18 @@ account satisfies MaybeAccount<MockData>;
...
 ```ts
-type MyAccountData = { name: string; age: number };
+type TokenData = { mint: Address };
+type MintData = { supply: bigint };
 
-const myAccounts: Account<MyAccountData | Uint8Array, Address>[];
+const myAccounts: [Account<TokenData | Uint8Array, '1111..'>, Account<MintData | Uint8Array, '2222..'>] = [
+    {} as Account<TokenData | Uint8Array, '1111..'>,
+    {} as Account<MintData | Uint8Array, '2222..'>,
+];
```
trevor-cortex, line 309: The `{} as Account<…>` tuple reads like a typetest rather than user-facing docs (the README skill asks for realistic snippets). The natural source of a heterogeneous tuple is `fetchJsonParsedAccounts`, so — assuming the typetest I suggested in the review body passes — something like this would be both realistic and show off the fix:

```ts
type TokenData = { mint: Address };
type MintData = { supply: bigint };

const accounts = await fetchJsonParsedAccounts<[TokenData, MintData]>(rpc, [tokenAddress, mintAddress]);
assertAccountsDecoded(accounts);

// Each element keeps its own address and data type.
accounts[0] satisfies MaybeAccount<TokenData>;
accounts[1] satisfies MaybeAccount<MintData>;
```

Same applies to the docblock `@example` in `decode-account.ts`.

packages/accounts/src/__typetests__/decode-account-typetest.ts
```diff
@@ -43,3 +68,66 @@ type MockDataDecoder = Decoder<MockData>;
...
 }
+
+{
+    // It narrows a homogeneous array without requiring a tuple annotation.
+    const accounts = [] as Account<MockData | ReadonlyUint8Array, Address>[];
+    assertAccountsDecoded(accounts);
+    accounts satisfies Account<MockData, Address>[];
+}
+
+// [DESCRIBE] assertAccountsDecoded per-element types
```
trevor-cortex, line 79: Three more cases I'd add under this describe (details in the review body): (1) the return type of `fetchJsonParsedAccounts<[MockData, OtherMockData], ['1111', '2222']>` — an intersection of two mapped tuples with `MaybeAccount<…> | MaybeEncodedAccount<…>` elements, which is the main real-world input for this helper; (2) a `readonly` tuple input, asserting readonly-ness is preserved; (3) a call from inside a generic `<TData extends object>` function, asserting the result is still usable as `Account<TData>[]`. Each of those exercises a different part of the mapped/conditional machinery that the current tests don't reach.

trevor-cortex (commented):
## Summary

Replaces the three `assertAccountsDecoded` overloads (single shared `TData`/`TAddress`) with a single signature generic over the whole array/tuple type. A homomorphic mapped type (`AssertedDecodedAccounts`) walks each element, strips `ReadonlyUint8Array`/`Uint8Array` from its `data` type via `DecodedAccountData`, and re-wraps it as `Account<…>` or `MaybeAccount<…>` depending on which shape the element had. The `accounts: AssertedDecodedAccounts<TAccounts> | TAccounts` union is the standard trick to satisfy "predicate type must be assignable to parameter type" for a mapped-type predicate. Runtime body is untouched. Typetests use the `Equal` helper (copied with its docblock per `CLAUDE.md`) to lock down exact per-element types for `Account` and `MaybeAccount` tuples, plus a homogeneous-array regression test.

The type-level design is sound and I walked through the conditional logic for the `Account`, `MaybeAccount`, and `Account & { exists: true }` cases — it resolves the way the tests claim. A few things worth a look before merging:

## Things to watch

**1. `-readonly` in `AssertedDecodedAccounts` makes a readonly input mutable after the assertion.** The constraint accepts `readonly (...)[]`, but the mapped type strips `readonly`, so `declare const xs: readonly [Account<…>, Account<…>]; assertAccountsDecoded(xs);` narrows `xs` to a mutable tuple (a mutable tuple is assignable to a readonly one, so the narrowing takes the candidate). That's a small soundness hole the old overloads didn't have (they rejected readonly arrays outright). Homomorphic mapped types preserve `readonly` on their own, so unless `-readonly` was needed to get something to compile, I'd drop it and add a typetest with a `readonly` tuple input. See inline.

**2. The primary real-world input isn't covered by a typetest.** The main producer of heterogeneous account tuples is `fetchJsonParsedAccounts`, whose return type is an *intersection of two mapped tuples*, with each element being `MaybeAccount<TData & { parsedAccountMeta? }, Addr> | MaybeEncodedAccount<Addr>`. Modern TS does distribute homomorphic mapped types over intersections of tuples, and the `[TAccount] extends [MaybeAccount<infer TData, …>]` branch should union the `infer` candidates into `(TData & meta) | Uint8Array` before `Exclude` strips the bytes — so I *think* this works — but I couldn't verify it without running tsc and it's exactly the kind of shape where mapped-type narrowing can silently fall back to `declared & candidate`. I'd add a typetest along the lines of:

```ts
type Fetched = Awaited<ReturnType<typeof fetchJsonParsedAccounts<[MockData, OtherMockData], ['1111', '2222']>>>;
const accounts = {} as Fetched;
assertAccountsDecoded(accounts);
accounts[0].exists && accounts[0].data satisfies MockData & { parsedAccountMeta?: … };
accounts[1].exists && accounts[1].data satisfies OtherMockData & { … };
```

If that passes, it'd also make a far more realistic README/docblock example than the `{} as Account<…>` tuple (the README skill says snippets should be realistic).

**3. Generic callers.** Under the old overloads a caller like `function f<TData extends object>(xs: Account<TData | Uint8Array>[]) { assertAccountsDecoded(xs); xs satisfies Account<TData>[]; }` got a clean `Account<TData>[]`. Now the element type becomes `Account<DecodedAccountData<TData | Uint8Array>>` with a deferred conditional inside. I believe `satisfies Account<TData>[]` still passes (TS knows `Exclude<TData, X>`'s constraint is `TData`), but hover types get uglier and there may be edge cases where assignability *to* the narrowed type breaks. A one-block typetest inside a generic function would settle it.

**4. Explicit type arguments are a breaking type change.** Anyone calling `assertAccountsDecoded<MyData>(accounts)` or `assertAccountsDecoded<MyData, MyAddress>(accounts)` will now fail to compile, since the single type parameter is the array type. Runtime is unchanged so `patch` may still be the right call under the repo's policy, but flagging it so you can decide deliberately — worth a mention in the changeset either way.

## Known limitations (not blocking, just for the record)

- An element that is `Account<…> & { exists: true }` (e.g. after `assertAccountsExist`) matches the `MaybeAccount` branch, then the `Account` branch, and comes out as plain `Account<…>` — the `exists: true` marker is dropped. Same as the old behaviour, so not a regression. Preservable with one more branch if you care.
- A *union* element type (`Account<A|U8,'1'> | Account<B|U8,'2'>` in a plain array, as opposed to a tuple) still collapses to `Account<A|B, '1'|'2'>` because the `[TAccount]` wrapping disables distribution. Distributing would break `MaybeAccount` handling, so this is a reasonable trade-off, but it's the same class of bug as #49 for a different input shape.

## For subsequent reviewers

The interesting bits are all in `AssertedDecodedAccount` (`decode-account.ts` L118–124). Check the order of the two `extends` tests and convince yourself a plain `Account` (no `exists` key) genuinely fails the `MaybeAccount` match — it does, because both union members of `MaybeAccount` require `exists`. Everything else is mechanical.

o-mid:
Addressed the review:

- Dropped `-readonly` so a readonly tuple stays readonly, with a typetest.
- Switched `DecodedAccountData` to `Extract<Exclude<TData, ReadonlyUint8Array>, object>`.
- Kept `exists: true` when the input already has it (the `assertAccountsExist` shape).
- Commented the `[…]` wraps and the `AssertedDecodedAccounts | TAccounts` parameter union.
- Replaced the `{} as Account` README and docblock example with `fetchJsonParsedAccounts`.
- Added typetests for the `fetchJsonParsedAccounts` tuple, a readonly input, and a generic `Account<TData>[]` caller.

`pnpm --filter @solana/accounts test:typecheck` and the decode-account unit tests pass.

PR #2041, opened 2026-09-08
Add fatal, ignoreBOM and removeNullCharacters options to the UTF-8 codec

lorisleiva:
The UTF-8 codec silently substitutes `U+FFFD` for invalid input (lone surrogates when encoding, malformed byte sequences when decoding), strips a leading byte order mark and strips null characters from decoded strings. These are sensible defaults for padded fixed-size strings but make the codec lossy for formats such as borsh, where a `String` must be valid UTF-8 and may contain `U+0000` or a leading `U+FEFF`.

```ts
const codec = getUtf8Codec();
codec.decode(new Uint8Array([0x61, 0x00, 0x62]));       // "ab": the null character is gone.
codec.decode(new Uint8Array([0xef, 0xbb, 0xbf, 0x61])); // "a": the byte order mark is gone.
codec.decode(new Uint8Array([0xff]));                   // "\ufffd": no error.
codec.encode('\ud800');                                 // 0xefbfbd: the lone surrogate became U+FFFD.
```

This PR adds an optional `Utf8CodecConfig` with three levers. Defaults are unchanged on node and browsers; on React Native, whose text-encoding polyfill never stripped a leading byte order mark, the default now matches the other platforms.

```ts
const codec = getUtf8Codec({ fatal: true, ignoreBOM: true, removeNullCharacters: false });
codec.decode(new Uint8Array([0x61, 0x00, 0x62]));       // "a\0b"
codec.decode(new Uint8Array([0xef, 0xbb, 0xbf, 0x61])); // "\ufeffa"
codec.decode(new Uint8Array([0xff]));                   // Throws SolanaError (SOLANA_ERROR__CODECS__INVALID_UTF8_BYTES, offset 0).
codec.encode('\ud800');                                 // Throws SolanaError (SOLANA_ERROR__CODECS__INVALID_UTF8_STRING, index 0).
```

With `fatal`, decoding relies on the platform `TextDecoder` and only runs a package-side validator to locate the malformed sequence for the error context, or up front on platforms whose polyfill does not reflect the `fatal` attribute (React Native). Differential tests check that validator against the platform implementation on 20k random inputs, and benchmarks in `__benchmarks__/run.ts` put the overhead of `fatal` on node within 4–7%.


lorisleiva:
* **#2041** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2041?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2041?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/codecs-strings/src/utf8.ts
```diff
@@ -60,22 +98,51 @@ export const getUtf8Encoder = (): VariableSizeEncoder<string> => {
...
  * @see {@link getUtf8Codec}
  */
-export const getUtf8Decoder = (): VariableSizeDecoder<string> => {
+export const getUtf8Decoder = (config: Utf8CodecConfig = {}): VariableSizeDecoder<string> => {
     let textDecoder: TextDecoder;
     return createDecoder({
         read(bytes, offset) {
-            const value = (textDecoder ||= new TextDecoder()).decode(bytes.slice(offset));
-            return [removeNullCharacters(value), bytes.length];
+            textDecoder ||= new TextDecoder('utf-8', { fatal: config.fatal ?? false });
```
trevor-cortex: Non-blocking / possible follow-up: `TextDecoder` defaults to `ignoreBOM: false`, so a leading `EF BB BF` is stripped from the decoded string even with `{ fatal: true, removeNullCharacters: false }`. For the borsh "lossless `String`" use case that's one remaining round-trip failure (`encode('\ufeffa')` → `decode` → `'a'`), and the RN polyfill doesn't strip it, so behaviour also differs by platform. If you'd rather not add an `ignoreBOM` option in this PR, a line in the `@remarks` would at least make the caveat discoverable.

packages/codecs-strings/src/utf8.ts
```diff
@@ -29,11 +61,16 @@ import { removeNullCharacters } from './null-characters';
  *
  * @see {@link getUtf8Codec}
  */
-export const getUtf8Encoder = (): VariableSizeEncoder<string> => {
+export const getUtf8Encoder = (
+    config: Omit<Utf8CodecConfig, 'removeNullCharacters'> = {},
```
trevor-cortex: Nit: `Pick<Utf8CodecConfig, 'fatal'>` says the same thing more directly, and it stays correct if a decoder-only option (e.g. `ignoreBOM`) is added to `Utf8CodecConfig` later — with `Omit` that would silently leak into the encoder's accepted config.

packages/codecs-strings/src/assertions.ts
```diff
@@ -29,3 +35,118 @@ export function assertValidBaseString(alphabet: string, testValue: string, given
...
+            if (lead === 0xf4) max = 0x8f; // Above U+10FFFF.
+        } else {
+            return index;
+        }
+        const start = index;
+        index++;
+        for (let i = 0; i < count; i++, index++) {
+            const byte = bytes[index];
+            const inRange = i === 0 ? byte >= min && byte <= max : byte >= 0x80 && byte <= 0xbf;
+            if (index >= length || !inRange) {
```
trevor-cortex: Nit: `bytes[index]` is read before the `index >= length` check. It works because `undefined >= min` is `false` so `inRange` ends up `false`, but checking the bound first (`if (index >= length) return start;` then compute `inRange`) reads more clearly and doesn't rely on `undefined` comparison semantics.

trevor-cortex (commented):
## Summary

**Verdict: LGTM** (I don't have approve permissions on this repo, hence COMMENT.)

Adds an optional `Utf8CodecConfig` to `getUtf8Encoder` / `getUtf8Decoder` / `getUtf8Codec` with two levers: `fatal` (throw `SolanaError` on lone surrogates when encoding and on malformed byte sequences when decoding, instead of substituting `U+FFFD`) and `removeNullCharacters` (opt out of stripping `\u0000` from decoded output). Defaults are unchanged. Two new error codes (`8078026` / `8078027`) are added in sequence in the codecs range with typed context and messages. A package-side validator (`findMalformedUtf8SequenceOffset`) is used both to locate the offending offset for error context and as the primary check on platforms whose `TextDecoder` doesn't honour `fatal`.

This is well put together. The validator's lead/continuation-byte table matches Unicode Table 3-7 (E0→A0..BF, ED→80..9F, F0→90..BF, F4→80..8F, C0/C1/F5+ rejected outright), the differential tests against the platform give me confidence it agrees with WHATWG, and the runtime `textDecoder.fatal` gate is the right way to detect the RN polyfill — I checked `fastestsmallesttextencoderdecoder` and its `TextDecoder` is literally `function TextDecoder(){}` with no `fatal` getter and constructor args ignored, so `!textDecoder.fatal` correctly routes to the up-front validator there.

## Things to watch

- **BOM stripping remains a lossy behaviour** (non-blocking, possible follow-up). `new TextDecoder('utf-8', { fatal })` still defaults to `ignoreBOM: false`, so a leading `EF BB BF` is silently dropped: `getUtf8Codec({ fatal: true, removeNullCharacters: false }).decode(new Uint8Array([0xef, 0xbb, 0xbf, 0x61]))` yields `"a"`, not `"\ufeffa"`. Since the PR's motivation is a lossless codec for borsh `String`, that's one remaining case where `encode` → `decode` doesn't round-trip. It's also a pre-existing platform inconsistency — the RN polyfill does *not* strip the BOM. Worth either an `ignoreBOM` option or at least a mention in the `@remarks`; left an inline note where it would go.
- **Encoder validation lives in `write`, not `getSizeFromValue`.** That's fine for `encode()` since `write` always runs, and it avoids validating twice. Just noting it so nobody assumes `getSizeFromValue` is a validation point.
- **`decodeFatal`'s catch path** reports `findMalformedUtf8SequenceOffset(...)` as the offset. If the platform ever rejects something the validator accepts, the context would carry `offset: -1`. The 20k-sample differential test makes this unlikely; just flagging it as the one place where a validator/platform disagreement would surface.

## For subsequent reviewers

- Error codes are appended in order (`8078026`, `8078027`) and the `SolanaErrorCode` union stays alphabetically sorted. Context shapes mirror existing byte-carrying errors (`bytes: ReadonlyUint8Array` as in `INVALID_PATTERN_MATCH_BYTES`).
- The `utf8-polyfill-test.ts` file mocks `@solana/text-encoding-impl` with `fatal`-less classes and deletes `String.prototype.isWellFormed` for the file's VM context — that exercises both fallback branches (`assertIsWellFormedUtf8Bytes` up front, and the manual surrogate loop). It's restored in `afterAll`, and Jest's per-file contexts mean it can't leak.
- Changeset marks both `@solana/codecs-strings` and `@solana/errors` as `minor`; consistent with a new public option + new error codes under the fixed-version group.

Two small inline nits below; neither blocks.

packages/codecs-strings/src/utf8.ts
```diff
@@ -60,22 +108,62 @@ export const getUtf8Encoder = (): VariableSizeEncoder<string> => {
...
+                ignoreBOM: config.ignoreBOM ?? false,
+            });
+            let value = config.fatal
+                ? decodeFatal(textDecoder, bytes, offset)
+                : textDecoder.decode(bytes.slice(offset));
+            // Polyfills that ignore the `ignoreBOM` option do not reflect it
+            // either, nor do they strip the byte order mark; do it ourselves.
+            if (!config.ignoreBOM && textDecoder.ignoreBOM === undefined && value.charCodeAt(0) === 0xfeff) {
+                value = value.slice(1);
+            }
```
trevor-cortex, line 126: Just to make it explicit for the record: this branch changes the *default* behaviour on React Native. The polyfill never stripped a leading BOM, so `getUtf8Decoder().decode([0xef, 0xbb, 0xbf, 0x61])` used to return `"\ufeffa"` there and now returns `"a"`, matching node/browsers. That's the correct fix, but since the changeset frames the defaults as unchanged, a short mention (e.g. "on React Native, a leading byte order mark is now stripped by default, consistent with other platforms") would save someone a surprise.

trevor-cortex (commented):
## Re-review

**Verdict: LGTM** (COMMENT for lack of approve rights.)

Re-fetched the full diff. Changes since my last pass:

- **`ignoreBOM` option added**, mirroring the `TextDecoder` option name and semantics (`true` = keep the BOM). The decoder now constructs `new TextDecoder('utf-8', { fatal, ignoreBOM })`, and for polyfills that don't reflect `ignoreBOM` (checked via `=== undefined`, which is correct — compliant implementations expose a boolean getter) the leading `U+FEFF` is stripped package-side when `ignoreBOM` is falsy. Tests cover default stripping, preservation, non-leading BOM being untouched, the `encode → decode` round-trip, and the polyfill fallback in both directions. README, module docs, `@remarks` and the changeset are all updated to match.
- **Validator bounds check reordered** so `index >= length` is tested before `bytes[index]` is read. Addressed.
- Encoder config went with `Omit<Utf8CodecConfig, 'ignoreBOM' | 'removeNullCharacters'>` — fine, that's a style call and it's correct as written.

The three options compose cleanly (BOM strip → null strip ordering matches platform semantics, since a BOM after a leading `\0` isn't "leading" either way), and the polyfill test's mocked `TextDecoder` now mimics the real polyfill's non-stripping behaviour by delegating to the global decoder with `ignoreBOM: true`, so the fallback branch is genuinely exercised.

## One thing for reviewers to be aware of

The PR description says defaults are unchanged, which is true on node and browsers. On React Native, though, the default decoder previously did *not* strip a leading BOM (the polyfill never did), and after this PR it does. That's the right outcome — it removes a platform inconsistency — but it is a small behaviour change for RN users of `getUtf8Decoder()` with default config, so it might deserve a line in the changeset. Left an inline note on the spot.

Nothing else from me — nice work.

PR #2042, opened 2026-09-08
Add a requireSizePrefix option to the array, map and set codecs

lorisleiva:
When a collection's size is stored as a prefix, decoding an exhausted byte array yields an empty collection instead of failing. That is a deliberate default: it lets a program append a `Vec` to an existing account layout and still decode accounts created before the change. It is also lenient in a way that some formats cannot accept: borsh, for instance, requires the length prefix to be present, so a truncated buffer must fail rather than decode to `[]`.

```ts
getArrayCodec(getU8Codec()).decode(new Uint8Array([])); // []
getArrayCodec(getU8Codec(), { requireSizePrefix: true }).decode(new Uint8Array([])); // Throws: missing size prefix.
```

This PR adds an opt-in `requireSizePrefix` option to the array, map and set codec configs. When set, the decoder no longer short-circuits on an exhausted byte array and the prefix codec's own byte-length assertion throws. Fixed and `"remainder"` sizes are unaffected, and the default is unchanged; the previously untested default is now covered explicitly.

### DIFF
--- a/packages/codecs-data-structures/src/__tests__/array-test.ts
+++ b/packages/codecs-data-structures/src/__tests__/array-test.ts
@@ -1,7 +1,12 @@
 import { addCodecSizePrefix, fixCodecSize, offsetCodec, resizeCodec } from '@solana/codecs-core';
 import { getU8Codec, getU16Codec, getU32Codec, getU64Codec } from '@solana/codecs-numbers';
 import { getUtf8Codec } from '@solana/codecs-strings';
-import { SOLANA_ERROR__CODECS__INVALID_NUMBER_OF_ITEMS, SolanaError } from '@solana/errors';
+import {
+    SOLANA_ERROR__CODECS__CANNOT_DECODE_EMPTY_BYTE_ARRAY,
+    SOLANA_ERROR__CODECS__INVALID_BYTE_LENGTH,
+    SOLANA_ERROR__CODECS__INVALID_NUMBER_OF_ITEMS,
+    SolanaError,
+} from '@solana/errors';
 
 import { getArrayCodec } from '../array';
 import { b } from './__setup__';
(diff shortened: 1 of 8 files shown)


lorisleiva:
* **#2042** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2042?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2042?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/codecs-data-structures/src/__tests__/array-test.ts
```diff
@@ -135,6 +140,45 @@ describe('getArrayCodec', () => {
...
+    });
+
+    it('can require the size prefix to be present', () => {
+        const strict = { requireSizePrefix: true } as const;
+
+        // Missing prefix.
+        expect(() => array(u8(), strict).read(b(''), 0)).toThrow(
+            new SolanaError(SOLANA_ERROR__CODECS__CANNOT_DECODE_EMPTY_BYTE_ARRAY, { codecDescription: 'u32' }),
+        );
+        expect(() => array(u8(), strict).read(b('ff'), 1)).toThrow(SolanaError);
```
trevor-cortex: Nit, optional: the neighbouring assertions pin the exact `SolanaError` (code + context), while this one only checks the class. For consistency you could assert `new SolanaError(SOLANA_ERROR__CODECS__CANNOT_DECODE_EMPTY_BYTE_ARRAY, { codecDescription: 'u32' })` here too — `bytes.length - offset === 0` at offset 1, so that's the error it raises.

trevor-cortex (commented):
**Verdict: LGTM ✅** (posting as a comment — the bot account lacks approve permission on this repo.)

## Summary

Adds an opt-in `requireSizePrefix?: boolean` to `ArrayCodecConfig`, `MapCodecConfig` and `SetCodecConfig`. The only runtime change is a single extra condition in `getArrayDecoder.read`: the "exhausted bytes → empty collection" short-circuit is now skipped when the flag is set, letting the prefix codec's own `assertByteArrayIsNotEmptyForCodec` / `assertByteArrayHasEnoughBytesForCodec` throw. Map and set already forward `config as object` to `getArrayDecoder`, so no further plumbing is needed. Default behaviour is unchanged and now explicitly covered by tests. Docblocks, README and a `minor` changeset are included.

## Verification notes

- Traced the error paths the tests assert against: `b('')` @0 hits `CANNOT_DECODE_EMPTY_BYTE_ARRAY` (`codecDescription: 'u32'`), `b('0300')` @0 passes the not-empty check and hits `INVALID_BYTE_LENGTH` with `bytesLength: 2, expected: 4`. Both match `numberDecoderFactory` → `codecs-core/assertions`.
- `{ requireSizePrefix: true, size: 0 }` and `size: 'remainder'` correctly bypass the flag since the guard is `typeof size === 'object'`.
- Checked siblings for consistency: `addDecoderSizePrefix` already throws on a missing prefix, and `getNullableDecoder` is only lenient when `prefix: null` is explicitly set. So array/map/set were the only lenient-by-default collection codecs; this PR doesn't leave an obvious inconsistency behind.
- `minor` bump matches the repo's "features → minor" policy in `CLAUDE.md`.

- The flag lives on the shared config type and is therefore accepted (and ignored) by `getArrayEncoder` / `getMapEncoder` / `getSetEncoder`. That mirrors how `description` is encoder-only yet accepted by the decoders, so it's consistent with the existing pattern rather than something to fix here.
- Follow-up outside this repo: the Codama JS renderer emits `getArrayCodec(...)` / `getMapCodec(...)` / `getSetCodec(...)` calls, so borsh-strict clients would want a way to pass `requireSizePrefix: true` through from there.

PR #2046, opened 2026-09-09
Document the per-account input pattern used by generated instruction builders

lorisleiva:
This PR adds the per-account type parameter pattern — the shape the Codama JS renderer now emits (codama-idl/renderers-js#215) — to the `ResolvedInstructionAccountMeta` integration typetests, asserting the same input matrix as the two existing patterns and additionally pinning that excess property checks are preserved on the `input` parameter (a misspelled optional account is a compile error), whilst wider objects and spreads carrying extra properties still compile. A contrast case documents that the single `TInput` pattern forgoes those checks. The `ResolvedInstructionAccountMeta` and `InstructionAccountInputAddress` docblocks are updated to lead with the per-account shape, and stale comments describing the intersection anchor pattern as the shape the renderer must emit are corrected.

### DIFF
--- a/packages/program-client-core/src/__typetests__/instruction-input-resolution-typetest.ts
+++ b/packages/program-client-core/src/__typetests__/instruction-input-resolution-typetest.ts
@@ -370,8 +370,11 @@ type Equal<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ?
     // The following replicates the full type machinery of a generated program client — the
     // instruction type mapping address strings to IDL-declared metas, the input type, and
     // the instruction builder — in order to assert the final `accounts` tuple end-to-end at
-    // realistic call sites. It is the executable specification of the signature shape the
-    // Codama JS renderer must emit in order to use `ResolvedInstructionAccountMeta`.
+    // realistic call sites. Three builder signature shapes are able to use
+    // `ResolvedInstructionAccountMeta` and are asserted below against the same matrix:
+    // an intersection anchor (`Input<...> & TInput`), a single `TInput` type parameter, and
+    // one type parameter per account. The last one is the shape emitted by the Codama JS
+    // renderer and is the executable specification of its output.
     type MockInstruction<
         TAccountPayer extends AccountMeta<string> | string = string,
         TAccountOwner extends AccountMeta<string> | string = string,
@@ -399,12 +402,12 @@ type Equal<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ?
         payer: InstructionSignerInput<TAccountPayer>;
         rent?: InstructionAccountInput<TAccountRent> | null;
     };
-    // IMPORTANT: the parameter must intersect the concrete input type with `TInput`
-    // (`MockInput<...> & TInput`). Referencing the address type parameters only in
-    // `TInput`'s constraint makes their inference fall back to `string` — see the
-    // canary test at the end of this describe block. Additionally, `TInput` must default
-    // to the concrete input type so that call sites providing explicit type arguments
-    // keep working — see the dedicated test below.
+    // INTERSECTION ANCHOR PATTERN: the parameter intersects the concrete input type with
+    // `TInput` (`MockInput<...> & TInput`). Referencing the address type parameters only in
+    // `TInput`'s constraint makes their inference fall back to `string` — see the canary
+    // test further down this describe block. Additionally, `TInput` must default to the
+    // concrete input type so that call sites providing explicit type arguments keep
+    // working — see the dedicated test below.
     const getMockInstructionBuilder = null as unknown as <
         TAccountPayer extends string,
         TAccountOwner extends string,
@@ -544,10 +547,11 @@ type Equal<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ?
         void result;
     }
 
-    // ALTERNATIVE PATTERN: instruction builders may instead capture the caller's input in
-    // a single `TInput` type parameter and recover each account's address type parameter
+    // SINGLE `TInput` PATTERN: instruction builders may instead capture the caller's input
+    // in a single `TInput` type parameter and recover each account's address type parameter
     // using `InstructionAccountInputAddress`. This block asserts the same matrix for that
-    // signature shape.
+    // signature shape. Note that it forgoes excess property checks on the input — see the
+    // contrast test at the end of this describe block.
     const getSingleGenericBuilder = null as unknown as <TInput extends MockInput>(
         input: TInput,
     ) => MockInstruction<
@@ -641,8 +645,8 @@ type Equal<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ?
     // CANARY: referencing the address type parameters only in `TInput`'s constraint does
     // NOT infer them — they fall back to `string`. This asserts the *broken* behaviour on
     // purpose: if a future TypeScript version starts inferring through constraints, this
-    // test will fail, signalling that the renderer's intersection parameter pattern above
-    // could be simplified.
+    // test will fail, signalling that the intersection anchor pattern above could be
+    // simplified.
     {
         const getConstraintOnlyBuilder = null as unknown as <
             TAccountMint extends string,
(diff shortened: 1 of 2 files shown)


lorisleiva:
* **#2046** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2046?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2046?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/program-client-core/src/__typetests__/instruction-input-resolution-typetest.ts
```diff
@@ -654,6 +658,173 @@ type Equal<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ?
...
+        TAccountOwner extends InstructionAccountInput | InstructionSignerInput =
+            | InstructionAccountInput
+            | InstructionSignerInput,
+        TAccountMint extends InstructionAccountInput = InstructionAccountInput,
+        TAccountRent extends InstructionAccountInput = InstructionAccountInput,
+    > = {
+        mint: TAccountMint;
+        owner: TAccountOwner;
+        payer: TAccountPayer;
+        rent?: TAccountRent;
```
trevor-cortex, line 682: The other two shapes in this block declare `rent?: InstructionAccountInput<TAccountRent> | null` and assert `rent: null` in their optional-account test; here `null` is no longer accepted and the test below swaps in `rent: undefined`. The runtime still treats `null` as "absent" (`ResolvedInstructionAccount['value']` includes `null`, `getAccountMetaFactory` checks `!account.value`).

Since this is the executable spec of the renderer's output, does codama-idl/renderers-js#215 emit `rent?: TAccountRent | null` or `rent?: TAccountRent`? If the former, this should be `rent?: TAccountRent | null` and the `rent: null` case can be restored in the matrix — inference from `null` yields no candidate for `TAccountRent`, so it falls back to the constraint and resolves to `ReadonlyAccount<string>` just like the omitted case. If the latter, that's a caller-visible change from the previous generated shape and probably deserves a sentence in the PER-ACCOUNT comment so it doesn't read as an oversight.

packages/program-client-core/src/__typetests__/instruction-input-resolution-typetest.ts
```diff
@@ -654,6 +658,173 @@ type Equal<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ?
...
+    {
+        type Result = ReturnType<typeof getPerAccountBuilder>;
+        true satisfies Equal<Result['accounts'][0], AccountSignerMeta<string> & WritableSignerAccount<string>>;
+        true satisfies Equal<Result['accounts'][1], ReadonlyAccount<string>>;
+        true satisfies Equal<Result['accounts'][2], WritableAccount<string>>;
+    }
+
+    // CONTRAST: the single `TInput` pattern above does NOT perform excess property checks,
+    // since the fresh object literal is inferred as `TInput` itself. This asserts that
+    // limitation on purpose so that the difference between the two patterns is documented.
```
trevor-cortex: Nit: this describe block now has three shapes, and the intersection anchor (`MockInput<...> & TInput`) forgoes excess property checks for the same reason — the literal is inferred into `TInput`, so `rnet` becomes a known property of the intersection. Either reword to "the `TInput`-based patterns" (and optionally add a `getMockInstructionBuilder({ ..., rnet: address5 })` line here so the contrast is pinned for both), or make it explicit that the intersection anchor is out of scope for this contrast.

trevor-cortex (commented):
## Summary

**Verdict: LGTM** (submitting as a comment since this account can't formally approve here).

Adds a third builder signature shape — one type parameter per account, holding that account's *input value* — to the `ResolvedInstructionAccountMeta` integration typetests, and reframes the docblocks so that this shape (what the Codama JS renderer now emits) leads, with the single-`TInput` shape as the alternative. The new block asserts the same resolution matrix as the two existing patterns and additionally pins two properties the other shapes can't offer: excess property checks survive on fresh object literals (`rnet` is a compile error), and wider objects / spreads still pass. A `ReturnType` test covers the no-inference fallback, and a CONTRAST test pins that the single-`TInput` shape does *not* get excess property checks. Stale comments describing the intersection anchor as "the shape the renderer must emit" are corrected.

I walked through the type-level claims and they hold:

- **Constraint fallback** (omitted `rent`, `rent: undefined`, `ReturnType`): with no inference candidate the parameter falls back to its constraint (`InstructionAccountInput`, or `InstructionSignerInput` for `payer`). `InstructionAccountInputAddress` distributes over that union and every member carries a `string` brand → `string`. `ResolvedInstructionAccountMeta` is non-distributive, and neither `[InstructionAccountInput]` nor `[InstructionSignerInput]` as a whole extends `{ role }` or `TransactionSigner`, so it deterministically yields `TAddress` → the IDL-declared meta. The `rent: undefined` case works because `packages/tsconfig/base.json` doesn't enable `exactOptionalPropertyTypes`, so `undefined` matches the optional property's implicit `| undefined` and contributes no candidate.
- **Excess property checks**: the instantiated parameter type is a plain object type, so freshness-based checks apply. In the `TInput` shapes the literal *is* the inferred `TInput` (or a constituent of the intersection), so `rnet` becomes a known property and no error fires. Spread-sourced properties are exempt from excess checks by design, so the `{ ...wider, owner: signer9 }` case is correct as written.
- **Docblock examples**: both compile as written; `TransferInput` gaining a defaulted type parameter keeps the bare `TransferInput` reference in the second example valid.

No changeset needed — test-only plus docblock changes, per the repo's guidance.

## Things to watch out for

- **`null` for optional accounts.** The two existing shapes declare `rent?: InstructionAccountInput<TAccountRent> | null` and the matrix exercises `rent: null`. The new `PerAccountInput` declares `rent?: TAccountRent` with `TAccountRent extends InstructionAccountInput`, so `null` is rejected and the test substitutes `rent: undefined`. The runtime still supports `null` (`ResolvedInstructionAccount['value']` includes `null`, and `getAccountMetaFactory` checks `!account.value`). Since this block is presented as the executable spec of the renderer's output, it'd be good to confirm which shape the renderer actually emits — see inline comment.
- **"Two patterns" in the CONTRAST comment.** There are three shapes in this describe block now, and the intersection anchor shape forgoes excess property checks for the same reason the single-`TInput` one does. Minor wording nit, see inline.

- The `PerAccountInput` type alias carries defaults on its type parameters that the builder never relies on (it always passes all four). They mirror the renderer's emitted input type, not a requirement of the test — don't flag as dead.
- The CONTRAST block intentionally has no `@ts-expect-error`: if a future TS version started performing excess property checks through inferred `TInput`, that call would error and the typetest would fail, which is the signal it's designed to give.

packages/program-client-core/src/__typetests__/instruction-input-resolution-typetest.ts
lorisleiva: Good catch on the inconsistency, but the premise is slightly off: the renderer never accepted null for optional accounts — the pre-PR output was ata?: Address<TAccountAta>; and it's now ata?: TAccountAta;, so there's no caller-visible change here. The | null only ever lived in the two mock inputs of this typetest, which I added to exercise the helper against null values rather than to mirror the renderer. So the per-account block is actually the accurate one. I've added a sentence to the comment so it doesn't read as an oversight, and kept the rent: null cases on the other two shapes for the helper coverage.

trevor-cortex (commented):
## Re-review

**Verdict: LGTM** (comment review — this account can't formally approve here).

Both points from my previous pass are addressed in the latest revision:

- **`null` for optional accounts** — the PER-ACCOUNT comment now states explicitly that, matching the renderer's output, optional accounts are omitted or `undefined` rather than `null`, and that the `| null` on the two `MockInput`-based shapes exists only to exercise the helper's `null` handling. That's exactly the clarification I was after; the `undefined` case in the matrix is the right one for this shape.
- **CONTRAST wording** — now reads "the `TInput`-based patterns" and pins *both* `getMockInstructionBuilder` and `getSingleGenericBuilder` with the `rnet` literal. I checked the intersection anchor case: the fresh literal is inferred into `TInput`, so `rnet` is a known property of `MockInput<...> & TInput` and no excess property error fires, while the address brands still infer from the `MockInput<...>` side. Compiles as intended.

The describe-block header was also expanded to enumerate all three shapes and name the per-account one as the renderer's executable spec, which makes the file self-describing. No changes to `instruction-input-resolution.ts` since the last pass.

Nothing further from me.

PR #2048, opened 2026-09-10
Document the `@solana/program-client-core` package

lorisleiva:
This PR replaces the one-line README of `@solana/program-client-core` with a full one. It leads with the inputs accepted by the instruction builders of generated program clients — `InstructionAccountInput` and `InstructionSignerInput`, including address wrappers such as web3.js's `PublicKey`, program derived addresses and explicit role overrides — and the rules by which they resolve to account metas, then documents every exported type and function following the structure of the other package READMEs.

### DIFF
--- a/packages/program-client-core/README.md
+++ b/packages/program-client-core/README.md
@@ -11,4 +11,193 @@
 
 # @solana/program-client-core
 
-This package contains types and utilities for building Solana program clients. This is mainly used by the JavaScript Codama renderer to generate Kit-compatible program clients.
+This package contains types and utilities for building Solana program clients. It is mainly used by the [JavaScript Codama renderer](https://github.com/codama-idl/renderers-js) to generate Kit-compatible program clients. It can be used standalone, but it is also exported as part of Kit [`@solana/kit`](https://github.com/anza-xyz/kit/tree/main/packages/kit) under the `@solana/kit/program-client-core` subpath.
+
+You will rarely need to import from this package yourself. However, it defines what the instruction builders of generated program clients accept as inputs and how these inputs are resolved into account metas, which is useful to know when using such clients.
+
+## Instruction inputs
+
+The instruction builders of generated program clients accept an input object containing the accounts and arguments of the instruction. The types below describe what may be provided for each account, depending on whether the program's IDL declares it as a signer.
+
+### `InstructionAccountInput`
+
+This type represents the accepted inputs for a non-signer instruction account. Namely, one of the following:
+
+- An `Address` — the most common case.
+- Any object exposing an `address` property — see `HasAddress` in `@solana/addresses`. This lets third-party wrappers around addresses be passed directly, as long as they expose a Kit `Address` under an `address` property. Note that `TransactionSigner` objects satisfy this shape too, in which case they merely act as address carriers for non-signer accounts.
+- A `ProgramDerivedAddress` — i.e. an `[address, bump]` tuple.
+- An `AccountNonSignerMeta` from `@solana/instructions` — i.e. `{ address, role }` — to explicitly override the role derived from the program's IDL, e.g. to mark an account as writable when the IDL declares it as readonly.
+
+```ts
+import { AccountRole } from '@solana/instructions';
+
+// Given a generated instruction builder whose `mint` account is a non-signer account.
+getMintToInstruction({ mint: mintAddress /* ... */ }); // An address.
+getMintToInstruction({ mint: addressWrapper /* ... */ }); // A third-party wrapper exposing an `address` property.
+getMintToInstruction({ mint: mintPda /* ... */ }); // A program derived address.
+getMintToInstruction({ mint: { address: mintAddress, role: AccountRole.READONLY } /* ... */ }); // A role override.
+```
+
+### `InstructionSignerInput`
+
+This type represents the accepted inputs for a signer instruction account. Namely, one of the following:
+
+- A `TransactionSigner` — the most common case.
+- An `AccountSignerMeta` from `@solana/signers` — i.e. `{ address, role, signer }` — to explicitly override the role derived from the program's IDL, e.g. to mark a signer account as writable when the IDL declares it as readonly.
+
+If the account's signature is provided by other means — e.g. when the transaction is signed by a multisig or by a wallet that adds the signature later — use `createNoopSigner()` from `@solana/signers` to satisfy the requirement.
+
+```ts
+import { createNoopSigner } from '@solana/signers';
+
+// Given a generated instruction builder whose `authority` account is a signer account.
+getMintToInstruction({ authority: authoritySigner /* ... */ }); // A transaction signer.
+getMintToInstruction({ authority: createNoopSigner(authorityAddress) /* ... */ }); // A signature provided by other means.
+```
+
+Accounts that may or may not be signers — e.g. an authority that can either be a signer or a multisig account — accept both `InstructionAccountInput` and `InstructionSignerInput`. Providing a `TransactionSigner` for such an account marks it as a signer.
+
+### Resolution rules
+
+The account metas of the resulting instruction are determined from the provided inputs as follows, in order of precedence:
+
+| Provided input                                                                  | Resulting account meta                                                                                               |
+| ------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- |
+| An account meta — i.e. carrying an explicit `role`                              | The provided role is used as-is, regardless of the flags declared by the IDL. Any attached `signer` is preserved.    |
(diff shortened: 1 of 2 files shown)


lorisleiva:
* **#2048** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2048?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2048?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/program-client-core/README.md
```diff
@@ -11,4 +11,187 @@
...
+## Instruction inputs
+
+The instruction builders of generated program clients accept an input object containing the accounts and arguments of the instruction. The types below describe what may be provided for each account, depending on whether the program's IDL declares it as a signer.
+
+### `InstructionAccountInput`
+
+This type represents the accepted inputs for a non-signer instruction account. Namely, one of the following:
+
+- An `Address` — the most common case.
+- Any object exposing an `address` property — see `HasAddress` in `@solana/addresses`. This lets third-party wrappers around addresses, such as the `PublicKey` class of other frameworks, be passed directly. Note that `TransactionSigner` objects satisfy this shape too, in which case they merely act as address carriers for non-signer accounts.
```
trevor-cortex: Nit on the `PublicKey` example: legacy `@solana/web3.js`'s `PublicKey` doesn't expose an `address` property (only `toBase58()` / `toBytes()` / etc.), so it does **not** satisfy `HasAddress` and readers coming from web3.js may try it and hit a type error. The source docblock deliberately says "a framework's address wrapper class" without naming `PublicKey`.

Suggest either dropping the `PublicKey` mention or making the requirement explicit, e.g.:

> This lets third-party wrappers around addresses be passed directly, as long as they expose a Kit `Address` under an `address` property.

(`AccountNonSignerMeta` from `@solana/instructions`, `HasAddress` from `@solana/addresses` — both confirmed against the imports in `instruction-input-resolution.ts`.)

trevor-cortex (commented):
## Summary

Replaces the one-line README of `@solana/program-client-core` with a full one. It opens with what generated instruction builders accept for each account (`InstructionAccountInput` / `InstructionSignerInput`), lays out the runtime resolution rules as a precedence table with the two failure modes, and then documents every export under `## Types` and `## Functions`, matching the layout of `packages/instructions/README.md` and friends.

**Verdict: LGTM** (submitting as a comment because this account can't approve on the repo). Three non-blocking nits inline.

I cross-checked the prose against `src/instruction-input-resolution.ts`, `self-fetch-functions.ts`, `self-plan-and-send-functions.ts` and `instructions.ts`:

- The four rows of the resolution table map 1:1 onto the branches of `getAccountMetaFactory()` (explicit role → `TransactionSigner` with `isSigner !== false` → signer as address carrier → writable flag), and the `'either'` default for a missing `isSigner` is correct.
- Every error code named in the README is the one actually thrown by the corresponding helper.
- The examples for `getAddressFromResolvedInstructionAccount()`, `getResolvedInstructionAccountAsProgramDerivedAddress()` and `getResolvedInstructionAccountAsTransactionSigner()` correctly pass `resolved*.value`, matching the signatures. (Side note, not for this PR: the `@example` blocks in the source docblocks for those three functions pass the whole `resolvedMint` / `resolvedMetadata` / `resolvedAuthority` object instead of `.value`, so they contradict the signature. Worth a follow-up cleanup.)
- The `@solana/kit/program-client-core` subpath exists in `packages/kit/package.json`.
- Intro sentence ("It can be used standalone, but it is also exported as part of Kit…") is the same wording used by the other package READMEs.

Docs-only, so no changeset required.

## Things to watch

All non-blocking, details inline:

1. The `PublicKey` reference in the `InstructionAccountInput` bullet could be read as "legacy web3.js `PublicKey` works out of the box", which it doesn't (no `address` property). Suggest tightening the wording.
2. `ResolvedInstructionAccountMeta` never says what the third type parameter defaults to, or that non-signer accounts should leave it off — that's the one part of the source docblock the README drops, and it's the detail a renderer author needs.
3. `getAddressFromResolvedInstructionAccount()` is the only function entry that doesn't name its error code.

Nothing to run; the content is accurate against `main` as of this review. If you only have time for one section, check the resolution table and the "failure cases" bullets — they're the parts most likely to drift if `getAccountMetaFactory()` changes later.

packages/program-client-core/README.md
```diff
@@ -11,4 +11,187 @@
...
+
+```ts
+const resolvedMint: ResolvedInstructionAccount = { value: mintAddress, isSigner: false, isWritable: true };
+```
+
+### `ResolvedInstructionAccountMeta`
+
+This type helper computes the account meta type produced by an instruction account based on the input provided for it, mirroring the [resolution rules](#resolution-rules) applied at runtime by `getAccountMetaFactory()`. Generated instruction builders use it to accurately type the accounts of the instructions they return.
+
+It takes the type of the provided input, the address type parameter of the account and, optionally, the meta type to produce when a `TransactionSigner` is provided. When the input carries no static information — e.g. when the builder's type parameters fall back to their constraints — it deterministically resolves to the address type parameter, which generated instruction types map to the account meta declared by the program's IDL.
```
trevor-cortex: This is the one place where the README loses information relative to the source docblock. `TSignerMeta` defaults to `TAddress`, and that default is what makes a `TransactionSigner` act as a plain address carrier for non-signer accounts — i.e. for an account the IDL declares `isSigner: false`, the renderer should *omit* the third argument. As written, "optionally, the meta type to produce when a `TransactionSigner` is provided" doesn't tell the reader what happens when they don't provide it, and the example only shows the signer/`'either'` case.

Suggested tweak:

> …and, optionally, the meta type to produce when a `TransactionSigner` is provided. The latter defaults to the address type parameter, so leaving it out — as generated builders do for non-signer accounts — makes signers act as plain address carriers.

Optionally add a second line to the example, e.g. `type MintMeta<TInput extends InstructionAccountInput> = ResolvedInstructionAccountMeta<TInput, InstructionAccountInputAddress<TInput>>;` with a `// Non-signer account.` comment.

packages/program-client-core/README.md
```diff
@@ -11,4 +11,187 @@
...
+
+This function ensures a resolved instruction input is neither `null` nor `undefined` and returns it. It throws a `SolanaError` with code `SOLANA_ERROR__PROGRAM_CLIENTS__RESOLVED_INSTRUCTION_INPUT_MUST_BE_NON_NULL` otherwise, using the provided input name in the error message.
+
+```ts
+const authority = getNonNullResolvedInstructionInput('authority', maybeAuthority);
+```
+
+### `getAddressFromResolvedInstructionAccount()`
+
+This function extracts the address from the value of a resolved instruction account, whether it is an `Address`, a `ProgramDerivedAddress` or any object exposing an `address` property — such as a `TransactionSigner`, an account meta or a third-party address wrapper. It throws if the value is `null` or `undefined`.
```
trevor-cortex: Every other function entry in this section names the `SolanaError` code it throws; this one just says "It throws if the value is `null` or `undefined`." Since it delegates to `getNonNullResolvedInstructionInput()`, the code is `SOLANA_ERROR__PROGRAM_CLIENTS__RESOLVED_INSTRUCTION_INPUT_MUST_BE_NON_NULL` — worth naming it for consistency.

trevor-cortex (commented):
## Re-review

All three nits from my previous pass are addressed, and the side-note about the source docblocks was picked up too:

- **`PublicKey` wording (was L27)** — the bullet now reads "as long as they expose a Kit `Address` under an `address` property" and no longer names web3.js's `PublicKey`. The example comment was updated to match ("A third-party wrapper exposing an `address` property"). ✅
- **`ResolvedInstructionAccountMeta` default (was L94)** — now explicitly says `TSignerMeta` defaults to the address type parameter and that generated builders leave it out for non-signer accounts, with a new `MintMeta` example showing the two-parameter form alongside the existing `AuthorityMeta` one. Both examples type-check against the `[TInput] extends [...]` definition in source. ✅
- **Missing error code on `getAddressFromResolvedInstructionAccount()` (was L149)** — now names `SOLANA_ERROR__PROGRAM_CLIENTS__RESOLVED_INSTRUCTION_INPUT_MUST_BE_NON_NULL`, consistent with the other entries. ✅
- **Docblock `@example`s in `instruction-input-resolution.ts`** — the three examples now pass `resolved*.value`, matching the function signatures. Comment-only change, so still no changeset needed. ✅

I re-read the full README diff against the current source once more. Nothing new to flag: the new `MintMeta` example uses the correct `InstructionAccountInputAddress<TInput>` for the second parameter, and the `InstructionAccountInputAddress` entry's "no brand resolves to `string`" claim matches the fallthrough branch of the conditional type.

**Verdict: LGTM** — submitting as a comment since this account can't approve on the repo.

PR #2054, opened 2026-09-14
Add tsconfig false for react example app config

mcintyre94:
#### Problem

`pnpm dev` was failing for the react-app example, because of an update to rolldown

```sh
Error: Error during dependency optimization:

Build failed with 1 error:

[RESOLVE_ERROR] Could not resolve './chunk-4HY25MPG.browser.mjs' in ../../packages/react/dist/index.browser.mjs
   ╭─[ ../../packages/react/dist/index.browser.mjs:1:54 ]
   │
 1 │ import { useIsomorphicLayoutEffect, useLatest } from './chunk-4HY25MPG.browser.mjs';
   │                                                      ───────────────┬──────────────
   │                                                                     ╰──────────────── Tsconfig not found
───╯
```

#### Summary of Changes

- Update react-app config (only necessary because of the way it uses workspace packages)
- Add a smoke test to make sure this keeps working as dependabot etc updates tooling

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


mcintyre94:
@trevor-cortex

examples/react-app/vite.config.ts
```diff
@@ -26,6 +26,9 @@ export default defineConfig(({ mode }) => ({
         // that both the app and the plugin reference.
         // This is only an issue in our example because of the linked workspace.
         include: ['@solana/react'],
+        rolldownOptions: {
+            tsconfig: false,
+        },
```
trevor-cortex, line 34: The `include` above has a great explanatory comment; this one has none, and `tsconfig: false` looks arbitrary without context. Worth a short note along the lines of:

```ts
// Rolldown auto-discovers `tsconfig.json` by default (walking up from each module). For the
// linked `@solana/react` workspace package that walk-up lands on the monorepo's tsconfigs and
// breaks pre-bundling. Optimized deps are already-compiled JS, so no tsconfig is needed here.
rolldownOptions: {
    tsconfig: false,
},
```

Including the actual error message you hit would also help anyone bisecting a future breakage.

examples/react-app/dev-server-test.mjs
```diff
@@ -0,0 +1,71 @@
...
+        } finally {
+            clearTimeout(forceStop);
+            if (!passed) t.diagnostic(logs);
+        }
+    });
+
+    async function checkModules() {
+        const baseUrl = await ready.promise;
+        // The ready banner and index.html can both succeed before optimization fails.
+        for (const path of ['src/main.tsx', 'node_modules/.vite/deps/@solana_react.js']) {
```
trevor-cortex, line 60: `node_modules/.vite/deps/@solana_react.js` couples the test to Vite's cache dir and flat-id naming scheme, both of which are internal details that could shift in a Vite/rolldown major. It's the pragmatic choice given `main.tsx` doesn't import `@solana/react` directly (so you can't just parse the rewritten import URL out of its transformed output), but I'd add a brief comment saying so, e.g. `// Vite's optimized-dep cache path; if this 404s after a Vite bump, check whether the cache dir or flat id naming changed before assuming pre-bundling broke.`

trevor-cortex (commented):
**Verdict: LGTM** (I can't formally approve on this repo, so posting as a comment.)

## Summary

Fixes `pnpm dev` for `examples/react-app` after a rolldown bump by setting `optimizeDeps.rolldownOptions.tsconfig: false`, and adds a Node `node:test` smoke test (`test:dev`) that boots the Vite dev server with `--force`, waits for the `Local:` banner, and asserts that both `src/main.tsx` and the pre-bundled `@solana/react` chunk are served as 200 JS while the server is still alive. CI runs the smoke test right after `pnpm build` on both Node matrix entries with a 2-minute cap.

The fix itself looks right. Rolldown enabled tsconfig auto-discovery by default (beta.60) and later changed the walk-up semantics to "first `tsconfig.json` that owns the file", so when Vite pre-bundles the linked `@solana/react` from `packages/react/dist/...`, rolldown wanders into the monorepo's tsconfigs. Since dep pre-bundling only ever processes already-compiled JS from `node_modules`, disabling tsconfig loading there loses nothing — the app's own source still goes through `vite:oxc` with the app tsconfig as before. `tsconfig` is a documented `string | boolean` rolldown input option, so `false` is a supported value rather than a hack.

The test is nicely defensive: detached process group + `process.kill(-pid)` so pnpm/sh/vite all get cleaned up, SIGKILL fallback after 5s, `t.signal` on the fetches, racing `checkModules()` against the process exiting, and dumping logs via `t.diagnostic` only on failure. `Promise.withResolvers` matches the repo convention in `CLAUDE.md`.

## Things to watch

- The test hardcodes Vite's optimized-dep cache path (`node_modules/.vite/deps/@solana_react.js`). That's a Vite implementation detail, so if this step starts failing after a future Vite/rolldown bump, first check whether the cache dir or flat-id naming changed before assuming pre-bundling actually regressed. Noted inline.
- `tsconfig: false` currently has no comment, while the neighbouring `include` has a very thorough one. A one-liner recording the actual error would save the next person a bisect. Noted inline.
- The test is POSIX-only (`spawn('pnpm')` without a shell, negative-pid kill). The repo's other scripts already rely on bash-isms, so this isn't a regression, just worth knowing.

- No changeset needed: CI config plus a private example package.
- The smoke test isn't wired into Turborepo, so locally it only works after a `pnpm build` (or `pnpm turbo compile:js compile:typedefs` from the example dir). That's documented in the file header. If it ever becomes annoying it could move into `turbo.json` as a `test:dev` task with `dependsOn: ["^compile:js"]`, but it's fine as-is for a CI guard.
- I couldn't reproduce the original failure from here, so I'm trusting the PR description on the root cause; the CI run on this PR is the real verification that the fix + test work together.

examples/react-app/vite.config.ts
mcintyre94: Added a comment

examples/react-app/dev-server-test.mjs
mcintyre94: Added a comment

lorisleiva (approved):
Nice! Love the regression test.

PR #2061, opened 2026-09-15
Add sentinel size strategy to array-like codecs

lorisleiva:
This PR adds codec support for Codama's `sentinelCountNode` by introducing a fourth `size` strategy on the `array`, `set`, and `map` codecs. Passing a `{ __kind: 'sentinel', sentinel, strategy? }` object ends the collection when the bytes at the next item boundary match the given sentinel, compared at boundaries only.

```ts
const codec = getArrayCodec(getU8Codec(), { size: { __kind: 'sentinel', sentinel: new Uint8Array([0]) } });
codec.encode([1, 2, 3]);
// 0x01020300
//   |      └-- The sentinel that marks the end of the array.
//   └-- 3 items of 1 byte each.
```

Unlike `addCodecSentinel`, the sentinel's bytes may occur inside an item without terminating the collection, since the comparison only happens at item boundaries.

The optional `strategy` mirrors the spec's `sentinelCountStrategy` enumeration: `required` (default) writes the sentinel and demands it when decoding, `optional` writes it but tolerates a buffer that ends without it, and `omitted` never writes it and stops at the end of the buffer.

```ts
const omitted = { __kind: 'sentinel', sentinel: new Uint8Array([0]), strategy: 'omitted' } as const;
getArrayCodec(getU8Codec(), { size: omitted }).encode([1, 2, 3]);
// 0x010203 -- no sentinel written; decoding stops at the end of the buffer.
```

A new `SOLANA_ERROR__CODECS__SENTINEL_MISSING_AT_END_OF_BYTES` error is thrown only under `required` when the buffer is exhausted without the terminator.

The codec deliberately does not enforce the spec's two validity constraints (no item may begin with the sentinel's bytes; under `optional`/`omitted` the sentinel must be no wider than the smallest item), matching `sentinelCountNode` which leaves these to the IDL author. Both constraints are documented on the exported types and in the README, with tests illustrating the premature-stop and skipped-tail failure modes that arise when they are violated.

Spec reference: codama-idl/spec#156 (adds `sentinelCountNode`).


lorisleiva:
* **#2061** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2061?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2061?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/codecs-data-structures/src/array.ts
```diff
@@ -204,6 +290,31 @@ export function getArrayDecoder<TTo>(item: Decoder<TTo>, config: ArrayCodecConfi
...
+                    if (offset + sentinel.length > bytes.length) {
+                        // Not enough bytes remain to hold the sentinel.
+                        if (strategy === 'required') {
+                            throw new SolanaError(SOLANA_ERROR__CODECS__SENTINEL_MISSING_AT_END_OF_BYTES, {
+                                hexSentinel: hexBytes(sentinel),
+                                sentinel,
+                            });
+                        }
+                        break;
+                    }
```
trevor-cortex, line 310: Non-blocking design question. Invariant 2 ("under `optional`/`omitted` the sentinel must be no wider than the smallest item") exists only because this branch conflates two situations: *no bytes remain* and *some bytes remain but fewer than the sentinel*. Under `optional`/`omitted`, the second case can't be a sentinel by definition, so it could just fall through to `item.read` instead of breaking:

```ts
const remaining = bytes.length - offset;
if (remaining <= 0 || (strategy === 'required' && remaining < sentinel.length)) {
    if (strategy === 'required') throw ...;
    break;
}
if (remaining >= sentinel.length && containsBytes(bytes, sentinel, offset)) {
    offset += sentinel.length;
    break;
}
const [value, newOffset] = item.read(bytes, offset);
```

With that, the `01022a` / `ffff` case in the test decodes to `[1, 2, 42]` and the invariant (plus its README/docblock paragraph and the test) goes away. Trade-off: trailing garbage shorter than the sentinel would now surface as an item-decoder error rather than being silently dropped, and it diverges from the spec's stated constraint — so if byte-for-byte parity with codama-rs / other renderers is the goal, keep as is. Just flagging so it's a deliberate choice.

packages/codecs-data-structures/src/array.ts
```diff
@@ -204,6 +290,31 @@ export function getArrayDecoder<TTo>(item: Decoder<TTo>, config: ArrayCodecConfi
...
+            if (isSentinelSize(size)) {
+                const { sentinel, strategy = 'required' } = size;
+                while (true) {
+                    if (offset + sentinel.length > bytes.length) {
+                        // Not enough bytes remain to hold the sentinel.
+                        if (strategy === 'required') {
+                            throw new SolanaError(SOLANA_ERROR__CODECS__SENTINEL_MISSING_AT_END_OF_BYTES, {
+                                hexSentinel: hexBytes(sentinel),
+                                sentinel,
+                            });
```
trevor-cortex, line 307: Nit: `INVALID_NUMBER_OF_ITEMS` and `OFFSET_OUT_OF_RANGE` in this package both carry `codecDescription`, and `config.description` is already in scope here (it's used a few lines up in the encoder's `assertValidNumberOfItemsForCodec` call). Adding `codecDescription: config.description ?? 'array'` — and possibly `offset` / `bytesLength` — would make this a lot easier to debug when it fires deep inside a struct. Would need matching additions in `context.ts` / `messages.ts`.

packages/codecs-data-structures/src/array.ts
```diff
@@ -204,6 +290,31 @@ export function getArrayDecoder<TTo>(item: Decoder<TTo>, config: ArrayCodecConfi
...
+                        break;
+                    }
+                    if (containsBytes(bytes, sentinel, offset)) {
+                        // The sentinel is present; consume it and stop.
+                        offset += sentinel.length;
+                        break;
+                    }
+                    const [value, newOffset] = item.read(bytes, offset);
+                    offset = newOffset;
+                    array.push(value);
```
trevor-cortex, line 318: Pre-existing hazard shared with `remainder`, not a regression: if `item` is zero-width (e.g. `getUnitDecoder()`, or a `fixCodecSize(..., 0)`), `newOffset === offset` and this loop never terminates unless the bytes at `offset` happen to be the sentinel. A one-line guard here (`if (newOffset === offset) throw ...`) would cover both strategies cheaply, but I'd also be fine leaving it out of this PR.

trevor-cortex (commented):
## Summary

Adds a fourth `size` strategy to `getArrayCodec` / `getSetCodec` / `getMapCodec`: `{ __kind: 'sentinel', sentinel, strategy? }`. The decoder loops item-by-item, checking `containsBytes(bytes, sentinel, offset)` at each item boundary; the encoder appends the sentinel unless `strategy === 'omitted'`. `required` (default) throws the new `SOLANA_ERROR__CODECS__SENTINEL_MISSING_AT_END_OF_BYTES` when the buffer runs out; `optional` / `omitted` stop silently. Existing `typeof size === 'object'` checks are replaced with `isPrefixSize` / `isSentinelSize` narrowing so the prefix path isn't confused by the new object shape. Sets and maps get it for free since they delegate to the array codec.

The implementation is correct as far as I can tell — I traced the loop against every test case plus a few extra edge cases (offset > 0, multi-byte sentinel with u16 items, empty input under each strategy) and they all check out. `fixedSize` / `maxSize` correctly fall through to `null` / `undefined` since `computeArrayLikeCodecSize` only handles numeric sizes. Error code slot (8078028) follows sequentially, union/context/messages entries are alphabetised, and the changeset looks CLI-generated. Docs and tests are thorough, including the two failure modes for the unenforced invariants, which is exactly what I'd want to see.

I'd approve this as-is (the bot can't formally approve on this repo); everything below is non-blocking.

## Things to watch out for

- **Invariant 2 is arguably avoidable at the decoder level.** See inline — under `optional` / `omitted`, treating "fewer bytes than the sentinel remain, but > 0" as "not a sentinel, read an item" instead of "stop" would make the short-tail footgun disappear. It's a conscious spec-parity vs. robustness call; either answer is fine, but worth deciding explicitly.
- **Zero-width items spin forever.** `getArrayDecoder(getUnitDecoder(), { size: sentinel })` on any input that doesn't start with the sentinel never advances `offset`. `remainder` has the same pre-existing hazard, so this isn't a regression, but a `newOffset === offset` guard would be cheap. Inline note.
- **Empty sentinel** (`new Uint8Array([])`) degenerates: `containsBytes` always matches, so the decoder immediately returns `[]` and the encoder silently drops every item. Probably worth an `if (sentinel.length === 0) throw` at construction (or a mention in the docblock). Low priority.
- **Error context is thinner than siblings.** Other codec errors in this package carry `codecDescription`; this one doesn't, even though `config.description` is in scope. Inline note.

## Docs nits (unchanged lines, so noting here)

- `getArrayCodec`'s `@remarks` block (around L414–418 in `array.ts`) still lists only the three original strategies.
- `SetCodecConfig` and `MapCodecConfig` docblocks in `set.ts` / `map.ts` also list only three options.
- README: the set section gained a sentinel example but the map section's equivalent snippet didn't. Trivial, and the "see the array codec" link covers it.

## For subsequent reviewers

Nothing needs a second pair of eyes on correctness. The main thing worth a human opinion is the design question on the `optional` / `omitted` short-tail behaviour — whether Kit should match the spec's stated constraint literally or be strictly more lenient than the spec requires.

packages/codecs-data-structures/src/array.ts
lorisleiva: @trevor-cortex Have you seen the original feature spec though? https://github.com/codama-idl/spec/issues/154

This needs to be usable on a program such as `token-2022`.

trevor-cortex: I hadn't — thanks, that settles it. Withdrawn.

The token-2022 extensions region is exactly the case my suggestion breaks: `try_for_each_tlv_extension_type` stops at the first `Uninitialized` header **or when fewer than two bytes remain**, and the unused tail can be any length. With a 1-byte tail, my version would fall through to `item.read` on an `extension` (`u16 type + u16 length + value`) and throw on the leftover byte — precisely the failure the "remaining < size(sentinel) → stop" rule exists to prevent. Same story for the 32-byte zero-padded `ExtraAccountMeta` seed slot.

The `01022a` / `ffff` test I was pointing at is the inverse violation (sentinel wider than the item), and the spec's "Is anything lost by rule 2?" section already argues that layout is broken in every strategy regardless. So the current code matches the spec's decoding algorithm line-for-line, rule 2 is what makes the positional check sufficient, and the documented invariant + test are the right way to handle it. Nothing to change here.

packages/codecs-data-structures/src/array.ts
lorisleiva: Will cover in a follow-up PR.

packages/codecs-data-structures/src/__tests__/array-test.ts
```diff
@@ -118,6 +120,117 @@ describe('getArrayCodec', () => {
...
+        const wide = { __kind: 'sentinel', sentinel: b('ffff') } as const;
+        expect(array(u8(), { size: wide }).encode([42, 1, 2])).toStrictEqual(b('2a0102ffff'));
+        expect(array(u8(), { size: wide }).read(b('2a0102ffff'), 0)).toStrictEqual([[42, 1, 2], 5]);
+
+        // The sentinel is only compared at item boundaries, so it may legitimately appear inside an item.
+        // Here the u16 `256` encodes (little-endian) to `0001`, whose second byte is `00` — part of the
+        // `0000` sentinel — yet decoding does not stop mid-item.
+        const wideItem = { __kind: 'sentinel', sentinel: b('0000') } as const;
+        expect(array(u16(), { size: wideItem }).encode([256, 512])).toStrictEqual(b('000100020000'));
+        expect(array(u16(), { size: wideItem }).read(b('000100020000'), 0)).toStrictEqual([[256, 512], 6]);
```
mcintyre94: I find this test a bit confusing, because `0001` and `0002` don't contain the `0000` sentinel.

Would it work for this test to encode big endian and use `00` as the sentinel?

packages/codecs-data-structures/src/array.ts
```diff
@@ -15,21 +16,99 @@ import {
...
+ *    item.** Otherwise a trailing region shorter than the sentinel but large enough to hold a valid
+ *    item would be skipped: decoding stops as soon as fewer bytes than the sentinel remain, so that
+ *    final item would never be read. This cannot arise under `"required"` because a terminator is
+ *    always written.
+ *
+ * @example
+ * A single `0xff` byte is a poor sentinel for a list of public keys: roughly one key in 256 starts
+ * with `0xff`, so such a key would prematurely terminate the list. A sentinel as wide as an item —
+ * for instance the all-zero (default) public key — avoids this, since only that exact key can ever
+ * match the terminator.
```
mcintyre94: Nit: not sure if it's confusing to tag this `@example` / if the docs will render it expecting code

mcintyre94 (approved):
Nice!

packages/codecs-data-structures/src/array.ts
lorisleiva: I think it would have worked but you're right it's weird. I've pushed that as more information under the first invariant.

packages/codecs-data-structures/src/__tests__/array-test.ts
lorisleiva: Yeah good shout, that example was confusing. I've updated it so it better showcases the "sentinel within items is fine since it is checked at the boundaries" use-case.

PR #2067, opened 2026-09-17
fix(codecs-numbers): reject truncated and overlong shortU16 encodings

latent-9:
Supersedes #1986, which the stale bot closed during review limbo — reopening it is permission-blocked (`ReopenPullRequest: FORBIDDEN`), so re-submitting the identical 2 commits (`bc820fd`, `a3e06ae`) with the verification below.

## Problem

The `shortU16` decoder loop had no upper bound, violating the codec's own documented contract of a **1 to 3 byte** encoding (`maxSize: 3`). Two concrete failure modes:

1. **Overlong chains accepted**: inputs with continuation bits keep decoding, producing values far outside the `0..65535` domain.
2. **Truncated input silently misread**: past the end of the buffer, `bytes[offset]` is `undefined`; `undefined & 0b10000000 === 0` terminates the loop, so truncated input decoded to garbage with an offset beyond `bytes.length` (e.g. `decode([0x80])` → `[0, 2]`).

## Change

Bound the read loop at three bytes with per-byte availability asserts (`INVALID_BYTE_LENGTH`, like every other number decoder); a continuation bit after three bytes throws `NUMBER_OUT_OF_RANGE`.

## Verification (executed, 10/10)

Ran the actual pre/post decoder code with the real `@solana/codecs-core` + `@solana/errors` packages:
- Old: `[0x80]`→`[0,2]`, `[0x80,0x80]` silent, `[0xff,0xff,0xff,0x00]`→2097151
- New: exact claimed codes + contexts; valid decodes unchanged; full 0..65535 round-trip sweep intact.

### DIFF
--- a/packages/codecs-numbers/src/__tests__/short-u16-test.ts
+++ b/packages/codecs-numbers/src/__tests__/short-u16-test.ts
@@ -1,3 +1,9 @@
+import {
+    SOLANA_ERROR__CODECS__INVALID_BYTE_LENGTH,
+    SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE,
+    SolanaError,
+} from '@solana/errors';
+
 import { getShortU16Codec } from '../short-u16';
 import { assertRangeError, assertValid, assertValidEncode } from './__setup__';
 
@@ -52,6 +58,37 @@ describe('getShortU16Codec', () => {
             const bytes = codec.encode(i);
             expect(codec.decode(bytes)).toBe(i);
         }
+    });
+
+    it('throws when the buffer ends before the continuation chain terminates', () => {
+        expect.hasAssertions();
+        // Continuation bit set, but no further bytes.
+        expect(() => shortU16().decode(new Uint8Array([0x80]))).toThrow(
+            new SolanaError(SOLANA_ERROR__CODECS__INVALID_BYTE_LENGTH, {
+                bytesLength: 1,
+                codecDescription: 'shortU16',
+                expected: 2,
+            }),
+        );
+        expect(() => shortU16().decode(new Uint8Array([0x80, 0x80]))).toThrow(
+            new SolanaError(SOLANA_ERROR__CODECS__INVALID_BYTE_LENGTH, {
+                bytesLength: 2,
+                codecDescription: 'shortU16',
+                expected: 3,
+            }),
+        );
+    });
+
+    it('rejects continuation chains that exceed the three-byte encoding', () => {
+        expect.hasAssertions();
+        expect(() => shortU16().decode(new Uint8Array([0xff, 0xff, 0xff, 0x00]))).toThrow(
+            new SolanaError(SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE, {
+                codecDescription: 'shortU16',
+                max: MAX,
+                min: MIN,
+                value: 0x7f | (0x7f << 7) | (0x7f << 14),
+            }),
+        );
     });
 
     it('has the right sizes', () => {
--- a/packages/codecs-numbers/src/short-u16.ts
+++ b/packages/codecs-numbers/src/short-u16.ts
@@ -1,4 +1,5 @@
 import {
+    assertByteArrayHasEnoughBytesForCodec,
     combineCodec,
     createDecoder,
     createEncoder,
@@ -8,6 +9,8 @@ import {
     VariableSizeDecoder,
     VariableSizeEncoder,
 } from '@solana/codecs-core';
+
+import { SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE, SolanaError } from '@solana/errors';
 
 import { assertNumberIsBetweenForCodec } from './assertions';
 
@@ -90,7 +93,8 @@ export const getShortU16Decoder = (): VariableSizeDecoder<number> =>
         read: (bytes: ReadonlyUint8Array | Uint8Array, offset): [number, Offset] => {
             let value = 0;
             let byteCount = 0;
-            while (++byteCount) {
+            while (++byteCount <= 3) {
+                assertByteArrayHasEnoughBytesForCodec('shortU16', byteCount, bytes, offset);
                 const byteIndex = byteCount - 1;
                 const currentByte = bytes[offset + byteIndex];
                 const nextSevenBits = 0b1111111 & currentByte;
@@ -98,10 +102,17 @@ export const getShortU16Decoder = (): VariableSizeDecoder<number> =>
                 value |= nextSevenBits << (byteIndex * 7);
                 if ((currentByte & 0b10000000) === 0) {
                     // This byte does not have its continuation bit set. We're done.
-                    break;
+                    return [value, offset + byteCount];
                 }
             }
-            return [value, offset + byteCount];
+            // Every encoded shortU16 fits in at most three bytes. A continuation bit on
+            // what would be the fourth byte implies a value outside the u16 domain.
+            throw new SolanaError(SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE, {
+                codecDescription: 'shortU16',
+                max: 65535,
+                min: 0,
+                value,
+            });
         },
     });
 
--- a/.changeset/quiet-shorts-refuse.md
+++ b/.changeset/quiet-shorts-refuse.md
@@ -1,0 +1,5 @@
+---
+'@solana/codecs-numbers': patch
+---
+
+Fixed `getShortU16Decoder()` silently accepting malformed input: a truncated buffer decoded to garbage with an offset past the end of the byte array, and continuation chains longer than the documented three-byte encoding produced values outside the u16 domain. Truncated input now throws `SOLANA_ERROR__CODECS__INVALID_BYTE_LENGTH` and overlong chains throw `SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE`, matching the guards every other number decoder already uses.


lorisleiva:
@trevor-cortex

packages/codecs-numbers/src/short-u16.ts
```diff
@@ -9,6 +10,8 @@ import {
     VariableSizeEncoder,
 } from '@solana/codecs-core';
 
+import { SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE, SolanaError } from '@solana/errors';
```
trevor-cortex, line 12: Nit: the blank line between the `@solana/codecs-core` block and this import will most likely trip `simple-import-sort` — the repo keeps all `@solana/*` package imports in a single group (compare `codecs-data-structures/src/array.ts`). `pnpm style:fix` should collapse it.

trevor-cortex (commented):
## Summary

Bounds the `shortU16` decoder's read loop at three bytes. Each iteration now asserts the byte is actually present (`INVALID_BYTE_LENGTH`, same helper the fixed-size number decoders use), and a continuation bit on the third byte throws `NUMBER_OUT_OF_RANGE`. Two tests cover truncation at 1 and 2 bytes and the 3-continuation-bytes case; patch changeset for `@solana/codecs-numbers`.

The core change is correct and I traced the loop by hand: valid 1/2/3-byte encodings return at the same offsets as before, `[0x80]` and `[0x80, 0x80]` now hit the length assert with the exact contexts the tests claim, and `[0xff, 0xff, 0xff]` (with or without a trailing byte) reaches the throw. The existing 0..65535 round-trip sweep guards against regressions on the happy path. Nice that the overlong check doesn't depend on a fourth byte existing.

## Things to watch

**The u16 bound is only half-enforced.** The new error says the value must be in `0..65535`, but a 3-byte encoding whose third byte has no continuation bit is still returned unchecked: `[0x80, 0x80, 0x04]` decodes to `65536` and `[0xff, 0xff, 0x7f]` to `2097151`. Agave's reference decoder (`short_vec.rs`) has a distinct `Overflow` error for exactly this case alongside `ByteThreeContinues`. Since `assertNumberIsBetweenForCodec` is already imported in this file, closing the gap is a one-liner before the `return` — see inline. I'd fold it into this PR so the decoder's contract matches the error it now throws, rather than leaving a second "shortU16 accepts out-of-range values" fix for later.

**Import grouping** — a blank line was introduced between the two `@solana/*` imports in `short-u16.ts`. Elsewhere in the repo (e.g. `codecs-data-structures/src/array.ts`) package imports sit in one group with no blank line, which is what `simple-import-sort` enforces, so `test:lint` will likely flag it. `pnpm style:fix` should sort it out.

## Notes for other reviewers / possible follow-ups (non-blocking)

- The `getShortU16Codec` docblock (unchanged, ~L130–140) still describes the third byte as using "all 8 bits" and lists a 3-byte range of `16,384 to 4,194,303`. That was already inaccurate (the implementation masks every byte to 7 bits, and the encoder rejects anything above 65535), but this PR is explicitly about tightening the decoder's contract, so it's a natural moment to correct that comment to `16,384 to 65,535` and drop the "all 8 bits" line. Adding `@throws` entries to the decoder docblock would also fit the repo's docblock guidance.
- Agave additionally rejects non-canonical "aliased" encodings (a `0x00` byte anywhere after the first, e.g. `[0x80, 0x00]` for `0`). Kit still accepts those. Probably fine to leave — it's a stricter canonicality rule rather than a domain violation — but worth knowing if anyone wants byte-for-byte parity with the validator later.
- Behavioural note for the changelog: inputs that previously decoded (to garbage) now throw. Patch is the right bump for a fix; just flagging that any consumer relying on lenient decoding of malformed data will start seeing errors.

packages/codecs-numbers/src/short-u16.ts
```diff
@@ -90,18 +93,26 @@ export const getShortU16Decoder = (): VariableSizeDecoder<number> =>
...
+                assertByteArrayHasEnoughBytesForCodec('shortU16', byteCount, bytes, offset);
                 const byteIndex = byteCount - 1;
                 const currentByte = bytes[offset + byteIndex];
                 const nextSevenBits = 0b1111111 & currentByte;
                 // Insert the next group of seven bits into the correct slot of the output value.
                 value |= nextSevenBits << (byteIndex * 7);
                 if ((currentByte & 0b10000000) === 0) {
                     // This byte does not have its continuation bit set. We're done.
-                    break;
+                    return [value, offset + byteCount];
```
trevor-cortex, line 109: This is the only exit for a *complete* 3-byte encoding, and `value` here can be anything up to `2^21 - 1`. So `[0x80, 0x80, 0x04]` still returns `65536` and `[0xff, 0xff, 0x7f]` returns `2097151`, even though the throw below now advertises a `0..65535` domain for this codec. Agave's `short_vec` decoder rejects these as `Overflow`, separately from the `ByteThreeContinues` case you're handling.

Since `assertNumberIsBetweenForCodec` is already imported here, this closes the gap:

```suggestion
                    // This byte does not have its continuation bit set. We're done.
                    assertNumberIsBetweenForCodec('shortU16', 0, 65535, value);
                    return [value, offset + byteCount];
```

It's a no-op for 1- and 2-byte encodings (max 16383) so the round-trip sweep is unaffected. Could gate on `byteCount === 3` if you'd rather skip the comparison on the hot path, but it's two integer compares either way.

packages/codecs-numbers/src/__tests__/short-u16-test.ts
```diff
@@ -54,6 +60,37 @@ describe('getShortU16Codec', () => {
...
+        expect.hasAssertions();
+        expect(() => shortU16().decode(new Uint8Array([0xff, 0xff, 0xff, 0x00]))).toThrow(
+            new SolanaError(SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE, {
+                codecDescription: 'shortU16',
+                max: MAX,
+                min: MIN,
+                value: 0x7f | (0x7f << 7) | (0x7f << 14),
+            }),
+        );
+    });
```
trevor-cortex, line 92: If you take the range-check suggestion in `short-u16.ts`, it'd be worth pinning the boundary here too so the decoder's domain is actually asserted rather than only the 4th-byte case:

```ts
// Third byte terminates, but the value overflows u16.
expect(() => shortU16().decode(new Uint8Array([0x80, 0x80, 0x04]))).toThrow(
    new SolanaError(SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE, { ...rangeErrorValues, value: MAX + 1 }),
);
```

`rangeErrorValues` is already defined at the top of the file, so the other two cases in this block could reuse it as well instead of spelling out `codecDescription`/`max`/`min` again.

latent-9:
@trevor-cortex addressed, thanks for the sharp trace:

- **Range gap closed**: `assertNumberIsBetweenForCodec('shortU16', 0, 65535, value)` before the in-loop return, mirroring the encoder. Verified by execution: `[0x80,0x80,0x04]`→`NUMBER_OUT_OF_RANGE {value: 65536}`, `[0xff,0xff,0x7f]`→`{value: 2097151}`, while max-valid `[0xff,0xff,0x03]`→65535 and the full 0..65535 round-trip sweep stay green, plus the three previous regression cases unchanged.
- **Import grouping** fixed (single `@solana/*` group) — that was the exact lint failure in CI.
- **Docblock** corrected (low 7 bits, 16,384–65,535) + regression test for the 65536 case.

Left the aliased-encoding leniency as-is per your non-blocking note.

koriyoshi2041:
The current `bf1e1309` head is still red in both Node jobs on `@solana/codecs-numbers#test:lint`:

- `short-u16.ts:1`: `simple-import-sort/imports`
- `short-u16-test.ts:123`: `jest/no-conditional-in-test`
- `short-u16-test.ts:124,126,198`: `jest/no-conditional-expect`

The production behavior and exhaustive coverage look useful; this appears limited to reshaping the test assertions and applying the repository's import sort before rerunning CI.

packages/codecs-numbers/src/__tests__/short-u16-test.ts
```diff
@@ -62,4 +112,107 @@ describe('getShortU16Codec', () => {
...
+                for (let c = 0; c < 0x04; c += 1) {
+                    buf[0] = a;
+                    buf[1] = b;
+                    buf[2] = c;
+                    const want = (a & 0x7f) | ((b & 0x7f) << 7) | (c << 14);
+                    expect(codec.read(buf, 0)).toEqual([want, 3]);
+                }
+            }
+        }
+    });
```
lorisleiva: I believe this test is redundant (see line 55 above)

latent-9: @lorisleiva you're right, the sweep at line 55 already covers it -- dropping the exhaustive block and the lint-fix commit, reverting to 74c71f1

latent-9:
@lorisleiva You're right — the sweep in the first test already covers those values. Dropped the exhaustive enumeration tests; kept the targeted boundary pins and the import-sort fix.

koriyoshi2041:
Verified the updated `bad233a5` head: both Node build/test jobs and the aggregate `all-web3-checks` gate are green now. The targeted 65,536 boundary pin remains, while the lint-hostile exhaustive test was removed and the import order is fixed. Thanks for closing this out.

packages/codecs-numbers/src/short-u16.ts
```diff
@@ -90,18 +92,31 @@ export const getShortU16Decoder = (): VariableSizeDecoder<number> =>
...
             }
-            return [value, offset + byteCount];
+            // Every encoded shortU16 fits in at most three bytes. A continuation bit on
+            // what would be the fourth byte implies a value outside the u16 domain.
+            throw new SolanaError(SOLANA_ERROR__CODECS__NUMBER_OUT_OF_RANGE, {
+                codecDescription: 'shortU16',
+                max: 65535,
+                min: 0,
+                value,
+            });
```
lorisleiva: Unless I'm mistaken this only happens when we end up with more than 3 bytes to decode. Therefore, I believe the error thrown should be a `SOLANA_ERROR__CODECS__INVALID_BYTE_LENGTH` instead. The `value` here isn't even quite correct because it refers to the "value we've accumulated so far".

packages/codecs-numbers/src/short-u16.ts
```diff
@@ -90,18 +92,31 @@ export const getShortU16Decoder = (): VariableSizeDecoder<number> =>
...
                 const nextSevenBits = 0b1111111 & currentByte;
                 // Insert the next group of seven bits into the correct slot of the output value.
                 value |= nextSevenBits << (byteIndex * 7);
                 if ((currentByte & 0b10000000) === 0) {
                     // This byte does not have its continuation bit set. We're done.
-                    break;
+                    // Every byte only contributes 7 bits, so a terminated
+                    // 3-byte encoding can still exceed the u16 domain
+                    // (e.g. [0x80, 0x80, 0x04] is 65536). Reject it here so
+                    // the decoder never returns a value the encoder refuses.
```
lorisleiva: Remove this comment please.

packages/codecs-numbers/src/short-u16.ts
```diff
@@ -90,18 +92,31 @@ export const getShortU16Decoder = (): VariableSizeDecoder<number> =>
...
+                    assertNumberIsBetweenForCodec('shortU16', 0, 65535, value);
```
lorisleiva: Let's add a local (non-exported) constant in this file instead of a magic number:

```ts
const MAX_SHORT_U16_VALUE = 65535;
```

lorisleiva (changes requested):

latent-9:
Thanks for the review! All three points are addressed in fa6e525a: the overlong-chain case now throws `SOLANA_ERROR__CODECS__INVALID_BYTE_LENGTH` (the chain demands a fourth byte, so it is a byte-length problem, and the accumulated value was indeed misleading in the error context), the explanatory comments are gone, and the magic number is replaced with a local `MAX_SHORT_U16_VALUE` constant.

One note: the `bytesLength: 4` in the error context is the length the encoding demands, not the length of the input — `[0xff, 0xff, 0xff]` (exactly three bytes) hits the same throw. Happy to adjust if you'd rather it read differently.

lorisleiva (approved):
Thanks!

PR #2071, opened 2026-09-18
Fix realloc message packer for exact multiples of the realloc limit

lorisleiva:
This PR fixes `getReallocMessagePackerInstructionPlan` emitting a 0-byte instruction as its last chunk whenever `totalSize` is an exact multiple of the 10,240-byte realloc limit. The size of the last instruction was computed as `totalSize % REALLOC_LIMIT`, which is zero in that case, so the account ended up one full chunk short of the requested size:

```
totalSize   before            after
10_240      [0]               [10240]
20_480      [10240, 0]        [10240, 10240]
15_000      [10240, 4760]     [10240, 4760]   (unchanged)
0           []                []              (unchanged)
```

The chunk sizes are now derived by walking the remaining size down in steps of at most `REALLOC_LIMIT`:

```ts
const instructions: Instruction[] = [];
for (let remaining = totalSize; remaining > 0; remaining -= REALLOC_LIMIT) {
    instructions.push(getInstruction(Math.min(REALLOC_LIMIT, remaining)));
}
```

Tests cover a single full chunk, two full chunks and a zero total size alongside the existing partial-chunk case.


lorisleiva:
* **#2071** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2071?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2071?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Summary

Fixes an off-by-one-chunk bug in `getReallocMessagePackerInstructionPlan`: when `totalSize` was an exact multiple of `REALLOC_LIMIT`, `totalSize % REALLOC_LIMIT` evaluated to `0` for the last instruction, so the plan emitted a useless 0-byte realloc and left the account one full chunk short. The `ceil`/`%` arithmetic is replaced with a simple loop that walks `remaining` down by `REALLOC_LIMIT` and emits `min(REALLOC_LIMIT, remaining)` per step. The docblock is updated to match, and three tests cover the previously-broken cases (single full chunk, two full chunks, zero size) alongside the existing partial-chunk test. A `patch` changeset is included for `@solana/instruction-plans`.

## Verification

I traced the new loop by hand:

| `totalSize` | emitted sizes |
|---|---|
| `0` | `[]` |
| `10_240` | `[10240]` |
| `15_000` | `[10240, 4760]` |
| `20_480` | `[10240, 10240]` |

All match the PR description. The zero-size test relies on `getMessagePackerInstructionPlanFromInstructions([])` behaving sensibly — I checked, and `done()` is `0 >= 0 === true` while `packMessageToCapacity` hits the `ALREADY_COMPLETE` guard first, so the assertions are correct.

One incidental improvement worth noting (not a concern): the old code would have thrown a `RangeError` from `new Array(-1)` for a sufficiently negative `totalSize` (e.g. `-20_000`, since `Math.ceil(-1.95) === -1`). The loop now returns an empty plan for any non-positive `totalSize`, which is a more graceful failure mode for an input that's nonsensical anyway.

## Notes for other reviewers

- The loop is the entire behavioural change; the rest is docs, tests and the changeset. Nothing else in the file is touched.
- `patch` is the right bump per `CLAUDE.md` (bug fix), and the changeset targets the correct package.
- Tests follow the same shape as the existing `getReallocMessagePackerInstructionPlan` tests in the file (same `createInstruction`/`message` helpers, same `toStrictEqual` on `instructions`), so nothing new to learn there.

LGTM — I'd approve, but my token lacks approve rights on this repo, so posting as a comment.

mcintyre94 (approved):
Good spot!

PR #2073, opened 2026-09-18
Add helpers and a generic rejection error for custom message packers

lorisleiva:
This PR makes it practical to write a custom `MessagePacker` without reimplementing the built-in packers' internals. Today a packer that wants to refuse a candidate message for a reason other than byte size (for example a per-transaction realloc budget) has to throw `MESSAGE_CANNOT_ACCOMMODATE_PLAN` with fabricated byte counts, and has to copy the instruction-count validation and the size arithmetic by hand.

Three helpers are now exported from `@solana/instruction-plans`: `resolveMaxInstructionsPerTransaction` validates and defaults the `maxInstructions` option, `assertMaxInstructionsPerTransaction` throws `MAX_INSTRUCTIONS_PER_TRANSACTION_EXCEEDED` when a message would hold too many instructions, and `assertMessageCanAccommodateSize` throws `MESSAGE_CANNOT_ACCOMMODATE_PLAN` with correctly computed byte counts from sizes the caller already has. The two built-in packers use them.

A new `SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_REJECTED_BY_PACKER` error with a free-text `reason` lets a packer refuse a message for any other constraint. The planner treats it like the existing capacity errors, so rejecting a candidate opens a new transaction message while rejecting a fresh message still propagates. The set of errors that mean "use a new candidate" is now exposed as `isMessagePackerErrorThatRequiresNewCandidate`, which the planner uses in place of its inline six-code chain.

```ts
packMessageToCapacity: (message, config) => {
    const maxInstructions = resolveMaxInstructionsPerTransaction(config?.maxInstructions);
    assertMaxInstructionsPerTransaction(message.instructions.length + 1, maxInstructions);
    if (budgetLeft(message) <= 0) {
        throw new SolanaError(SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_REJECTED_BY_PACKER, {
            reason: 'the account cannot grow any further in this transaction',
        });
    }
    const nextMessage = appendTransactionMessageInstruction(nextInstruction, message);
    assertMessageCanAccommodateSize({
        currentSize: getTransactionMessageSize(message),
        nextSize: getTransactionMessageSize(nextMessage),
        sizeLimit: getTransactionMessageSizeLimit(nextMessage),
    });
    return nextMessage;
};
```


lorisleiva:
* **#2073** <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2073?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit/2073?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/errors/src/messages.ts
```diff
@@ -552,6 +553,8 @@ export const SolanaErrorMessages: Readonly<{
     [SOLANA_ERROR__INVARIANT_VIOLATION__INVALID_TRANSACTION_PLAN_KIND]: 'Invalid transaction plan kind: $kind.',
     [SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_PACKER_ALREADY_COMPLETE]:
         'No more instructions to pack; the message packer has completed the instruction plan.',
+    [SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_REJECTED_BY_PACKER]:
+        'The message packer rejected the provided transaction message: $reason. Pack the next instruction(s) into a new transaction message instead.',
```
trevor-cortex: Minor wording thought: the trailing "Pack the next instruction(s) into a new transaction message instead." reads as advice to the planner, but the only time an end user actually *sees* this message is when the planner already tried a fresh message and the packer rejected that too — at which point the advice is exactly what just failed. Consider trimming to something like:

```ts
'The message packer rejected the transaction message: $reason.'
```

or rephrasing the second sentence to describe the situation ("even a new transaction message was rejected") rather than prescribe a fix. Non-blocking.

packages/instruction-plans/src/message-packer-errors.ts
```diff
@@ -0,0 +1,191 @@
...
+
+type MessagePackerErrorThatRequiresNewCandidate = SolanaError<
+    | typeof SOLANA_ERROR__INSTRUCTION_PLANS__MAX_INSTRUCTIONS_PER_TRANSACTION_EXCEEDED
+    | typeof SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_CANNOT_ACCOMMODATE_PLAN
+    | typeof SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_REJECTED_BY_PACKER
+    | typeof SOLANA_ERROR__TRANSACTION__TOO_MANY_ACCOUNT_ADDRESSES
+    | typeof SOLANA_ERROR__TRANSACTION__TOO_MANY_ACCOUNTS_IN_INSTRUCTION
+    | typeof SOLANA_ERROR__TRANSACTION__TOO_MANY_INSTRUCTIONS
+    | typeof SOLANA_ERROR__TRANSACTION__TOO_MANY_SIGNER_ADDRESSES
+>;
```
trevor-cortex: Consider exporting this type (with a docblock, per the repo's docblock rule). Narrowing via the guard works fine without it, but consumers who want to type a handler parameter — e.g. `function onCapacityError(error: MessagePackerErrorThatRequiresNewCandidate)` — currently have no way to name the union and would have to spell out all seven codes themselves. It'd also give the API reference a place to document *which* codes are in the set, rather than that living only in the guard's prose. Non-blocking.

trevor-cortex (commented):
**Verdict: LGTM** — approving in spirit (this account can't formally approve on the repo). Nits below are all non-blocking.

## Summary

Extracts the instruction-count / size validation that the built-in packers relied on into a public `message-packer-errors.ts` module (`resolveMaxInstructionsPerTransaction`, `assertMaxInstructionsPerTransaction`, `assertMessageCanAccommodateSize`), adds a new `SOLANA_ERROR__INSTRUCTION_PLANS__MESSAGE_REJECTED_BY_PACKER` error with a free-text `reason`, and replaces the planner's inline six-code `isSolanaError` chain with an exported `isMessagePackerErrorThatRequiresNewCandidate` guard that also covers the new code. `max-instructions.ts` is folded into the new module.

I traced the refactors and they're behaviour-preserving:

- `resolveMaxInstructionsPerTransaction` = old `assertValid…` + `resolveMaxInstructions` fused. The planner now resolves once up front and `TraverseContext.maxInstructionsPerTransaction` becomes a plain `number`, which removes the repeated `resolveMaxInstructions(...)` calls at every assertion site. Same semantics, less noise.
- In `getMessagePackerInstructionPlanFromInstructions`, the restructured `if (index === instructionIndex) assert… else if (nextSize > sizeLimit) …` is equivalent to the old nested check, and `numBytesRequired` / `numFreeBytes` compute to the same values (`nextSize - originalMessageSize`, `sizeLimit - originalMessageSize`).
- `max-instructions.ts` was never re-exported from `index.ts`, so dropping `DEFAULT_MAX_INSTRUCTIONS_PER_TRANSACTION`'s `export` isn't a public break.
- Rejection semantics in the planner check out: `selectAndMutateCandidate` swallows `MESSAGE_REJECTED_BY_PACKER` and moves on (covering both the `traverseMessagePacker` predicate and the `fitEntirePlanInsideMessage` path for non-divisible / parallel-nested packers), while `createNewMessage` doesn't catch, so a fresh-message rejection propagates. Both planner tests exercise exactly those two outcomes.

## Nits (non-blocking)

- **`MessagePackerInstructionPlan` docblock example is now inconsistent.** The `MessagePacker` type's example was updated to add the `if (!isMessagePackerErrorThatRequiresNewCandidate(error)) throw error;` guard, but the near-identical "Using a message packer plan" example in the `MessagePackerInstructionPlan` docblock (`instruction-plan.ts` ~L205–215, unchanged) still has a bare `catch` with just comments. Worth mirroring the guard there too so the two examples don't teach different things.
- **Consider exporting `MessagePackerErrorThatRequiresNewCandidate`.** See inline.
- **Error message wording.** See inline on `messages.ts`.
- **Test gap (optional):** the planner tests cover `traverseMessagePacker`, but not the `fitEntirePlanInsideMessage` route where a rejection is swallowed and the plan falls back to normal traversal. Something like `sequentialInstructionPlan([A, nonDivisibleSequentialInstructionPlan([rejectingPacker([B])])])` → expect `sequential([Tx: A], [Tx: B])` would pin that down, since it's the path most likely to regress if someone later narrows the catch in `selectAndMutateCandidate`.

- No README change needed here — the package README just links to solanakit.com docs, and per `CLAUDE.md` docs can't reference unpublished types, so the docs follow-up has to land post-release.
- The changeset bumps `@solana/errors` and `@solana/instruction-plans` as `minor`, which matches new-feature precedent.
- The linear packer only adopted the max-instructions helpers, not `assertMessageCanAccommodateSize` — that's correct, its size math has the `+1`/`-1` shortU16 leeway adjustments that don't map onto the generic helper.

trevor-cortex (commented):
**Re-review: LGTM, nothing further** — approving in spirit (this account can't formally approve on the repo).

All four points from the previous pass are addressed in the current head:

- **Error message** trimmed to `'The message packer rejected the provided transaction message: $reason.'` — no more misleading "pack into a new message" advice at the point where a fresh message was already rejected.
- **`MessagePackerErrorThatRequiresNewCandidate`** is now exported with a docblock that enumerates each code in the union and what it means. The guard's return type narrows to it.
- **`MessagePackerInstructionPlan` docblock example** now carries the same `if (!isMessagePackerErrorThatRequiresNewCandidate(error)) throw error;` guard as the `MessagePacker` one, and the `MessagePacker` type docblock gained a paragraph pointing custom packers at the three helpers and the new error.
- **`fitEntirePlanInsideMessage` path** is now covered by the new "falls back to a new transaction message when a message packer rejects the parent candidate of a non-divisible plan" test (`sequential([A, nonDivSeq([rejectingPacker([B])])])` → `sequential([Tx: A], [Tx: B])`). That pins down the `selectAndMutateCandidate` catch for the nested route, which was the one most likely to regress silently.

Re-checked the rest of the diff (`transaction-planner.ts`, `__setup__.ts`, `codes.ts`, `context.ts`, changeset) — unchanged from the prior pass, no regressions. Planner tests now cover all three outcomes: reject-candidate-via-`traverseMessagePacker`, reject-candidate-via-`fitEntirePlanInsideMessage`, and reject-fresh-message propagates.

Good to merge from my side.

mcintyre94 (approved):
LGTM! I found these helpers a little odd because they're just comparing two numbers and throwing an error, but I think they make it much more straightforward to write a packer and it wouldn't make sense to pass any other input (like the transaction message or something). So I think this makes sense!

